PR #83, opened 2016-05-23
Add correctness tests

mstemm:
Add overall statistics reporting to falco to count the number of rules that triggered and add correctness tests using the test running framework [Avocado](https://avocado-framework.github.io/). A set of positive and negative trace files are available on s3 and are pulled down by travis and run against falco, which simply verifies the positive traces have a detected event and the negative traces do not detect anything.

Based on these tests, do a pass of rule cleanups to improve fn/fp coverage.

This is somewhat blocked on a sysdig bug I ran into (https://github.com/draios/sysdig/issues/598), although I'd be ok merging this and re-enabling the rule once the sysdig bug is fixed.

@henridf


userspace/falco/falco.cpp
```diff
@@ -28,6 +28,14 @@ extern "C" {
 #include "utils.h"
 #include <yaml-cpp/yaml.h>
 
+bool g_terminate = false;
+//
+// Helper functions
+//
+static void signal_callback(int signal)
```
henridf: ha, i had this initially then took it out as it wasn't necessary at the time...

henridf:
🎉 we have tests!
👍

PR #84, opened 2016-05-24
Quote path variables that may contain spaces.

mstemm:
Make sure that references to variables that may be paths (which in turn
may contain spaces) are quoted, so cmake won't break on the spaces.

This fixes https://github.com/draios/falco/issues/79.

@henridf

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,7 +42,7 @@ set(PROBE_DEVICE_NAME "sysdig")
 
 set(CMD_MAKE make)
 
-set(SYSDIG_DIR ${PROJECT_SOURCE_DIR}/../sysdig)
+set(SYSDIG_DIR "${PROJECT_SOURCE_DIR}/../sysdig")
 
 include(ExternalProject)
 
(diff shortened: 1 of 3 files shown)


henridf:
👍

CMakeLists.txt
```diff
@@ -151,7 +151,7 @@ ExternalProject_Add(lpeg
                     DEPENDS luajit
 		    URL "http://s3.amazonaws.com/download.draios.com/dependencies/lpeg-1.0.0.tar.gz"
                     URL_MD5 "0aec64ccd13996202ad0c099e2877ece"
-                    BUILD_COMMAND LUA_INCLUDE=${LUAJIT_INCLUDE} ${PROJECT_SOURCE_DIR}/scripts/build-lpeg.sh
+                    BUILD_COMMAND LUA_INCLUDE=${LUAJIT_INCLUDE} "${PROJECT_SOURCE_DIR}/scripts/build-lpeg.sh"
```
elfring: How do you think about to enclose also this include parameter by double quotes?

userspace/falco/CMakeLists.txt
```diff
@@ -1,12 +1,12 @@
-include_directories(${PROJECT_SOURCE_DIR}/../sysdig/userspace/libsinsp/third-party/jsoncpp)
+include_directories("${PROJECT_SOURCE_DIR}/../sysdig/userspace/libsinsp/third-party/jsoncpp")
```
elfring: Would it be better to call this CMake command only once with a longer parameter list?

elfring:
Thanks for [your improvement of build scripts](https://github.com/draios/falco/commit/18f4a203385360d7c96b8aec0e8f12b9bb81ad29 "More quoting for build scripts").

CMakeLists.txt
mstemm: I could be wrong, but I though it wasn't necessary as LUAJIT_INCLUDE is defined in terms of LUAJIT_SRC, which is quoted:

```
set(LUAJIT_INCLUDE "${LUAJIT_SRC}")
```

userspace/falco/CMakeLists.txt
mstemm: I'm not sure it's a big difference, is it?

CMakeLists.txt
elfring: I suggest to apply quoting rules consistently. It does not matter that the previous assignment of the mentioned variable is fine while its use for an other parameter might be a bit unsafe.

userspace/falco/CMakeLists.txt
elfring: I suggest to consider another coding style issue. Do you eventually care also for run time characteristcs of CMake commands?

PR #87, opened 2016-06-01
Update fbash rules to use proc.sname.

mstemm:
Update fbash rules to use proc.sname instead of proc.aname and to rely
on sessions instead of process ancestors.

I also wanted to add details on the address/port being listened to but
that's blocked on https://github.com/draios/falco/issues/86.

Along with this change, there are new positive trace files
installer-bash-starts-network-server.scap and
installer-bash-starts-session.scap that test these updated rules.


mstemm:
Tests passed, so going ahead and merging.

PR #98, opened 2016-07-08
Add list support to rules file.

mstemm:
Once sysdig adds support for handling "in (...)" filter expressions as
set membership tests, it will be advantageous to combine lists of items
together into a single list so they can all be checked in a single set
membership test.

This commit adds support for a new yaml item type "list" containing a
field "name" and field "items" containing a list of items. These are
represented as a yaml list, which allows yaml to handle some of the
initial parsing with the list items maintained natively in lua.

When parsing macro or rule conditions, replace all references to a list
name with the list items as a comma separated string.

Modify the falco rules to switch to lists whenever possible. The
new convention is to use the suffix _binaries for lists of program names
and _procs for macros that define a filter expression using the list.

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -68,76 +68,76 @@
 - macro: linux_so_dirs
   condition: ubuntu_so_dirs or centos_so_dirs or fd.name=/etc/ld.so.cache
 
-- macro: coreutils_binaries
-  condition: >
-    proc.name in (truncate, sha1sum, numfmt, fmt, fold, uniq, cut, who,
+- list: coreutils_binaries
+  items: [
+    truncate, sha1sum, numfmt, fmt, fold, uniq, cut, who,
     groups, csplit, sort, expand, printf, printenv, unlink, tee, chcon, stat,
-    basename, split, nice, yes, whoami, sha224sum, hostid, users, stdbuf,
+    basename, split, nice, "yes", whoami, sha224sum, hostid, users, stdbuf,
     base64, unexpand, cksum, od, paste, nproc, pathchk, sha256sum, wc, test,
     comm, arch, du, factor, sha512sum, md5sum, tr, runcon, env, dirname,
     tsort, join, shuf, install, logname, pinky, nohup, expr, pr, tty, timeout,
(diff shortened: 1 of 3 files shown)


rules/falco_rules.yaml
```diff
@@ -307,13 +305,13 @@
 # sshd, sendmail-msp, sendmail attempt to setuid to root even when running as non-root. Excluding here to avoid meaningless FPs
 - rule: non_sudo_setuid
   desc: an attempt to change users by calling setuid. sudo/su are excluded. user "root" is also excluded, as setuid calls typically involve dropping privileges.
-  condition: evt.type=setuid and evt.dir=> and not user.name=root and not userexec_binaries and not proc.name in (sshd, sendmail-msp, sendmail)
+  condition: evt.type=setuid and evt.dir=> and not user.name=root and not proc.name in (userexec_binaries, sshd, sendmail-msp, sendmail)
```
henridf: so a list `(a, b, c)` can contain a mix of lists and individual elements, and the result is a list where the included lists have been automatically expanded? e.g. `(a, b, [c, d, e])` is the same thing as `(a, b, d, c, e)`?

mstemm: Currently, lists cannot contain references to other lists. I had started on that support, but it made the code more complex and there wasn't a need for it with the current set of binaries.

Macros/rules, like this one, can contain a mix of individual items and list references, and you're right that the list reference is expanded directly to a single top-level set of items.

rules/falco_rules.yaml
```diff
@@ -68,76 +68,76 @@
...
 # privileges or as a different user.
-- macro: userexec_binaries
-  condition: proc.name in (sudo, su)
+- list: userexec_binaries
+  items: [sudo, su]
 
-- macro: user_mgmt_binaries
-  condition: (login_binaries or passwd_binaries or shadowutils_binaries)
+- list: user_mgmt_binaries
+  items: [login_binaries, passwd_binaries, shadowutils_binaries]
```
mstemm: This contradicts what I said in reply to @henridf's question. I'll go add support for list items containing list references.

mstemm:
@henridf if you have other feedback lemme know, but I'll go ahead and merge this now.

PR #103, opened 2016-07-28
Falco engine

mstemm:
These are the changes to make it possible to embed falco in another program. See the individual commits for more details on the changes.

This depends on https://github.com/draios/sysdig/pull/640.

@luca3m


userspace/engine/falco_common.h
```diff
@@ -0,0 +1,69 @@
...
+// be of this type.
+//
+
+struct falco_exception : std::exception
+{
+	falco_exception()
+	{
+	}
+
+	~falco_exception() throw()
```
luca3m: should the destructor be virtual?

aleks-f: whenever class is meant to be inherited from, destructor should be virtual. I'd remove that destructor exception specification, too.

mstemm: I was just following the example of sinsp_exception. Should I change both to be virtual and to drop the exception specification?

luca3m: I noticed that it's the same for sinsp_exception, let's change both

aleks-f: sorry, my bad - I didn;t pay attention; won't go because it inherits from `std::exception` - it's a wart we have to carry.

mstemm: You're right. I had thought we could remove the exception specification in the derived class, but I looked a the c++ standard and this example is explicitly mentioned:

```
struct B {
virtual void f() throw (int, double);
virtual void g();
};
struct D: B {
void f(); // ill-formed
void g() throw (int); // OK
};
```

PR #125, opened 2016-10-03
Add the new pmatch operator.

mstemm:
Make changes to the lua-specific rule parser/compiler to handle the
pmatch operator.


nebi-frame:
what does it do can you give an example?

mstemm:
Here's an example: https://github.com/falcosecurity/falco/blob/dev/rules/falco_rules.yaml#L1169. It matches a complete path name against a set of path prefixes. It returns true if the complete path has one of the prefixes as a prefix.

PR #145, opened 2016-10-26
Added envvar SYSDIG_SKIP_LOAD to Dockerfile to skip kernel module manipulation

carlsverre:
This helps when running on a system which has the module loaded, but getting
access to the module file is hard for some reason.  Since I know that the right
version of the module is loaded I just want Falco to connect.

I tested this with this run command:

docker run -e SYSDIG_SKIP_LOAD=1 -it -v /dev:/host/dev -v /proc:/host/proc --privileged falco

And it successfully connected to Sysdig and started printing out warnings for my
system.

### DIFF
--- a/docker/dev/docker-entrypoint.sh
+++ b/docker/dev/docker-entrypoint.sh
@@ -1,13 +1,17 @@
 #!/bin/bash
 #set -e
 
-echo "* Setting up /usr/src links from host"
+# Set the SYSDIG_SKIP_LOAD variable to skip loading the sysdig kernel module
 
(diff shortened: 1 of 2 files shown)


mstemm:
Hi, I'm ooto until next Wednesday but I'll take a look at it then!

mstemm:
These changes look great. Could you review our contributor license agreement (CLA) and update your commit with a signed-off message? The instructions are here: https://github.com/draios/falco#what-do-i-need-to-do-in-order-to-contribute-code.

carlsverre:
Done :)  Thanks!

PR #153, opened 2016-12-02
Adding DNF as non-alerting for RPM and package management

djcross:
DNF is the present package management for fedora. RHEL/CentOS will likely follow.  Tested on Fedora 24.


mstemm:
Thanks so much for the update to the ruleset! Merging.

PR #154, opened 2016-12-02
If we're excluding the process sudo we should also ignore it's children

djcross:
Otherwise we see alerts such as ` command=unix_chkpwd user nullok parent=sudo`

I'm do have a slight concern here that I could be opening up a vector for false negatives. I _think_ we should be good and can confirm that running for instance `sudo passwd` will still alert.

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -346,7 +346,7 @@
 
 - rule: User mgmt binaries
   desc: activity by any programs that can manage users, passwords, or permissions. sudo and su are excluded. Activity in containers is also excluded--some containers create custom users on top of a base linux distribution at startup.
-  condition: spawned_process and proc.name in (user_mgmt_binaries) and not proc.name in (su, sudo) and not container and not proc.pname in (cron_binaries, systemd, run-parts)
+  condition: spawned_process and proc.name in (user_mgmt_binaries) and not proc.name in (su, sudo) and not container and not proc.pname in (cron_binaries, sudo, systemd, run-parts)
   output: "User management binary command run outside of container (user=%user.name command=%proc.cmdline parent=%proc.pname)"
   priority: WARNING
 


mstemm:
With this rule, `sudo passwd` will no longer trigger. I tested it by running sysdig as `sudo sysdig "evt.type = execve and evt.dir=< and proc.name in (passwd) and not proc.name in (su, sudo) and not proc.pname in (cron_binaries, sudo, systemd, run-parts)"` and then running `sudo passwd`. Sysdig (as expected) didn't detect anything.

I think this could indeed present too big an opening that could be exploited. Anyone could change users/groups/etc. at will simply by running them via sudo.

Were you running into false positives that this would fix? Maybe we can find another rule change to address the false positives without allowing general use of these programs via sudo.

djcross:
Testing on Fedora 24 I get alerts with this new rule:

/etc/falco_rules.yaml:

`  condition: spawned_process and proc.name in (user_mgmt_binaries) and not proc.name in (su, sudo) and not container and not proc.pname in (cron_binaries, sudo, systemd, run-parts)`

`sudo passwd danx`

```
09:45:30.700987596: Warning User management binary command run outside of container (user=root command=unix_chkpwd danx nullok parent=passwd)
09:45:30.709749742: Warning User management binary command run outside of container (user=root command=unix_chkpwd danx chkexpiry parent=passwd) 
```

The alerts I had intended to suppress by this update are the following:

`sudo passwd danx`

```
09:54:19.269198950: Warning User management binary command run outside of container (user=root command=unix_chkpwd danx nullok parent=sudo)     
09:54:21.743268371: Warning User management binary command run outside of container (user=root command=unix_chkpwd danx nullok parent=sudo)                                                                                         09:54:21.752917089: Warning User management binary command run outside of container (user=root command=unix_chkpwd danx chkexpiry parent=sudo)
```

However running through now I can indeed see the following which has become suppressed:

`09:54:21.758933000: Warning User management binary command run outside of container (user=root command=passwd danx parent=sudo)`

As you said, perhaps a stricter rule addition for this would work. I'll give it more of a think.

mstemm:
I think this rule should always trigger whenever anyone runs `sudo passwd danx`. That's the kind of potentially suspicious activity we want to detect.

However, a lot of administrative activities end up doing things to a system that look exactly like suspicious activities. For example, think of  "Write below binary dir". If you were in the middle of installing a new package that installs files below /bin, /usr/bin, etc. you'd expect that rule to trigger, even though this instance is not a true indication of a problem.

We eventually want to have a way to indicate to falco that administrative tasks are ongoing (or have it detect that itself by carefully watching for administrative programs spawning). https://github.com/draios/falco/issues/149 tracks that. But for now I think the best thing to do is accept some level of noisiness.

djcross:
yeah, I had wanted to surpress `command=unix_chkpwd danx nullok parent=sudo` as `command=passwd danx parent=sudo` is the useful alert to me.  I think it would be best to just close off this PR as I obviously went down the wrong path here. Cheers!

PR #169, opened 2016-12-16
Add systemd as a login binary

jcoetzee:
SSH'ing into an Ubuntu 16.04 box triggers a bunch of "Sensitive file opened for reading by non-trusted program" errors caused by systemd. Not sure if this should replace or compliment systemd-logind?

falco-CLA-1.0-signed-off-by: Jonathan Coetzee <jon@thancoetzee.com>


mstemm:
Yeah, I would think that systemd-logind would be the one handing actions at login too. This change seems safe, however. Thanks!

PR #171, opened 2016-12-17
Add cmp as trusted binary to read sensitive files

jcoetzee:
cmp will periodically compare /etc/shadow with the backup stored in /var/backups

`Warning Sensitive file opened for reading by non-trusted program (user=root name=cmp command=cmp -s shadow.bak /etc/shadow file=/etc/shadow) container=host (id=host)`


mstemm:
cmp is pretty generic, so I'm not sure we want to generally allow it to run. Do you know what program is running cmp? Maybe we could look for that instead as a parent using `proc.aname`, which looks up the tree of process ancestors.

mstemm:
Sorry for the noise, I was syncing dev with master as a part of the 0.5.0 release. If you re-fork and reapply the simple change the noise will go away.

mstemm:
I wonder if this related pr might address the problem you saw: https://github.com/draios/falco/pull/192.

mstemm:
I think #192 fixed this, so I'm going to close the PR for now. Let us know if you have additional changes and please add them in a new PR, thanks!

PR #182, opened 2017-01-03
Update third party libs to address security vulnerabilities

mstemm:
Update third-party libraries to address security vulnerabilities.


mstemm:
Results of security scan after these changes:

screen shot 2017-01-03 at 5 46 54 pm
screen shot 2017-01-03 at 5 47 04 pm
screen shot 2017-01-03 at 5 47 12 pm

PR #200, opened 2017-01-27
Rule updates related to other security products

mstemm:
@juju4, here's a PR that merges most of the changes you suggested with the changes I had already made. 

Generally, if you have sysdig trace files you can share for any of
  these cases, it would be very useful. I can add them to the suite of
  regression tests.

Here are some notes on the differences between the two PRs:
  - I addressed linux-image-4.4 somewhat differently, by having a macro parent_linux_image_upgrade_script which looks for linux-image-* and allows it to run shells. I'm guessing the only FP you noticed was linux-image-4.4 spawning shells?
  - I'm assuming critical-stack is short for critical-stack-intel? From what I could see while running it, the only setuid it does is to drop privileges from root. Given that, the current Non sudo setuid rule should not trigger. Am I missing something?
  - From what I could tell by reading the source code, it's not necessary to add cloud-init to a list dev_creation_binaries as the only file it writes to below /dev is /dev/console. Adding /dev/console to the exclusions list should be enough. Is this correct?
  - For the list of host intrusion detection systems, I made a few changes:
     - Of the files in the list, I only saw the program aide accessing sensitive files.
     - I only saw the aide wrapper binaries aide.wrapper and update-aide.con and syslog-summary spawning shells. Notably not logcheck.
     - Based on that, I only kept a list aide_wrapper_binaries containing aide.wrapper and update-aide.con and added that to the shell spawning rules. The only thing in hids_binaries is aide, and was added to Read sensitive file.  I also added syslog-summary to known_shell_spawn_binaries.
     - I renamed hids_procs (and a few other *_procs in other places) to hids_binaries--generally I use the extension _binaries for lists of programs, and _procs for macros that check that a proc.name (or proc.pname) is in the list of binaries. Also, only aide is in this list.
  - For the list of network intrusion detection systems, I also added a bro_running_python to capture python scripts run by broctl that spawn shells.
  - For the list of monitoring procs, I kept icinga2, nrpe, and check_sar_perf.(py), but I didn't keep the more general scripting programs like perl/php, etc. Since perl and php are so generic I'd rather now allow them to spawn shells. If you know of specific contexts in which python/php/etc scripts are being run, we can add exceptions for those. Look at things like ansible_running_python/bro_running_python for examples.
  - I didn't see orchestration_procs/testing_procs being used anywhere so I didn't include it. I did add ansible/ansible-playboo to the list of shell spawners.
  - From looking at the postfix source, I didn't see the postfix program showq changing its uid via setuid. Do you have a repro case for that? If so, I'll add it to the mail_binaries list.
  - The additions to Read sensitive file untrusted (i.e. iptables, ps, ...) were already covered by the list read_sensitive_files_binaries, so I left them in the list. I did add hids_binaries (which only contains aide, as I mentioned before)
  - In Write below rpm database, rather than allow all python programs, I used the macro ansible_running_python.
  - I made several changes to Run shell in container:
     - Do you have an example of containers that run package management commands as they are run. I definitely know of images that use package management commands to install additional software on top of a base image, but I'd think that would be at image creation time, not container run time.
     - I didn't include monitoring_binaries in the set of container shell spawners as I think it's not all that common to deploy nagios or icinga2 in containers. It looks like nagios containers are community-built, and the icinga2 container says its for dev/test/demo use only.
     - Do you have an example of a container that runs init, install-info, ifup, mkinitramfs, cloud-init, or man-db in a container? Like package_mgmt_procs, I would think it would be more at image creation time.

Wanna take a look at the changes and let me know what you think?

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -99,10 +99,13 @@
   items: [setup-backend, dragent, sdchecks]
 
 - list: docker_binaries
-  items: [docker, dockerd, exe]
+  items: [docker, dockerd, exe, docker-compose]
 
 - list: k8s_binaries
   items: [hyperkube, skydns, kube2sky, exechealthz]
+
+- list: lxd_binaries
+  items: [lxd, lxcfs]
 
 - list: http_server_binaries
   items: [nginx, httpd, httpd-foregroun, lighttpd]
@@ -118,8 +121,8 @@
 - list: package_mgmt_binaries
   items: [
     dpkg, dpkg-preconfigu, dnf, rpm, rpmkey, yum, frontend,
-    apt, apt-get, apt-add-reposit, apt-auto-remova, apt-key,
-    preinst
+    apt, apt-get, aptitude, add-apt-reposit, apt-auto-remova, apt-key,
+    preinst, update-alternat, unattended-upgr
     ]
 
 - macro: package_mgmt_procs
@@ -139,11 +142,26 @@
 - list: user_mgmt_binaries
   items: [login_binaries, passwd_binaries, shadowutils_binaries]
 
+- list: dev_creation_binaries
+  items: [blkid]
+
+- list: aide_wrapper_binaries
+  items: [aide.wrapper, update-aide.con]
+
+- list: hids_binaries
+  items: [aide]
+
+- list: nids_binaries
+  items: [bro, broctl]
+
+- list: monitoring_binaries
+  items: [icinga2, nrpe, npcd, check_sar_perf.]
+
 - macro: system_procs
   condition: proc.name in (coreutils_binaries, user_mgmt_binaries)
 
 - list: mail_binaries
-  items: [sendmail, sendmail-msp, postfix, procmail, exim4]
+  items: [sendmail, sendmail-msp, postfix, procmail, exim4, pickup, showq]
 
 - macro: sensitive_files
   condition: fd.name startswith /etc and (fd.name in (/etc/shadow, /etc/sudoers, /etc/pam.conf) or fd.directory in (/etc/sudoers.d, /etc/pam.d))
@@ -209,6 +227,9 @@
 - macro: python_running_denyhosts
   condition: proc.name=python and (proc.cmdline contains /usr/sbin/denyhosts or proc.cmdline contains /usr/local/bin/denyhosts.py)
 
+- macro: parent_bro_running_python
+  condition: proc.pname=python and proc.cmdline contains /usr/share/broctl
+
 # As a part of kernel upgrades, dpkg will spawn a perl script with the
 # name linux-image-N.N. This macro matches that.
 - macro: parent_linux_image_upgrade_script
@@ -231,7 +252,7 @@
                           package_mgmt_binaries, ssl_mgmt_binaries, dhcp_binaries,
                           ldconfig.real, ldconfig, confd, gpg, insserv,
                           apparmor_parser, update-mime, tzdata.config, tzdata.postinst,
-                          systemd-machine, debconf-show)
+                          systemd-machine, debconf-show, rollerd, bind9.postinst)
     and not proc.pname in (sysdigcloud_binaries)
     and not fd.directory in (/etc/cassandra, /etc/ssl/certs/java)
     and not ansible_running_python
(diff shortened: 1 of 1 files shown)


juju4:
Here my comments.
* I'm remerging with my file and deploying on a few systems to validate for next week. would be nice if consider_containers macro was included too :)
```
- macro: consider_containers
#  condition: not container
  condition: evt.num>0
```
one issue pcmdline is not available in released falco, right? else can't validate rules. my current falco is 0.5.0

* linux-image. yes. mostly using ubuntu xenial which only have linux 4.4, yaketty has 4.8 but not using.
examples
```
falco: Warning Shell spawned by untrusted binary (user=root shell=sh parent=linux-image-4.4 cmdline=sh -c /usr/sbin/update-init
ramfs -c -k 4.4.0-53-generic >&2)
falco: Warning Shell spawned by untrusted binary (user=root shell=sh parent=linux-image-4.4 cmdline=sh -c readlink -q -m "/boot
/initrd.img-4.4.0-53-generic")
falco: Warning Shell spawned by untrusted binary (user=root shell=sh parent=linux-image-4.4 cmdline=sh -c basename "/boot//abi-
4.4.0-53-generic")
falco: Warning Shell spawned by untrusted binary (user=root shell=sh parent=linux-image-4.4 cmdline=sh -c readlink -q -m "/boot
/vmlinuz-4.4.0-53-generic")
falco: Warning File below /etc opened for writing (user=root name=apt-auto-remova command=apt-auto-remova /etc/kernel/postinst.
d/apt-auto-removal 4.4.0-53-generic /boot/vmlinuz-4.4.0-53-generic file=/etc/apt/apt.conf.d/01autoremove-kernels.dpkg-new)
```
* yes, https://intel.criticalstack.com/
```
falco: Warning Unexpected setuid call by non-sudo, non-root program (user=critical-stack parent=systemd command=critical-stack- --debug pull --loop uid=critical-stack)
```
Yes file has no setuid or special bit. not sure, why I got this alert many times
* I believe it's this alert
```
falco: Warning File created below /dev by untrusted program (user=root command=cloud-init /usr/bin/cloud-init modules --mode=final file=/dev/console)
```
* HIDS: yes, aide is the one checksumming files. could be other tools still like ossec or tripwire but it isn't on the systems I test currently. Right many are about spawning shells
* monitoring procs.yes agreed. It's a problem of older monitoring (which are still widely used) that are spawing shells too much.
* postfix showq. found this alert
```
falco: Warning Unexpected setuid call by non-sudo, non-root program (user=postfix parent=master command=showq -t unix -u -c uid=postfix)
```
* on "run shell inside containers", it might be interesting to differentiate docker and lxd...
In my view, recommended way to use docker is one process and ephemeral container and as such you are right.
lxd is more a system container like a mini-VM, as such it can live long, be patched and modified like a normal system. with that, you deploy inside lxd like any VM, para-VM or bare-metal systems. so icinga2, init and so on.

mstemm:
I can add the consider_containers macro, but I hadn't modified any of the container-based rules to use it. Is that something you had been doing in your rules file?

Right, pcmdline is in the dev branch,  but I haven't done a release yet with it. I was waiting for this PR to do a release, how about I release right after this merges? You can also try out the sysdig/falco:dev tag in docker hub and/or the dev channel for packages, we build that every day.

I'll add "critical-stack-" to the setuid list, it's setuid()ing for you :)

I'll add showq to the setuid list as well.

I see what you mean about lxd. I think that's something where we would want to do a full pass over the container rules, so I think I'll defer it to https://github.com/draios/falco/issues/203 if that's ok.

So to summarize, I'll add critical-stack- and showq to the setuid list, you're ok with the remaining differences, you'll do some additional testing, and then I'll merge?

juju4:
In my file, I add consider_containers macro for multiple rules but I ask myself if I should re-test without as false-positives cleaning is a lot more advanced it was.

there is no urgency for pcmdline. not too widely use for now so minor inconvenience during merge.
ok critical-stack- and showq and yes lxd need more review.

another little change, add npcd (pnp4nagios) to monitoring
```
- list: monitoring_binaries
  items: [icinga2, nrpe, npcd, check_sar_perf.]
```

Thanks

mstemm:
I'm going to go ahead and merge this now. If you have other changes to make, please go ahead and file another PR or issue. Thanks so much for the rule updates!

PR #204, opened 2017-01-31
CMakeLists: add dependencies to lyaml project (fix #130)

### LINKED ISSUE #130: lyaml build shall depends upon luajit
On release 0.3.0:

Build of lyaml failed when testing if luajit-prefix/src/luajit is a Lua interpreter.

Adding a dependency upon juajit for lyaml in the CMakeLists.txt allos the compilation to go further

ret2libc:
I am not sure whether it's a good idea requiring `autoreconf` or it would be better to add also that as an ExternalProject.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,7 +53,7 @@ option(USE_BUNDLED_DEPS "Enable bundled dependencies instead of using the system
 
 #
 # zlib
-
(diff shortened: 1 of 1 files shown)


mstemm:
I think keeping autoreconf as a dependency is fine. Thanks!

Issue #60, opened 2016-05-12
Categorize rules into profiles


mstemm:
Categorize the existing set of rules into profiles like 'production server', 'desktop', 'devops management', etc, using the tags we'll define in https://github.com/draios/falco/issues/58.

mstemm:
This was fixed in https://github.com/draios/falco/issues/206.

PR #206, opened 2017-02-04
Add tags

### LINKED ISSUE #59: Add ability to run only those rules matching one or more tags
Add the ability to run only those rules matching one or more tags. This will allow us to come up with profiles for 'production server', 'desktop', etc.

mstemm:
Changes to add tags to falco rules and enable/disable sets of rules based on tags. See commits for details.

This, along with the corresponding changes in https://github.com/draios/sysdig/pull/746, fixes #59 and #60.


mstemm:
I know you guys aren't as familiar with the falco code, but since there are corresponding changes on the sysdig side I thought I'd have you take a look here as well.

mstemm:
The TravisCI failure is because it's building against mainline sysdig, which doesn't have the changes in https://github.com/draios/sysdig/pull/746 yet.

mstemm:
I'm going to make some additional changes to support globbing on rule names in addition to tag matches. Let me make those changes before you take a look.

mstemm:
Ok, I've done the additional changes. Any feedback welcome!

userspace/engine/falco_engine.cpp
```diff
@@ -107,20 +108,51 @@ void falco_engine::load_rules_file(const string &rules_filename, bool verbose, b
...
+	{
+		ruleset_id = find_ruleset_id(*ruleset);
+	}
+
+	m_evttype_filter->enable_tags(tags, enabled, ruleset_id);
+}
 
+uint16_t falco_engine::find_ruleset_id(std::string &ruleset)
+{
+	auto it = m_known_rulesets.find(ruleset);
```
bertocci: Similar comments to the sysdig review: lower_bound, insert with a hint, and no second find to cut this down from 3 map walks to just 1

userspace/engine/falco_engine.h
```diff
@@ -59,12 +69,25 @@ class falco_engine : public falco_common
 	};
 
 	//
+	// Return the ruleset id corresponding to this ruleset name,
+	// creating a new one if necessary. If you provide any ruleset
+	// to enable_rule/enable_rule_by_tag(), you should look up the
+	// ruleset id and pass it to process_event().
+	//
+	uint16_t find_ruleset_id(std::string &ruleset);
```
bertocci: const ruleset

userspace/engine/falco_engine.h
```diff
@@ -47,9 +48,18 @@ class falco_engine : public falco_common
...
+	// can later be passed as an argument to process_event(). This
+	// allows for different sets of rules being active at once.
 	//
-	void enable_rule(std::string &pattern, bool enabled);
+	void enable_rule(std::string &pattern, bool enabled, std::string *ruleset = NULL);
+
+	//
+	// Enable/Disable any rules with any of the provided tags (set, exact matches only)
+	//
+	void enable_rule_by_tag(std::set<std::string> &tags, bool enabled, std::string *ruleset = NULL);
```
bertocci: const tags

userspace/engine/falco_engine.h
```diff
@@ -120,6 +144,8 @@ class falco_engine : public falco_common
 	inline bool should_drop_evt();
 
 	falco_rules *m_rules;
+	uint16_t m_next_ruleset_id;
+	std::map<string, uint16_t> m_known_rulesets;
```
bertocci: Is it possible to have an empty string as the key in a map of strings? It feels weird, but I'm pretty sure it works. If you populate m_known_rulesets[""] = 0 in the constructor, this simplifies a lot of the code here and in falco.cpp. No more passing naked pointers, checking if ruleset is not NULL, etc.

userspace/engine/falco_engine.h
```diff
@@ -47,9 +48,18 @@ class falco_engine : public falco_common
...
 	//
-	// Enable/Disable any rules matching the provided pattern (regex).
+	// Enable/Disable any rules matching the provided pattern
+	// (regex). If ruleset is non-NULL, enable/disable these
+	// rules in the context of the provided ruleset. The ruleset
+	// can later be passed as an argument to process_event(). This
+	// allows for different sets of rules being active at once.
 	//
-	void enable_rule(std::string &pattern, bool enabled);
+	void enable_rule(std::string &pattern, bool enabled, std::string *ruleset = NULL);
```
mstemm: Yep.

userspace/engine/falco_engine.h
mstemm: I thought about that, but the main reason I went with a string pointer is that it allows the current falco integration in the agent, which doesn't care about rulesets, to remain unchanged. Maybe I should make new methods that provide a ruleset string and have the current methods with no argument call the new methods.

userspace/engine/falco_engine.cpp
mstemm: will do.

userspace/engine/falco_engine.h
bertocci: Could be a wrapper in falco, but agent changes are also fine. I don't think there are many calls. Either way works. Now's the time to change the API if it makes implementation much easier.

mstemm:
The sysdig changes were merged and the tests are passing now. I've also addressed all feedback. Wanna take a look again?

PR #210, opened 2017-02-22
Enable compilation on OS X

ret2libc:
* enable HAS_CAPTURE and driver compilation only on Linux
* set linker flags to correctly compile luajit for os x (as done in sysdig)


mstemm:
The changes look great, thanks!

PR #217, opened 2017-03-06
Create new list 'rabbitmq_binaries' and add it to 'shell in container' rule

dkerwin:
Example event when running rabbitmq in k8s pod:

```
Shell spawned in a container other than entrypoint (user=<NA>
k8s_rabbitmq.123345678_rabbitmq-0_data-2e88ef72-0283-11e7-81fd-00242129f019
(id=00242129f019) shell=sh parent=erl_child_setup cmdline=sh -c exec
/bin/sh -s unix:cmd)
```

Bonus: Removed pg_ctl as it is not defined anywhere in the ruleset.

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -157,6 +157,9 @@
 - list: monitoring_binaries
   items: [icinga2, nrpe, npcd, check_sar_perf.]
 
+- list: rabbitmq_binaries
+  items: [erl_child_setup]
+
 - macro: system_procs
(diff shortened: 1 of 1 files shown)


mstemm:
We want to keep pg_ctl, I believe. This is not a list name but simply the program [pg_ctl](https://www.postgresql.org/docs/9.5/static/app-pg-ctl.html).

Also, once I merged your other PR it caused a conflict with this one.

Can you add pg_ctl back in, resolve the conflicts, and re-open another PR?

If you're busy I can also make identical commits myself and make a PR, but I'd rather you have credit for it if possible.

PR #220, opened 2017-03-06
Add support for gitlab omnibus containers/pod

dkerwin:
Add support for gitlab omnibus containers/pod (https://docs.gitlab.com/omnibus/README.html). 

Example events:
```
Shell spawned in a container other than entrypoint (user=<NA> k8s_gitlab.XXX (id=YYYY) shell=sh parent=sshd cmdline=sh -c /opt/gitlab/embedded/service/gitlab-shell/bin/gitlab-shell key-42)

Shell spawned in a container other than entrypoint (user=<NA> k8s_gitlab.XXX (id=YYY) shell=sh parent=git cmdline=sh -c git-upload-pack '/var/opt/gitlab/git-data/repositories/example/demo_service.git' git-upload-pack '/var/opt/gitlab/git-data/repositories/example/demo_service.git')
```

Might make sense to create a bunch of dedicated gitlab rules as my approach is pretty generic.

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -113,6 +113,9 @@
 - list: db_server_binaries
   items: [mysqld]
 
+- list: gitlab_binaries
+  items: [gitlab-shell, git]
+
 - macro: server_procs
   condition: proc.name in (http_server_binaries, db_server_binaries, docker_binaries, sshd)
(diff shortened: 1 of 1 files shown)


mstemm:
Thanks for the improvements!

Issue #188, opened 2017-01-09
Add container-based tests to travis


mstemm:
Now that we have the ability to build a container from a local package, we should be able to test that the built container runs successfully.

mstemm:
Fixed by #224.

PR #224, opened 2017-03-21
Own driver

mstemm:
Start packaging (and building when necessary) a falco-specific kernel
module in falco releases. Previously, falco would depend on sysdig and
use its kernel module instead.

Also add some travis tests that test packaged builds.

See commits for details on changes.


mstemm:
Once merged, this fixes #188.

PR #256, opened 2017-06-27
WIP: Modify jenkins slave macros

dkerwin:
Remove /tmp from slave.jar commandline as this will not always match (eg jenkinsci/jnlp-slave). Add jenkinsci/jnlp-slave to list of shell spawning containers.


mstemm:
You said WIP. Did you think this was ready to merge as-is, or did you want to make more changes first?

PR #278, opened 2017-09-25
Rework config file handling

mstemm:
These changes allow for a local rules file that will be preserved across
upgrades and allows the main rules file to be overwritten across upgrades.

- Move all config/rules files below /etc/falco/
- Add a "local rules" file /etc/falco/falco_rules.local.yaml. The intent
  is that it contains modifications/deltas to the main rules file
  /etc/falco/falco_rules.yaml. falco_rules.yaml can be overwritten on
  upgrade, while falco.yaml/falco_rules.local.yaml are
  flagged as config files and not overwritten on upgrade.
- Change the handling of the config item "rules_file" in falco.yaml to
  allow a list of files. By default, this list contains:
  [/etc/falco/falco_rules.yaml, /etc/falco/falco_rules.local.yaml].

This fixes https://github.com/draios/falco/issues/268.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,7 @@ if(NOT DEFINED FALCO_VERSION)
 endif()
 
 if(NOT DEFINED FALCO_ETC_DIR)
-	set(FALCO_ETC_DIR "/etc")
+	set(FALCO_ETC_DIR "/etc/falco")
 endif()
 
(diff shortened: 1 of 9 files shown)


mstemm:
I still need to fix redhat rpm packaging--several non-config files are being flagged with the `%config` attribute in the rpm spec, because they are being installed using absolute paths. In order to do that, I want to make some changes on the sysdig agent side first.

PR #295, opened 2017-11-03
docker socket and /dev should not be read-only in daemonset

agilgur5:
- these aren't read-only in any other examples


mstemm:
Good point! We made some changes since this PR was proposed to split the falco daemonset into rbac and non-rbac methods, so I'll make the change in a separate PR #325.

agilgur5:
@mstemm it is courtesy to ensure contributors are credited with their changes. I'm personally quite copious in projects I maintain to ensure contributors receive their due credit, even if I have made changes to their changes (I would be committer and they author), and particularly so when there are conflicts that occurred months after their PR. Just my 2 cents...

mstemm:
Sorry, you're right--I should have waited for an updated PR that had the same changes applied to the updated files, which got moved to new subdirectories. I only went ahead with making the same changes on the updated files myself, as it was a fairly small change. I'll make sure I stay with the original PRs in the future.

agilgur5:
Ok. In the past, I've also just taken commits and amended or rebased them myself, so closing the PR, but leaving the contributor as the author of the commit that gets put into master (examples are probably better than words: https://github.com/agilgur5/react-signature-canvas/pull/14#issuecomment-354422891). It's something I've seen often in projects that follow a rebase style (e.g. Django)

PR #296, opened 2017-11-03
use https://kubernetes.default instead of just https://kubernetes

agilgur5:
- otherwise it doesn't work if it's in a different namespace


mstemm:
Since this PR was proposed, we made some other changes to document rbac and non-rbac methods for deploying falco, so this PR has conflicts. I carried the change over in a new PR https://github.com/draios/falco/pull/325, so I'll close this one.

agilgur5:
Same comment as https://github.com/draios/falco/pull/295#issuecomment-367154258

mstemm:
Indeed, I'll make sure I stay with the original PRs in the future.

PR #307, opened 2017-12-12
Rule updates vdec

mstemm:
Several additional changes from secure customers.


mstemm:
I think the test failures are actually due to https://github.com/draios/falco/issues/308, which if true is a longstanding bug. The changes in https://github.com/draios/sysdig/pull/993 exposed the test failure. Although from looking at the test I think it was also incorrect.

Merging these rule changes anyway.

PR #321, opened 2018-02-09
Rule updates 2018 02.v1

mstemm:
Changes to reduce false positives as well as handling suggested use cases from customers.

@mattpag could you take a look at the last 2 commits (ssh/nodeport support?)


mstemm:
yeah, that was a typo on my part. I had changed it for testing, but hadn't changed it back. It's fixed now.

PR #335, opened 2018-03-12
Use distinct names for file and program output pointers.

jmcarp:
And use consistent indentation.

I'm pretty sure this is behind the issue I mentioned earlier today in Slack: I got strange behavior (opening multiple copies of the same program) with `keep_alive` program output, but only when I was also using file output.

cc @wjwoodson @sharms


mstemm:
Oh, fantastic! I'll merge.

PR #373, opened 2018-05-30
Changed container macro to is_container

xaocon:
The name change should make it easier to find and read.

falco-CLA-1.0-signed-off-by: Evan Pitstick <nerdx00@gmail.com>

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -293,7 +293,7 @@
 # container=%container.info user=%user.name). The output will change
 # based on the context and whether or not -pk/-pm/-pc was specified on
 # the command line.
-- macro: container
+- macro: is_container
   condition: container.id != host
 
 - macro: interactive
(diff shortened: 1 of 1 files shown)


xaocon:
Not sure why it says it's missing the sign-off. It's there.

xaocon:
I should have rebased instead of merging but everything seems to be passing now. I've got the sign-off in there so my guess is that the checker is looking for sysdig sign-off instead of falco like in it's details.

mstemm:
Thanks--I want to figure out how widespread the use of the container macro is in user rules files, as this change would break those files, before merging. I'll try to check our sysdig secure customers first.

mstemm:
It looks like the macro `container` is already pretty widely used in custom/user rules files, so changing the name of the macro from `container` to `is_container` would break those  user rules files. I guess you could *also* add an is_container macro that just pointed to macro, but that seems a bit redundant.

What do you think?

mstemm:
We'll go ahead and close this PR for now. Feel free to make other changes if you'd like, though.

PR #379, opened 2018-06-07
Enable all rules

mstemm:
Make sure rules are proactively enabled rather than only disabled. This unbreaks falco for the dev branch.


mstemm:
Need to merge https://github.com/draios/sysdig/pull/1145 also.

PR #389, opened 2018-07-10
Kubernetes response engine

nestorsalceda:
A response engine for Falco that allows to process security events executing playbooks to respond to security threats.

I also ping @mstemm for being aware of the merge.

Thanks!


kubernetes-response-engine/README.md
```diff
@@ -0,0 +1,18 @@
+# Kubernetes Response Engine for Sysdig Falco
```
mstemm: Can we move this to a lower directory, something like `integrations/alert_handlers`? @bencer mentioned that he had other additions like a helm chart, so we're looking for a place for that as well.

kubernetes-response-engine/playbooks/README.md
```diff
@@ -0,0 +1,164 @@
+# Playbooks
```
mstemm: I was initially confused, thinking that this was related to ansible. You might want to ask around to see if others had the same confusion.

kubernetes-response-engine/deployment/falco-config/falco.yaml
```diff
@@ -0,0 +1,102 @@
+# File(s) or Directories containing Falco rules, loaded at startup.
```
mstemm: Instead of copying the entire falco config, can you just include a patch or highlight the changes that need to be made from the main falco config to add the additional file output? I'm worried that with a second copy it will become out-of-date with respect to the main config file as we make changes to the main config file.

kubernetes-response-engine/deployment/falco-config/falco_rules.local.yaml
```diff
@@ -0,0 +1,38 @@
+####################
+# Your custom rules!
+####################
```
mstemm: Is this rules file really related to the integration, or is it for some specific example of suspicious behavior? I wouldn't mix up the two, otherwise people will think the integration only is related to cryptomining.

kubernetes-response-engine/deployment/falco/falco-daemonset.yaml
```diff
@@ -0,0 +1,84 @@
+apiVersion: extensions/v1beta1
```
mstemm: Same here--does this differ from the existing falco daemonset configs already at examples/k8s-using-daemonset ?

mstemm:
Overall my biggest piece of feedback is that it copies lots of other configs/etc (probably so you could have a completely self-contained example). But I think the big contribution here is the python code that expresses actions as modules. I think the PR should only have that, and then have instructions how to combine it with a falco config, nats integration, etc to get a working example.

kubernetes-response-engine/playbooks/README.md
nestorsalceda: Yes, you are not the unique who pointed us to this misunderstanding. 

A few days ago, @KnoxAnderson suggested to use this name because is well known among security teams.

https://docs.microsoft.com/en-us/azure/security-center/security-center-playbooks
https://blogs.cisco.com/security/using-a-playbook-model-to-organize-your-information-security-monitoring-strategy

So we chose the term "Playbook", perhaps a possible alternative would be "Security Playbooks"?

kubernetes-response-engine/deployment/falco-config/falco_rules.local.yaml
nestorsalceda: Good catch, is an specific example of suspicious behavior.

Thanks!

kubernetes-response-engine/README.md
nestorsalceda: Yeah, I love the integrations subdirectory. 

I'm going to place under integrations/kubernetes_response_engine.

Thanks!

kubernetes-response-engine/deployment/falco-config/falco.yaml
nestorsalceda: As long as the new Falco Helm Chart is merged (I hope in a while), we can remove the falco and falco_config directory from deployment. So this change will dissapear.

Thanks!

kubernetes-response-engine/deployment/falco/falco-daemonset.yaml
nestorsalceda: Same than above, when the new Falco Helm Chart is merged this change will dissapear.

Thanks!

nestorsalceda:
Hey @mstemm I think I have addressed all your concerns. A quick summary is:

* I have moved top level directory to integrations/kubernetes-response-engine
* I have removed all Falco configuration files because they are going to live in the Helm Chart

What do you think? It looks good to you?

Thanks!

integrations/kubernetes-response-engine/playbooks/specs/infrastructure/kubernetes_client_spec.py
```diff
@@ -0,0 +1,65 @@
+from mamba import description, context, it, before
```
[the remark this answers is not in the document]
nestorsalceda: I love to write tests :metal:

nestorsalceda:
Cool! Do you need anything else for merging the PR?

Thanks!

PR #392, opened 2018-07-13
[SMAGENT-968] Ensure the /lib/modules symlink to /host/lib/modules is set correctly

bertocci:
If /lib/modules exists in the base image, the symlink will get created at
/lib/modules/modules. This removes any existing empty directory but will
fail if we try to remove a non-empty /lib/modules. (Punting on how to
handle non-empty base image dirs for now)

I'm propagating the changes from draios/agent-rhel#3. They aren't needed immediately for the debian base image, but it doesn't hurt, protects against future changes, and keeps the dockerfiles uniform.

### DIFF
--- a/docker/dev/Dockerfile
+++ b/docker/dev/Dockerfile
@@ -48,7 +48,11 @@ RUN curl -s https://s3.amazonaws.com/download.draios.com/DRAIOS-GPG-KEY.public |
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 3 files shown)


codytubbs:
```
RUN test ! -e /lib/modules/* || ( mkdir -p /lib/native_modules \
 && mv /lib/modules/* /lib/native_modules/. ) \
 && ln -sF $SYSDIG_HOST_ROOT/lib/modules /lib/.
```

That should do it.

PR #405, opened 2018-08-10
Add support for container metaevent to detect container spawning

mattpag:
A new macro `container_started` contains both the new and the old check.
Also, returns true only for execve enter events with `vpid=1`, not for both directions.

This depends on https://github.com/draios/sysdig/pull/1190.

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -299,6 +299,9 @@
 - macro: container
   condition: container.id != host
 
+- macro: container_started
+  condition: (evt.type = container or (evt.type=execve and evt.dir=< and proc.vpid=1))
+
 - macro: interactive
   condition: >
     ((proc.aname=sshd and proc.name != sshd) or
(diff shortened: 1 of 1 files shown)


rules/falco_rules.yaml
```diff
@@ -299,6 +299,9 @@
 - macro: container
   condition: container.id != host
 
+- macro: container_started
+  condition: (evt.type = container or (evt.type=execve and evt.dir=< and proc.vpid=1))
```
Kaizhe: Just want to drop a note here, if the container is brought up with flag `--pid=host` or `--pid=<container id or name>`, the vpid wont be equal to 1. We might miss the containers brought up with pid namespace configured. For more reference: https://docs.docker.com/engine/reference/run/#imagedigest

mfdii:
@mattpag @mstemm What's the status on this? Can we merge?

mstemm:
We need to wait, the related sysdig code wasn't ever merged and now there's a new approach for emitting container events: https://github.com/draios/sysdig/pull/1283. That work needs to complete and we also need to change sysdig to emit "container" events when containers are created.

rules/falco_rules.yaml
mstemm: Yeah container.id = host allows you to know that you're in a container. proc.vpid=1 was the best way we had to denote that a container was started. Now we'll have a more reliable way using the container event.

mstemm:
Closing in favor of https://github.com/falcosecurity/falco/pull/550. I cherry-picked the commit from this PR to #550.

PR #407, opened 2018-08-25
Falco probe builder with Docker

nestorsalceda:
This week I have been working in the feedback I received in https://github.com/draios/sysdig/pull/1199

We have a problem with our builder, because is hard to install a recent version of Clang + LLVM, so I have extracted the builder to a Docker container which extends from Falco image which already ships valid Clang and LLVM versions

I also have splitted the builder in several files, which allows us to debug better the builders, it also supports parallel builds and adding new builders is done with just one file. 

I also moved to Falco repository because this solution is more focused on Falco and this allows different development/release pace among Falco and other Sysdig components.


gianlucaborello:
I took a very quick look at this (so I haven't run it, just looked) and I like this approach, I think it has good potential. Now, since I like it I think we should use it for everything (Falco, Sysdig, Agent + kernel module, eBPF), not just this specific intersection of features you're looking for now (Falco + eBPF) :-)

In particular, a few comments that I think you might want to consider:

- Code duplication: it's obvious that this code contains a lot of duplication with the builder in the sysdig repo. I think this is a recipe for problems in the future, those scripts change quite frequently as new quirks are discovered by the agent team, and keeping them in sync is going to be error prone. So, I think ultimately all this logic should be applied to the sysdig repo, where it belongs, close to the actual probe sources.

- Builder scripts: No major comments, the split makes sense.

- Docker builder container: I understand the approach you've taken and how it favors the delivery of this feature, but to be honest it all feels too tightly coupled to me. Rather than having a single build container that is based on Falco and our current runtime distribution of choice (Debian sid), I would make separate generic container builders for the flavors we need (e.g. Centos 6 + Debian?), standardize the set of inputs they need to receive, and then make sure each builder script calls the specific relevant container, rather than the falco-probe-builder. This seems more scalable to me, especially for the broader goal if being used by every other product.

- More parameterization: As I said in the last point I really think this should support building probes for agent/sysdig/falco, and not just falco. In order to do this, if we have generic docker builder containers, you would need to add a few other parameters, such as the probe name, and probe code location (not limited to the git repo like we're doing, you could for example copy the probe sources from the binary tarballs we distribute in S3 along with the deb/rpm packages), etc. Overall, I think the approach would be very clean.

What do you think? I understand this might fall into pure engineering work and you might not want to spend your time on this, but this is what I think.

Also, assume I would just say "This looks good to me, let's merge it". What would you expect after this? Would you expect that we keep serving the agent and sysdig probes from our legacy script, and we would switch to this just for falco, and then "merge" the output directories of the two approaches in our S3 bucket? 

Another point in favor of doing the broader work is that, as we discussed a few weeks ago (and please correct me if I'm wrong), we are not in a terrible rush to release this, right? For eBPF, we have proper workarounds for both COS and Minikube involving runtime download of headers, whereas for the kernel module we are currently covered by our legacy infrastructure that precompiles the probes, so there is no impending doom. Am I correct?

Thanks

PR #410, opened 2018-09-04
Add base64 encoding and snap length support

zt2:
sysdig-CLA-1.0-signed-off-by: Yue Feng <ztz5651483@gmail.com>
falco-CLA-1.0-signed-off-by: Yue Feng <ztz5651483@gmail.com>

### DIFF
--- a/userspace/engine/formats.cpp
+++ b/userspace/engine/formats.cpp
@@ -115,7 +115,27 @@ int falco_formats::format_event (lua_State *ls)
 
 		if(s_json_output)
 		{
-			s_inspector->set_buffer_format(sinsp_evt::PF_JSON);
+			switch(s_inspector->get_buffer_format())
+			{
+				case sinsp_evt::PF_NORMAL:
+					s_inspector->set_buffer_format(sinsp_evt::PF_JSON);
+					break;
+				case sinsp_evt::PF_EOLS:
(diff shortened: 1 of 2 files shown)


zt2:
Have a look plz...
@mstemm

mstemm:
The changes look good, and thanks for the contribution! I wanted to double check on the use case, though. Normally, most/all of the events that have an %evt.buffer that would be a candidate for display as base64 are also events that falco ignores. So I guess you have your own rules where you do use those events, and then also run falco with the `-A` flag?

zt2:
> The changes look good, and thanks for the contribution! I wanted to double check on the use case, though. Normally, most/all of the events that have an %evt.buffer that would be a candidate for display as base64 are also events that falco ignores. So I guess you have your own rules where you do use those events, and then also run falco with the `-A` flag?

Yes, I use falco with `-A` and some custom rules to capture special system events. I'd like to use falco instead sysdig that's because rules in falco can be managed easily.  But some nice features in sysdig are missed in falco, like `%evt.buffer` (it can not display non ascii characters) and `snaplen` (can not custom `%evt.buffer` length).

There are other users has the same feature request please see https://github.com/draios/falco/issues/363

mstemm:
Ok, I just wanted to make sure you knew about the required configuration. Merging, and thanks again!

PR #411, opened 2018-09-05
Add an integration with Phantom

nestorsalceda:
This PR adds a playbook to our response engine which creates an incident in Phantom when Falco raises an alert.


nestorsalceda:
Hey @mstemm thanks for the review! I have rebased the branch on top of dev :)

PR #414, opened 2018-09-11
Add a playbook which starts to capturing data using Sysdig and uploads capture to a s3 bucket

nestorsalceda:
This PR adds a playbook to our response engine which starts to capture data from a pod using sysdig when Falco raises an alert.

### DIFF
--- a/integrations/kubernetes-response-engine/playbooks/playbooks/infrastructure.py
+++ b/integrations/kubernetes-response-engine/playbooks/playbooks/infrastructure.py
@@ -69,7 +69,7 @@ class KubernetesClient:
     def start_sysdig_capture_for(self, pod_name, event_time,
                                  duration_in_seconds, s3_bucket,
                                  aws_access_key_id, aws_secret_access_key):
-        job_name = 'sysdig-capturer-{}-{}'.format(pod_name, event_time)
+        job_name = 'sysdig-{}-{}'.format(pod_name, event_time)
 
         node_name = self.find_node_running_pod(pod_name)
         namespace = self._find_pod_namespace(pod_name)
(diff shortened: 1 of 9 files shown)


integrations/kubernetes-response-engine/playbooks/playbooks/infrastructure.py
```diff
@@ -64,6 +65,152 @@ def add_label_to_pod(self, name, label, value):
...
+                                                   s3_bucket,
+                                                   aws_access_key_id,
+                                                   aws_secret_access_key)
+
+        return self._batch_v1.create_namespaced_job(namespace, body)
+
+    def _build_sysdig_capture_job_body(self, job_name, node_name,
+                                       duration_in_seconds, s3_bucket,
+                                       aws_access_key_id, aws_secret_access_key):
+        return client.V1Job(
```
mstemm: I understand why it's advantageous to run the sysdig program itself as a pod within k8s, but I would think that the delay it would take to create the pod, possibly pull the image, etc to be too long. Did you consider just invoking sysdig and assuming it's available on the host?

nestorsalceda: Yes, that was my first thought but I also would need to install the AWS CLI tools in the host for the S3 uploading. Also, perhaps in a future I would need the GCloud tools also for uploading to other storage, so I prefer to package all of this as a Docker image.

The drawback is the delay to create the pod, pull ... But I think I can minimize that delay pulling the image when deploying the engine, like some kind of warm-up.

Thanks for the feedback, I truly appreciate it! :)

nestorsalceda:
I have rebased it @mstemm, thanks!

PR #429, opened 2018-09-28
fix deprecated statement.

tao12345666333:
xref: https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

Signed-off-by: Jintao Zhang <zhangjintao9020@gmail.com>

### DIFF
--- a/docker/dev/Dockerfile
+++ b/docker/dev/Dockerfile
@@ -1,6 +1,6 @@
 FROM debian:unstable
 
-MAINTAINER Sysdig <support@sysdig.com>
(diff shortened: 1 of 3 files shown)


mstemm:
Thanks--it's a minor thing, but for falco we want the CLA sign-off message to say `falco-CLA-1.0-signed-off-by:` instead of `sysdig-CLA-1.0-signed-off-by:`. Could you update your commit message?

(We'll be changing the process for committers soon, but for now we're continuing to use the old process).

tao12345666333:
@mstemm  Thanks， it's my mistake. I have updated the commit message.

PR #445, opened 2018-10-20
adding few executables in corresponding groups

juju4:
- userexec: critical-stack (https://criticalstack.com/)

- hids: aide/debian wrappers, logcheck, osquery, ossec
* Warning Sensitive file opened for reading by non-trusted program (user=root program=osqueryd command=osqueryd

- postconf (ubuntu/debian)
* Error File below /etc opened for writing (user=root command=postconf -e inet_protocols=all parent=postfix.postins pcmdline=postfix.postins -e /var/lib/dpkg/info/postfix.postinst configure  file=/etc/postfix/main.cf.tmp program=postconf gparent=frontend ggparent=dpkg gggparent=apt-get)

- ufw, cloud-init
* Error File below /etc opened for writing (user=root command=ufw /usr/sbin/ufw disable parent=bash pcmdline=bash  file=/etc/ufw/ufw.conf program=ufw gparent=sshd ggparent=sshd gggparent=systemd)
* Error File below a monitored directory opened for writing (user=root command=cloud-init /usr/bin/cloud-init init file=/root/.ssh/authorized_keys parent=init pcmdline=init  gparent=lxd)

Thanks


mstemm:
Thanks for the updates! A couple of inline comments.

rules/falco_rules.yaml
```diff
@@ -852,7 +852,7 @@
                           gen_resolvconf., update-ca-certi, certbot, runsv,
                           qualys-cloud-ag, locales.postins, nomachine_binaries,
                           adclient, certutil, crlutil, pam-auth-update, parallels_insta,
-                          openshift-launc, update-rc.d)
+                          openshift-launc, update-rc.d, ufw, cloud-init)
```
mstemm: For ufw, could we limit the scope to both the ufw program as well as the directory to which it writes? With something like this:

```
- macro: ufw_writing_config
  condition: proc.name=ufw and fd.directory=/etc/ufw
```
And then add ufw_writing_config to the (long) list of exceptions in write_etc_common.

For cloud-init, it looks like your example involves writing below /root/.ssh, right? In which case you want to add it as an exception to the `Write below monitored dir` rule instead. Again, ideally the exception would tie together the program as well as the file/directory being written.

I have some other rule updates pending, so if you'd like I can merge this PR and then make those changes in my rule updates branch once I rebase.

juju4: no problem. good to me. Thanks @mstemm!

PR #465, opened 2018-11-12
Fix AWS permissions for Kubernetes Response Engine

nestorsalceda:
Honor the principle of least privilege when giving permissions for Kubernetes Response Engine and RBAC and use a system account created specifically for AWS and Kubernetes Response Engine.


mfdii:
The`cluster-role.yaml` is in the `deployment` and the `cluster-role-binding.yaml` is in `deployment/aws`. There doesn't seem to be anything that is AWS specific and in theory we should have the CNCF response engine stack use a similar user/roles if possible.

Regardless, I feel like the cluster-role and cluster-role-binding files should live in the same directory.

PR #468, opened 2018-11-15
Avoid going past end of ruleset/etag arrays

mstemm:
It's possible to call event_tags_for_ruleset/evttypes_for_ruleset for a
ruleset that hasn't been loaded. In this case, it's possible to go past
the end of the m_rulesets array.

After fixing that, it's also possible to go past the end of the
event_tags array in event_tags_for_ruleset().

So in both cases, check the index against the array size before
indexing.


mstemm:
This fixes https://github.com/falcosecurity/falco/issues/469.

PR #471, opened 2018-11-20
Allow SSL for k8s audit endpoint

mfdii:
Allow enabling SSL for the Kubernetes audit log web server. This required adding two new configuration options: `webserver.ssl_enabled` and `webserver.ssl_certificate`. To enable SSL add the below to the `webserver` section of the `falco.yaml` config:
```
webserver:
  enabled: true
  listen_port: 8765s
  k8s_audit_endpoint: /k8s_audit
  ssl_enabled: true
  ssl_certificate: /etc/falco/falco.pem
```
Note that the port number has an `s` appended to indicate SSL for the port which is how civetweb expects SSL ports be denoted. We could change this to dynamically add the `s` if `ssl_enabled: true`.

The `ssl_certificate` is a combination SSL Certificate and corresponding key contained in a single file. You can generate a key/cert as follows:
```
$ openssl req -newkey rsa:2048 -nodes -keyout key.pem -x509 -days 365 -out certificate.pem
$ cat certificate.pem key.pem > falco.pem
$ sudo cp falco.pem /etc/falco/falco.pem
```

### DIFF
--- a/userspace/falco/configuration.cpp
+++ b/userspace/falco/configuration.cpp
@@ -34,6 +34,7 @@ falco_configuration::falco_configuration()
 	  m_webserver_enabled(false),
 	  m_webserver_listen_port(8765),
 	  m_webserver_k8s_audit_endpoint("/k8s_audit"),
+	  m_webserver_ssl_enabled(false),
 	  m_config(NULL)
 {
 }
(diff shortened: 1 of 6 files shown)


userspace/falco/configuration.cpp
```diff
@@ -32,8 +32,9 @@ using namespace std;
 falco_configuration::falco_configuration()
 	: m_buffered_outputs(true),
 	  m_webserver_enabled(false),
-	  m_webserver_listen_port(8765),
+	  m_webserver_listen_port("8765"),
 	  m_webserver_k8s_audit_endpoint("/k8s_audit"),
+	  m_webserver_ssl_enabled(false),
```
mstemm: As you mentioned in the comments, I think it's a bit confusing to have to change two things to enable ssl (setting webserver.ssl_enabled *and* adding a 's' to the port definition). How about we do as you suggested and add 's' based on the value of webserver.ssl_enabled?

mstemm:
All the other changes look good. Let's just clean up the port handling for ssl.

userspace/falco/configuration.cpp
mfdii: If we assume `webserver.ssl_enabled: true` means any ports specified will be SSL, that means you won't be able to listen on a non-SSL port and a SSL port at the same time. Generally I can't think why you'd want the falco web service to listen on multiple ports for this use case, and best practice should be that the kubernetes audit events are sent encrypted in transit as they could have sensitive information. If you're ok with this limitation I'll make the change.

mfdii: change pushed

PR #475, opened 2018-11-26
Add /etc mount for falco container

### LINKED ISSUE #473: Fix Kubernetes daemonset example to include /host/etc
In order for the eBPF probe loader to work properly on COS and minikube, /etc from the host needs to be mounted inside the Falco container.

The below needs added to the [daemonset](https://github.com/falcosecurity/falco/blob/dev/integrations/k8s-using-daemonset/k8s-with-rbac/falco-daemonset-configmap.yaml#L37):

```
            - mountPath: /host/usr
              name: usr-fs
              readOnly: true
            - mountPath: /host/etc
              name: etc-fs
              readOnly: true
            - mountPath: /etc/falco
              name: falco-config
...
        - name: usr-fs
          hostPath:
            path: /usr
        - name: etc-fs
          hostPath:
            path: /etc
        - name: falco-config
          configMap:
            name: falco-config
```

Opening this to remind myself to fix this.

mfdii:
Fixes #473.

This is required to support COS and Minikube as the probe loader script looks for the OS version by reading files in /etc.

I also added the environment variables to enable the eBPF probe, but left them commented out.

### DIFF
--- a/integrations/k8s-using-daemonset/k8s-with-rbac/falco-daemonset-configmap.yaml
+++ b/integrations/k8s-using-daemonset/k8s-with-rbac/falco-daemonset-configmap.yaml
@@ -18,6 +18,13 @@ spec:
           image: falcosecurity/falco:latest
           securityContext:
             privileged: true
+# Uncomment the 3 lines below to enable eBPF support for Falco.
+# This allows Falco to run on Google COS.
+# Leave blank for the default probe location, or set to the path
+# of a precompiled probe.
+#          env:
+#          - name: SYSDIG_BPF_PROBE
+#            value: ""
           args: [ "/usr/bin/falco", "-K", "/var/run/secrets/kubernetes.io/serviceaccount/token", "-k", "https://kubernetes.default", "-pk"]
           volumeMounts:
             - mountPath: /host/var/run/docker.sock
@@ -35,6 +42,9 @@ spec:
               readOnly: true
             - mountPath: /host/usr
               name: usr-fs
+              readOnly: true
+            - mountPath: /host/etc/
+              name: etc-fs
               readOnly: true
             - mountPath: /etc/falco
               name: falco-config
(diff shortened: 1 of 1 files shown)


mstemm:
Do you think it would be more secure to explicitly mount `/etc/os-release` and `/etc/VERSION` instead of all of `/etc/` ? You did create the mount read-only, which is good, but someone who got into the falco daemonset pod could read stuff like /etc/passwd (and maybe even shadow, depending on how it's running) from the host.

mfdii:
Ok, did as you suggested and I am only mounting the files Falco needs. 

There's a bug in the probe loader script in the logic used to detect COS vs Minikube for eBPF support. I need to submit a PR to fix that before we should merge this.

mfdii:
Required probe loader script fix in [draios/sysdig/1273](https://github.com/draios/sysdig/pull/1273).

gianlucaborello:
(Commenting on this since I saw it referenced from https://github.com/draios/sysdig/pull/1273)

Are we sure we want to introduce that much granularity in the mount paths, which need to be propagated to sysdig, agent as well as traditional command lines?

What you seem to be wanting to prevent is already possible in a very easy way: just go into a Falco container and run `cat /host/proc/1/root/etc/shadow`, you'll easily get access to the host file system even if you didn't mount `/etc`.

And this is without even counting that we are a privileged container, have full access to the Docker socket, and have the ability to compile and inject custom code into the kernel, we are basically already root++ on the host, so mounting `/etc` basically doesn't make a difference. 

From my point of view, the only reason why we are not mounting the entire host file system, thus greatly simplifying the command line invocation, is just because we don't want to keep those mount points busy when containers churn, otherwise I'd consider it reasonable, and more practical, to do that as well (other projects, such as Google's cadvisor, do exactly that and mount the entire host's `/`).

Feel free to ignore this, just wanted to provide my perspective.

PR #477, opened 2018-11-27
SMAGENT-1175: Add -Wextra -Werror to build

adalton:
As part of SMAGENT-1175, we added the option to build the agent as well
as sysdig with `-Wextra` and `-Werror`.  This change keeps oss-falco in
sync with that change, based on a BUILD_WARNINGS_AS_ERRORS option, which
is disabled by default.

Enabling this identified an error where we were throwing exceptions from a
destructor, which C++11 doesn't allow.  I changed the exception to a log+assert.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,8 @@ cmake_minimum_required(VERSION 2.8.2)
 
 project(falco)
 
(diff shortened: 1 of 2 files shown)


mstemm:
Closing in favor of https://github.com/falcosecurity/falco/pull/479.

PR #478, opened 2018-11-27
Add optional support for -Wextra and -Werror

adalton:
The `-Wextra` compile-time option will enable additional diagnostic
warnigns.  The `-Werror` option will cause the compiler to treat warnings
as errors.  This change adds a build time option,
`BUILD_WARNINGS_AS_ERRORS`, to conditionally enable those flags.  Note
that depending on the compiler you're using, if you enable this option,
compilation may fail (some compiler version have additional warnings
that have not yet been resolved).
    
Testing with these options in place identified a destructor that was
throwing an exception.  C++11 doesn't allow destructors to throw
exceptions, so those throw's would have resulted in calls to
`terminate()`.  I replace them with an error log and a call to `assert()`.


mstemm:
Closing in favor of https://github.com/falcosecurity/falco/pull/479.

PR #450, opened 2018-10-29
K8s audit evts

mstemm:
Changes to support [K8s Audit Events](https://kubernetes.io/docs/tasks/debug-application-cluster/audit/#audit-backends) as an event source in addition to system calls.

https://github.com/falcosecurity/falco/wiki/K8s-Audit-Event-Support generally describes the feature.

This depends on https://github.com/draios/sysdig/pull/1254.

This pretty much covers https://github.com/falcosecurity/falco/issues/184.

PR #485, opened 2018-12-17
Fix missing arguments

zt2:
Hi,

I found some arguments was missed in `falco.cpp`, so I re-add it in `userspace/falco/falco.cpp` and `userspace/engine/formats.cpp`


mstemm:
Thanks for noticing this! I accidentally removed it when resolving rebase conflicts in #450.

PR #488, opened 2018-12-26
Remove kubernetes-response-engine from system:masters

nestorsalceda:
As long as we are using rbac for allowing actions on several resources,
we can restrict this a bit more.


mstemm:
I'll defer to your judgement on the change :)

PR #496, opened 2019-01-17
Add falco service k8s

mstemm:
Docs changes that create a falco service when used with k8s, and use falco running within k8s to show off audit log support.

@markyjackson-taulia these instructions will probably be very useful to you. There are some caveats related to kernel modules that I can discuss with you offline.


mstemm:
I'm going to merge now just so I can incorporate the changes in a 0.13.1 release, but we can address any feedback right afterward.

PR #507, opened 2019-01-29
update program_output example

daixiang0:
Fix below issue:
when config with mail:
```
program_output:
  enabled: true
  keep_alive: false
  program: mail -s "Falco Notification" 764524258@qq.com
```
```
16:07:24.426355331: Informational Container with sensitive mount started (user=root command=runc:[1:CHILD] init k8s.ns=<NA> k8s.pod=<NA> container=95d852f177a5 image=sysdig/falco-event-generator@sha256:8be9e858ed798bb41c22601ba80f552244f50e452505a2eb34c3e1c82369b343 mounts=/var/lib/kubelet/pods/641309ac-25f8-11e9-8f73-1e00130014eb/volumes/kubernetes.io~secret/default-token-bp7qk:/var/run/secrets/kubernetes.io/serviceaccount:ro:false:rprivate,/var/lib/kubelet/pods/641309ac-25f8-11e9-8f73-1e00130014eb/etc-hosts:/etc/hosts::true:rprivate,/var/lib/kubelet/pods/641309ac-25f8-11e9-8f73-1e00130014eb/containers/falco-event-generator/76eff74a:/dev/termination-log::true:rprivate) k8s.ns=<NA> k8s.pod=<NA> container=95d852f177a5
sh: 1: mail: not found

```
If not, falco can not send mail.

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -137,4 +137,4 @@ webserver:
 program_output:
   enabled: false
   keep_alive: false
-  program: mail -s "Falco Notification" someone@example.com
+  program: "jq '{text: .output}' | curl -d @- -X POST https://hooks.slack.com/services/XXX"


mfdii:
This PR needs to be against `dev`, not `master`.

daixiang0:
The apt source seems have problem @mfdii

mstemm:
I think it wouldn't be very likely that you would want to run the normal falco container with a program output that could send mail directly. You'd probably need email server configuration that also configured an outbound mail gateway and probably credentials to contact that gateway. You could always create a new image that started with falco using `FROM falco:latest` and added mailutils and copied in the necessary configuration.

Of course, when running directly on the host, say via debian or rpm packages, you could rely on the host's configuration for mail servers, etc.

daixiang0:
For k8s env, i think it is possible to send mail from container, and in production env it should be default action i think.

By the way, in https://github.com/falcosecurity/falco/blob/dev/falco.yaml, send mail is default action. As you said, maybe add a comment here is better. If so, i would add comment about mail part.

mfdii:
Sending mail is not a default action, it’s an example and as mark points out it you would probably never want to do this. We should remove the example in favor of the slack example.

daixiang0:
@mfdii let us change those part in :)

daixiang0:
any update?

daixiang0:
@mfdii hi, any input?

mstemm:
So now the only change in the PR is that the example program output command line refers to jq and curl, right? That's the change you intended?

daixiang0:
yes, since that mail from container directly is not good enough, why show this example, change it is better, right?

mstemm:
I see that the PR was still based against the master branch. I cherry-picked the commit onto a PR targeted towards the dev branch (https://github.com/falcosecurity/falco/pull/541).

daixiang0:
@mstemm thanks very much.
A qustion, which branch should be based against?

PR #509, opened 2019-01-30
kh: improve mount on /var/lib/kubelet rule

Kaizhe:


Kaizhe:
Thank you @JPLachance

PR #512, opened 2019-02-01
Sync timezone

daixiang0:
The default timezone is UTC:
```
root@falco-daemonset-k852s:/# ls -l /etc/localtime
lrwxrwxrwx 1 root root 27 Dec 26 00:00 /etc/localtime -> /usr/share/zoneinfo/Etc/UTC
```
If not sync timezone, the output time is always wrong.


mfdii:
We use `dev` as the default branch, not `master`. In the dev branch we've [started mounting `/etc/`](https://github.com/falcosecurity/falco/blob/dev/integrations/k8s-using-daemonset/k8s-with-rbac/falco-daemonset-configmap.yaml#L46) of the host in order to access files to detect the operating system.  

Please use the daemonset config in `dev` as it's the latest.

PR #514, opened 2019-02-01
Add additional k8s rbac resources

mstemm:
Falco also needs to list/watch replicasets, daemonsets, and deployments,
so add them to the resources list.

### DIFF
--- a/integrations/k8s-using-daemonset/k8s-with-rbac/falco-account.yaml
+++ b/integrations/k8s-using-daemonset/k8s-with-rbac/falco-account.yaml
@@ -15,7 +15,7 @@ metadata:
     role: security
 rules:
   - apiGroups: ["extensions",""]
-    resources: ["nodes","namespaces","pods","replicationcontrollers","services","events","configmaps"]
+    resources: ["nodes","namespaces","pods","replicationcontrollers","replicasets","services","daemonsets","deployments","events","configmaps"]
     verbs: ["get","list","watch"]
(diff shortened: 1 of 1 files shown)


integrations/k8s-using-daemonset/k8s-with-rbac/falco-account.yaml
```diff
@@ -15,7 +15,7 @@ metadata:
     role: security
 rules:
   - apiGroups: ["extensions",""]
-    resources: ["nodes","namespaces","pods","replicationcontrollers","services","events","configmaps"]
+    resources: ["nodes","namespaces","pods","replicationcontrollers","replicasets","services","daemonsets","deployments","events","configmaps"]
```
JPLachance: `replicasets`, `daemonsets` and `deployments` are now in the `apps` API group. I think we need to extend the `apiGroups` with `apps` in order for this to work 🙂

mstemm:
@JPLachance pointed out that many resources e.g. deployments/daemonsets/replicasets are also available via the `apps/v1` endpoint. The k8s metadata code in oss sysdig (which is used by oss falco) still uses the `extensions/v1beta1` endpoints. Eventually those endpoints will go away, so the sysdig code will need to switch to the new endpoints.

mstemm:
Filed https://github.com/draios/sysdig/issues/1308 on sysdig to track updating it to use the new endpoints.

mstemm:
Checked with how we do our internal RBAC and these choices are correct given the paths we use, so merging.

PR #515, opened 2019-02-02
Sync timezone

daixiang0:
If not, the time of warning is meaningless.

### DIFF
--- a/integrations/k8s-using-daemonset/k8s-with-rbac/falco-daemonset-configmap.yaml
+++ b/integrations/k8s-using-daemonset/k8s-with-rbac/falco-daemonset-configmap.yaml
@@ -46,6 +46,9 @@ spec:
             - mountPath: /host/etc/
               name: etc-fs
               readOnly: true
+            - mountPath: /etc/localtime
+              name: etc-localtime
+              readOnly: true
             - mountPath: /etc/falco
               name: falco-config
       volumes:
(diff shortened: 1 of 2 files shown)


mstemm:
This PR brings up a good point, notably that falco's log messages don't have any time zone information:

```
Wed Feb  6 14:01:39 2019: Falco initialized with configuration file /mnt/sf_mstemm/work/src/falco/falco.yaml
Wed Feb  6 14:01:39 2019: Loading rules from file /etc/falco/falco_rules.yaml:
Wed Feb  6 14:01:39 2019: Loading rules from file /etc/falco/falco_rules.local.yaml:
Wed Feb  6 14:01:39 2019: Loading rules from file /etc/falco/k8s_audit_rules.yaml:
```

However, I think a better way to make the time zone part of the log messages unambiguous is to change their format to ISO 8601, which includes time zone information. When run locally, the time will be whatever time zone used by `/etc/localtime`. When run in containers, the date/time will always be UTC as that's what /etc/localtime is in our debian:unstable container.

mstemm:
Actually it's more complicated than I thought, as there isn't any %evt.time.xxx field in sysdig that includes time zone information. So let me ask around to ask if there was a reason for that (like performance). I may end up adding an `%evt.time.iso8601` option to sysdig and start using it in falco as well.

daixiang0:
That could be better, but for now, mount localtime can be a workaroud.
I think even when `%evt.time.iso8601` in falco, mount localtime is a good choice for user to get real time.

daixiang0:
Oh, i saw that all thing in `/dev` has mounted, let us open a issue to track it.

PR #518, opened 2019-02-07
Change log timestamp to ISO8601 w/ timezone

mstemm:
This prevents ambiguity when, say, running falco in a container, which
may have a different /etc/localtime than the host system.

### DIFF
--- a/userspace/falco/logger.cpp
+++ b/userspace/falco/logger.cpp
@@ -30,10 +30,16 @@ const static struct luaL_reg ll_falco [] =
 };
 
 int falco_logger::level = LOG_INFO;
+bool falco_logger::time_format_iso_8601 = false;
 
 void falco_logger::init(lua_State *ls)
 {
 	luaL_openlib(ls, "falco", ll_falco, 0);
+}
+
+void falco_logger::set_time_format_iso_8601(bool val)
+{
+	falco_logger::time_format_iso_8601 = val;
 }
 
 void falco_logger::set_level(string &level)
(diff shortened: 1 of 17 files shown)


daixiang0:
Fix #527

mfdii:
@lorenzo-david can you review for @mstemm

mstemm:
Actually, this still requires some work before it's ready to review.

userspace/falco/logger.cpp
```diff
@@ -123,9 +129,18 @@ void falco_logger::log(int priority, const string msg)
 		}
 
 		std::time_t result = std::time(nullptr);
-		string tstr = std::asctime(std::localtime(&result));
-		tstr = tstr.substr(0, 24);// remove trailling newline
-		fprintf(stderr, "%s: %s", tstr.c_str(), copy.c_str());
+		if(falco_logger::time_format_iso_8601)
+		{
+			char buf[sizeof "YYYY-MM-DDTHH:MM:SS-0000"];
+			strftime(buf, sizeof(buf), "%FT%T%z", std::gmtime(&result));
```
lorenzo-david: Might be worth checking return value of strftime.

userspace/falco/logger.cpp
```diff
@@ -123,9 +129,18 @@ void falco_logger::log(int priority, const string msg)
...
+			fprintf(stderr, "%s: %s", buf, msg.c_str());
+		}
+		else
+		{
+			string tstr = std::asctime(std::localtime(&result));
```
lorenzo-david: I guess std::asctime can return empty tstr in case of error.

userspace/falco/logger.cpp
mstemm: ok will do.

userspace/falco/logger.cpp
mstemm: yep will fix.

PR #523, opened 2019-02-08
Add HTTP output handler

mfdii:
Adding a simple HTTP handler that allows you to post JSON output directly to a HTTP endpoint instead of the current method of using the program output and curl. 

This was implemented in C++ and the outputs.lua calls back into C++ to handle the the http post using libcurl. 

This can be enabled by setting the `http_output` option in `falco.yaml` to `enabled: true`.
```
http_output:
  enabled: true
  url: http://some.url/you/post/to
```

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -138,3 +138,7 @@ program_output:
   enabled: false
   keep_alive: false
   program: mail -s "Falco Notification" someone@example.com
+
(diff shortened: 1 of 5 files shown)


mfdii:
@mstemm this is failing on gcc-6, just fyi. I know you're working on it

PR #537, opened 2019-02-15
Update the Puppet module

JPLachance:
Greetings!

I noticed the Puppet module is a little outdated compared to this repository. I also noticed the installation was not working because the DEB repository was not managed by this module. This PR should fix that!

A list of the changes:
\* Apply puppet-lint recommendations
\* Update the README since the project moved from `draios` to `falcosecurity` in GitHub
\* Move parameters in their own file
\+ Add the DEB repository automatically (based on https://github.com/garethr/garethr-sysdig/blob/master/manifests/install.pp)
\+ Add the EPEL repository automatically (based on https://github.com/garethr/garethr-sysdig/blob/master/manifests/install.pp)
\+ Add a `logrotate` configuration
\* Update the configuration file with all the latest updates

🆘 Please note that I'm not a Puppet expert 😄

falco-CLA-1.0-contributing-entity: Coveo Solutions Inc.
falco-CLA-1.0-signed-off-by: Jean-Philippe Lachance <jplachance@coveo.com>

### DIFF
--- a/integrations/puppet-module/sysdig-falco/manifests/install.pp
+++ b/integrations/puppet-module/sysdig-falco/manifests/install.pp
@@ -3,13 +3,13 @@ class falco::install inherits falco {
   case $::osfamily {
     'Debian': {
       apt::source { 'sysdig':
-        location          => 'http://download.draios.com/stable/deb',
-        release           => 'stable-$(ARCH)/',
-        repos             => '',
-        required_packages => 'debian-keyring debian-archive-keyring',
-        key               => 'EC51E8C4',
-        key_source        => 'https://s3.amazonaws.com/download.draios.com/DRAIOS-GPG-KEY.public',
-        include_src       => false,
+        location => 'http://download.draios.com/stable/deb',
+        release  => 'stable-$(ARCH)/',
+        repos    => '',
+        key      => {
+          source => 'https://s3.amazonaws.com/download.draios.com/DRAIOS-GPG-KEY.public',
+          id     => 'D27A72F32D867DF9300A241574490FD6EC51E8C4'
+        },
       }
 
       ensure_packages(["linux-headers-${::kernelrelease}"])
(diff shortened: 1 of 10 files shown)


integrations/puppet-module/sysdig-falco/manifests/install.pp
```diff
@@ -1,6 +1,61 @@
 # == Class: falco::install
 class falco::install inherits falco {
+  case $::osfamily {
+    'Debian': {
+      apt::source { 'sysdig':
+        location          => 'http://download.draios.com/stable/deb',
+        release           => 'stable-$(ARCH)/',
+        repos             => '',
+        required_packages => 'debian-keyring debian-archive-keyring',
```
mstemm: I installed using puppet 4, which in turn uses puppetlabs-apt v6.3.0. In that version, `required_packages`, `key_source`, and `include_src` are all deprecated and attempts to use them results in an error:

```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Apt::Source[sysdig]:
  has no parameter named 'required_packages'
  has no parameter named 'key_source'
  has no parameter named 'include_src' at /etc/puppetlabs/code/environments/production/modules/falco/manifests/install.pp:5 on node ubuntu
```

How about we remove them and just state that puppet >= 4 is required?

JPLachance: Okay! So, I tried to stay compatible with Puppet 3. The latest version of `apt` that is compatible with Puppet 3 is 2.4.0, but it turns out there is a bug in that version of `apt`: https://tickets.puppetlabs.com/browse/MODULES-4913

So, do we upgrade to Puppet 4? On my side, I'll override apt to make it work, but here, we can break the compatibility with Puppet 3 if you want 🙂

JPLachance: Okay, done! The patched version of apt is 4.1.0, I upgraded to 4.5.1, which is compatible with Puppet 4.7.0. I did test 4.5.1, it works, and I'm kind of running out of time to test 6.3.0 😄 If you have the time, please do, otherwise, I think we can merge it as is 🙂

PR #539, opened 2019-02-19
Move kubernetes-response-engine to falcosecurity/kubernetes-response-engine

nestorsalceda:
As long as Falco and Response Engine have different release cycle, they
are separated.


mfdii:
The directory and the README for the directory is being deleted as part of this change. As we don't know how many places might be linking to this directory, we at least need to keep the `integrations/kubernetes-response-engine/README.md` directory/file with a note that the code has moved to the new location.

PR #559, opened 2019-03-09
kh: exclude rancher events

Kaizhe:


mstemm:
The rancher changes look good to me. Let's keep the PR open for a bit longer just in case we have more rules changes to make.

Issue #545, opened 2019-02-27
Note when alerts are correlated with system call drops


mstemm:
Falco's internal state can get confused in the face of dropped system calls, and dropped system calls can be common when doing a fetch of container information. We should identify periods where there are dropped system calls and flag any falco events during that time as potentially tainted by the dropped system calls.

justinzhf:
@mstemm hi! I am a graduate student of Nanjing University of Aeronautics and Astronautics. My major is data mining, to be exact, research on anomaly detection of complex system. I have finished a paper about power system anomaly detection. I am interested in this idea of GSoC 2019(https://github.com/cncf/soc/blob/master/README.md#falco-integration-with-aiml-platforms). I really want to do something about this idea, could you tell me how can I get started? Thanks you.

justinzhf:
@mstemm @mstemm hi! I am a graduate student of Nanjing University of Aeronautics and Astronautics. My major is data mining, to be exact, research on anomaly detection of complex system. I have finished a paper about power system anomaly detection. I am interested in this idea of GSoC 2019(https://github.com/cncf/soc/blob/master/README.md#falco-integration-with-aiml-platforms). I really want to do something about this idea, could you tell me how can I get started? Thank you.

mstemm:
@justinzhf your comment doesn't really have anything to do with this github issue. The way to apply for google summer of code projects is via https://summerofcode.withgoogle.com/. It looks like applications will be accepted between March 25 and April 9. 

I'll keep your comment and this reply up for a few days but I'll delete them afterward as it's not really related to this github issue.

mstemm:
Fixed by https://github.com/falcosecurity/falco/pull/561.

PR #561, opened 2019-03-13
Flag excess drops

mstemm:
Changes to monitor for dropped system calls/kernel buffer overflows and take configurable actions when this occurs.

Certainly related to and might be a fix for https://github.com/falcosecurity/falco/issues/545.


falco.yaml
```diff
@@ -63,6 +63,26 @@ priority: debug
 # buffered. Defaults to false
 buffered_outputs: false
 
+# Falco uses a shared buffer between the kernel and userspace to pass
+# system call informaation. When falco detects that this buffer is
```
Kaizhe: typo: informaation -> information

userspace/falco/overflow.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+		case ACT_ALERT:
+			m_outputs->handle_msg(now,
+					      falco_outputs::PRIORITY_CRITICAL,
+					      msg,
+					      rule,
+					      output_fields);
+			break;
+
+		case ACT_EXIT:
```
Kaizhe: do you happen to sort `m_actions`? if `ACT_EXIT` comes before `ACT_ALERT`, then the alert will be missing, right?

mattpag: +1, we should either sort it or explicitly say in the comment that the actions will be performed in order.

falco.yaml
```diff
@@ -63,6 +63,26 @@ priority: debug
...
+#   - "ignore": do nothing. If an empty list is provided, ignore is assumed.
+#   - "log": log a CRITICAL message noting that the buffer was full.
+#   - "alert": emit a falco alert noting that the buffer was full.
+#   - "exit": exit falco with a non-zero rc.
+#
+# The rate at which overflow log/alert messages are emitted is
+# governed by a token bucket. The rate corresponds to one message
+# every 30 seconds with a burst of 10 messages.
+
+kernel_buffer_overflow:
```
mattpag: For all occurrences (code/comments/logs) I would suggest using the term "ring buffer", maybe even "sysdig ring buffer", instead of "kernel buffer".

userspace/falco/overflow.cpp
mstemm: Good point. I think I'll move the exit to outside the loop so all the actions will be taken, and then the exit will happen.

Kaizhe: 👍

falco.yaml
mstemm: ok fixed

falco.yaml
mstemm: New name: "syscall event drops"

mstemm:
Also added a wiki page documenting the feature: https://github.com/falcosecurity/falco/wiki/Actions-For-Dropped-System-Call-Events and notes to the falco config wiki page for the new config options.

PR #563, opened 2019-03-14
Puppet write below /etc and new user_known_x macros!

JPLachance:
Greetings!

Falco is currently whitelisting multiple processes to write below `/etc` and Puppet was not one of them. Adding Puppet to the list of allowed `proc.name` will remove multiple false positives.

On top of that, I added the `user_known_write_root_conditions` and the `user_known_non_sudo_setuid_conditions` macros, so users can add their own condition without overriding the official Falco rules file.

Changes:
+ Add "puppet" in the list of known `proc.name` writing below etc because Puppet often manages configurations
+ Add the user_known_write_root_conditions macro to allow custom conditions in the "Write below root" rule
+ Add the user_known_non_sudo_setuid_conditions to allow custom conditions in the "Non sudo setuid" rule

falco-CLA-1.0-contributing-entity: Coveo Solutions Inc.
falco-CLA-1.0-signed-off-by: Jean-Philippe Lachance <jplachance@coveo.com>

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -909,7 +909,7 @@
                           gen_resolvconf., update-ca-certi, certbot, runsv,
                           qualys-cloud-ag, locales.postins, nomachine_binaries,
                           adclient, certutil, crlutil, pam-auth-update, parallels_insta,
-                          openshift-launc, update-rc.d)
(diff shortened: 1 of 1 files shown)


mstemm:
Thanks again for the changes!

PR #567, opened 2019-03-28
exclude oc on writing below root

Kaizhe:


Kaizhe:
@mstemm done, thanks!

PR #573, opened 2019-04-04
Rule updates 2019 04.v2

mstemm:
Several changes to address FPs based on user feedback.

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -904,7 +904,7 @@
   condition: (proc.name=prometheus-conf and fd.name startswith /etc/prometheus/config_out)
 
 - macro: openshift_writing_conf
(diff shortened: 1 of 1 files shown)


rules/falco_rules.yaml
```diff
@@ -172,6 +172,9 @@
 - list: gitlab_binaries
   items: [gitlab-shell, gitlab-mon, gitlab-runner-b, git]
 
+- list: dynatrace_binaries
```
Kaizhe: how bad it is if we don't exclude `dynatrace` here? I am still debating with myself whether we should include the security vendors in the open source version.

omer-sd: I don’t think it should be in the public repo. We can recommend this to our customers personally.

PR #575, opened 2019-04-04
rule update: add MITRE tags for rules

Kaizhe:
Added MITRE tags for default falco rules

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -1924,7 +1924,7 @@
   desc: >
     Detect grep private keys or passwords activity.
   condition: >
-    (spawned_process and
+    (spawned_process and container and
      ((grep_commands and private_key_or_password) or
       (proc.name = "find" and (proc.args contains "id_rsa" or proc.args contains "id_dsa")))
     )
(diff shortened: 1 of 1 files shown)


Kaizhe:
At the top level, we have the following categories are covered:

`Execution`, `Persistence`, `Privilege Escalation`, `Credential Access`, `Discovery`, `Lateral Movement`, `Exfiltration`, `Command and Control (in sub cat Port Knocking)`

https://attack.mitre.org/matrices/enterprise/linux/

KnoxAnderson:
Maybe something like Mitre-Execution @Kaizhe

Kaizhe:
👌

rules/falco_rules.yaml
```diff
@@ -1107,7 +1107,7 @@
     Sensitive file opened for reading by trusted program after startup (user=%user.name
     command=%proc.cmdline parent=%proc.pname file=%fd.name parent=%proc.pname gparent=%proc.aname[2])
   priority: WARNING
-  tags: [filesystem]
+  tags: [filesystem, Credential Access]
```
mstemm: I think this should be `mitre_credential_access` right?

PR #578, opened 2019-04-10
Rule updates 2019 04.v4 mitre persistence

Kaizhe:


rules/falco_rules.yaml
```diff
@@ -1910,14 +1910,19 @@
 - macro: grep_commands
   condition: (proc.name in (grep_binaries))
 
+- macro: grep_more
+  condition: (never_true)
```
Kaizhe: this need to be turn on based on requirements

rules/falco_rules.yaml
```diff
@@ -343,11 +343,11 @@
     WARNING
   tag: [file, mitre_persistence]
 
-- rule: Schedule Cron Jobs in Container
+- rule: Schedule Cron Jobs
   desc: Detect cron jobs scheduled in container
```
mstemm: Also remove container from the description.

rules/falco_rules.yaml
```diff
@@ -343,11 +343,11 @@
...
   condition: >
     ((open_write and fd.name startswith /etc/cron) or
-     (spawned_process and proc.name = "crontab")) and container
+     (spawned_process and proc.name = "crontab"))
   output: >
     Cron jobs were scheduled to run inside container (user=%user.name command=%proc.cmdline
```
mstemm: Also remove container from output.

rules/falco_rules.yaml
```diff
@@ -1910,14 +1910,19 @@
 - macro: grep_commands
   condition: (proc.name in (grep_binaries))
 
+- macro: grep_more
```
mstemm: Can you add a comment indicating what this does (namely a less restrictive search for things that might be passwords/ssh/etc) ?

rules/falco_rules.yaml
```diff
@@ -1936,17 +1941,84 @@
...
   tags: [process, mitre_credential_access]
 
+- list: log_directories
+  items: [/var/log, /dev/log]
+
+- list: log_files
+  items: [syslog, auth.log, secure, kern.log, cron, user.log, dpkg.log, last.log, yum.log, access_log, mysql.log, mysqld.log]
+
+- macro: access_log_files
+  condition: (fd.directory pmatch (log_directories) or fd.filename in (log_files))
```
mstemm: If you're trying to match a specific directory against a set of other directories, you can just use `fd.directory in (log_directories)`. You only need to use pmatch when you're trying to match a full path against a set of directories.

rules/falco_rules.yaml
```diff
@@ -1936,17 +1941,84 @@
...
+
+- rule: Clear Log Activities
+  desc: Detect clearing of critical log files
+  condition: >
+    ((open_write and access_log_files and evt.arg.flags contains "O_TRUNC") or
+     (remove and access_log_files))
```
mstemm: As we just discussed, unlinks don't have an active fd--the path is just an argument to the syscall. We should limit rules like this to only writes until we can efficiently break up path arguments into directory and/or filename components.

rules/falco_rules.yaml
```diff
@@ -1936,17 +1941,84 @@
...
+  output: >
+    Log files were removed (user=%user.name command=%proc.cmdline file=%fd.name)
+  priority:
+    WARNING
+  tags: [file, mitre_defense_evasion]
+
+- list: data_remove_commands
+  items: [shred, mkfs, mke2fs]
```
mstemm: Maybe `dd` also?

rules/falco_rules.yaml
```diff
@@ -1936,17 +1941,84 @@
...
+
+- macro: clear_data_procs
+  condition: (proc.name in (data_remove_commands))
+
+- rule: Remove Data from Disk
```
mstemm: Maybe add "bulk" to this to indicate it's a bulk operation?

rules/falco_rules.yaml
```diff
@@ -1936,17 +1941,84 @@
...
+  desc: Detect process running to clear data from disk
+  condition: spawned_process and clear_data_procs
+  output: >
+    Data has been removed from disk (user=%user.name command=%proc.cmdline file=%fd.name) 
+  priority:
+    WARNING
+  tags: [process, mitre_persistence]
+
 - rule: Delete Bash History
   desc: Detect bash history deletetion
```
mstemm: Typo deletetion -> deletion

rules/falco_rules.yaml
```diff
@@ -1936,17 +1941,84 @@
...
+    NOTICE
+  tag: [process, mitre_persistence]
+
+- list: exclude_hidden_directories
+  items: [/root/.cassandra]
+
+- macro: check_container_only
+  condition: (never_true or container)
+  
+- rule: Create Hidden Files or Directories
```
mstemm: This will be very noisy, I think we should disable it by default.

PR #579, opened 2019-04-11
Rule updates 2019 04.v5

mstemm:


rules/falco_rules.yaml
```diff
@@ -329,20 +346,117 @@
...
+
+- rule: Unexpected outbound connection destination
+  desc: Detect any outbound connection to a destination outside of an allowed set of ips, networks, or domain names
   condition: >
-    open_write and fd.filename in (user_context_files)
+    consider_all_outbound_conns and outbound and not
+    ((fd.sip in (allowed_destination_ipaddrs)) or
+     (fd.snet in (allowed_destination_networks)) or
+     (fd.sip.name in (allowed_destination_domains)))
+  output: Disallowed outbound connection destination (command=%proc.cmdline connection=%fd.name user=%user.name)
```
Kaizhe: should we put both `sip` and `cip` in the output as we are showing the connection detail

mstemm: %fd.name has both halves, here's an example: 
```15:48:15.536202984: Notice Disallowed outbound connection destination (command=curl http://www.github.com connection=10.0.2.15:35873->192.30.255.113:80 user=mstemm)```

File: rules/falco_rules.yaml [hunk 2]
```diff
@@ -1659,6 +1792,57 @@
...
+# (proc.env contains "HTTP_PROXY=http://my.http.proxy.com ")
+# The trailing space is intentional so avoid matching on prefixes of
+# the actual proxy.
+- macro: allowed_ssh_proxy_env
+  condition: (always_true)
+
+- rule: Program run with disallowed http proxy env
+  desc: An attempt to run a program with a disallowed HTTP_PROXY environment variable
+  condition: >
+    spawned_process and not allowed_ssh_proxy_env and proc.env icontains HTTP_PROXY
```
Kaizhe: Should we add a list of commands like `curl` and `wget` in this case?

mstemm: I couldn't think of a way to get a good list though, and I didn't want to limit it to only curl/wget.

rules/falco_rules.yaml
Kaizhe: good!

rules/falco_rules.yaml
Kaizhe: then this will be a little bit noisy, as not all the commands in the shell are relevant to the PROXY env variable. We can have a list and add more later.

mstemm: Yeah I struggled between looking for the execs (bad when lots of execs pass around the bad env) and the connects (bad when there are a lot of connects). So I guess I'll add a list.

PR #582, opened 2019-04-16
Fix errors when building via docker from MacOS

mfdii:
When using the GitHub CMake integration, it expects a real user to be present via /etc/passwd. There are ways around this when building on a linux host via the Falco docker builder container, but on a Mac it gets a bit harder. Moving cmake to use curl and pull the yamlcpp tarball via S3 removes the dependency on a local user existing. 

Also, when building via the docker builder, I got warnings (which were treated as errors by GCC) about variable initialization order. Reordered the variables in question to get get rid of the errors, however I'm not sure if we need to check the compiler flags we are using in the Falco docker build container as I haven't seen this before.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -241,8 +241,6 @@ else()
 	message(STATUS "Using bundled yaml-cpp in '${YAMLCPP_SRC}'")
 	set(YAMLCPP_LIB "${YAMLCPP_SRC}/libyaml-cpp.a")
 	set(YAMLCPP_INCLUDE_DIR "${YAMLCPP_SRC}/include")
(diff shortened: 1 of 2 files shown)


mfdii:
@mstemm fixed my multiple commit mess. should be ready for review.

CMakeLists.txt
```diff
@@ -244,8 +244,8 @@ else()
 	# Once the next version of yaml-cpp is released (first version not requiring
```
mstemm: Can you remove these comments as they aren't relevant anymore?

mfdii: done.

Issue #588, opened 2019-04-29
aws-cni emits excessive logs


whiskeyjimbo:
aws-cni for eks hits rule `Change thread namespace - an attempt to change a program/thread\'s namespace (commonly done as a part of creating a container) by calling setns.` quite a bit causing an excessive amount of log messages.

Simple solution would be to add `aws-cni` to: https://github.com/falcosecurity/falco/blob/0e1c436d14fb468dd1530115ed19e43faa0a8331/rules/falco_rules.yaml#L1429 is that the correct solution though? or one could continue with an override.

mfdii:
Yes, adding `aws-cni` there would be the correct solution to reduce the FPs. Would you be able to submit a PR, else we can get it into the next round of rule updates?

mstemm:
Can you include a full alert example? I'd like to see the context include cmdline, etc.

fntlnz:
@whiskeyjimbo can you provide more info on this?

leodido:
Closing this since missing info. Comment (and we'll re-open) in case there are details about this.

PR #590, opened 2019-05-01
Rule updates 2019 05.v1

mstemm:


mfdii:
Could you add a fix for the FPs generated by `aws-cni`? #588

rules/falco_rules.yaml
```diff
@@ -1029,6 +1029,13 @@
                    fd.name startswith "/etc/rancher-dns")
              )
 
+- macro: jboss_in_container_writing_passwd
+  condition: >
+    ((proc.cmdline="run-java.sh /opt/jboss/container/java/run/run-java.sh"
```
Kaizhe: the `proc.cmdline` looks strange, the `args[1]` is the abs path of the `exe`?

mstemm: It does look strange, but the policy event did have that as the full cmdline. Here's an example:

```
File below /etc opened for writing (user=<NA> command=run-java.sh /opt/jboss/container/java/run/run-java.sh parent=java pcmdline=java -javaagent:/opt/jboss/container/jolokia/jolokia.jar=config=/opt/jboss/container/jolokia/etc/jolokia.properties....
```

And for reference, the output field of that rule is:

```
output: "File below /etc opened for writing (user=%user.name command=%proc.cmdline parent=%proc.pname pcmdline=%proc.pcmdline file=%fd.name program=%proc.name gparent=%proc.aname[2] ggparent=%proc.aname[3] gggparent=%proc.aname[4])"
```

PR #593, opened 2019-05-03
Default Rule adjustments

natalysheinin:
## Overview

Correcting typo in `- rule: Launch Suspicious Network Tool in Container` and updating and including the google accounts daemon in the `Read sensitive file untrusted` rule.

## Reference

New versions of gke have the google_accounts daemon running stuff at
`proc.aname[3]`

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -643,7 +643,8 @@
 - macro: run_by_google_accounts_daemon
   condition: >
     (proc.aname[1] startswith google_accounts or
-     proc.aname[2] startswith google_accounts)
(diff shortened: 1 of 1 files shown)


Kaizhe:
@natalysheinin thank you for the input. Could you please also include a sample output(the FP case) with the original rules ? I think this will be served a good record to this PR.

mfdii:
@Kaizhe I spoke with @natalysheinin at Kubecon. Because the rules have been modified to remove the alert,they no longer has a sample to send us. Can we merge this?

Kaizhe:
@mfdii thanks for confirm that.

PR #613, opened 2019-05-18
rule update: add exception for write below etc 

Kaizhe:
Add exception for `write below etc`:

`etcd-manager` will consistently updates dns: `/etc/hosts` file. This happens when k8s cluster created by `kops`

add new rule: `create symlinks over sensitive files and directories`

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -69,6 +69,9 @@
 - macro: spawned_process
   condition: evt.type = execve and evt.dir=<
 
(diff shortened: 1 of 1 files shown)


Kaizhe:
> The approach looks good, but I wanted to make sure that you only want to cover symlinks of the whole directory and not paths below it, right?

Yes @mstemm, for now the target would either be a list of sensitive files or directories (directory name only, not sub directories or files under it)

PR #617, opened 2019-05-23
fix egrep rule and ncat rule

dms1lva:
fixed the grep binaries list that had `egre` instead of `egrep`
added the ncat command line arguments `-c` and `-e` that can be used to spawn remote shells

Launching this : ` ncat -l localhost 4443 -e "/bin/ls"`

would not match anything when running: `sudo sysdig "proc.name = \"ncat\" and (proc.args contains \"--sh-exec\" or proc.args contains \"--exec\")"`

using `sudo sysdig "proc.name = \"ncat\" and (proc.args contains \"--sh-exec\" or proc.args contains \"--exec\" or proc.args contains \"-e\")"` fixes this

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -2114,7 +2114,8 @@
   condition: >
     spawned_process and container and
     ((proc.name = "nc" and (proc.args contains "-e" or proc.args contains "-c")) or
-     (proc.name = "ncat" and (proc.args contains "--sh-exec" or proc.args contains "--exec" or proc.args contains "-e" or proc.args contains "-c"))
+     (proc.name = "ncat" and (proc.args contains "--sh-exec" or proc.args contains "--exec" or proc.args contains "-e " 
+                              or proc.args contains "-c " or proc.args contains "--lua-exec"))
     )
   output: >
     Netcat runs inside container that allows remote code execution (user=%user.name
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -2114,7 +2114,7 @@
   condition: >
     spawned_process and container and
     ((proc.name = "nc" and (proc.args contains "-e" or proc.args contains "-c")) or
-     (proc.name = "ncat" and (proc.args contains "--sh-exec" or proc.args contains "--exec"))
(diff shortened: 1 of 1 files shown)


rules/falco_rules.yaml
```diff
@@ -2114,7 +2114,7 @@
   condition: >
     spawned_process and container and
     ((proc.name = "nc" and (proc.args contains "-e" or proc.args contains "-c")) or
-     (proc.name = "ncat" and (proc.args contains "--sh-exec" or proc.args contains "--exec"))
+     (proc.name = "ncat" and (proc.args contains "--sh-exec" or proc.args contains "--exec" or proc.args contains "-e" or proc.args contains "-c"))
```
mstemm: How about "-e " and "-c " to make sure you don't match args that start with -c, like -chat?

Kaizhe: since we will evaluate the `proc.name` first,  the existing evaluation for arguments will be evaluated later, then it should be fine. 

I have a minor request: can you please also add `--lua-exec` in the argument evaluation?

rules/falco_rules.yaml
```diff
@@ -2151,7 +2151,7 @@
   tags: [network, process, mitre_discovery, mitre_exfiltration]
 
 - list: grep_binaries
-  items: [grep, egre, fgrep]
+  items: [grep, egrep, fgrep]
```
Kaizhe: Thanks for catching that!

rules/falco_rules.yaml
dms1lva: I believe @mstemm's suggestion is valid because ncat has a `--chat` option, which gets caught by my current proposed fix. I'll fix it.

Yes @Kaizhe I'll add the `--lua-exec` option also, good catch!

rules/falco_rules.yaml
dms1lva: You're welcome! :)

mstemm:
The cause of the test failures was a recent sysdig change that I fixed with https://github.com/falcosecurity/falco/pull/646, and not due to the rules, so I'll go ahead and merge.

PR #625, opened 2019-05-27
Allow Ansible to run using Python 3

cnorthwood:
Some newer distros default to Python 3 by default, not 2, which causes Ansible to trigger these rules.


nibalizer:
lgtm

mstemm:
The cause of the test failures was a different sysdig change that was fixed in https://github.com/falcosecurity/falco/pull/635. I'll go ahead and merge these changes.

PR #634, opened 2019-05-30
Add rules for running with IBM Cloud Kubernetes Service

nibalizer:
IBM Cloud Kubernetes Service is a hosted k8s from IBM. (Sometimes referred to as IKS)


mstemm:
The cause of the test failures was a different sysdig change that was fixed in #635. I'll go ahead and merge these changes.

PR #637, opened 2019-05-31
Rule updates 2019 05.v3

Kaizhe:


rules/falco_rules.yaml
```diff
@@ -1104,7 +1104,7 @@
   condition: (proc.name=prometheus-conf and fd.name startswith /etc/prometheus/config_out)
 
 - macro: openshift_writing_conf
-  condition: (proc.name=oc and fd.name startswith /etc/origin/node)
+  condition: (proc.name=oc and (fd.name startswith /etc/origin/node or fd.name startswith /etc/sysconfig/origin-node))
```
fntlnz: Should this include also `/etc/sysconfig/atomic-openshift-node` and `/etc/sysconfig/atomic-openshift-master` ?

Kaizhe: The change was wrong, I will undo it.

PR #648, opened 2019-06-05
rule update: update image check from rancher_write_conf macro

Kaizhe:


rules/falco_rules.yaml
```diff
@@ -1039,9 +1039,9 @@
                proc.cmdline startswith "agent.py /opt/datadog-agent")
               and fd.name startswith "/etc/dd-agent")
 
+# TODO: have a better way to identify the image repo (public vs internal)
 - macro: rancher_writing_conf
-  condition: (container.image.repository in (rancher_images)
-              and proc.name in (lb-controller,rancher-dns,healthcheck,rancher-metadat)
+  condition: (proc.name in (lb-controller,rancher-dns,healthcheck,rancher-metadat)
```
mstemm: I think the reason I qualified the macro with the image is that the process name 'healthcheck' was really generic. Maybe we could continue to have the image check just for that process?

Kaizhe: I agree `healthcheck` is too generic. How about removing `healthcheck` for now.

Kaizhe: added issue: https://github.com/falcosecurity/falco/issues/649

mstemm: Thanks! I think you can close the issue now given your later commits.

rules/falco_rules.yaml
```diff
@@ -1040,11 +1040,13 @@
...
-              and (fd.name startswith "/etc/haproxy" or
-                   fd.name startswith "/etc/rancher-dns")
-             )
+  condition: (((proc.name=healthcheck and container.image.repository contains "rancher/healthcheck") or
+               (proc.name=lb-controller and container.image.repository contains "rancher/lb-service-haproxy") or
+               (proc.name=rancher-dns and container.image.repository contains "rancher/dns")) and 
+              (fd.name startswith "/etc/haproxy" or fd.name startswith "/etc/rancher-dns"))
+
+- macro: rancher_writing_root
+  condition: (proc.name=rancher-metadat and container.image.repository contains "rancher/metadata" and fd.name startswith "/answers.json")
```
fntlnz: the process name here seems wrong

mstemm: That's because process names that falco sees are truncated at 16 characters. It's a kernel limit: http://lxr.linux.no/linux+v2.6.37/include/linux/sched.h#L245

fntlnz: Thanks for the clarification Mark. I have concerns about this but I'll open a separate issue once I have more info.

PR #651, opened 2019-06-06
wip: Clean up privileged/sens mount container rules

mstemm:
Previously, the exceptions for Launch Privileged Container/Launch
Sensitive Mount Container came from a list of "trusted" images and/or a
macro that defined "trusted" containers. We want more fine-grained
control over the exceptions for these rules, so split them into
exception lists/macros that are specific to each rule. This defines:

 - privileged_images: only those images that are known to require
   privileged=true
 - privileged_containers: uses privileged_images and (for now) still
   allows all openshift images
 - user_privileged_containers: allows user exceptions
 - sensitive_mount_images: only those images that are known to perform
   sensitive mounts
 - sensitive_mount_containers: uses sensitive_mount_images
 - user_sensitive_mount_containers: allows user exceptions

For backwards compatibility purposes only, we keep the trusted_images
list and user_trusted_containers macro and they are still used as
exceptions for both rules. Comments recommend using the more
fine-grained alternatives, though.

While defining these lists, also do another survey to see if they still
require these permissions and remove them if they didn't.

Removed:
 - quay.io/coreos/flannel
 - consul

Moved to sensitive mount only:
 - gcr.io/google_containers/hyperkube
 - datadog
 - gliderlabs/logspout

Finally, get rid of the k8s audit-specific lists of privileged/sensitive
mount images, relying on the ones in falco_rules.yaml.

```release-note
Refactor rules/exception lists for sensitive mount, privileged, hostnetwork images to have separate lists for each instead of a combined "trusted" list that was used for all three.
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -1671,7 +1671,7 @@
 # NOTE: This list is only provided for backwards compatibility with
 # older local falco rules files that may have been appending to
 # trusted_images. To make customizations, it's better to add images to
-# either privileged_images or sensitive_mount_images.
+# either privileged_images or falco_sensitive_mount_images.
 - list: trusted_images
   items: []
 
@@ -1696,7 +1696,7 @@
           registry.access.redhat.com/sematext/logagent]
 
 # These container images are allowed to run with --privileged
-- list: privileged_images
+- list: falco_privileged_images
   items: [
     sysdig/agent, sysdig/falco, sysdig/sysdig,
     gcr.io/google_containers/kube-proxy, calico/node,
@@ -1704,11 +1704,11 @@
     docker/ucp-agent, sematext_images
     ]
 
-- macro: privileged_containers
+- macro: falco_privileged_containers
   condition: (openshift_image or
               user_trusted_containers or
               container.image.repository in (trusted_images) or
-              container.image.repository in (privileged_images) or
+              container.image.repository in (falco_privileged_images) or
               container.image.repository startswith istio/proxy_ or
               container.image.repository startswith quay.io/sysdig)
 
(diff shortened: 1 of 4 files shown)


rules/falco_rules.yaml
```diff
@@ -1652,40 +1652,91 @@
...
-- macro: trusted_containers
+- macro: privileged_containers
   condition: (openshift_image or
+              user_trusted_containers or
               container.image.repository in (trusted_images) or
+              container.image.repository in (privileged_images) or
               container.image.repository startswith istio/proxy_ or
               container.image.repository startswith quay.io/sysdig)
 
+# Add conditions to this macro (probably in a separate file,
```
fntlnz: I like this idea

rules/falco_rules.yaml
```diff
@@ -1652,40 +1652,91 @@
...
-    rook/toolbox, cloudnativelabs/kube-router, consul, mesosphere/mesos-slave,
-    datadog/docker-dd-agent, datadog/agent, docker/ucp-agent, gliderlabs/logspout
+    sysdig/agent, sysdig/falco, sysdig/sysdig,
+    gcr.io/google_containers/kube-proxy, calico/node,
+    rook/toolbox, cloudnativelabs/kube-router, mesosphere/mesos-slave,
+    docker/ucp-agent, sematext_images
     ]
 
-- macro: trusted_containers
+- macro: privileged_containers
```
Kaizhe: This should all into topic, given customers might have defined the macro/list names duplicated with the default rule sets. Should we have some reserved words like `falco_*`

leodido: +1 for reserved prefixes.

rules/falco_rules.yaml
leodido: Me too

rules/falco_rules.yaml
mstemm: Sure, I'll start using a `falco_` prefix for falco-provided lists/macros. Not changing the name of any existing list/macro though.

PR #652, opened 2019-06-06
rule update: improve rancher macro

Kaizhe:


rules/falco_rules.yaml
```diff
@@ -1040,13 +1043,16 @@
...
-               (proc.name=rancher-dns and container.image.repository contains "rancher/dns")) and 
+  condition: ((proc.name in (healthcheck, lb-controller, rancher-dns)) and
+              (container.image.repository contains "rancher/healthcheck" or
+               container.image.repository contains "rancher/lb-service-haproxy" or
+               container.image.repository contains "rancher/dns") and
               (fd.name startswith "/etc/haproxy" or fd.name startswith "/etc/rancher-dns"))
 
 - macro: rancher_writing_root
-  condition: (proc.name=rancher-metadat and container.image.repository contains "rancher/metadata" and fd.name startswith "/answers.json")
+  condition: (proc.name=rancher-metadat and
```
fntlnz: is this process name right?

Kaizhe: yes, like @mstemm pointed out in my previous PR, the falco has a limit for the process name which is 15.

fntlnz: ok thank you

PR #675, opened 2019-06-17
rule update: add back trusted_containers list for backport compatibility

Kaizhe:
Signed-off-by: kaizhe <derek0405@gmail.com>

**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
 /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flaky-test

> If contributing rules or changes to rules, please make sure to uncomment the appropriate kind

> /kind rule/update
 /kind rule/create

> /area engine
 /area rules
> /area deployment
> /area integrations
> /area examples

For backport compatibility

Fixes #

```release-note
N/A
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -1677,12 +1677,12 @@
 - list: trusted_images
   items: []
 
-# NOTE: This list is only provided for backwards compatibility with
+# NOTE: This macro is only provided for backwards compatibility with
 # older local falco rules files that may have been appending to
-# trusted_containers. To make customizations, it's better to add containers to
-# user_trusted_containers, user_priivleged_containers or user_sensitive_mount_containers.
-- list: trusted_containers
-  items: []
+# trusted_images. To make customizations, it's better to add containers to
+# user_trusted_containers, user_privileged_containers or user_sensitive_mount_containers.
+- macro: trusted_containers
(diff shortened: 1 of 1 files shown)


rules/falco_rules.yaml
```diff
@@ -1677,6 +1677,10 @@
 - list: trusted_images
   items: []
 
+# Backport compatibility
```
mstemm: Can you add a longer comment block like the one for trusted_images that specifically says this is only for backwards compatibility and instead the changes should be made to *_privileged_images or _sensitive_mount_images if possible, and *_privileged_containers/_sensitive_mount_containers as a backup?

rules/falco_rules.yaml
```diff
@@ -1677,6 +1677,10 @@
...
+- list: trusted_containers
```
mstemm: I think trusted_containers used to be a macro, right? In that case it should be a macro here and also it should be added to all the rules that might rely on that list.

Kaizhe: so it should be 
```
- macro: trusted_containers
  condition: container.image.repository in (trusted_images)
```

PR #687, opened 2019-06-25
wip: unit test setup

### LINKED ISSUE #686: Unit test setup

Unit test via [Catch2](https://github.com/catchorg/Catch2).

**Why is this needed**:

Do I really need to fill this? LOL

fntlnz:
**What type of PR is this?**

/kind feature

This PR adds support to use catch2 as a testing framework, initially it allows only to test falco_engine but it will be expanded to other areas as soon as we extract functionalities into libraries.

There's no actual tests in this PR since this one is to setup the environment that will then be used in other PRs checking in the tests.

Here's how a unit test will look like:

```cpp
#include <catch.hpp>

using namespace std;
using Catch::Matchers::Equals;

TEST_CASE("someFunction check if false is false") {
    REQUIRE_FALSE(false);
}
```

Tests will be organized under the `tests` folder.

For example, `falco_engine` tests will go under `tests/engine` using the same structure as for `userspace`.

For every other type of assertion and details, refer to the the [catch2 documentation](https://github.com/catchorg/Catch2/blob/master/docs/Readme.md#top).

Fixes #686

image

```release-note
new: unit test setup
```


leodido:
/hold
I propose to hold it until we integrate them in Travis.

PR #709, opened 2019-07-08
ka.useragent field

### LINKED ISSUE #698: Add a new field ka.useragent
For convenient reference:

jevt.value[/userAgent] -> ka.useragent

**Why is this needed**:
It's nice to have :)

fntlnz:
**What type of PR is this?**

/kind feature
/kind cleanup

NONE

/area engine

- Adds the `ka.useragent` field as alias of jevt.value[/userAgent]
- Doing this also required another commit to do the necessary formatting according to the new `.clang-format`

Fixes #698 

```release-note
fields: kubernetes audit ka.useragent alias
```

### DIFF
--- a/userspace/engine/falco_engine_version.h
+++ b/userspace/engine/falco_engine_version.h
@@ -24,4 +24,4 @@ limitations under the License.
 // This is the result of running "falco --list -N | sha256sum" and
 // represents the fields supported by this version of falco. It's used
 // at build time to detect a changed set of fields.
-#define FALCO_FIELDS_CHECKSUM "9b5557ec8f16f5606a1544573b152d211d5212f653ee039146836a17266ff449"
+#define FALCO_FIELDS_CHECKSUM "ceb069d9f9b2d4ebcc5de39bddc53b7af2e6b8f072edc293668fd6ac4e532413"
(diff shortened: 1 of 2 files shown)


fntlnz:
To test this, this rule can be used: 

```yaml
- rule: Only allow hyperkube as useragent
  desc: Test doing an useragent check for k8s audit
  condition: ka.useragent != "hyperkube"
  priority: ERROR
  output: "unexpected client useragent: useragent=%ka.useragent " 
  source: k8s_audit
```

Here is a detection example:

```
Events detected: 434
Rule counts by severity:
   ERROR: 434
Triggered rules by rule name:
   Only allow hyperkube as useragent: 434
   - event drop detected: 44 occurrences
   - num times actions taken: 11
```

userspace/engine/falco_engine_version.h
```diff
@@ -24,4 +24,4 @@ limitations under the License.
 // This is the result of running "falco --list -N | sha256sum" and
 // represents the fields supported by this version of falco. It's used
 // at build time to detect a changed set of fields.
-#define FALCO_FIELDS_CHECKSUM "9b5557ec8f16f5606a1544573b152d211d5212f653ee039146836a17266ff449"
+#define FALCO_FIELDS_CHECKSUM "ceb069d9f9b2d4ebcc5de39bddc53b7af2e6b8f072edc293668fd6ac4e532413"
```
mstemm: When you update the checksum, that almost always means new fields were added to either sysdig (for syscall events) or falco (for k8s audit events). When that happens, you should also bump the `FALCO_ENGINE_VERSION` define on line 22. That allows us to know the times at which the engine was upgraded to have new fields. When you actually *use* those fields, you should update the rules file that uses them to have a `- required_engine_version: NNN` where NNN is the current value of `FALCO_ENGINE_VERSION`.

mstemm:
In the future, I think it would be a good idea to separate whitespace/formatting changes from other changes. It's hard to tell which changes are simply syntax and which change behavior.

fntlnz:
@mstemm We did the changes in two separate commits, however I agree that doing in two different PRs would've been even better!

userspace/engine/falco_engine_version.h
fntlnz: good to know, is this documented anywhere or should we do it?

mstemm: From a user perspective, there are some notes at https://falco.org/docs/rules/#versioning. Not much for developer specific notes, though.

PR #711, opened 2019-07-08
ignore container probes

### LINKED ISSUE #310: ignore k8s readiness and liveness exec probes
Falco detects liveness and readiness exec probes as "Non sudo setuid" and individually whitelisting all the probes as rules is quite tedious. Since probes get called very frequently, this produces a ton of alerts. It would be great if Falco could by default ignore commands launched from readiness and liveness probes -- I'm not sure how to write a rule that would do this based on the existing output without being too permissive (`user` is `<NA>`, Pod is different per probe, command is different per probe).

agilgur5:
- support for this check was added in sysdig and engine v3
- k8s and other container healthchecks, liveness, and readiness probes
  should be ignored by certain rules like "Non sudo setuid"

Signed-off-by: Anton Gilgur <agilgur5@gmail.com>

**What type of PR is this?**

/kind bug
/kind rule-update

/area rules
/area integrations

Container probes currently fire off alerts, and since they are frequent, fire off lots of noisy false positive alerts. The support introduced in sysdig to detect these alerts means they can now be ignored in the default rules.

Fixes #310 

/assign @mstemm 

`required_engine_version` has never been changed before, so I'm not sure what that process entails or how that should work (or if 3 is correct -- that's the one you said, so I just went with that).
You also mentioned in #310 that `Terminal shell in container` might also need an exception, but I've not gotten a false positive on that one before. I wrote the container probe check as a macro so it could easily be added anywhere though (might want to move the macro up higher or something then from where it currently is).

Not sure if my comment on the macro should include some other details or what

Yes? I think so, since engine version update and rules update.

```release-note
- `required_engine_version` was updated
- A new `is_container_probe` macro was added
- The "Non sudo setuid" rule was updated to ignore container probes (via above macro)
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -24,7 +24,7 @@
 # change to this rules file, we'll uncomment this line and set it to
 # the falco engine version in use at the time.
 #
-#- required_engine_version: 2
+- required_engine_version: 4
 
 # Currently disabled as read/write are ignored syscalls. The nearly
 # similar open_write/open_read check for files being opened for
@@ -2118,6 +2118,15 @@
 - macro: user_known_non_sudo_setuid_conditions
   condition: user.name=root
 
+# Various probes may be necessary to manage and orchestrate containers and
+# these probes should be excepted out of certain rules.
+# See https://github.com/falcosecurity/falco/issues/310
+- macro: is_container_probe
+  condition: >
+    (proc.is_container_healthcheck=true or
+    proc.is_container_liveness_probe=true or
+    proc.is_container_readiness_probe=true)
+
 # sshd, mail programs attempt to setuid to root even when running as non-root. Excluding here to avoid meaningless FPs
 - rule: Non sudo setuid
   desc: >
@@ -2134,6 +2143,7 @@
     and not java_running_sdjagent
     and not nrpe_becoming_nagios
     and not user_known_non_sudo_setuid_conditions
+    and not is_container_probe
   output: >
     Unexpected setuid call by non-sudo, non-root program (user=%user.name cur_uid=%user.uid parent=%proc.pname
     command=%proc.cmdline uid=%evt.arg.uid container_id=%container.id image=%container.image.repository)


rules/falco_rules.yaml
```diff
@@ -2104,6 +2104,15 @@
 - macro: user_known_non_sudo_setuid_conditions
   condition: user.name=root
 
+# Various probes may be necessary to manage and orchestrate containers and
+# these probes should be excepted out of certain rules.
+# See https://github.com/falcosecurity/falco/issues/310
+- macro: is_container_probe
+  condition: >
+    proc.is_container_healthcheck or
```
mstemm: Can you surround the condition with parentheses? It makes sure you don't end up with ambiguity when using the macro in larger logical expressions.

agilgur5: Yup, added it in the latest version. Would've thought they'd be auto-added around macros o.o

rules/falco_rules.yaml
```diff
@@ -24,7 +24,7 @@
 # change to this rules file, we'll uncomment this line and set it to
 # the falco engine version in use at the time.
 #
-#- required_engine_version: 2
+- required_engine_version: 3
```
```suggestion
- required_engine_version: 4
```

leodido: Or just rebase on top of the `dev` branch.

mstemm:
Can you remind me of a container that had a health check/liveness probe that also performed setuids? Was this one of your own containers or a commonly available one?

rules/falco_rules.yaml
agilgur5: Ok. Is 4 the minimum necessary for this check?

agilgur5:
> Can you remind me of a container that had a health check/liveness probe that also performed setuids? Was this one of your own containers or a commonly available one?

I'm a little confused, based on the addition to sysdig and your response in #310 I thought this was something that was reproduced (esp. since I hadn't heard of the "Terminal in Shell" one).

If it hasn't been repro'd, might want to remove `is_container_healthcheck` from the macro as I don't recall experiencing a false positive on that one (though most healthchecks are over HTTP, not exec -- it may also have false positives with exec commands). It wasn't "a container" that performed setuids, it was Kubernetes performing the exec commands. Say if our readiness check checked for the existence of a file or volume, that would cause Falco to fire off alerts in the default ruleset and I would have to add custom ignores the specific command for the specific container. I don't remember which specific exec commands anymore but it didn't matter what they were.

rules/falco_rules.yaml
agilgur5: ?

leodido: Yes

agilgur5: ok done

mstemm:
Sorry for the delay on reviewing/merging this. I just want to do some double-checking to make sure that the setuid that occurs by runc itself as it creates the probe will be detected as being done by a healthcheck/liveness/readiness probe. I'm more certain that the exec()d health check liveness readiness probe would be properly detected, but I'm guessing the setuid is occurring before that point.

fntlnz:
@agilgur5 I checked out this PR and got a compilation error.

```
Thu Aug  8 10:01:58 2019: Falco initialized with configuration file falco.yaml
Thu Aug  8 10:01:58 2019: Loading rules from file rules/falco_rules.yaml:
Thu Aug  8 10:01:58 2019: Runtime error: Error loading rules: ...ts/falcosecurity/falco/userspace/engine/lua/compiler.lua:198: Compilation error when compiling "(proc.is_container_healthcheck or proc.is_container_liveness_probe or proc.is_container_readiness_probe)
": 32: syntax error, unexpected 'or', expecting 'pmatch', 'in', 'endswith', 'startswith', 'glob', 'icontains', 'contains', '>', '<', '>=', '<=', '!=', '==', '=', 'exists'. Exiting.
```

It is because there's no condition in the variables contained `is_container_probe` , I assume you wanted to do something like:

```
- macro: is_container_probe
  condition: >
    (proc.is_container_healthcheck=true or
    proc.is_container_liveness_probe=true or
    proc.is_container_readiness_probe=true)

```


It would be good if you can rebase this PR with master (there had been a lot of changes since you first submitted it).

Also I'm not able to reproduce the warning triggered for liveness probes in my environment, I'm using this pod definition:

```
apiVersion: v1
kind: Pod
metadata:
  labels:
    test: liveness
  name: liveness-exec
spec:
  containers:
  - name: liveness
    image: k8s.gcr.io/busybox
    args:
    - /bin/sh
    - -c
    - touch /tmp/healthy; sleep 30; rm -rf /tmp/healthy; sleep 600
    livenessProbe:
      exec:
        command:
        - cat
        - /tmp/healthy
      initialDelaySeconds: 5
      periodSeconds: 5
```


It's comparable to what you want to do?

Thank you

fntlnz:
@agilgur5 are you still working on this? Let me know if you need help!

agilgur5:
@fntlnz just updated everything. The issue and PR hadn't been responded to in a while, so apologies that this got backlogged.

> It is because there's no condition in the variables contained `is_container_probe` , I assume you wanted to do something like:

Yep, my bad, that is indeed what I wanted to do, updated.

> It would be good if you can rebase this PR with master (there had been a lot of changes since you first submitted it).

Sure, there's no conflicts, but I guess that makes compiling/testing easier if it's rebased already. Done.

agilgur5:
> Also I'm not able to reproduce the warning triggered for liveness probes in my environment, I'm using this pod definition:
> ....
> It's comparable to what you want to do?

That does in fact look similar to something that triggered it. It was triggered by basically every liveness exec probe.
I'm not currently using Falco and don't have access to the code or environment that was triggering alerts, so I can't give an example myself. Perhaps we can get some folks from the issue thread who upvoted or commented to give some reproducible examples: @JPLachance , @pms1969 could either of you provide an example probe that triggers the alert?

JPLachance:
Hello @agilgur5, @fntlnz,

I have many pods in my environment with liveness and readiness probe an I don't see a clear link between those probes and the `Non sudo setuid` events I reported in #310.

Like I wrote in the [issue](https://github.com/falcosecurity/falco/issues/310#issuecomment-468000618), I have many event like this one:
``` 
{
  "container.id": "host",
  "evt.arg.uid": "root",
  "evt.time": 1551294467843706541,
  "k8s.ns.name": null,
  "k8s.pod.name": null,
  "proc.cmdline": "<NA>",
  "proc.pname": null,
  "user.name": null,
  "user.uid": 4294967295
}
```

With a little hindsight, I think such events are a sign of Falco struggling to fetch all the info he need to process events (most fields are `<NA>` or `null`).

I don't know how to make a liveness/readiness probe trigger the `Non sudo setuid` rule, sorry 🙁

fntlnz:
we still want to investigate it stalebot!

fntlnz:
Ok I lost a bit track on this @agilgur5 - my notifications didn't work as I expected - I'll do another review soon! Thanks for the effort!

ianhundere:
Checking in to see if this has been resolved or not.

agilgur5:
@ianhundere do you have a reproduction of this issue? This PR was stalled more generically but it also was not reproduced as per the comments above (and I'm a few years past using Falco). I would recommend reading through comments before making your own and notifying folks on the thread.

ianhundere:
We're just getting varies `Notice` alerts due to our RabbitMQ firing healthchecks. Was looking to see if there was a way to ignore said probes. No worries though, thanks for the heads up.

agilgur5:
If you can give a reproduction here or in #310 , that would certainly help with moving either forward. I don't have one anymore, but did experience `Non sudo setuid` alerts on probes some time ago, as per my comments.

PR #715, opened 2019-07-09
Allow regex chars rule names

mstemm:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind flaky-test

> /kind rule-update

> /kind rule-create

/area engine

> /area rules

> /area deployment

> /area integrations

> /area examples

Escape regex characters in rule names before passing them to the falco engine.

Fixes #

```release-note
Escape regex characters in rule names before passing them to the falco engine.
```

### DIFF
--- a/test/falco_tests.yaml
+++ b/test/falco_tests.yaml
@@ -456,13 +456,13 @@ trace_files: !mux
       - open_from_cat
     trace_file: trace_files/cat_write.scap
 
-  disabled_rules_using_regex:
+  disabled_rules_using_substring:
     detect: False
     rules_file:
       - rules/empty_rules.yaml
       - rules/single_rule.yaml
     disabled_rules:
-      - "open.*"
+      - "open_from"
     trace_file: trace_files/cat_write.scap
 
   disabled_rules_using_enabled_flag:
--- a/test/falco_tests.yaml
+++ b/test/falco_tests.yaml
@@ -86,6 +86,15 @@ trace_files: !mux
       - rules/rule_names_with_spaces.yaml
     trace_file: trace_files/cat_write.scap
 
+  rule_names_with_regex_chars:
+    detect: True
+    detect_level: WARNING
+    rules_file:
+      - rules/rule_names_with_regex_chars.yaml
+    detect_counts:
+      - 'Open From Cat ($\.*+?()[]{}|^)': 8
+    trace_file: trace_files/cat_write.scap
+
   multiple_rules_first_empty:
     detect: True
     detect_level: WARNING
--- a/test/rules/rule_names_with_regex_chars.yaml
+++ b/test/rules/rule_names_with_regex_chars.yaml
@@ -1,0 +1,25 @@
+#
+# Copyright (C) 2016-2018 Draios Inc dba Sysdig.
+#
+# This file is part of falco.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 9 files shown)


test/falco_tests.yaml
```diff
@@ -86,6 +86,15 @@ trace_files: !mux
       - rules/rule_names_with_spaces.yaml
     trace_file: trace_files/cat_write.scap
 
+  rule_names_with_regex_chars:
+    detect: True
+    detect_level: WARNING
+    rules_file:
+      - rules/rule_names_with_regex_chars.yaml
+    detect_counts:
+      - 'Open From Cat (^$\.*+?()[]{}|)': 8
```
leodido:
```suggestion
      - 'Open From Cat ($\.*+?()[]{}|^)': 8
```

What do you think of using the same order the chars are in the regex?

test/rules/rule_names_with_regex_chars.yaml
```diff
@@ -0,0 +1,25 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+- macro: is_cat
+  condition: proc.name=cat
+
+- rule: Open From Cat (^$\.*+?()[]{}|)
```
leodido:
```suggestion
- rule: Open From Cat ($\.*+?()[]{}|^)
```

userspace/engine/rules.cpp
```diff
@@ -195,7 +197,22 @@ int falco_rules::enable_rule(lua_State *ls)
...
 	bool enabled = (lua_tonumber(ls, -1) ? true : false);
 
-	rules->enable_rule(rule, enabled);
+	// Escape any regex special characters in the rule name
+	std::string sanitized;
+
+	try {
+		std::regex escape_chars { R"([$\.*+?()[\]{}|^])" };
+
+		sanitized = std::regex_replace(rule, escape_chars, R"(\$&)" );
```
leodido:
```suggestion
		sanitized = std::regex_replace(rule, escape_chars, "$&");
```

test/falco_tests.yaml
mstemm: Sure, I reordered it originally when using a regex. Now that I'm using find_first_of() I'm using the same order for tests and code.

test/rules/rule_names_with_regex_chars.yaml
mstemm: I went the other way, changing the code, but did make them consistent.

userspace/engine/rules.cpp
mstemm: Thanks, moot now that I switched to find_first_of.

mstemm:
Strangely, there's something where the regression tests pass on my machine, but not when run in the falco-tester container. I'm investigating.

mstemm:
The problem was that the regex_match call at https://github.com/falcosecurity/falco/blob/dev/userspace/engine/ruleset.cpp#L228 doesn't actually work with gcc 4.8.2. This is noted on the c++11 compatibility notes for gcc 4.8.2: https://gcc.gnu.org/onlinedocs/gcc-4.8.2/libstdc++/manual/manual/status.html#status.iso.2011.

We can fix this by either upgrading compilers to a newer gcc version, which would also probably mean switching to a different base system for the falco-builder image, or by changing the semantics of `void enable_rule(const std::string &pattern, bool enabled, const std::string &ruleset);` to be a substring match instead of a regex match. I think I'd lean towards upgrading gcc versions.

mstemm:
Actually, I'm going to change the semantics to be a substring match. The only consumer of this api was internal so it's ok to make that change.

mstemm:
I'm going to close this PR and open a new one that is specifically targeted towards replacing regex patterns with substrings in enable_rule().

mstemm:
Closed in favor of https://github.com/falcosecurity/falco/pull/743.

PR #759, opened 2019-08-01
Fix webserver errors

leodido:
**What type of PR is this?**

/kind bug

NONE

Fixes #

I'd like to add (me or someone else) unit tests to this PR too. Anyway it could be merged also just as is.

```release-note
Improvements to how the webserver handles incoming invalid inputs
```

### DIFF
--- a/userspace/engine/falco_engine.cpp
+++ b/userspace/engine/falco_engine.cpp
@@ -365,46 +365,54 @@ unique_ptr<falco_engine::rule_result> falco_engine::process_k8s_audit_event(json
 
 bool falco_engine::parse_k8s_audit_json(nlohmann::json &j, std::list<json_event> &evts)
 {
-	// If the Kind is EventList, split it into individual events.
-	if(j.value("kind", "<NA>") == "EventList")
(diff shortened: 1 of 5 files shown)


userspace/engine/falco_engine.cpp
```diff
@@ -365,46 +365,54 @@ unique_ptr<falco_engine::rule_result> falco_engine::process_k8s_audit_event(json
...
 
+				std::string tmp;
+				sinsp_utils::ts_to_string(ns, &tmp, false, true);
+
+				evts.back().set_jevt(je, ns);
+			}
+
+			return true;
+		}
+		else if(j.value("kind", "<NA>") == "Event")
```
Kaizhe: Is the input always `EventList` I mean, always come as a list of values. If that's the case, this block seems not necessary

leodido: Not sure about this. Imho the way to go is unit testing it so we can be sure (to some degree). Infact I setup the unit test env also for `userspace/falco` in this PR.

PR #762, opened 2019-08-05
fix: make sure that when deleting shell history the system call is taken into account

fntlnz:

**What type of PR is this?**

/kind bug
/kind rule-update

/area rules

When deleting a shell history file we can't rely on the name of the command deleting it.
This PR uses the rmdir, unlink and unlinkat syscalls instead to reliably detect if an history file had been deleted.

NONE

```release-note
rules: Delete Bash History is renamed to Delete or rename shell history
rules, Delete or rename shell history: When deleting a shell history file now the syscalls are taken into account rather than just the commands deleting the files
rules, Delete or rename shell history: history deletion now supports fish and zsh in addition to bash
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -2381,13 +2381,29 @@
     WARNING
   tags: [process, mitre_persistence]
 
-- rule: Delete Bash History
-  desc: Detect bash history deletion
+- rule: Delete or rename shell history
+  desc: Detect shell history deletion
   condition: >
-    ((spawned_process and proc.name in (shred, rm, mv) and proc.args contains "bash_history") or
-     (open_write and fd.name contains "bash_history" and evt.arg.flags contains "O_TRUNC"))
+    (modify and (
+      evt.arg.name contains "bash_history" or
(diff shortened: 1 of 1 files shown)


rules/falco_rules.yaml
```diff
@@ -2381,13 +2381,15 @@
...
   tags: [process, mitre_persistence]
 
-- rule: Delete Bash History
-  desc: Detect bash history deletion
+- rule: Delete or rename shell history
+  desc: Detect shell history deletion
   condition: >
-    ((spawned_process and proc.name in (shred, rm, mv) and proc.args contains "bash_history") or
-     (open_write and fd.name contains "bash_history" and evt.arg.flags contains "O_TRUNC"))
+    modify and evt.arg[1] contains "bash_history" or
```
Kaizhe: the `modify` works fine for the `mv` or `rm` scenarios, however, there is another scenario is not covered: write with `O_TRUNC` flag.

fntlnz: You mean open followed by a write with `O_TRUNC` right?

PR #765, opened 2019-08-13
fix(rules/Set Setuid or Setgid bit): use chmod syscalls instead of chmod command

fntlnz:

**What type of PR is this?**

/kind bug

/kind rule-update

/area rules

The "Set Setuid or Setgid bit" rule is using a check on the invocation of the `chmod` command in order to check if a suid or sgid flag is applied to a file.
That method is ineffective because anyone using the raw syscall or glibc can bypass that.

For example:
```bash
#include <sys/stat.h>
#include <fcntl.h>

int main ()
{
  fchmodat(AT_FDCWD, "/tmp/yo", S_ISUID, 0);
}
```

This PR uses the newly added `fchmod, chmod, fchmodat` syscalls (to sysdig)
in order to avoid that.

NONE

Only works if sysdig PR https://github.com/draios/sysdig/pull/1472 is merged. It adds support to the chmod syscalls we are using here.

```release-note
rule(Set Setuid or Setgid bit): syscalls are used to detect suid and sgid
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -72,6 +72,9 @@
 - macro: create_symlink
   condition: evt.type in (symlink, symlinkat) and evt.dir=<
 
+- macro: chmod
+  condition: evt.type in (chmod, fchmod, fchmodat)
+
 # File categories
 - macro: bin_dir
   condition: fd.directory in (/bin, /sbin, /usr/bin, /usr/sbin)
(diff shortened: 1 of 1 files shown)


fntlnz:
@Kaizhe I noticed that the previous rule used a macro `consider_all_chmods` to disable it. I added the same macro I'm just unsure if we still need it. Can you confirm why we need it? Thanks

mstemm:
I would also increment the required engine version, as you depend on the changes in https://github.com/draios/sysdig/pull/1472 to be effective. Once the sysdig PR is merged you’ll have to increment the falco engine and checksum anyway.

Kaizhe:
@fntlnz good changes, couple requests here:
1. Post the rule triggered output here to make sure the rule still works as expected.
2. I would like to to turn it on, so replace `never_true` to `always_true`
3. Add a customized macro `user_known_chmod_applications`

fntlnz:
Here is the outputs @Kaizhe 

## Rule output examples


### chmod
Syscall

```
chmod("/tmp/test", 04000)
```
Output

```
19:30:43.073551187: Notice Setuid or setgid bit is set via chmod (fd=<NA> filename=/tmp/test mode=S_ISUID user=ubuntu command=a.out container_id=host container_name=host image=<NA>:<NA>)
```

### fchmodat

Syscall

```
fchmodat(AT_FDCWD, "/tmp/ciao", 04000)
```

Output

```
19:31:04.851600216: Notice Setuid or setgid bit is set via chmod (fd=<NA> filename=/tmp/ciao mode=S_ISUID user=ubuntu command=a.out container_id=host container_name=host image=<NA>:<NA>)
```

### fchmod

```
openat(AT_FDCWD, "/tmp/test", O_RDWR)   = 3
fchmod(3, 04000)
```

Syscall

```
19:31:52.570751037: Notice Setuid or setgid bit is set via chmod (fd=<f>/tmp/test filename=<NA> mode=S_ISUID user=ubuntu command=a.out container_id=host container_name=host image=<NA>:<NA>)
```

Issue #818, opened 2019-09-06
The Falco (gRPC) API


kris-nova:
**What do we mean by "API"*?

The language here implies a documented way of interfacing with Falco, regardless of the programming language. 

This is **NOT** a traditional C/C++ API that we would see in a `.h` file like we would see in `linux` or other similar applications. 

This is more of a high-level spec or contract (probably implemented via gRPC) that users can count on using, while interacting with the `falco` core engine. This is similar to a `REST` API. 

The goals would be:

 - Any language can consume this via a well-known protocol like `gRPC`
 - We version this "API" over time, as we make breaking and non-backwards compatible changes
 - This is more "documentation" than "implementation" (think inputs/outputs) 

**What would you like to be added**:

The concept of an officially supported `falco` API spec/documentation via a protocol. 

This implies **alot**. Let's look at the `envoy` project for a quick example. The entire application is configurable via an API. You can see some documentation [here](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api). The program is also written in C++ and is commonly vendored throughout the cloud native ecosystem. We should look at enabling our users in the same way.

**Why is this needed**:

So that non-C++ programers can easily interface, control, and take advantage of the `falco` system. This would enable a powerful (and extremely fast) eventing system built on `falco` using the `sysdig` libraries `libsinsp` and `libscap`. 

So that users who wish to take advantage of the `falco` OSS tool can begin to build out their own inputs/outputs/rules for `falco` (either via `falcoctl`* or via their own device)

So that we can begin to define the boundaries for `falco` and what the engine is and isn't responsible for. 

**Notable references for this**: 

 #811 
 #789 
 #785 
 #528 
 #529 
 #531

leodido:
First gRPC server impl done in #822

Issue #811, opened 2019-09-01
falcoctl


kris-nova:
**What would you like to be added**:

A new CLI tool called `falcoctl`.

This would be a new repository.

Here is a [proof of concept](https://github.com/kris-nova/falcoctl/tree/init)

**Why is this needed**:

Think of `falcoctl` as a `.go` program that is statically linked so that 

> operators can enjoy Falco

Issif:
I totally agree. If it's OK for others, I propose to migrate your POC under `falcosecurity/falcoctl` and move forward. It makes more sense either if we think about #809.

mfdii:
It should be called `falconer` as it’s purpose is to tame and control Falco.

kris-nova:
Can we name it `falconer` and pronounce it as `falcon E - R` in honor of `kube C - T - L`?

leodido:
Agree ! 🦅 🦅 🦅

kris-nova:
Notes from slack:

```
1 - How do we install the Kubernetes resources needed for Falco? Do we have a new operator? Do we put this logic in the CLI tool? Do we parse YAML dynamically from local disk or URL? Obviously we need to allow folks to change bits in their specific Falco install (for instance kernel module vs eBPF probe) 

2 - What are we going to do with rules/outputs and using those as composable building blocks for Falco? How will we store them? How will we maintain them?

3 - What on earth is going to take responsibility of reconfiguring and restarting the API server. This is a bold task. 

4. Vendoring in Go still sucks, this isn't new

5. We shouldn't use init()


```

kris-nova:
Another thought

```
 6. How do we handle `falconer`/`falcoctl` interacting with different versions of Kubernetes?
```

### Version flags

We could have a flag to specify which version of Kubernetes to use (this might be a dependency nightmare) 

Example

```
falconer --k8s=1.10.8 install falco # Installs this concrete version
falconer --k8s=1.15 install falco # Installs the most recent 1.15.* version
```

### Release different binaries 

This would allow us some freedom of building in custom logic for each release (relevant to #809) as we push this to GA.

We could branch our code for various versions of Kubernetes, but again, this would be a pain to keep up with. Maybe we could only support Kubernetes LTS? 

```
falco1.14 install falco
falco 1.13 install falco
```

Issif:
[...]

PR #822, opened 2019-09-06
gRPC server and gRPC outputs service

### LINKED ISSUE #528: ROADMAP: Improved Falco Outputs
**Description:** The goal behind this idea is to improve the available options for sending alerts from Falco when a security violation occurs inside a container. Currently outputs are limited to stdout, files, syslog, and executing a program. We’d like to offer more output options such as: NATS.io, Kafka, gRPC, Google Pub/Sub, AWS SNS, HTTPs Webhooks, etc.
Also, currently, Falco only supports a single program output. We want it to support multiple program_outputs instead.
**Recommended Skills:** C/C++ experience, working with external libraries, working knowledge of message queues and modern Pub/Sub systems.
**Mentor(s):** Mark Stemm (@mstemm), Loris Degioanni (@ldegio), Michael Ducy (@mfdii)

### LINKED ISSUE #791: Review the Falco signal handler

I noticed that on my system sending a `SIGINT` to the Falco process took a few seconds for our signal handler to respond. This isn't a deal breaker because we have 30 seconds by default in Kubernetes to respond. But we should be more responsive with signals passed to Falco. 

**Why is this needed**:

Output from a `SIGINT` signal sent to Falco

```
[nova nova]# time falco
Tue Aug 27 07:22:38 2019: Falco initialized with configuration file /home/nova/go/src/github.com/falcosecurity/falco/falco.yaml
Tue Aug 27 07:22:38 2019: Loading rules from file /etc/falco/falco_rules.yaml:
Tue Aug 27 07:22:38 2019: Loading rules from file /etc/falco/falco_rules.local.yaml:
Tue Aug 27 07:22:38 2019: Loading rules from file /etc/falco/k8s_audit_rules.yaml:
Tue Aug 27 07:22:38 2019: Loading rules from file /etc/falco/rules.d/falco_rules.local.yaml:
Tue Aug 27 07:22:39 2019: Loading rules from file /etc/falco/rules.d/falco_rules.yaml:
Tue Aug 27 07:22:39 2019: Starting internal webserver, listening on port 8765
^CTue Aug 27 07:22:45 2019: SIGHUP Received, restarting...
Events detected: 0
Rule counts by severity:
Triggered rules by rule name:
   - num times actions taken: 0

real	0m7.135s
user	0m0.523s
sys	0m0.218s
```

*Note*: I executed the program, then immediately sent `ctl+c`

leodido:
**What type of PR is this?**

/kind feature

NONE

This PR adds a gRPC server to falco in its own thread.
Implementors can use it to create and extend the falco API.
The only implemented call by this PR is the subscribe call that can be used to subscribe to falco output events directly.

The code this PR contains refers to the initial proposal but implements a streaming server rather than a bidirectional client/server paradigm.

- Original proposal: #789 
- Fixes #528 
- Fixes #791 
- Refs to #818 (umbrella issue for various related topics)

This PR is still in **WIP**. Me and @fntlnz are still working on it.

/assign @leodido 
/assign @fntlnz 

Work we still need to do:

- [x] Implement abstract structure to handle various kind of gRPC services
- [x] Implement streaming service - ie., `subscribe`
- [x] Get the outputs through `handle_event` in falco_outputs.cpp
- [x] Separate the sending of events from the sending of messages in Lua too
- [x] Proto definitions
- [x] Mutual TLS authentication
- [x] Make it work for k8s_audit
- [x] Source is not showed in the outputs, investigate why
- [x] Make sure all the threads have their cleaning logic (signal handlers etc..)
- [x] Make all the parameters configurable
    - [x] certificates (private key, root cert etc.)
	- [x] the grpc server
        - [x] enable
        - [x] threadiness
        - [x] bind address + bind port
    - [x] the grpc output itself
- [x] Do some name and file refactoring, right now everything is in grpc_server.cpp and grpc_server.h, we can do better
- [x] Go SDK
    - [x] Write the Go library
    - [x] Usage examples

Next items for which we need to open issues:
- [ ] Unit tests
- [ ] Integration tests (the same way are done for other features?)
- [ ] Add tags in the `response` struct for `falco_outputs.proto` (this also requires a client change) 
- [ ] Catch gRPC status and logs (?) (for example when input certificates are wrong it segfaults, handle it)
- [ ] Documentation
    - [ ] Falco gRPC server configuration and usage
    - [ ] Falco gRPC outputs configuration and usage
    - [ ] gRPC API documentation
    - [ ] How to generate the certificates for mutual TLS
    - [ ] SDK examples
- [ ] Implement other output queuing mechanism, right now we only support a shared queue in a round robin fashion

Some notes about this implementation:

- ~We feel like it is not PoC quality even if it's a PoC, needs some refactoring but overall we think the design is solid enough. It's even thread safe thanks to TBB;~ Lots of things happened since this.
- When multiple clients Subscribe to the outputs, messages are dispatched in a  Round Robin  fashion. Single queue, multiple consumers - We can have other strategies here, like a queue for every consumer (fanout) or even more sophisticated things. This first approach seems good enough now, all future approaches will be "configurable" so, we are open to different strategies but this is the default one for now

```release-note
Falco gRPC api server implementation, contains a subscribe method to subscribe to outputs from any gRPC capable language
```


Issif:
I've several concerns about that gRPC output :

- A simple queue with several consumers (messages are distributed in RRD mod for avoiding duplicates) is great. Even it's this queue is used, we still need a queue which gathers all messages, we talked about idea of `falcoctl` that can consumes events by gRPC, it needs to get all messages, not 1 over N.
- On a k8s cluster, we have several falco instances, and they can come/go with node creations/deletions. How our consumers can be aware of new falcos to be able to connect asap?
- We can think about a "sidecard container" (`falcosidekick` for ex.), running in same POD than falco and make it connecting internally on loopback. It should work, we only need to accept to have several containers more. It's only possible with internal consumers of course.
- We could also imagine a mechanism where all falcos are declaring to each others (a kind of cluster). When a consumer is connected to a falco gRPC, it retrieves list of other instances (and updates after).  Broadcast/multicast are often unavailable, we might need a registration somewhere (etcd, consul, etc).

Issif:
I've several concerns about that gRPC output :

- A simple queue with several consumers (messages are distributed in RRD mod for avoiding duplicates) is great. Even it's this queue is used, we still need a queue which gathers all messages, we talked about idea of `falcoctl` that can consumes events by gRPC, it needs to get all messages, not 1 over N.
- On a k8s cluster, we have several falco instances, and they can come/go with node creations/deletions. How our consumers can be aware of new falcos to be able to connect asap?
- We can think about a "sidecard container" (`falcosidekick` for ex.), running in same POD than falco and make it connecting internally on loopback. It should work, we only need to accept to have several containers more. It's only possible with internal consumers of course.
- We could also imagine a mechanism where all falcos are declaring to each others (a kind of cluster). When a consumer is connected to a falco gRPC, it retrieves list of other instances (and updates after).  Broadcast/multicast are often unavailable, we might need a registration somewhere (etcd, consul, etc).

kris-nova:
I think we want to let this PR through and assign the Go SDK to me? Can we push our existing Go code through a new repo like #811?

fntlnz:
@Issif I think all your concerns are all resolved because:

1. The single queue mechanism is the "Round robin" (shared queue between consumers) mechanism, this feature is still ongoing, as we wrote in the notes, we need an alternative implementation that does Fan-out (one queue per consumer).
2. That is a client side mechanism that must be implemented in the SDKs, not a concern for this implementation, it's how client server works. See also point 4 of this list, with that mechanism we can avoid that problem by simply not connecting directly to the single falco instance.
3. That can be done as a different proposal, here we are implementing the server, this conversation must be 100% on that, while I agree with the intent I want to keep this very light and focused on what we are doing, so if you have an idea I would be happy to read it in a proposal and review a PoC of it. Please do it! :grinning: 
4. On an API point of view (not just outputs), we are filling a proposal to do a gRPC server that can aggregate all the Falco instances APIs so that we can call actions on multiple servers using a thing like the falcoctl without connecting to all of them. IMHO Falco instances don't need and don't want (it's a security software) to talk each other, also this is not a primary-secondary architecture. Here is the issue we opened to fill the proposal #824

Issif:
@fntlnz you made my day (at least my morning 😉 ). A big 👍 for the whole.

leodido:
With the commits pushed today we are really near to the completion of a first minimum viable gRPC server for falco outputs.

userspace/engine/formats.cpp
```diff
@@ -265,3 +266,34 @@ int falco_formats::format_event (lua_State *ls)
...
+
+int falco_formats::resolve_tokens(lua_State *ls)
+{
+	if (!lua_isstring(ls, -1) ||
+	    !lua_isstring(ls, -2)) {
+		lua_pushstring(ls, "Invalid arguments passed to resolve_tokens()");
+		lua_error(ls);
+	}
+	gen_event* evt = (gen_event*)lua_topointer(ls, 1);
+	const char *source = (char *) lua_tostring(ls, 2); // TODO(fntlnz, leodido): do we need this one?
```
leodido: We don't. Remove

userspace/engine/formats.cpp
```diff
@@ -265,3 +266,34 @@ int falco_formats::format_event (lua_State *ls)
...
+
+
+int falco_formats::resolve_tokens(lua_State *ls)
+{
+	if (!lua_isstring(ls, -1) ||
+	    !lua_isstring(ls, -2)) {
+		lua_pushstring(ls, "Invalid arguments passed to resolve_tokens()");
+		lua_error(ls);
+	}
+	gen_event* evt = (gen_event*)lua_topointer(ls, 1);
```
leodido: Damn spaces lol

userspace/falco/configuration.cpp
```diff
@@ -148,7 +148,23 @@ void falco_configuration::init(string conf_filename, list<string> &cmdline_optio
 		m_outputs.push_back(http_output);
 	}
 
-	if (m_outputs.size() == 0)
+	m_grpc_enabled = m_config->get_scalar<bool>("grpc", "enabled", false);
+	m_grpc_bind_address = m_config->get_scalar<string>("grpc", "bind_address", "0.0.0.0:5060");
+	m_grpc_threadiness = m_config->get_scalar<uint32_t>("grpc", "threadiness", 8); // todo > limit it to avoid overshubscription? std::thread::hardware_concurrency()
```
leodido: Investigate threads oversubscription

userspace/falco/falco.cpp
```diff
@@ -34,6 +34,7 @@ limitations under the License.
 #include <unistd.h>
 #include <getopt.h>
 
+
```
leodido: Remove this newline

userspace/engine/formats.cpp
```diff
@@ -265,3 +266,34 @@ int falco_formats::format_event (lua_State *ls)
...
+	const char *source = (char *) lua_tostring(ls, 2); // TODO(fntlnz, leodido): do we need this one?
+	const char *format = (char *) lua_tostring(ls, 3);
+	string sformat = format;
+
```
leodido: Remove this new line

userspace/falco/falco.cpp
```diff
@@ -269,7 +272,12 @@ uint64_t do_inspect(falco_engine *engine,
...
 
-		if (g_terminate || g_restart)
+		if(g_terminate)
+		{
+			falco_logger::log(LOG_INFO, "SIGINT received, exiting...\n");
+			break;
+		}
+		else if (g_restart)
 		{
 			falco_logger::log(LOG_INFO, "SIGHUP Received, restarting...\n");
```
leodido: lowercase "Received"

userspace/falco/falco.cpp
```diff
@@ -1213,6 +1239,10 @@ int falco_init(int argc, char **argv)
 		result = EXIT_FAILURE;
 
 		webserver.stop();
+		if (grpc_server_thread.joinable()) {
```
leodido: Format this piece of code

userspace/falco/falco_output.proto
```diff
@@ -0,0 +1,64 @@
...
+message response {
+  google.protobuf.Timestamp time = 1;
+  priority priority = 2;
+  source source = 3;
+  string rule = 4;
+  string output = 5;
+  map<string, string> output_fields = 6;
+}
+
+// todo: investigate outputting tags
```
leodido: We did this. Remove

userspace/falco/falco_outputs.cpp
```diff
@@ -298,3 +302,65 @@ int falco_outputs::handle_http(lua_State *ls)
 	}
 	return 1;
 }
+
+// TODO(fntlnz, leodido): verify if this works with k8s_audit as source
```
leodido: Remove this todo. It's already written in the PR checklist.

userspace/falco/grpc_server.cpp
```diff
@@ -0,0 +1,271 @@
...
+
+		// Complete the processing
+		response res;
+		(srv->*m_process_func)(*m_stream_ctx, m_req, res); // subscribe()
+	}
+	else
+	{
+		// Handle the edge case when `m_request_func` event failed
+		// which means `m_stream_ctx` was not set
+		// todo > log error
```
leodido: We need a thread safe logger ...

userspace/falco/grpc_server_impl.h
```diff
@@ -0,0 +1,46 @@
...
+	~falco_grpc_server_impl() = default;
+
+	void shutdown();
+
+protected:
+	bool is_running();
+
+	void subscribe(const stream_context& ctx, const request& req, response& res);
+
+
```
leodido: Remove whitelines

fntlnz:
@leodido I'm not sure this fixes #791 - I'm keeping sending SIGTERM and it logs SIGINT. Need to re-check this.

fntlnz:
This does not work with k8s_audit yet, making it happen!

fntlnz:
This is the script we are using right now to generate the MTLS certificates.
This will need to be included in the documentation and optionally we can make a falcoctl command to generate those.

```bash
#!/bin/bash

# Generate valid CA
openssl genrsa -passout pass:1234 -des3 -out ca.key 4096
openssl req -passin pass:1234 -new -x509 -days 365 -key ca.key -out ca.crt -subj  "/C=SP/ST=Italy/L=Ornavasso/O=Test/OU=Test/CN=Root CA"

# Generate valid Server Key/Cert
openssl genrsa -passout pass:1234 -des3 -out server.key 4096
openssl req -passin pass:1234 -new -key server.key -out server.csr -subj  "/C=SP/ST=Italy/L=Ornavasso/O=Test/OU=Server/CN=localhost"
openssl x509 -req -passin pass:1234 -days 365 -in server.csr -CA ca.crt -CAkey ca.key -set_serial 01 -out server.crt

# Remove passphrase from the Server Key
openssl rsa -passin pass:1234 -in server.key -out server.key

# Generate valid Client Key/Cert
openssl genrsa -passout pass:1234 -des3 -out client.key 4096
openssl req -passin pass:1234 -new -key client.key -out client.csr -subj  "/C=SP/ST=Italy/L=Ornavasso/O=Test/OU=Client/CN=localhost"
openssl x509 -passin pass:1234 -req -days 365 -in client.csr -CA ca.crt -CAkey ca.key -set_serial 01 -out client.crt

# Remove passphrase from Client Key
openssl rsa -passin pass:1234 -in client.key -out client.key
```

fntlnz:
Ok this is almost ready, the only blocker we have right now is that `falco_formats::resolve_tokens` does not work for `k8s_audit` events when called by lua in `mod.grpc`. Investigating that.

falco.yaml
```diff
@@ -166,4 +166,15 @@ program_output:
...
+
+grpc:
+  enabled: false
+  bind_address: "0.0.0.0:5060"
+  threadiness: 8
+  private_key: "/etc/falco/certs/server.key"
+  cert_chain: "/etc/falco/certs/server.crt"
+  root_certs: "/etc/falco/certs/ca.crt"
+
+grpc_output:
```
kris-nova: Was this a design choice? Why do we have `grpc_output` separate from `grpc`?

userspace/falco/configuration.cpp
kris-nova: I think we should parse this dynamically from `/proc`?

userspace/falco/configuration.cpp
```diff
@@ -148,7 +148,23 @@ void falco_configuration::init(string conf_filename, list<string> &cmdline_optio
...
+	// todo(fntlnz,leodido) > chose correct paths
+	m_grpc_private_key = m_config->get_scalar<string>("grpc", "private_key", "/etc/falco/certs/server.key");
+	m_grpc_cert_chain = m_config->get_scalar<string>("grpc", "cert_chain", "/etc/falco/certs/server.crt");
+	m_grpc_root_certs = m_config->get_scalar<string>("grpc", "root_certs", "/etc/falco/certs/ca.crt");
```
kris-nova: I think any hard coded strings should be `--flags`?

userspace/falco/grpc_context.h
```diff
@@ -0,0 +1,62 @@
...
+#ifdef GRPC_INCLUDE_IS_GRPCPP
+#include <grpcpp/grpcpp.h>
+#else
+#include <grpc++/grpc++.h>
+#endif
+
+const std::string meta_session = "session_id";
+const std::string meta_request = "request_id";
+
+class context
```
kris-nova: If this new class is only for `gRPC` then when aren't we labeling it as such?

```
class grpc_context
```

userspace/falco/grpc_server.cpp
```diff
@@ -0,0 +1,253 @@
...
+#include <grpcpp/grpcpp.h>
+#else
+#include <grpc++/grpc++.h>
+#endif
+
+#include "logger.h"
+#include "grpc_server.h"
+#include "grpc_context.h"
+
+template<>
```
kris-nova: Good job on the polymorphism - strong +1

falco.yaml
leodido: Yes Kris.

They are separated because we want the Falco Outputs feature to be separated from future gRPC features. In order to be able to selectively enable one thing or the other.

userspace/falco/configuration.cpp
leodido: I think that they are not exactly the same thing. There are various things we should consider reading this info from `/proc` - eg., hyperthreading enabled or not, processors turned off.

Thus, at maximum I would consider using `/proc/cpuinfo` processors count only as fallback when `hardware_concurrency()` returns 0 (which means not implemented).

userspace/falco/configuration.cpp
leodido: These are defaults that (should) match the default values we have in the `falco.yaml` config file.

userspace/falco/grpc_context.h
leodido: You're right Kris :)

But we'd like to use namespaces for this! WDYT?

fntlnz:
Ok after today's changes this PR is ready! :tada: I'm opening issues and an umbrella issue for the remaining work we need to do for documentation and tooling

userspace/falco/grpc_context.h
kris-nova: I love namespaces!

leodido: Done!

userspace/falco/output.proto
```diff
@@ -0,0 +1,28 @@
...
+
+import "google/protobuf/timestamp.proto";
+import "schema.proto";
+
+package falco.output;
+
+option go_package = "github.com/falcosecurity/client-go/pkg/api/output";
+
+service service {
+  rpc subscribe(request) returns (stream response);
```
mstemm: These names are pretty generic, can you change them to something like subscribe_alerts so it's clear that the client is subscribing to all falco alerts, and not other streams of info?

userspace/falco/lua/output.lua
```diff
@@ -152,70 +169,86 @@ end
 function mod.http_reopen()
 end
 
+function mod.grpc(event, rule, source, priority, priority_num, msg, format, options)
```
mstemm: I think it's fine to do it this way, but I always thought routing events through lua just so they could be written to a log file/stdout/back to c for http output was a bit awkward. Did you consider having grpc outputs entirely separate from lua and using the json_event_formatter directly?

userspace/falco/grpc_server.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+	m_completion_queue = builder.AddCompletionQueue();
+	m_server = builder.BuildAndStart();
+	falco_logger::log(LOG_INFO, "Starting gRPC server at " + m_server_addr + "\n");
+
+	// The number of contexts is multiple of the number of threads
+	// This defines the number of simultaneous completion queue requests of the same type (service::AsyncService::Request##RPC)
+	// For this approach to be sufficient server::IMPL have to be fast
+	int context_num = m_threadiness * 10;
+	REGISTER_STREAM(output::request, output::response, subscribe, subscribe, context_num)
```
mstemm: If we want to add a second grpc method (like I want to do for coverting k8s psps into falco rules files), how will this work with REGISTER_STREAM and the contexts?

userspace/falco/output.proto
leodido: Yes but they are namespaced. Thus being scoped to a namespace they at the end result in pretty significant name imho, eg., `falco::output::service`.

userspace/falco/grpc_server.cpp
leodido: It depends on how your proto looks like. You can use `REGISTER_STREAM()` if you have a stream or you will have to make the right macro for different use cases.

fntlnz: @mstemm once this is merged we will add a macro to process unary requests as well.
We have in plan to do a top level `version` service to get the falco server version from grpc.

Doesn't make sense for us to add that in this PR, but will be there for when you have to add the psp stuff!

userspace/falco/lua/output.lua
fntlnz: I know, the purpose of this was to keep things as simple as possible as first implementation. We'd like to do that as a refactoring.

fntlnz:
@mstemm I agree that this is lacking tests, at the planning meetings we agreed to do them in a subsequent iteration over this code to let everyone start integrating with this as soon as possible while we work. It's in the "Next action items" section in the PR description

Issue #824, opened 2019-09-09
To fill a proposal for a gRPC aggregation service


leodido:
This is a summary of an initial idea that me and @fntlnz want to fill as a proposal.

So this will be expanded. We are just dumping initial idea we had so that we can fill the proposal later.

**What would you like to be added**:

Assuming a k8s cluster having more Falcos deployed I'd like to have a mechanism able to connect and subscribe to all of them proxying all the events coming from them.
The goal is to have an external server which aggregates the events coming from the Falcos deployed.
Ideally this gRPC aggregator service can be made in Go.

An image is worth a thousand words.

![gRPC aggregator service for Falcos](https://user-images.githubusercontent.com/120051/64515397-d0e08280-d2ec-11e9-993b-b7f911352ecf.png)

This mechanism doesn't add any logic on the existing Falco gRPC API.

It just add a layer to query, discover, and instructing all of them using an unique Falco ID to discern the different Falco instances.

For example the subscribe call (the outputs gRPC API) will return the normal output plus an ID (gRPC metadata) telling to the consumer from which the output comes.

**Why is this needed**:

We need an aggregator able to relay all the events of various Falcos so that them acts like a single entity for external clients connecting to them.

- Falco instances can be very volatile (can go up and down). We want a system that has knowledge of all of them. This can be done for example using the Kubernetes discovery mechanism or in a non-Kubernetes system using a configuration file or multicast.
- When using a client (eg., `falcoctl`) one cannot connect to all of the Falco instances. Let's say that we have 100 Falcos, it's extremely hard to know what it's going on all of them.

/assign @leodido 
/assign @fntlnz

kris-nova:
Let's check out Envoy for this!

kris-nova:
We probably need to think about two things with this model:

# Order

I do NOT think we guarantee order what so ever with model. A client gets alerts as they trickle through.

# Falco meta information

We probably wan't to include details about the falco instance as it relates to Kubernetes. What node is this running on? What is the name of the pod? How long has falco been running.

leodido:
Yes Kris I agree with you.

About the metadata we have already created the mechanism to include such
info as gRPC metadata (in the current PR about gRPC streaming server).

Thanks for bringing this up!

fntlnz:
@kris-nova I agree for order, for the Falco meta information part that was covered in the issue by the sentence:

[...]

PR #908, opened 2019-10-30
proposal: Falco API

leodido:
**What type of PR is this?**

/kind design

/kind documentation

/area proposals

Writes down a proposal as the outcome of various discussions between me, @fntlnz, @kris-nova, and the whole community about giving Falco an API.

_will link to issues later_

**WIP**

In the coming days me and @fntlnz are going to complete it.

```release-note
NONE
```

### DIFF
--- a/proposals/20191030-api.md
+++ b/proposals/20191030-api.md
@@ -2,13 +2,13 @@
 
 ## Summary
 
-With this proposal we inted to present the intent of creating a Falco API.
+This is a proposal to better structure the Falco API.
 
-This proposal also intends to list the set of the gRPC services such API will expose.
+It will contain a list of gRPC services that can be added to have a first stable version of the API for v1.0.0.
 
 ## Motivation
 
-We want users to interface with Falco outputs, inputs, rules, and configurations via thirdy-part clients.
+We want to enable users to use thirdy-party clients to interface with Falco outputs, inputs, rules, and configurations.
 
 Such ability would enable the community to create a whole set of OSS tools, built on top of Falco.
 
@@ -35,7 +35,7 @@ Some examples, already in place, are:
 - Receive Falco events with a well-defined contract over wire --> **outputs**
 - Receive Falco drops with a well-defined contract over wire --> **drops**
 - Receive current Falco version and related meta information (commit hash, built type, etc.) --> **version**
-- Config Falco via API (CRUD) -> **configs**
+- Configure Falco via API (CRUD) -> **configs**
 - Inject and/or modify Falco rules via API (CRUD) --> **rules**
 - Send input events to Falco over wire --> **inputs**
 
@@ -130,7 +130,7 @@ This translates in having the following set of `proto` files.
 
 #### Benefits
 
-We think this structure makes the implementation more easy to write and to maintain because it automatically matches to abstractions - ie., one service by RPC type, a pool of contexts by RPC type - implemented in the code.
+We think this structure makes the implementation easier to write and to maintain because it automatically matches the abstractions - ie., one service by RPC type, a pool of contexts by RPC type - implemented in the code.
 
 #### Limitations
 
--- a/proposals/20191030-api.md
+++ b/proposals/20191030-api.md
@@ -25,23 +25,27 @@ Some examples, already in place, are:
 
 - To design and implement an API to make Falco more configurable
 - To design and implement an API to make Falco more consumable
+- To have an API so that users are able to interface with Falco
 - To design a set of simple contract interfaces
-- To have a secure by default Falco API
+- To have an API which is secure by default
+- To implement this API in C++ using gRPC as a first implementation
 
 ## Non-Goals
 
-- Replace the existing inputs: the **inputs** will be only for new source events users may want to add, initially
+- To replace the existing inputs: the **inputs** will be only for new source events users may want to add, initially
+- To document the API: to be approached as separate task
+- To create/update the API client: to be approached as a separate task
 
 ## Proposal
 
 ### Use cases
 
-- Receive Falco events with a well-defined contract over wire --> **outputs**
-- Receive Falco drops with a well-defined contract over wire --> **drops**
+- Receive Falco events with a well-defined contract --> **outputs**
+- Receive Falco drops with a well-defined contract --> **drops**
 - Receive current Falco version and related meta information (commit hash, built type, etc.) --> **version**
 - Configure Falco via API (CRUD) -> **configs**
 - Inject and/or modify Falco rules via API (CRUD) --> **rules**
-- Send input events to Falco over wire --> **inputs**
+- Send input events to Falco --> **inputs**
 
 ### Unary services
 
--- a/proposals/20191030-api.md
+++ b/proposals/20191030-api.md
@@ -26,7 +26,7 @@ Some examples, already in place, are:
 
 ## Non-Goals
 
--
+- Replace the existing inputs: the **inputs** will be only for new source events users may want to add, initially
 
 ## Proposal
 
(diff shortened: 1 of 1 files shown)


proposals/20191030-api.md
```diff
@@ -0,0 +1,144 @@
+# Falco APIs
+
+## Summary
+
+With this proposal we inted to present the intent of creating a Falco API.
```
fntlnz:
```suggestion
This is a proposal to better structure the Falco API.
```

proposals/20191030-api.md
```diff
@@ -0,0 +1,144 @@
...
+
+This proposal also intends to list the set of the gRPC services such API will expose.
```
fntlnz:
```suggestion
It will contain a list of gRPC services that can be added to have a first stable version of the API for v1.0.0.
```

proposals/20191030-api.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Motivation
+
+We want users to interface with Falco outputs, inputs, rules, and configurations via thirdy-part clients.
```
fntlnz:
```suggestion
We want to enable users to use thirdy-party clients to interface with Falco outputs, inputs, rules, and configurations.
```

proposals/20191030-api.md
```diff
@@ -0,0 +1,144 @@
...
+-
+
+## Proposal
+
+### Use cases
+
+- Receive Falco events with a well-defined contract over wire --> **outputs**
+- Receive Falco drops with a well-defined contract over wire --> **drops**
+- Receive current Falco version and related meta information (commit hash, built type, etc.) --> **version**
+- Config Falco via API (CRUD) -> **configs**
```
fntlnz:
```suggestion
- Configure Falco via API (CRUD) -> **configs**
```

proposals/20191030-api.md
```diff
@@ -0,0 +1,144 @@
...
+      rpc drops(request) returns (stream response);
+    }
+    service bidirectional {
+      ...
+    }
+    ```
+
+#### Benefits
+
+We think this structure makes the implementation more easy to write and to maintain because it automatically matches to abstractions - ie., one service by RPC type, a pool of contexts by RPC type - implemented in the code.
```
fntlnz:
```suggestion
We think this structure makes the implementation easier to write and to maintain because it automatically matches the abstractions - ie., one service by RPC type, a pool of contexts by RPC type - implemented in the code.
```

proposals/20191030-api.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Proposal
+
+### Use cases
+
+- Receive Falco events with a well-defined contract over wire --> **outputs**
+- Receive Falco drops with a well-defined contract over wire --> **drops**
+- Receive current Falco version and related meta information (commit hash, built type, etc.) --> **version**
+- Configure Falco via API (CRUD) -> **configs**
+- Inject and/or modify Falco rules via API (CRUD) --> **rules**
```
abroglesc: This would be awesome but would still require a route into the kubernetes cluster (or instance) to configure these or a sidecar that is used for configuration.

I am fairly new to the Falco project but has any thought been given to a central management server for Falco agents to report into and fetch new configurations / rules dynamically rather than having to hit an API or reload rules manually? I know this would be a new project to be spun up of the management server but interested if this has been proposed / discussed in the past before

proposals/20191030-api.md
```diff
@@ -0,0 +1,118 @@
+# Falco APIs
+
+## Summary
+
+This is a proposal to better structure the Falco API.
```
kris-nova: I think we should spend a few minutes talking about what exactly the Falco API is. As well as highlighting the problem(s) that it would potentially be solving.

I know we go into detail later, but a sentence or two here might be useful.

proposals/20191030-api.md
```diff
@@ -0,0 +1,118 @@
...
+
+It will contain a list of gRPC services that can be added to have a first stable version of the API for v1.0.0.
```
kris-nova: So is gRPC the API? or is gRPC the first implementation of a way of exposing the API over a network socket?

I think maybe the API is simply the agreement of the shape of objects, and what they represent and do right?

proposals/20191030-api.md
```diff
@@ -0,0 +1,118 @@
...
+
+- To design and implement an API to make Falco more configurable
+- To design and implement an API to make Falco more consumable
+- To design a set of simple contract interfaces
+- To have a secure by default Falco API
+
+## Non-Goals
+
+- Replace the existing inputs: the **inputs** will be only for new source events users may want to add, initially
+
```
kris-nova: What about documentation? Is that a non-goal here? If so we should maybe call it out?

proposals/20191030-api.md
```diff
@@ -0,0 +1,118 @@
...
+## Proposal
+
+### Use cases
+
+- Receive Falco events with a well-defined contract over wire --> **outputs**
```
kris-nova: remove wire? again this goes back to my previous comment about the API being a shape and agreement and not an implementation.

For instance if a user wanted to vendor falco in C++ they would still potentially be using "an API"

proposals/20191030-api.md
leodido: Maybe something like [this](https://github.com/falcosecurity/falco/issues/824) can help with what are you suggesting?

kris-nova: WRT having any external configuration management, the work starts here with implementing a clean interface around the core of the program. 

I think we would all be very excited to see work like a sidecar or configuration management server come to life, and the Falco community would be happy to sponsor this work. 

If you have ideas for configuration management we could open up a new proposal (like this one) with your ideas, and get a repository and a Falco client set up so work can begin building out something like what you described :smile:

max-lobur:
My comment applies to any configuration API on k8s. Everything going towards gitops right now, because it's easier to use, it's auditable (e.g. SOC2, PCI), there's governance (GH protected branches), history and so on.

The falco flow with API will eventually be:

User -> git repo -> terraform CI job -> terraform falco provider -> falco API -> falco configuration engine -> falco.
Where API and terraform provider is a huge chunk of work because it's network interaction and it has to be properly secured, throttled and instrumented. 

I think the much easier k8s flow would be:
User -> git repo -> gitops (e.g. flux-cd) -> CRD object on a cluster -> falco configuration engine (e.g. falco-operator) -> falco.

This already fits into all our existing pipelines, it's actually already built and so on.

Speaking of 3rd party dynamic rules providers. Any security product which will support falco will run it's agent in k8s anyways (already do), and they will have the communication with their agent anyways, so it will be much easier for them to just inject CRDs, and it will be much easier to me just to set RBAC for CRDs instead of thinking of another API.

max-lobur:
I think this day configuration management of anything in k8s is pretty much solved, I'd be much more interested to see record-n-replay, a development of falco-rules catalog (e.g. the hub), adoption within security vendors, smarter rules and so on.

kris-nova:
Hey @max-lobur 

So I think you bring up some valid topics here, and to be clear the pipeline you suggested would completely work with this proposal. 

This work defines very clear, rigid, and secure-by-default interfaces for working with Falco. Regardless of higher level implementation (such as Kubernetes, or gitops) users would now have a clean way of plumbing data into and out of Falco. 

Once we have a contract for interfacing with Falco via C/C++ building things out like CRDs are a natural next step that we plan on taking. Or in a perfect world, that the community could help us with.

max-lobur:
I see, so it's mostly for C/C++ interface rather than for configuration management over network

kris-nova:
Yes exactly @max-lobur 

Right now we have logic in C/C++ that deals with `pulling` data from a source directly.

This proposal attempts to standardize the input/output mechanisms for the C/C++ code so that we can build on top of it.

One of the big features here is that we are run mutual TLS by default :sunglasses: and are building tools around making that process as user friendly as possible 

So the pipeline you proposed is a *fantastic* examples of how this new API would be useful

leodido:
Clarification about where/how we plan to track and version the `.proto` files.
cc @kris-nova 

Since it is very likely that also the engine (`userspace/engine`) and not only `userspace/falco` will use some of the proto definitions (eg., `schema.proto`) my plan is to mainly experiment with two approaches:

1. create a root level directory in Falco repository responsible to contain the proto files used across the repository (API, engine, etc.)
2. create another repository in the falcosecurity organization

I have not made a decision on this since I would like to investigate it hands-on while writing the code. But it will be option 1 or option 2.

Do you have suggestions/ideas/preferences about it?

kris-nova:
@leodido,

Sounds good to me 👌🏻

PR #927, opened 2019-11-18
feat: add hostname field in gRPC output

### LINKED ISSUE #923: Expose the node metadata in the Falco outputs API

Falco should expose the "node" metadata (host that originated an alert) to third-party consumers.

**Why is this needed**:

It would be really useful to solve use cases such as https://github.com/falcosecurity/falco-exporter/issues/10, where third-party services need to act every time an alarm is raised on a node. For example, to cordon a node if running in a Kubernetes cluster, or just take the affected node down using draino.

The "node" metadata should be propagated to [falco-exporter](https://github.com/falcosecurity/falco-exporter), that communicates with falco via gRPC, to be able to consume that info as a label in the Prometheus metrics.

aarroyoc:
Signed-off-by: Adrián Arroyo Calle <adrian.arroyocalle@gmail.com>

**What type of PR is this?**

/kind feature

/area integrations

This PR adds a new field (hostname) in gRPC event messages.

Fixes #923

In theory, it adds a new field without causing trouble to old clients. It also passes hostname to other outputs but it does nothing. This was done to keep consistency between methods signatures. Maybe we want to move the field to `outputFields`

```release-note
adds a hostname field to gRPC output
```

### DIFF
--- a/userspace/falco/falco_outputs.cpp
+++ b/userspace/falco/falco_outputs.cpp
@@ -146,6 +146,12 @@ void falco_outputs::handle_event(gen_event *ev, string &rule, string &source,
 
 	std::lock_guard<std::mutex> guard(m_ls_semaphore);
 	lua_getglobal(m_ls, m_lua_output_event.c_str());
+	char hostname[1024];
+	int err = gethostname(hostname, sizeof(hostname));
+	if(err != 0){
+		string err = "Failed to get hostname";
+		throw falco_exception(err);
+	}
 	if(lua_isfunction(m_ls, -1))
 	{
 		lua_pushlightuserdata(m_ls, ev);
@@ -154,8 +160,9 @@ void falco_outputs::handle_event(gen_event *ev, string &rule, string &source,
 		lua_pushstring(m_ls, falco_common::priority_names[priority].c_str());
 		lua_pushnumber(m_ls, priority);
 		lua_pushstring(m_ls, format.c_str());
+		lua_pushstring(m_ls, hostname);
 
-		if(lua_pcall(m_ls, 6, 0, 0) != 0)
+		if(lua_pcall(m_ls, 7, 0, 0) != 0)
 		{
 			const char* lerr = lua_tostring(m_ls, -1);
 			string err = "Error invoking function output: " + string(lerr);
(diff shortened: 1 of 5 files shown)


userspace/falco/falco_outputs.cpp
```diff
@@ -146,6 +146,12 @@ void falco_outputs::handle_event(gen_event *ev, string &rule, string &source,
 
 	std::lock_guard<std::mutex> guard(m_ls_semaphore);
 	lua_getglobal(m_ls, m_lua_output_event.c_str());
+	char hostname[1024];
+	int err = gethostname(hostname, sizeof(hostname));
```
fntlnz: While this might be good for most use cases, imagine this scenario.

You run this as a Daemonset on Kubernetes.

Then you consume the outputs somewhere else and try to get the hostname. 

You notice that you are getting the hostname of the container and not the one of the node.

My proposed solution, is to create a new environment variable for the hostname. Then we read that environment variable and if it's not present we get the system hostname.

In that way, when deploying in kubernetes we can just go and do something like this in the Pod spec:

```yaml
...
      containers:
        - name: falco
          image: falcosecurity/falco:latest
          env:
            - name: FALCO_GRPC_HOSTNAME
              valueFrom:
                fieldRef:
                  fieldPath: spec.nodeName
...
```

Please note that `FALCO_GRPC_HOSTNAME` is just a filler idea I had for the variable name, I'm totally unsure what name to pick !

aarroyoc: I've updated the PR to reflect that change

palmerabollo:
Hi @leodido, thanks for your feedback. Could you please point us to a better place to setup the env var? Thanks!

userspace/falco/falco_outputs.cpp
```diff
@@ -146,6 +146,19 @@ void falco_outputs::handle_event(gen_event *ev, string &rule, string &source,
...
 	std::lock_guard<std::mutex> guard(m_ls_semaphore);
 	lua_getglobal(m_ls, m_lua_output_event.c_str());
+	std::string hostname;
+	char* env_hostname = getenv("FALCO_GRPC_HOSTNAME");
+	if(env_hostname == NULL){
+		char c_hostname[256];
+		int err = gethostname(c_hostname, 256);
+		if(err != 0){
+			string err = "Failed to get hostname";
+			throw falco_exception(err);
```
fntlnz: You can pass the error string directly here

fntlnz:
@palmerabollo @aarroyoc - the environment variable setup is all done in `falco.cpp` - see what we do with the k8s api env variable for example here: https://github.com/falcosecurity/falco/blob/dev/userspace/falco/falco.cpp#L1112-L1116

Once you have that, we have two options:
- pass it as a member when creating `falco_outputs`
- pass it to [do_inspect](https://github.com/falcosecurity/falco/blob/dev/userspace/falco/falco.cpp#L220) so that when it comes to call [handle_event](https://github.com/falcosecurity/falco/blob/dev/userspace/falco/falco.cpp#L328) we can give the information to the outputs.

Wdyt?

aarroyoc:
I've added it in falco_outputs because I think it's more of an initialization value

userspace/falco/falco_outputs.cpp
leodido: Let's merge the functionality then we can pick battle on names (I really like such battles Lorenzo knows ahaha).

palmerabollo:
@fntlnz does it looks good to you?

fntlnz:
Looks good! Thanks @aarroyoc for the pr and for makin the requested changes ! This is a very needed and cool improvement ! Merging !

PR #941, opened 2019-11-29
Add a simple user_known_package_manager_in_container_conditions macro

JPLachance:
**What type of PR is this?**

/kind feature

/kind rule-update

/area rules

Changes I made:
\+ Add a simple user_known_package_manager_in_container_conditions macro
\* Use the user_known_package_manager_in_container_conditions macro in the "Launch Package Management Process in Container" rule

This is particularly useful when we need to ignore a Jenkins pod running jobs in a Kubernetes cluster.

I did not file an issue.

```release-note
rules: Added a user_known_package_manager_in_container macro to allow users to specify conditions that match a legitimate use case for using a package management process in a container.
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -2255,7 +2255,13 @@
 - macro: network_tool_procs
   condition: (proc.name in (network_tool_binaries))
 
-- macro: user_known_package_manager_in_container_conditions
+# In a local/user rules file, create a condition that matches legitimate uses
+# of a package management process inside a container.
+#
+# For example:
+# - macro: user_known_package_manager_in_container
+#   condition: proc.cmdline="dpkg -l"
+- macro: user_known_package_manager_in_container
   condition: (never_true)
 
 # Container is supposed to be immutable. Package management should be done in building the image.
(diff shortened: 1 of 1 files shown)


Kaizhe:
@JPLachance can you elaborate more about the use case? Do you mean something like a Jenkins runs `apt-get` command inside the pod?

JPLachance:
Hello @Kaizhe,

In a Jenkins job, we often perform `pip install some-lib` or `pip install -r requirements.txt`.

Since both `pip` and `pip3` are members of the `package_mgmt_binaries` list, each time the job runs, it triggers an event.

By doing:
```YAML
- macro: user_known_package_manager_in_container_conditions
  condition: >
    container.image.repository="my-jenkins-image-repository"
```
I'm able to remove the noise.

The `dpkg -l` command is also firing the `Launch Package Management Process in Container` rule, but `dpkg -l` is not doing any modification to the system.

@fntlnz, how would you explain that in the release notes? My first guess is: `Added a user_known_package_manager_in_container_conditions to allow users to specify conditions that match a legitimate use case for using a package management process in a container`

rules/falco_rules.yaml
```diff
@@ -2255,11 +2255,19 @@
 - macro: network_tool_procs
   condition: (proc.name in (network_tool_binaries))
 
+- macro: user_known_package_manager_in_container_conditions
```
Kaizhe: Can you rename the macro to `user_known_package_manager_in_container`, the name is too long. And I couldn't think of any better :(

Kaizhe:
@JPLachance can you please address the my comment? it's a minor one, other than that, looks good to me. Thanks!

rules/falco_rules.yaml
JPLachance: Fixed!

fntlnz:
Changes looks very good to me, since also @Kaizhe said lgtm and you addressed his comment about the naming let’s merge !

PR #943, opened 2019-12-01
rule update: Add rules to detect access to GCE Instance Metadata

rung:

**What type of PR is this?**

/kind feature
/kind rule-create

/area rules

- The motivation of this new rule is to detect access to GCE Instance Metadata.
  - SSRF is still a threat to GKE, so I think it's good to have the rule.
    - (Ref: https://hackerone.com/reports/341876)
- This rule doesn't turn on by default. To use this rule, a user has to overwrite gce_metadata_containers on local/user rules file.

Fixes #

- "169.254.169.254" is a metadata server on GKE(GCE).
  - (Ref: https://cloud.google.com/compute/docs/storing-retrieving-metadata, https://github.com/googleapis/google-cloud-go/blob/master/compute/metadata/metadata.go)
  - The IP is the same as metadata on EC2(AWS). So we can use the same rule as EC2 ("Contact EC2 Instance Metadata Service From Container"), but for backward compatibility, I think it's not good to change the rule name and macro name("ec2_metadata_containers"). So I would like to add this rule.

NONE

```release-note
rules: add rules to detect access to GCE Instance Metadata
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -2210,26 +2210,25 @@
   priority: NOTICE
   tags: [network, aws, container, mitre_discovery]
 
-# In a local/user rules file, you could override this macro to
-# explicitly enumerate the container images that you want to allow
-# access to GCE metadata. In this main falco rules file, there isn't
-# any way to know all the containers that should have access, so any
-# container is alllowed, by repeating the "container" macro. In the
-# overridden macro, the condition would look something like
-# (container.image.repository = vendor/container-1 or
-# container.image.repository = vendor/container-2 or ...)
+
+# This rule is not enabled by default, since this rule is for GCP only.
+# If you want to enable this rule, overwrite the first macro,
+# And you can filter the container that you want to allow access to metadata by overwriting the second macro.
+- macro: consider_gce_metadata_access
+  condition: (never_true)
+
 - macro: gce_metadata_containers
-  condition: container
+  condition: (k8s.ns.name = "kube-system")
 
 # On GCE instances, 169.254.169.254 is a special IP used to fetch
-# metadata about the instance. It may be desirable to prevent access
-# to this IP from containers.
+# metadata about the instance. The metadata could be used to get credentials by attackers.
 - rule: Contact GCE Instance Metadata Service From Container
   desc: Detect attempts to contact the GCE Instance Metadata Service from a container
-  condition: outbound and fd.sip="169.254.169.254" and container and not gce_metadata_containers
+  condition: outbound and fd.sip="169.254.169.254" and container and consider_gce_metadata_access and not gce_metadata_containers
   output: Outbound connection to GCE instance metadata service (command=%proc.cmdline connection=%fd.name %container.info image=%container.image.repository:%container.image.tag)
   priority: NOTICE
   tags: [network, gcp, container, mitre_discovery]
+
 
 # In a local/user rules file, you should override this macro with the
 # IP address of your k8s api server. The IP 1.2.3.4 is a placeholder
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -2222,7 +2222,7 @@
 
 # On GCP, AWS and Azure, 169.254.169.254 is a special IP used to fetch
 # metadata about the instance. The metadata could be used to get credentials by attackers.
-- rule: Contact Cloud Instance Metadata Service From Container
+- rule: Contact cloud metadata service from container
   desc: Detect attempts to contact the Cloud Instance Metadata Service from a container
   condition: outbound and fd.sip="169.254.169.254" and container and consider_metadata_access and not user_known_metadata_access
   output: Outbound connection to cloud instance metadata service (command=%proc.cmdline connection=%fd.name %container.info image=%container.image.repository:%container.image.tag)
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -2211,23 +2211,23 @@
   tags: [network, aws, container, mitre_discovery]
 
 
-# This rule is not enabled by default, since this rule is for GCP only.
+# This rule is not enabled by default, since this rule is for cloud environment(GCP, AWS and Azure) only.
 # If you want to enable this rule, overwrite the first macro,
 # And you can filter the container that you want to allow access to metadata by overwriting the second macro.
-- macro: consider_gce_metadata_access
+- macro: consider_metadata_access
   condition: (never_true)
 
-- macro: gce_metadata_containers
+- macro: user_known_metadata_access
   condition: (k8s.ns.name = "kube-system")
 
-# On GCE instances, 169.254.169.254 is a special IP used to fetch
+# On GCP, AWS and Azure, 169.254.169.254 is a special IP used to fetch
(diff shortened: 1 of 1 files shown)


fntlnz:
Hi @rung ! Thanks for your PR! I heard many falco users talking about having something like this! Would you be able to provide an output example and a simple reproduce step to trigger the rule?

I would expect something like:
- start a container
- do an http request to the container metadata


In that way we could open an issue to write an integration test for this if we ever have CI worker nodes on gce. Wdyt?

rung:
Hi, @fntlnz 
Thank you for your review!
I wrote output example and a simple reproduce step.

### Configuration
- Overwrite gce_metadata_containers on local/user rules file.
```
- macro: gce_metadata_containers
  condition: k8s.ns.name = "kube-system"
```
  - this means container other than "kube-system" namespace should not access to Metadata


### Procedure
1. Start a container on GKE
1. (Trigger)Http request to GCE Metadata from container
```
curl "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token" -H "Metadata-Flavor: Google"
```
- You can get access token on GCE Service Account using this command
- A record of metadata.google.internal is "169.254.169.254"

### Output Example
```
{"output":"15:09:47.374847124: Notice Outbound connection to GCE instance metadata service (command=curl http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token -H Metadata-Flavor: Google connection=10.4.0.9:43418->169.254.169.254:80 k8s.ns=default k8s.pod=centos-545965cf9-4ssc4 container=1022a8465482 image=centos:latest) k8s.ns=default k8s.pod=centos-545965cf9-4ssc4 container=1022a8465482 k8s.ns=default k8s.pod=centos-545965cf9-4ssc4 container=1022a8465482","priority":"Notice","rule":"Contact GCE Instance Metadata Service From Container","time":"2019-12-01T15:09:47.374847124Z", "output_fields": {"container.id":"1022a8465482","container.image.repository":"centos","container.image.tag":"latest","evt.time":1575212987374847124,"fd.name":"10.4.0.9:43418->169.254.169.254:80","k8s.ns.name":"default","k8s.pod.name":"centos-545965cf9-4ssc4","proc.cmdline":"curl http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token -H Metadata-Flavor: Google"}}
```

rules/falco_rules.yaml
```diff
@@ -2210,6 +2210,27 @@
...
+# In a local/user rules file, you could override this macro to
+# explicitly enumerate the container images that you want to allow
+# access to GCE metadata. In this main falco rules file, there isn't
+# any way to know all the containers that should have access, so any
+# container is alllowed, by repeating the "container" macro. In the
+# overridden macro, the condition would look something like
+# (container.image.repository = vendor/container-1 or
+# container.image.repository = vendor/container-2 or ...)
+- macro: gce_metadata_containers
+  condition: container
```
fntlnz: This macro has two responsibilities:
1. Enabling the rule
2. Filtering for container images

I think we would rather want two different macros: 
- one that does the enable/disable using the never_true condition [example](https://github.com/falcosecurity/falco/blob/dev/rules/falco_rules.yaml#L459) - this one will be used at the very start of the rule - in this way we disable without evaluating anything else 
- one to do the filtering (we can keep the one you already wrote, however I would put one of the examples you wrote in it instead of `container`

Then we document the rule by saying, if you want to enable this one you overwrite to `always_true` the first macro and then you overwrite the second macro with the images you want to filter on.

Wdyt ?

Thanks for the good work again, we are getting closer! I would really love to have this merged

fntlnz:
@rung thanks for the steps and the output! Makes the review process faster too :) added another comment with some nits!

rules/falco_rules.yaml
rung: Thank you for your review.
I thought your suggestion made my PR much better.

> 1. Enabling the rule
- I added "consider_gce_metadata_access" macro as flag of this rule.

> 2. Filtering for container images
- We can have default configuration to "gce_metadata_containers" now because we have "consider_gce_metadata_access".
  - so I set "(k8s.ns.name = "kube-system")" as default condition.
  - (I tried this rule on GKE, and there is no false positive)

- and I rewrote comments to simplify.

rung: We need different pre-configuration now.

### Configuration
- Overwrite consider_gce_metadata_access
```
- macro: consider_gce_metadata_access
  condition: (always_true)
```

rung:
@fntlnz Thank you for your comment!
I sent new commit to this PR. so could you review it?

rules/falco_rules.yaml
```diff
@@ -2210,6 +2210,26 @@
...
+# And you can filter the container that you want to allow access to metadata by overwriting the second macro.
+- macro: consider_gce_metadata_access
+  condition: (never_true)
+
+- macro: gce_metadata_containers
+  condition: (k8s.ns.name = "kube-system")
+
+# On GCE instances, 169.254.169.254 is a special IP used to fetch
+# metadata about the instance. The metadata could be used to get credentials by attackers.
+- rule: Contact GCE Instance Metadata Service From Container
```
Kaizhe: This rule looks similar to rule `Contact EC2 Instance Metadata Service From Container`. Can you make this rule a more generic one by removing `GCE`? With the conditions we may want to have two macros:
1. `consider_metadata_access`
2. `user_known_metadata_access`

rung: Hi, @Kaizhe 
Thank you for your review.
I didn't know Azure's Metadata's IP Address. but it seems Azure uses same IP Address (169.254.169.254) for metadata service. so I think the rule can be more generic too.
Ref: https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service

I will remove `GCE` as you suggest. Thank you.

rung: @Kaizhe I renamed the macros and rule names.
Could you review them again?

Kaizhe: The changes look great! Thank you for the contribution!

rules/falco_rules.yaml
```diff
@@ -2210,6 +2210,26 @@
...
+# And you can filter the container that you want to allow access to metadata by overwriting the second macro.
+- macro: consider_metadata_access
+  condition: (never_true)
+
+- macro: user_known_metadata_access
+  condition: (k8s.ns.name = "kube-system")
+
+# On GCP, AWS and Azure, 169.254.169.254 is a special IP used to fetch
+# metadata about the instance. The metadata could be used to get credentials by attackers.
+- rule: Contact Cloud Instance Metadata Service From Container
```
Kaizhe: One minor comment, please rename to rule to `Contact cloud metadata service from container`

Kaizhe: we have a loose naming convention, but this is something we try to follow.

rules/falco_rules.yaml
rung: Thank you for your review. I'm glad to contribute Falco.

rules/falco_rules.yaml
rung: @Kaizhe I renamed the name as you suggested.
Could you review it again?
Thank you!

PR #945, opened 2019-12-02
rule update: Add rules to detect raw packets creation

rung:

**What type of PR is this?**
/kind rule-create

/area rules

### What is this
- This rule detects new raw packets creation in container

### Why we need it
- Kubernetes set CAP_NET_RAW capability to container by default. so any pod can abuse raw packets like ARP Spoofing. This is a still threat on Kubernetes.
  - Ref: https://blog.aquasec.com/dns-spoofing-kubernetes-clusters
- There is no rule for raw packet on Falco now. so I would like to add this rule. What do you think?

Fixes #

NONE

```release-note
rules: add rules to detect raw packets creation
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -2617,7 +2617,7 @@
 
 - rule: Raw packet created in container
   desc: Detect new raw packets at the device driver (OSI Layer 2) level in a container. raw packets could be used to do ARP Spoofing by attacker.
-  condition: consider_raw_packet_communication and evt.type=socket and evt.arg[0] in (AF_PACKET, PF_PACKET) and container and not proc.name in (user_known_raw_packet_binaries)
+  condition: consider_raw_packet_communication and evt.type=socket and evt.arg[0]=AF_PACKET and container and not proc.name in (user_known_raw_packet_binaries)
   output: Raw packet was created in a container (user=%user.name command=%proc.cmdline container_id=%container.id container_name=%container.name image=%container.image.repository:%container.image.tag)
   priority: NOTICE
   tags: [network, mitre_discovery]
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -2604,7 +2604,25 @@
   output: "Docker or kubernetes client executed in container (user=%user.name %container.info parent=%proc.pname cmdline=%proc.cmdline)"
   priority: WARNING
   tags: [container, mitre_execution]
+
+
+# This rule is not enabled by default, as there are legitimate use
+# cases for raw packet. If you want to enable it, modify the
+# following macro.
+- macro: consider_raw_packet_creation
+  condition: (never_true)
+
+- list: raw_packet_binaries
+  items: []
+
+- rule: Raw packet was created in a container
+  desc: Detect new raw packets in a container. raw packets could be used to do ARP Spoofing by attacker.
+  condition: evt.type=socket and evt.arg[0]=AF_PACKET and container and not proc.name in (raw_packet_binaries)
+  output: Raw packet was created in a container (user=%user.name command=%proc.cmdline container_id=%container.id container_name=%container.name image=%container.image.repository:%container.image.tag)
+  priority: NOTICE
+  tags: [network, mitre_discovery]
   
+
 # Application rules have moved to application_rules.yaml. Please look
 # there if you want to enable them by adding to
 # falco_rules.local.yaml.
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -2609,16 +2609,16 @@
 # This rule is not enabled by default, as there are legitimate use
 # cases for raw packet. If you want to enable it, modify the
 # following macro.
-- macro: consider_raw_packet_communication
+- macro: consider_packet_socket_communication
   condition: (never_true)
 
-- list: user_known_raw_packet_binaries
+- list: user_known_packet_socket_binaries
   items: []
 
-- rule: Raw packet created in container
-  desc: Detect new raw packets at the device driver (OSI Layer 2) level in a container. raw packets could be used to do ARP Spoofing by attacker.
-  condition: consider_raw_packet_communication and evt.type=socket and evt.arg[0]=AF_PACKET and container and not proc.name in (user_known_raw_packet_binaries)
-  output: Raw packet was created in a container (user=%user.name command=%proc.cmdline container_id=%container.id container_name=%container.name image=%container.image.repository:%container.image.tag)
+- rule: Packet socket created in container
+  desc: Detect new packet socket at the device driver (OSI Layer 2) level in a container. Packet socket could be used to do ARP Spoofing by attacker.
+  condition: evt.type=socket and evt.arg[0]=AF_PACKET and consider_packet_socket_communication and container and not proc.name in (user_known_packet_socket_binaries)
+  output: Packet socket was created in a container (user=%user.name command=%proc.cmdline socket_info=%evt.args container_id=%container.id container_name=%container.name image=%container.image.repository:%container.image.tag)
   priority: NOTICE
   tags: [network, mitre_discovery]
   
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -2609,15 +2609,15 @@
 # This rule is not enabled by default, as there are legitimate use
 # cases for raw packet. If you want to enable it, modify the
 # following macro.
-- macro: consider_raw_packet_creation
+- macro: consider_raw_packet_communication
   condition: (never_true)
 
-- list: raw_packet_binaries
+- list: user_known_raw_packet_binaries
   items: []
 
-- rule: Raw packet was created in a container
-  desc: Detect new raw packets in a container. raw packets could be used to do ARP Spoofing by attacker.
-  condition: evt.type=socket and evt.arg[0]=AF_PACKET and container and not proc.name in (raw_packet_binaries)
+- rule: Raw packet created in container
+  desc: Detect new raw packets at the device driver (OSI Layer 2) level in a container. raw packets could be used to do ARP Spoofing by attacker.
+  condition: consider_raw_packet_communication and evt.type=socket and evt.arg[0] in (AF_PACKET, PF_PACKET) and container and not proc.name in (user_known_raw_packet_binaries)
   output: Raw packet was created in a container (user=%user.name command=%proc.cmdline container_id=%container.id container_name=%container.name image=%container.image.repository:%container.image.tag)
   priority: NOTICE
   tags: [network, mitre_discovery]


rung:
### Pre-Configuration
```
- macro: consider_raw_packet_creation
  condition: (always_true)
```

### Procedure step
- You can use tool like arpspoof and kube-dnsspoof for ARP Spoofing
  - I use kube-dnsspoof in this procedure

1. Download https://github.com/danielsagi/kube-dnsspoof/
2. Create pod
```
kubectl create -f pods/
```

3. Kubectl exec to hacker pod
```
kubectl exec -it hacker zsh
```

4. Do ARP Spoofing (Rule's trigger)
```
./exploit.py
```

### Sample Output
```
{"output":"17:53:23.882975839: Notice Raw packet was created in a container (user=root command=python3 ./exploit.py container_id=5b6d82ba1a95 container_name=k8s_hacker_hacker_default_a4eb89e7-1524-11ea-93b8-42010a80001b_0 image=danielsagi/kube-dnsspoof:latest) k8s.ns=default k8s.pod=hacker container=5b6d82ba1a95 k8s.ns=default k8s.pod=hacker container=5b6d82ba1a95 k8s.ns=default k8s.pod=hacker container=5b6d82ba1a95","priority":"Notice","rule":"Raw packet was created in a container","time":"2019-12-02T17:53:23.882975839Z", "output_fields": {"container.id":"5b6d82ba1a95","container.image.repository":"danielsagi/kube-dnsspoof","container.image.tag":"latest","container.name":"k8s_hacker_hacker_default_a4eb89e7-1524-11ea-93b8-42010a80001b_0","evt.time":1575309203882975839,"k8s.ns.name":"default","k8s.pod.name":"hacker","proc.cmdline":"python3 ./exploit.py","user.name":"root"}}
```

rules/falco_rules.yaml
```diff
@@ -2604,7 +2604,25 @@
   output: "Docker or kubernetes client executed in container (user=%user.name %container.info parent=%proc.pname cmdline=%proc.cmdline)"
   priority: WARNING
   tags: [container, mitre_execution]
+
+
+# This rule is not enabled by default, as there are legitimate use
+# cases for raw packet. If you want to enable it, modify the
+# following macro.
+- macro: consider_raw_packet_creation
```
Kaizhe: please rename the macro to `consider_raw_packet_communication`

and the macro is not used in the rule `Raw packet was created in a container`

rules/falco_rules.yaml
```diff
@@ -2604,7 +2604,25 @@
...
+  condition: (never_true)
+
+- list: raw_packet_binaries
+  items: []
+
+- rule: Raw packet was created in a container
+  desc: Detect new raw packets in a container. raw packets could be used to do ARP Spoofing by attacker.
+  condition: evt.type=socket and evt.arg[0]=AF_PACKET and container and not proc.name in (raw_packet_binaries)
```
Kaizhe: 1. can you rename the rule to `Raw socket created in container`
2. can you add another constraint to `evt.arg[1]=SOCK_RAW`? I haven't verified my self, but I would like to have the socket type as part of the condition.

rung: Hi, @Kaizhe 
Thank you for your review.

> 2. can you add another constraint to evt.arg[1]=SOCK_RAW?
> I haven't verified my self, but I would like to have the socket type as part of the condition.
- `AF_PACKET` is used for raw packet (layer 2). and socket_type could have `SOCK_RAW` and `SOCK_DGRAM` in my understanding.
  - This rule aims to catch a raw packet (Layer 2) connection to detect attacks like ARP Spoofing. So I couldn't add another constraint.
  - Ref: https://linux.die.net/man/7/packet
  - Ref: https://linux.die.net/man/7/raw

> 1. can you rename the rule to Raw socket created in container
- On Man page of packet(7), `raw packets` is used for AF_PACKET. so I used `raw packets`. what do you think about the naming?
  - And I agree `raw packets` is not clear well. So how about changing desc to this?
```
Detect new raw packets at the device driver (OSI Layer 2) level
 in a container. raw packets could be used to do ARP Spoofing by attacker.
```
  - Ref: https://linux.die.net/man/7/packet

rules/falco_rules.yaml
rung: Thank you for your review!
about naming, I wrote comments on another comment.
about rule's condition, I will fix it. Thank you for catching my mistake.

rules/falco_rules.yaml
Kaizhe: this is socket creation event (not receive or send), and I am not sure whether the ARP spoofing can be done via both `SOCK_RAW` and `SOCK_DGRAM`. I think it can be done with `SOCK_RAW` (https://opensourceforu.com/2015/03/a-guide-to-using-raw-sockets/) and further more, `PF_PACKET` seems similar to `AF_PACKET`. This quite depends on the underneath system.

Regarding the naming, I would suggest `raw socket created in container` if the `raw` type is the one must be in the condition.

rung: Thank you for your comment!

Yes, I should add PF_PACKET too! I'll fix it.

About `SOCK_DGRAM`, I found `SOCK_DGRAM` is used in some tools for ARP Spoofing.
(`SOCK_RAW` is popular for ARP Spoofing though, because `sockaddr_ll` struct is required to use `SOCK_DGRAM`)
Ref: https://github.com/smikims/arpspoof/blob/244d9eedd0dc9576b062bc01963d878560d574f5/arpspoof.c#L319

And naming, I think "raw socket" is a more common word. so I'll rename "raw packet" to "raw socket". and I'll add the explanation to desc

rung: @Kaizhe How about `packet socket created in container` for rule name?
When using `raw socket`, I think user could misunderstand L2 Layer as L3 Layer.
(i.e. `ping` uses raw socket(L3 Layer), but rule detects packet socket(L2 Layer), so user doesn't have to add `ping` to `raw_packet_binaries`. )
So I felt `packet socket` is more understandable.
Ref: https://linux.die.net/man/7/packet

rules/falco_rules.yaml
rung: I fixed the rule's condition. I will resolve this conversation for now.

rules/falco_rules.yaml
```diff
@@ -2604,7 +2604,25 @@
...
   tags: [container, mitre_execution]
+
+
+# This rule is not enabled by default, as there are legitimate use
+# cases for raw packet. If you want to enable it, modify the
+# following macro.
+- macro: consider_raw_packet_creation
+  condition: (never_true)
+
+- list: raw_packet_binaries
```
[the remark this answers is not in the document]
rung: I renamed the list name.

rules/falco_rules.yaml
rung: I sent new commit as you suggested. Thank you!
  - I renamed macro and list.
  - I fixed condition.
    - (added `PF_PACKET` and `consider_raw_packet_creation` macro).

- This is not resolved yet
   - Naming problem about `raw packet`.
   - Whether adding `SOCK_RAW` to condition.

rung: - I did investigation about socket_family. As far as I tested, Falco needs `AF_PACKET` only (`PF_PACKET` is not needed in condition).
  - because both are same value when calling syscall. and it seems both value is handled as `AF_PACKET`

### Investigation
#### PF_PACKET value on glibc
- AF_PACKET and PF_PACKET is same value.(17 = 0x11)
```
#define AF_PACKET	PF_PACKET
```
- Ref: https://sources.debian.org/src/glibc/2.29-3/sysdeps/unix/sysv/linux/bits/socket.h/

#### Confirmation when syscall
- both socket uses same value(0x11) for fist argument when syscall.
  - (Falco detected both socket() using AF_PACKET)
```
$ cat socktest.c 
#include <sys/socket.h>
#include <linux/if_packet.h>
#include <net/ethernet.h>
#include <arpa/inet.h>

int main() {
    int fd = socket(AF_PACKET, SOCK_RAW, htons(ETH_P_ARP));
    fd = socket(PF_PACKET, SOCK_RAW, htons(ETH_P_ARP));
}

$ gcc socktest.c -o socktest
$ sudo strace -e trace=socket -e raw=socket ./socktest
socket(0x11, 0x3, 0x608)                = 0x3
socket(0x11, 0x3, 0x608)                = 0x4
+++ exited with 0 +++
$ 
```

Kaizhe: @rung thanks for the further investigation! Really appreciate it!

rules/falco_rules.yaml
```diff
@@ -2604,7 +2604,25 @@
...
+# following macro.
+- macro: consider_raw_packet_communication
+  condition: (never_true)
+
+- list: user_known_raw_packet_binaries
+  items: []
+
+- rule: Raw packet created in container
+  desc: Detect new raw packets at the device driver (OSI Layer 2) level in a container. raw packets could be used to do ARP Spoofing by attacker.
+  condition: consider_raw_packet_communication and evt.type=socket and evt.arg[0]=AF_PACKET and container and not proc.name in (user_known_raw_packet_binaries)
```
Kaizhe: let's move `evt.type=sock` to the beginning of the condition for performance purpose.

rules/falco_rules.yaml
```diff
@@ -2604,7 +2604,25 @@
...
+# This rule is not enabled by default, as there are legitimate use
+# cases for raw packet. If you want to enable it, modify the
+# following macro.
+- macro: consider_raw_packet_communication
+  condition: (never_true)
+
+- list: user_known_raw_packet_binaries
+  items: []
+
+- rule: Raw packet created in container
```
Kaizhe: Let's rename to `Packet socket created in container` :)

rules/falco_rules.yaml
```diff
@@ -2604,7 +2604,25 @@
...
+  desc: Detect new raw packets at the device driver (OSI Layer 2) level in a container. raw packets could be used to do ARP Spoofing by attacker.
+  condition: consider_raw_packet_communication and evt.type=socket and evt.arg[0]=AF_PACKET and container and not proc.name in (user_known_raw_packet_binaries)
+  output: Raw packet was created in a container (user=%user.name command=%proc.cmdline container_id=%container.id container_name=%container.name image=%container.image.repository:%container.image.tag)
```
Kaizhe: can you add `socket_info=%evt.args` in the output section? Let's not have the `RAW_SOCK` in the condition this time. With the help of the output, we can identify the False Positive easily.

Kaizhe: And after looking into the ARP packet structure, I think `SOCK_DGRAM` is also able to spoof too.

rules/falco_rules.yaml
```diff
@@ -2604,7 +2604,25 @@
...
+# cases for raw packet. If you want to enable it, modify the
+# following macro.
+- macro: consider_raw_packet_communication
+  condition: (never_true)
+
+- list: user_known_raw_packet_binaries
+  items: []
+
+- rule: Raw packet created in container
+  desc: Detect new raw packets at the device driver (OSI Layer 2) level in a container. raw packets could be used to do ARP Spoofing by attacker.
```
```suggestion
  desc: Detect new raw packets at the device driver (OSI Layer 2) level in a container. Raw packets could be used to do ARP Spoofing by attacker.
```

rules/falco_rules.yaml
rung: Thank you :)
I renamed them including macro, list name and desc of the rule

rules/falco_rules.yaml
rung: Thank you for your comment.
I modified desc to use uppercase there.

rules/falco_rules.yaml
rung: I added.

### Output Example
```
{"output":"21:27:20.711269362: Notice Packet socket was created in a container (user=root command=socktest socket_info=domain=17(AF_PACKET) type=3 proto=1544  container_id=577c694601c8 container_name=k8s_ubuntu_ubuntu-9f765598b-jxwcd_default_e241e25f-1523-11ea-93b8-42010a80001b_1 image=ubuntu:latest) k8s.ns=default k8s.pod=ubuntu-9f765598b-jxwcd container=577c694601c8 k8s.ns=default k8s.pod=ubuntu-9f765598b-jxwcd container=577c694601c8 k8s.ns=default k8s.pod=ubuntu-9f765598b-jxwcd container=577c694601c8","priority":"Notice","rule":"Packet socket created in container","time":"2019-12-03T21:27:20.711269362Z", "output_fields": {"container.id":"577c694601c8","container.image.repository":"ubuntu","container.image.tag":"latest","container.name":"k8s_ubuntu_ubuntu-9f765598b-jxwcd_default_e241e25f-1523-11ea-93b8-42010a80001b_1","evt.args":"domain=17(AF_PACKET) type=3 proto=1544 ","evt.time":1575408440711269362,"k8s.ns.name":"default","k8s.pod.name":"ubuntu-9f765598b-jxwcd","proc.cmdline":"socktest","user.name":"root"}}
```

rules/falco_rules.yaml
rung: Thank you for checking it. I modified

### Output Example
```
{"output":"21:27:20.711269362: Notice Packet socket was created in a container (user=root command=socktest socket_info=domain=17(AF_PACKET) type=3 proto=1544  container_id=577c694601c8 container_name=k8s_ubuntu_ubuntu-9f765598b-jxwcd_default_e241e25f-1523-11ea-93b8-42010a80001b_1 image=ubuntu:latest) k8s.ns=default k8s.pod=ubuntu-9f765598b-jxwcd container=577c694601c8 k8s.ns=default k8s.pod=ubuntu-9f765598b-jxwcd container=577c694601c8 k8s.ns=default k8s.pod=ubuntu-9f765598b-jxwcd container=577c694601c8","priority":"Notice","rule":"Packet socket created in container","time":"2019-12-03T21:27:20.711269362Z", "output_fields": {"container.id":"577c694601c8","container.image.repository":"ubuntu","container.image.tag":"latest","container.name":"k8s_ubuntu_ubuntu-9f765598b-jxwcd_default_e241e25f-1523-11ea-93b8-42010a80001b_1","evt.args":"domain=17(AF_PACKET) type=3 proto=1544 ","evt.time":1575408440711269362,"k8s.ns.name":"default","k8s.pod.name":"ubuntu-9f765598b-jxwcd","proc.cmdline":"socktest","user.name":"root"}}
```

rung:
@Kaizhe @leodido 
Thank you for your review. I really appreciate it!
I sent new commit. so could you review it again?

rung:
@Kaizhe Thank you for lgtm! I'm really glad to add this rule.

I seems this PR is needed for you to approve additionally. Could you do it?

rung:
Thank you!

PR #947, opened 2019-12-02
Exclude exe_running_docker_save in the "Write below etc" rule

JPLachance:
**What type of PR is this?**
/kind bug
/kind rule-update

/area rules

The "Write below etc" rule is firing an event each time we spin a container. For example:
```JSON
{
    "priority": "Error",
    "output": "21:11:42.835233238: Error File below /etc opened for writing (user=<NA> command=exe /var/lib/docker/overlay2/b96d5259b14161da7fe7f3f0c036d723bbd937bc83cf6abdcac850210c397759/diff parent=<NA> pcmdline=<NA> file=/etc/group program=exe gparent=<NA> ggparent=<NA> gggparent=<NA> container_id=host image=<NA>) k8s.ns=<NA> k8s.pod=<NA> container=host k8s.ns=<NA> k8s.pod=<NA> container=host",
    "rule": "Write below etc",
    "output_fields": {
      "fd.name": "/etc/group",
      "proc.cmdline": "exe /var/lib/docker/overlay2/b96d5259b14161da7fe7f3f0c036d723bbd937bc83cf6abdcac850210c397759/diff",
      "container.id": "host",
      "evt.time": 1574889102835233300,
      "proc.name": "exe"
    }
  }
```
This event is surrounded by dozens like it.

Adding a simple `and not exe_running_docker_save` fixes the issue.

I did not file an issue, here is the PR instead!

```release-note
Fixed a false positive in the "Write below etc" rule by excluding "exe_running_docker_save". 
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -1266,7 +1266,9 @@
 
 - rule: Write below etc
   desc: an attempt to write to any file below /etc
-  condition: write_etc_common
+  condition: >
+    write_etc_common
+    and not exe_running_docker_save
   output: "File below /etc opened for writing (user=%user.name command=%proc.cmdline parent=%proc.pname pcmdline=%proc.pcmdline file=%fd.name program=%proc.name gparent=%proc.aname[2] ggparent=%proc.aname[3] gggparent=%proc.aname[4] container_id=%container.id image=%container.image.repository)"
   priority: ERROR
   tags: [filesystem, mitre_persistence]


fntlnz:
Thanks @JPLachance - changes lgtm but I’ll let @Kaizhe do a pass before we merge.
/assign @Kaizhe

Kaizhe:
@JPLachance just curious, do you enable `--userns-remap`?

rules/falco_rules.yaml
```diff
@@ -1266,7 +1266,9 @@
 
 - rule: Write below etc
   desc: an attempt to write to any file below /etc
-  condition: write_etc_common
+  condition: >
+    write_etc_common
+    and not exe_running_docker_save
```
Kaizhe: @JPLachance please add `exe_running_docker_save` inside `write_etc_common`

JPLachance: I can, but I'll open two PRs for the "Write below etc" rule really soon and the condition will become bigger.

I like it when the condition looks like:
```YAML
condition: >
  something and something_else or this
  and not a_condition_making_it_legit
  and not another_condition_making_it_legit
```

What do you think?

JPLachance:
@Kaizhe, what is `--userns-remap` exactly? Do you mean the [Docker option](https://docs.docker.com/engine/security/userns-remap/)?

I don't think we use it, unless it is the default value in KOPS (but I don't think it is).

Could it cause an issue?

rules/falco_rules.yaml
Kaizhe: `write_etc_common` actually looks like the way you suggested.

JPLachance: Ho! Okay, sorry, I miss-read your initial comment. On it!

JPLachance:
I'm confused. The `and not exe_running_docker_save` condition is already part of the `write_etc_common`. How did I got that alert? 🤔

PR #952, opened 2019-12-04
rule update: modify rule to detect connection to K8S API Server from a container

rung:

**What type of PR is this?**
/kind rule-update

/area rules

- What this PR does
  - Change default condition of "Contact K8S API Server From Container" rule
- Why we need it
  - To use this rule, Users have to set K8S API's IP address for every cluster now.
    - I want to use dynamic DNS resolution instead of manual IP address setting. By using updated rule, users can use this rule by default.
    - Ref: https://github.com/draios/sysdig/wiki/Domain-Name-Resolution-for-Connections

Fixes #

- Pod can access to K8S API using private ip address, so I stopped using "outbound" macro.
- K8S API might have global IP address too. but this rule detects connection to "kubernetes.default.svc.cluster.local" only.
  - "kubernetes.default.svc.cluster.local" is a name to Kubernetes API
  ```
  Pods can use the kubernetes.default.svc hostname to query the API server
  ```
    - Ref: https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-api/

### Procedure and Trigger
1. Start a container
1. Connect K8S API (Trigger)
```
curl -k https://$KUBERNETES_SERVICE_HOST:$KUBERNETES_PORT_443_TCP_PORT/api/
```

NONE

```release-note
rules: update "Contact K8S API Server From Container"
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -2240,13 +2240,7 @@
   tags: [network, container, mitre_discovery]
 
 
-# In a local/user rules file, you should override this macro with the
-# IP address of your k8s api server. The IP 1.2.3.4 is a placeholder
-# IP that is not likely to be seen in practice.
-- macro: k8s_api_server
-  condition: (fd.sip="1.2.3.4" and fd.sport=8080)
-
-# In a local/user rules file, list the container images that are
+# In a local/user rules file, list the namespace or container images that are
 # allowed to contact the K8s API Server from within a container. This
 # might cover cases where the K8s infrastructure itself is running
 # within a container.
@@ -2254,11 +2248,14 @@
   condition: >
     (container.image.repository in (gcr.io/google_containers/hyperkube-amd64,
      gcr.io/google_containers/kube2sky, sysdig/agent, sysdig/falco,
-     sysdig/sysdig))
+     sysdig/sysdig, falcosecurity/falco) or (k8s.ns.name = "kube-system"))
+
+- macro: k8s_api_server
+  condition: (fd.sip.name="kubernetes.default.svc.cluster.local")
 
 - rule: Contact K8S API Server From Container
   desc: Detect attempts to contact the K8S API Server from a container
-  condition: outbound and k8s_api_server and container and not k8s_containers
+  condition: evt.type=connect and evt.dir=< and (fd.typechar=4 or fd.typechar=6) and container and not k8s_containers and k8s_api_server
   output: Unexpected connection to K8s API Server from container (command=%proc.cmdline %container.info image=%container.image.repository:%container.image.tag connection=%fd.name)
   priority: NOTICE
   tags: [network, k8s, container, mitre_discovery]


Kaizhe:
@rung can you post the output here as well with your changes?

rung:
@Kaizhe 
This is output of this rule.

### Output sample
```
{"output":"05:35:37.855410748: Notice Unexpected connection to K8s API Server from container (command=curl -k https://10.70.0.1:443/api/ k8s.ns=default k8s.pod=ubuntu-9f765598b-jxwcd container=577c694601c8 image=ubuntu:latest connection=10.4.2.21:59880->10.70.0.1:443) k8s.ns=default k8s.pod=ubuntu-9f765598b-jxwcd container=577c694601c8 k8s.ns=default k8s.pod=ubuntu-9f765598b-jxwcd container=577c694601c8","priority":"Notice","rule":"Contact K8S API Server From Container","time":"2019-12-04T05:35:37.855410748Z", "output_fields": {"container.id":"577c694601c8","container.image.repository":"ubuntu","container.image.tag":"latest","evt.time":1575437737855410748,"fd.name":"10.4.2.21:59880->10.70.0.1:443","k8s.ns.name":"default","k8s.pod.name":"ubuntu-9f765598b-jxwcd","proc.cmdline":"curl -k https://10.70.0.1:443/api/"}}
```

rules/falco_rules.yaml
```diff
@@ -2240,25 +2240,22 @@
...
-     sysdig/sysdig))
+     sysdig/sysdig, falcosecurity/falco) or (k8s.ns.name = "kube-system"))
+
+- macro: k8s_api_server
+  condition: (fd.sip.name="kubernetes.default.svc.cluster.local")
 
 - rule: Contact K8S API Server From Container
   desc: Detect attempts to contact the K8S API Server from a container
-  condition: outbound and k8s_api_server and container and not k8s_containers
+  condition: evt.type=connect and evt.dir=< and (fd.typechar=4 or fd.typechar=6) and container and not k8s_containers and k8s_api_server
```
Kaizhe: Why not use `outbound`?

rung: Thank you for checking again!

As my special notes, K8S API could use private ip address. but `outbound` doesn't detect private ip address using `rfc_1918_addresses` macro. so I stopped using `outbound`
Ref: https://github.com/falcosecurity/falco/blob/418bcf2177b9c7f18f6a556871eb494624f69f2c/rules/falco_rules.yaml#L317-L327

Kaizhe: You're right, I forgot I made that changes already.

Kaizhe: everything looks good to me.

rung: Thank you for your review!

fntlnz:
Thanks again @rung ! Good job as always!

djsly:
Quick question, can we have a documentation on how we should add a list of pods that are allowed to talk to the API Server ?

we are getting > 50k alerts per 6 hours... where most of the communication are normal.

PR #955, opened 2019-12-04
rule(The docker client is executed in a container): modify condition to reduce false positive

rung:

**What type of PR is this?**
/kind rule-update

/area rules

- What this PR does
  - Modify condition of `The docker client is executed in a container` rule.
- Why we need it
  - On GKE, The rule has many false positives because default running container uses kubectl.
  - To reduce false positives, and to be able to set additional condition, I would like to add `user_known_k8s_client_container` macro to this rule.

Fixes #

- This Deployment is the cause of false positive on GKE.
  - https://github.com/GoogleCloudPlatform/k8s-stackdriver/tree/master/fluentd-gcp-scaler
  - `kubectl` is used on `scaler.sh`

```release-note
rule(The docker client is executed in a container): when executing the docker client, exclude containers running in the `kube-system` namespace to avoid false positives
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -2644,10 +2644,13 @@
 
 - list: k8s_client_binaries
   items: [docker, kubectl, crictl]
+
+- macro: user_known_k8s_client_container
+  condition: (k8s.ns.name = "kube-system")
   
 - rule: The docker client is executed in a container
   desc: Detect a k8s client tool executed inside a container
-  condition: spawned_process and container and proc.name in (k8s_client_binaries)
+  condition: spawned_process and container and not user_known_k8s_client_container and proc.name in (k8s_client_binaries)
   output: "Docker or kubernetes client executed in container (user=%user.name %container.info parent=%proc.pname cmdline=%proc.cmdline)"
   priority: WARNING
   tags: [container, mitre_execution]


fntlnz:
Thanks @rung - it looks good to me! I edited the release notes to match what you did since here we have a user facing change.

rung:
@fntlnz Thank you!

Kaizhe:
@rung this change is not good to me:
```
- macro: user_known_k8s_client_container
  condition: (k8s.ns.name = "kube-system")
```
if some bad guy was able to hack into any container in namespace `kube-system` and run `kubectl` (with some of the privileged service accounts), falco will not be able to detect it with your change.

I would propose to change like:
```
condition:  container.image.repository = gcr/fluentd
```

Kaizhe:
And this is GKE specific. Hopefully we will have the cloud provider info to leverage in the future.

rung:
@Kaizhe Thank you for your comment!

How about changing the condition to  this?
```
condition: (k8s.ns.name="kube-system" and container.image.repository=k8s.gcr.io/fluentd-gcp-scaler)
```

- The condition has namespace and image repository. So Falco user can get detections other than fluentd-gcp-scaler.  

> if some bad guy was able to hack into any container in namespace kube-system and run kubectl (with some of the privileged service accounts), falco will not be able to detect it with your change.

Make sense. On the other hand, if we don't specify namespaces, the attack surface could be extended imho.
Because RBAC of Kubernetes is set based on namespaces. And when we don't specify namespaces, user(non-admin) can deploy "gcr/fluentd-scaler" on their namespaces. So they can use "kubectl" using the container.

Kaizhe:
@rung adding `k8s.ns.name="kube-system" is fine with me. Please add comment to the macro saying this is a GKE scenario.

Thanks!

PR #957, opened 2019-12-05
rule(Write below etc): whitelist automount writing under /etc/mtab

marier-nico:
**Type of PR**
/kind rule-update
/area rules

This will reduce false-positives related to `automount` writing to `/etc/mtab`. For the sake of completeness (and to help anyone that might not be aware), "mtab lists currently mounted file systems and is used by the mount and unmount commands when you want to list your mounts or unmount all. It's not used by the kernel, which maintains its own list." (https://serverfault.com/questions/267609/how-to-understand-etc-mtab). `automount` is expected to write under `/etc/mtab`, but there was no macro to whitelist it.

**Which issue(s) this PR fixes**: I didn't open an issue for this.

```release-note
rule(Write below etc): allow automount to write to /etc/mtab
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -1153,6 +1153,9 @@
 - macro: etcd_manager_updating_dns
   condition: (container and proc.name=etcd-manager and fd.name=/etc/hosts)
 
+- macro: automount_using_mtab
+  condition: (proc.pname = automount and fd.name startswith /etc/mtab)
+
 # Add conditions to this macro (probably in a separate file,
 # overwriting this macro) to allow for specific combinations of
 # programs writing below specific directories below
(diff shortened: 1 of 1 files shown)


rules/falco_rules.yaml
```diff
@@ -1153,6 +1153,9 @@
 - macro: etcd_manager_updating_dns
   condition: (container and proc.name=etcd-manager and fd.name=/etc/hosts)
 
+- macro: automount_using_mtab
+  condition: (proc.pname = automount and fd.name startswith /etc/mtab)
```
```suggestion
  condition: (proc.pname=automount and fd.name=/etc/mtab)
```

WDYT?

leodido: After having spent some hours with Nicolas on this we decided to merge PR as is.

He used `startswith` because he had `automount` daemon writing to `/etc/mtab.[something]` files (which to me seems something happening when the `/etc/mtab` file cannot be written because it's locked).

In the next days he's going to observe again the original event in order to understand whether it is better to use equal or startswith here.

leodido:
Merging as is. But pls see note [above](https://github.com/falcosecurity/falco/pull/957#discussion_r354474660). cc @marier-nico

PR #966, opened 2019-12-09
build: add FALCO_COVERAGE CMake option

goller:
**What type of PR is this?**

/kind feature

/area build
/area tests

This adds a new CMake option `Falco_Coverage` that can be used to generate
code coverage analysis.  I used this while building up my test cases for PR #965 

With `cmake FALCO_COVERAGE=on` the `--coverage` option
is passed to both clang and gcc to help analyze untested
portions of the code base.  It produces gcov files.

These files can be analyzed by many tools such as lcov,
gcovr, etc.

Here is an example of one such tool, lcov:

```sh
 lcov --directory . --capture --output-file coverage.info
 lcov --extract coverage.info '/source/*' --output-file coverage.info
 genhtml coverage.info
```

Note: I put the `option` at the top of the CMakeLists file just because my style is to put
all options at the top of the for ease of use. What do you think?

I certainly could move that line near the conditional like all other options.

```release-note
NONE
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,8 @@ cmake_minimum_required(VERSION 3.3.2)
 
 project(falco)
 
+option(FALCO_Coverage "Build test suite with coverage information" OFF)
+
 if(NOT SYSDIG_DIR)
   get_filename_component(SYSDIG_DIR "${PROJECT_SOURCE_DIR}/../sysdig" REALPATH)
 endif()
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -633,6 +635,18 @@ endif()
 install(FILES falco.yaml
 	DESTINATION "${FALCO_ETC_DIR}")
 
+if(FALCO_Coverage)
+    if (NOT (("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU") OR ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")))
+        message(FATAL_ERROR "FALCO_Coverage requires GCC or Clang.")
```
leodido:
```suggestion
        message(FATAL_ERROR "FALCO_COVERAGE requires GCC or Clang.")
```

CMakeLists.txt
```diff
@@ -18,6 +18,8 @@ cmake_minimum_required(VERSION 3.3.2)
 
 project(falco)
 
+option(FALCO_Coverage "Build test suite with coverage information" OFF)
```
leodido:
```suggestion
option(FALCO_COVERAGE "Build test suite with coverage information" OFF)
```

Just for conformance with other option I would personally prefer having the option variable all caps. Just nitpicking you know me 🕺 😄 but I think it can come handy not having to remember the particular case :)

CMakeLists.txt
leodido: I do not have access to your fork @goller. Would you pls also fix the message here with the uppercase please? :)

PR #1041, opened 2020-02-06
Fix memory leaks in k8s audit and json events

### LINKED ISSUE #1040: memory leak while processing the audit endpoint
**Describe the bug**

Falco is not releasing memory when it is processing events on the k8s audit endpoint. We are running several Falco pods and one of them is processing the Kubernetes audit events, which are provided via this project: https://github.com/xebia/falco-eks-audit-bridge. We notice that all pods are fine, except the one that is receiving all the audit events. Below is graph of the memory usage:

Screenshot 2020-02-06 at 11 52 52

As you can see there is clear pattern where the pod gets OOM-killed after 1.5-2 hours. At the end we switched our falco-eks-audit-bridge to a different node and then we saw the exact same pattern with the Falco pod that is running on that node (it is offering the audit events via a localhost connection). 

**How to reproduce it**

We're still running some tests to confirm, but it should be easy to reproduce by just load-testing the audit endpoint with audit json logs.

**Expected behaviour**

Memory usage should remain stable, as it did in previous versions of Falco.

**Environment**

- Falco version:
0.19.0 with a 'minimal' build and ebpf probe.

- System info: release":"4.14.154-128.181.amzn2.x86_64"
- Cloud provider or hardware configuration: AWS
- Installation method: Kubernetes

fntlnz:
**What type of PR is this?**

/kind bug

Fixes the memory leak bug reported by many users on the Falco slack.

WRT I started doing some memory analysis on my dev environment and noticed a pattern I didn’t like:

All the snapshots I was able to take were showing a linear increase in memory usage

image

Here’s an evidence of the leak, in how audit events are processed.

image

The problem I observed is that whenever the add_extracted_value function (https://github.com/falcosecurity/falco/blob/master/userspace/engine/json_evt.cpp#L629-L639) is called it adds an element to a list of strings. After that, old elements are never flushed. That leads to the memory leak we can see in the massif output above.

After some debugging, I noticed that add_extracted_value was only called when json_event_filter_check has m_uses_paths set to true. However, that value is never explicitly set to true in that codebase.

Then I noticed that whenever a new filtercheck is created (code : https://github.com/falcosecurity/falco/blob/master/userspace/engine/json_evt.cpp#L1379) the function allocate_new is called.

However, the created filter check does not contain false as a value for m_uses_paths but a numerical value. The reason for this is that while doing type conversion.
And since in C++ only zero is false and everything else is true that’s why we call add_extracted and have the memory leak.

This is the situation after my fix, as you can see the memory is pretty much always allocated and immediately released after processing.

image

Fixes #1040

```release-note
fix: memory leak introduced in 0.18.0 happening while using json events and the kubernetes audit endpoint
```

### DIFF
--- a/userspace/engine/json_evt.cpp
+++ b/userspace/engine/json_evt.cpp
@@ -512,31 +512,27 @@ const json_event_filter_check::values_t &json_event_filter_check::extracted_valu
 
 bool json_event_filter_check::compare(gen_event *evt)
 {
-	json_event *jevt = (json_event *)evt;
+	auto jevt = (json_event *)evt;
 
 	uint32_t len;
 
-	const extracted_values_t *evalues = (const extracted_values_t *) extract(jevt, &len);
+	auto evalues = (const extracted_values_t *) extract(jevt, &len);
 	values_set_t setvals;
 
 	switch(m_cmpop)
 	{
 	case CO_EQ:
 		return evalues->second == m_values;
-		break;
 	case CO_NE:
 		return evalues->second != m_values;
-		break;
 	case CO_STARTSWITH:
 		return (evalues->first.size() == 1 &&
 			m_values.size() == 1 &&
 			evalues->first.at(0).startswith(*(m_values.begin())));
-		break;
 	case CO_CONTAINS:
 		return (evalues->first.size() == 1 &&
 			m_values.size() == 1 &&
 			evalues->first.at(0).contains(*(m_values.begin())));
-		break;
 	case CO_IN:
 		for(auto &item : evalues->second)
 		{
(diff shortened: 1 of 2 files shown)


fntlnz:
Here is the massif file of falco before the fix

[massif.out-before.zip](https://github.com/falcosecurity/falco/files/4165398/massif.out-before.zip)

And this is after:
[massif.out-after.zip](https://github.com/falcosecurity/falco/files/4165400/massif.out-after.zip)

fntlnz:
Artifacts to test this can be found here: https://circleci.com/gh/falcosecurity/falco/341#artifacts/containers/0

rajibmitra:
great findings ! Yes we should be release falco now.

Issue #1034, opened 2020-02-04
sysdig build dependency requires network access


rfrohl:
**Describe the bug**

The `sysdig` dependency during build is pulled from a git repository. This makes it kinda complicated to ship falco in a distribution. I was wondering if there is a specific reason that sysdig is pulled in via git.

**How to reproduce it**

build falco without network access.

**Expected behaviour**

CMake should pull the dependency from the system first and fall back to git.

**Environment**
Not sure if helpful: OBS build environment, trying to package falco for openSUSE.

- Falco version:
0.19.0
- System info:
not applicable
- Cloud provider or hardware configuration:
not applicable
- OS:
```
$ cat /etc/os-release 
NAME="openSUSE Tumbleweed"
# VERSION="20200128"
ID="opensuse-tumbleweed"
ID_LIKE="opensuse suse"
VERSION_ID="20200128"
PRETTY_NAME="openSUSE Tumbleweed"
CPE_NAME="cpe:/o:opensuse:tumbleweed:20200128"
```
- Kernel:
not applicable
- Installation method:
not applicable

**Additional context**

Would be willing to work on a fix, but wanted to get upstream feedback first.

leodido:
Hello @rfrohl, every third-party dependency is taken via the network.

Building without network access is not a goal for us at the moment.

In any case, we can discuss alternatives approaches.

A possible approach could be to provide CMake variables to specify paths for all third-party dependencies, to use them rather than git repositories.

Anyway, we discuss such matters into [community](https://github.com/falcosecurity/community) calls. Do you mind joining the next one (tomorrow) ?

fntlnz:
@rfrohl We take many dependencies from the network, can you please share what is your workflow?
From what I can see it looks like you have three different stages:

stage 1. `git clone` Falco from a machine with internet access
stage 2. Move the Falco dir to a build machine
stage 3. Try to build Falco but it fails because of no internet

We could implement a mechanism to allow you to retrieve dependencies at stage 1 so that when you copy you can just build.

rfrohl:
> We take many dependencies from the network, can you please share what is your workflow?

Sure, sorry that it was unclear in the initial report. It is basically a normal rpm build:
- take the released version (i.e. 0.19.0.tar.gz) and commit it to the source tracker.
- write spec file, which sets up the build environment and pulls all the dependencies from the distro (like protobuf, zlib, openssl...)
- this works for most (all other? ) things beside sysdig
[...]

PR #1054, opened 2020-02-22
Verify checksum after downloading sysdig tarball

afbjorklund:
Also it seemed that any of value of -DSYSDIG_VERSION
failed to propagate, from first cmake to second cmake.

That is, trying to follow instructions didn't have any effect:
```Cmake
# To update sysdig version for the next release, change the default below
# In case you want to test against another sysdig version just pass the variable - ie., `cmake -DSYSDIG_VERSION=dev ..`
```
In the case of a custom version, the checksum is ignored.

**What type of PR is this?**

/kind cleanup

/area build

Avoids warning when building, and checks against corrupt download errors.

`warning: did not verify file - no URL_HASH specified?`

The old commented out checksum is for some random `dev` version:

See 494edafdb19d7b3a6a432d89fbc31a6e82aa0ad6

```release-note
NONE
```

### DIFF
--- a/cmake/modules/sysdig.cmake
+++ b/cmake/modules/sysdig.cmake
@@ -22,7 +22,7 @@ endif()
 
 file(MAKE_DIRECTORY ${SYSDIG_CMAKE_WORKING_DIR})
 # cd /path/to/build && cmake /path/to/source
-execute_process(COMMAND "${CMAKE_COMMAND}" ${SYSDIG_CMAKE_SOURCE_DIR} WORKING_DIRECTORY ${SYSDIG_CMAKE_WORKING_DIR})
+execute_process(COMMAND "${CMAKE_COMMAND}" -DSYSDIG_VERSION=${SYSDIG_VERSION} ${SYSDIG_CMAKE_SOURCE_DIR} WORKING_DIRECTORY ${SYSDIG_CMAKE_WORKING_DIR})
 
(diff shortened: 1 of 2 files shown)


afbjorklund:
This official checksum works great with the workaround in https://github.com/falcosecurity/falco/issues/1034#issuecomment-583824253

Then I can see that whatever I provide, matches the expectations of falco.

cmake/modules/sysdig.cmake
```diff
@@ -22,7 +22,7 @@ endif()
 
 file(MAKE_DIRECTORY ${SYSDIG_CMAKE_WORKING_DIR})
 # cd /path/to/build && cmake /path/to/source
```
fntlnz: I think that `SYSDIG_CHECKSUM` doesn’t get propagated here too now that we have it ?

afbjorklund: Actually that was on purpose, the use case mentioned was something like `cmake -DSYSDIG_VERSION=dev`

And then it was hard to provide a checksum (you need to download the tarball, even if you know the commit)

If you want to checksum also branches/commits provided by the user, it is possible to add another parameter.

For now, it will just download the file and continue with a `no URL_HASH specified?` warning, just like before...

afbjorklund:
Thanks!

PR #1070, opened 2020-03-02
Add a macro to whitelist system binaries using the network

marier-nico:
**What type of PR is this?**

/kind feature

/area rules

This PR introduces two things : a list of allowed system binaries and a macro to allow certain system binaries to use the network under certain conditions.

The list is really just for convenience and is not meant to be overridden by users. It's something I noticed many other rules had and I figured it wouldn't hurt to add it here while I was adding the macro.

The macro is useful because someone might want to allow certain system procs to use the network under certain conditions. Our use-case is that we use a program output which publishes a message to an SNS topic in AWS to be picked up by our other tools. To do this, we use AWS CLI, which obviously needs to talk to the network. The issue is that Falco runs the command `sh -c aws cli ...` and that raises an event because `sh` is a shell binary. The macro resolves this because it allows us to whitelist the command we use (instead of whitelisting the entire `sh` binary).

No issues open for this.

```release-note
Add a macro to allow whitelisting system binaries using the network under specific conditions
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -2007,13 +2007,18 @@
   condition: >
     (fd.sockfamily = ip and (system_procs or proc.name in (shell_binaries)))
     and (inbound_outbound)
-    and not proc.name in (systemd, hostid, id)
+    and not proc.name in (known_system_procs_network_activity_binaries)
     and not login_doing_dns_lookup
   output: >
     Known system binary sent/received network traffic
     (user=%user.name command=%proc.cmdline connection=%fd.name container_id=%container.id image=%container.image.repository)
   priority: NOTICE
   tags: [network, mitre_exfiltration]
+
+# This list allows easily whitelisting system proc names that are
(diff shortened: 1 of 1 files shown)


fntlnz:
This is targeting the wrong branch,  changing the base. We are now using master.

fntlnz:
@Kaizhe PTAL again And thanks @marier-nico for sending over your first PR to Falco :tada:

leodido:
This is the second PR of @marier-nico ! So two times thanks 🤗🤗

marier-nico:
@leodido @fntlnz No problem! :smile: 

I'm not sure about this failing check though, I haven't made any code changes, could the config be invalid because of what I edited?

leodido:
/cc @leodido 

Want to take a look at this to understand why the CI is not being triggered

/hold

fntlnz:
Closing and reopening to let the CI handle this

fntlnz:
Approved, thanks @marier-nico

PR #1081, opened 2020-03-09
fix: use of "tag" instead of "tags" in default rules

vicenteherrera:
Signed-off-by: Vicente Herrera <vicenteherrera@vicenteherrera.com>

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind rule-update

> /kind rule-create

> /area build

> /area engine

> /area examples

/area rules

> /area integrations

> /area tests

> /area proposals

Fix rules that used wrong 'tag' keyword to use the right one, 'tags'.

Some rules where using 'tag' keyword instead of valid 'tags' keyword. That prevented using tags to filter them.

Fixes #

After this change, when using command line switches -T to disable rules by certain tag, or -t to only run rules with a certain tag, a different set with more rules will match as now they are more rules correctly tagged.

Yes

```release-note
fix: -T (to disable rules by certain tag) or -t (to only run rules with a certain tag) work now
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -450,7 +450,7 @@
     a shell configuration file has been modified (user=%user.name command=%proc.cmdline pcmdline=%proc.pcmdline file=%fd.name container_id=%container.id image=%container.image.repository)
   priority:
     WARNING
-  tag: [file, mitre_persistence]
+  tags: [file, mitre_persistence]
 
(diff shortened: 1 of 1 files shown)


fntlnz:
This fixed issue https://github.com/falcosecurity/falco/issues/1082

Issue #1121, opened 2020-03-26
Additional Falco rules, both for syscall and Kubernetes audit sources


vicenteherrera:
**Motivation**

We have been working on 6 rules that we needed using syscall and Kubernetes audit sources, and we want to share them with the community. They are completely generic, and any infrastructure monitored with Falco will benefit from them, so we think they are a better fit here than in the Cloud Native Security Hub project, that is more focused on specific applications and vulnerabilities.

**Feature**

Rules are:
* Full K8s Administrative Access: Detect any k8s operation by an administrator with full access.
* Ingress Object without TLS Certificate Created: Detect any attempt to create an ingress without TLS certification.
* Untrusted Node Successfully Joined the Cluster: Detect a node successfully joined the cluster outside of the list of allowed nodes.
* Untrusted Node Unsuccessfully Tried to Join the Cluster: Detect an unsuccessful attempt to join the cluster for a node not in the list of allowed nodes.
* Network Connection outside Local Subnet: Detect traffic to image outside local subnet.
* Outbound or Inbound Traffic not to Authorized Server Process and Port: Detect traffic that is not to authorized server process and port.

More details in Pull Request: https://github.com/falcosecurity/falco/pull/1122

**Alternatives**

We explored existing Falco and CloudNative Security Hub rules, not finding these kinds of detections.

**Additional context**

More details in Pull Request.

I don't know if for this kind of contribution, creating both an Issue and then a Pull Request is the best way to proceed, or just to do the Pull Request. 😉

PR #1122, opened 2020-03-26
Proposal for 2 new Falco rules for syscall source, and 4 for Kubernetes audit source.

vicenteherrera:
**What type of PR is this?**
/kind feature
/kind rule-create

/area rules

These proposed rules detect network and Kubernetes activity that, outside some whitelists, could indicate malicious activity. No other rules perform similar detections, and they are generic so everybody will benefit for having them in their Falco installation. See the proposed release notes for specific details.

Fixes https://github.com/falcosecurity/falco/issues/1121

The Issue is just a description of the proposal of the rules, I don't know if it would have been better to just create de pull request only.

Yes, new rules available in falco.yaml and k8s_audit_rules.yaml, one of the enabled by default: "Ingress Object without TLS Certificate Created"

```release-note

New Falco rules available:

* Full K8s Administrative Access: Detect any k8s operation by an administrator with full access.
* Ingress Object without TLS Certificate Created: Detect any attempt to create an ingress without TLS certification (rule enabled by default).
* Untrusted Node Successfully Joined the Cluster: Detect a node successfully joined the cluster outside of the list of allowed nodes.
* Untrusted Node Unsuccessfully Tried to Join the Cluster: Detect an unsuccessful attempt to join the cluster for a node not in the list of allowed nodes.
* Network Connection outside Local Subnet: Detect traffic to image outside local subnet.
Outbound or Inbound Traffic not to Authorized Server Process and Port: Detect traffic that is not to authorized server process and port.

```

### DIFF
--- a/rules/k8s_audit_rules.yaml
+++ b/rules/k8s_audit_rules.yaml
@@ -420,19 +420,23 @@
   tags: [k8s]
 
 
-
+# This list includes some of the default user names for an administrator in several K8s installations
 - list: full_admin_k8s_users
-  items: ["admin", "kubernetes-admin",  "kubernetes-admin@kubernetes", "default", "kubernetes-admin@cluster.local", "minikube-user"]
+  items: ["admin", "kubernetes-admin",  "kubernetes-admin@kubernetes", "kubernetes-admin@cluster.local", "minikube-user"]
 
-- macro: allowed_full_admin_users
-  condition: (k8s_audit_always_true)
+# This rules detect an operation triggered by an user name that is 
+# included in the list of those that are default administrators upon 
+# cluster creation. This may signify a permission setting too broader. 
+# As we can't check for role of the user on a general ka.* event, this 
+# may or may not be an administrator. Customize the full_admin_k8s_users 
+# list to your needs, and activate at your discrection.
 
 # # How to test:
 # # Execute any kubectl command connected using default cluster user, as:
 # kubectl create namespace rule-test
 
 - rule: Full K8s Administrative Access
-  desc: Detect any k8s operation by an administrator with full access.
+  desc: Detect any k8s operation by a user name that may be an administrator with full access.
   condition: >
     kevt 
     and non_system_user 
--- a/rules/k8s_audit_rules.yaml
+++ b/rules/k8s_audit_rules.yaml
@@ -418,3 +418,110 @@
   priority: DEBUG
   source: k8s_audit
   tags: [k8s]
+
+
+
+- list: full_admin_k8s_users
+  items: ["admin", "kubernetes-admin",  "kubernetes-admin@kubernetes", "default", "kubernetes-admin@cluster.local", "minikube-user"]
+
+- macro: allowed_full_admin_users
+  condition: (k8s_audit_always_true)
+
+# # How to test:
+# # Execute any kubectl command connected using default cluster user, as:
+# kubectl create namespace rule-test
+
+- rule: Full K8s Administrative Access
+  desc: Detect any k8s operation by an administrator with full access.
+  condition: >
+    kevt 
+    and non_system_user 
+    and ka.user.name in (admin_k8s_users) 
+    and not allowed_full_admin_users
+  output: K8s Operation performed by full admin user (user=%ka.user.name target=%ka.target.name/%ka.target.resource verb=%ka.verb uri=%ka.uri resp=%ka.response.code)
+  priority: WARNING
+  source: k8s_audit
+  tags: [k8s]
+
+
+
+- macro: ingress
+  condition: ka.target.resource=ingresses
+
+- macro: ingress_tls
+  condition: (jevt.value[/requestObject/spec/tls] exists)
+
+# # How to test:
+# # Create an ingress.yaml file with content:
+# apiVersion: networking.k8s.io/v1beta1
+# kind: Ingress
+# metadata:
+#   name: test-ingress
+#   annotations:
+#     nginx.ingress.kubernetes.io/rewrite-target: /
+# spec:
+#   rules:
+#   - http:
+#       paths:
+#       - path: /testpath
+#         backend:
+#           serviceName: test
+#           servicePort: 80
+# # Execute: kubectl apply -f ingress.yaml
+
+- rule: Ingress Object without TLS Certificate Created
+  desc: Detect any attempt to create an ingress without TLS certification.
+  condition: >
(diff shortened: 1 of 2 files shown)


rules/k8s_audit_rules.yaml
```diff
@@ -418,3 +418,110 @@
   priority: DEBUG
   source: k8s_audit
   tags: [k8s]
+
+
+
+- list: full_admin_k8s_users
+  items: ["admin", "kubernetes-admin",  "kubernetes-admin@kubernetes", "default", "kubernetes-admin@cluster.local", "minikube-user"]
```
Kaizhe: why `default` is admin user?

rules/k8s_audit_rules.yaml
```diff
@@ -418,3 +418,110 @@
...
+
+- macro: allowed_full_admin_users
+  condition: (k8s_audit_always_true)
+
+# # How to test:
+# # Execute any kubectl command connected using default cluster user, as:
+# kubectl create namespace rule-test
+
+- rule: Full K8s Administrative Access
```
Kaizhe: Full k8s administrative access should be based on the roles granted to the user, not the user name. I would put this rule on hold or disabled by default.

rules/k8s_audit_rules.yaml
```diff
@@ -418,3 +418,110 @@
...
+
+
+
+- macro: node
+  condition: ka.target.resource=nodes
+
+- macro: allow_all_k8s_nodes
+  condition: (k8s_audit_always_true)
+
+- list: allowed_k8s_nodes
```
Kaizhe: most likely the decision shouldn't be made by name. Let's create a list of trusted node labels and only nodes contains any trusted labels can join the cluster.

rules/k8s_audit_rules.yaml
vicenteherrera: K3s admin user is called "default", when /etc/rancher/k3s/k3s.yaml configuration file is created.

The best direct link I can find to share about this is: https://github.com/rancher/k3s/issues/140#issuecomment-469338111

rules/k8s_audit_rules.yaml
vicenteherrera: As Falco supported fields indicate, role information is only available in the Kubernetes audit event "when the request object refers to a cluster role binding" or "when the request object refers to a role/cluster role". https://falco.org/docs/rules/supported-fields/

So for any general event, we can only rely on user name to check this.

Also the rule is disabled by default, including a condition "and not allowed_full_admin_users" to a macro that is defined as "k8s_audit_always_true".

Kaizhe: The rules works only for k8s audit events. Relying on username is not the right thing to do. And as I suggest, this is something I want to put it on hold or disable by default.

Kaizhe: There is a `enabled` field in the rule you can specify.

rules/k8s_audit_rules.yaml
Kaizhe: I don't know whether falco supports k3s ? And also whether the audit events from k3s is compatible with k8s. If you can provide more information here, that would be great!

rules/k8s_audit_rules.yaml
vicenteherrera: The user name that executed the Kubernetes command is part of the k8s event in ka.user.name (but not the role). Another existing rule "Disallowed K8s User" also list allowed user names as a way to get a detection if a different one executes a command (this new rule is more or less the opposite).

I like more the clear semantics of the enabled field, but right now the only way to change a "enabled: false" to true would be to edit the original k8s_audit_rules.yaml. That would lead to problems when rules get updated. Macros, on the other hand, can be overwritten in falco_rules.loca.yaml or other files to enable rules more easily. I should post a feature request for a better way to enable existing disabled rules.

rules/k8s_audit_rules.yaml
vicenteherrera: Maybe they are not, and we overstepped here. k3s have an audit webhook. Let me make some test but it will take some time.

rules/k8s_audit_rules.yaml
Kaizhe: That's the intention for `enabled: false` and it is up to the user to enable it. In OSS community, we can worry less about the commercial product impact regarding patching rules.

rules/k8s_audit_rules.yaml
Kaizhe: it may be good to remove `default` in this PR, once you confirm falco support k3s and k3s audit is event is compatible with k8s, then we can add it back.

rules/k8s_audit_rules.yaml
vicenteherrera: Ok, but let me try to discuss this a little bit

So, you can add labels to a node at the same time that it joins the cluster. But we feel that if someone has the kind of access level to your cluster to add a node to it, he may well replicate the labels on the existing nodes to the new one, in an effort to make it look similar to the other ones.

The node name is a more unique way to represent trust in nodes. Someone may see other labels in existing nodes, but it's more difficult that he is aware of the specific list of allowed node names before attempting to join. And when that happens, Falco is going to alert you. 

Tell me what you think about this. ;-)

rules/k8s_audit_rules.yaml
vicenteherrera: Remove the `default` user, and clarified comments on the rule to make clear that we are detecting only usernames and not roles.

rules/k8s_audit_rules.yaml
vicenteherrera: OSS is my concern here. I don't think it should be encouraged to make people modify falco_rules.yaml or k8s_audit_rules.yaml directly for rules activation or customization, as it is going to make it harder for them to keep up with updates. I believe that's why falco_rules.local.yaml and macros with (never_true) and similar values exist on all other rules. `enabled:false` as it is defined now is not used in any other rule, they all rely on macros for activation. 
Just an opinion, of course, tell me if you still think it's necessary.

rules/k8s_audit_rules.yaml
Kaizhe: Managing by node name will be painful. As this is the first version, I will let it in and we can come back to improve it.

leodido:
/hold 

need to adjust release-note block
need review

leodido:
This is in!

/milestone 0.22.0

/hold cancel

PR #1137, opened 2020-04-09
rule(Write below root): use pmatch to check against known root direct…

Kaizhe:
…ories

Signed-off-by: kaizhe <derek0405@gmail.com>

**What type of PR is this?**

 /kind rule-update

 /area rules

When check against directories, we want to also check the subdirectories as well, so `in` is not the best option, use `pmatch` instead.

Fixes #

```release-note
rule(Write below root): use pmatch to check against known root directories
```


fntlnz:
I agree with @mstemm however this looks like a necessary evil for now since we don't really have a way to map everything.

PR #1151, opened 2020-04-17
Add secrets created deleted k8s audit rules

mstemm:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind rule-update

/kind rule-create

> /area build

> /area engine

> /area examples

/area rules

> /area integrations

> /area tests

> /area proposals

Add new rules to track creating/deleting K8s Secrets. kube-system and service account secrets are excluded.

Fixes #

```release-note
rule(K8s Secret Created): new rule to track the creation of Kubernetes secrets (excluding kube-system and service account secrets)
rule(K8s Secret Deleted): new rule to track the deletion of Kubernetes secrets (excluding kube-system and service account secrets)
```


rules/k8s_audit_rules.yaml
```diff
@@ -102,6 +102,12 @@
 - macro: role
   condition: ka.target.resource=roles
 
+- macro: secret
+  condition: ka.target.resource=secrets
+
+- macro: req_service_account_token
```
Kaizhe, line 108: Mark, we cannot set the audit level to `Request` or `RequestResponse` or the secret data will be recorded in the audit log.

mstemm: This isn't controlling the audit level though. This macro just looks for the type of the object for the request being a service account token.

You are right, though, that the recommended audit policy logs secrets at the RequestResponse level: https://github.com/falcosecurity/falco/blob/master/examples/k8s_audit_config/audit-policy.yaml#L59. How about we also change the recommended policy to log secrets at Metadata level and then change these rules to exclude requests starting with "system:" ?

Kaizhe: The macro reminds me that it is must from audit level `Request` or `RequestResponse`. How about having a `trusted_users` list?

mstemm: Yeah I'll change it to a users list.

PR #1158, opened 2020-04-23
Updates to the falco-driver-loader script

### LINKED ISSUE #1125: falco-driver-loader improvements proposal
**Motivation**

As part of the ongoing discussions about the [scope of Falco artifacts](https://github.com/falcosecurity/falco/issues/1114) we need an officially supported tool for building, downloading, and installing various drivers. Currently, the `falco-driver-loader` takes care of that, and it is already included in the new [stable binary distribution](https://dl.bintray.com/falcosecurity/bin/x86_64/). 

Furthermore, the source of `falco-driver-loader` needs to stay into the Falco repo,  so it can easily be kept in sync with Falco and the right parameters can be [substituted into the script](https://github.com/falcosecurity/falco/blob/master/scripts/falco-driver-loader#L387) during the building process. Also, having it in the same place helps with maintenance.

That being said, there are other tools that need to perform similar actions: for example [falcoctl installs the kernel module](https://github.com/falcosecurity/falcoctl/blob/master/pkg/kernelmoduleloader/kernel_module_loader.go) by replicating some of the logic already included in `falco-driver-loader`. Moreover, the installation logic can change over time and we will need a way to version this process too.

For these reasons, I would like to propose several improvements in order to make it usable by other tools and avoid duplicating the same logic in other places.

**Feature**

- [ ] Make explicit that `falco-driver-loader` is versioned alongside Falco (mostly a documentation issue)
- [ ] Do not unload the driver if is already present and the version matches (needed to avoid problems with manually loaded drivers)
- [ ] ~~Distribute another artifact that includes just `falco-driver-loader` so it can be quickly downloaded and used by both users and other tools (ie. `falcoctl`), the artifact will be versioned as same as Falco~~ :point_right: https://github.com/falcosecurity/falco/issues/1159
- [ ] Add a simple interface that allows executing single actions (ie. only download, only install, or only build) so it can be easily used by other tools (ie. `falcoctl`)
- [x] Update `falco-driver-loader` in order to download artifact from https://bintray.com/falcosecurity

**Alternatives**

**Additional context**

leodido:
**What type of PR is this?**

/kind cleanup

/kind feature

/area build

This PR refactors the `falco-driver-builder` in order to support the downloading from the upcoming "drivers build grid".

Fixes #1125 

Docs need to be updated in order to show to users an alternative approach of installing the Falco driver through the `falco-driver-loader` script (present into the container).

```release-note
update: containers look for prebuilt drivers on the Drivers Build Grid
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,9 +71,9 @@ set(CMAKE_CXX_FLAGS_RELEASE "-O3 -fno-strict-aliasing -DNDEBUG")
 include(GetFalcoVersion)
 
 set(PACKAGE_NAME "falco")
-set(PROBE_NAME "falco-probe")
+set(PROBE_NAME "falco")
 set(PROBE_DEVICE_NAME "falco")
-set(DRIVER_LOOKUP_URL "https://s3.amazonaws.com/download.draios.com")
+set(DRIVERS_REPO "https://bintray.com/falcosecurity/driver")
 if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
   set(CMAKE_INSTALL_PREFIX
       /usr
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,7 +73,7 @@ include(GetFalcoVersion)
 set(PACKAGE_NAME "falco")
 set(PROBE_NAME "falco")
 set(PROBE_DEVICE_NAME "falco")
-set(DRIVERS_REPO "https://bintray.com/falcosecurity/driver")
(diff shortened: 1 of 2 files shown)


leogr:
Hey @leodido

Side note (just a reminder for me):
Although this PR  mostly fixes #1125, one point is missing (about separately distributing the script as we discussed again during yesterday's community call).
Thus, I wanted just to let you know I'm going to open a separate issue that we can fix later, so there's nothing to modify in this PR.

Moreover, I'm really happy to see this PR :partying_face: 
Thank you

PR #1200, opened 2020-05-08
falco-driver-loader - break apart logic and improvements

### LINKED ISSUE #1125: falco-driver-loader improvements proposal

### LINKED ISSUE #1193: Break apart logic in `falco-driver-loader` or rename the script
**Motivation**

Right now our documentation and our calls use the term `loader` to define the behavior of [this script](https://github.com/falcosecurity/falco/blob/master/scripts/falco-driver-loader).

I think that the name could be improved to clarify exactly what the script is doing.

I also believe that the script should be dynamic and configurable to perform single tasks without making assumptions on behalf of the user. 

**Feature**

Right now the script performs multiple phases of logic within monolithic functions that perform more than one task.

For instance `load_kernel_module` does more than just `load` something. It also tries to download a kernel module, it also tries to compile a kernel module. 

I think we should take one of 2 paths

1) Rename the script to something that describes what it does.

```bash
./falcodriver-loader-downloader-compiler
```

2) Fix the script so it is modular and can be used as a single use case

```bash
# This will try to download first, and then attempt to compile if the download fails
./falcodriver module --download --compile 

# This will never download, and will only try to compile
./falcodriver bpf --compile

# This will first try to use DKMS to load the kernel module if it exists on the system, then download, then compile
./falcodriver module --load --download --compile

# This will do exactly what it does now and try to detect and succeed at all costs
./falcodriver # no arguments
```

**Alternatives**

**Additional context**

leogr:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind rule-update

> /kind rule-create

> /area build

> /area engine

> /area examples

> /area rules

> /area integrations

> /area tests

> /area proposals

See #1193

Fixes #1193
Fixes #1125 (again)

```shell
Usage:
  falco-driver-loader [driver] [options]

Available drivers:
  module        kernel module (default)
  bpf           eBPF probe

Options:
  --help        show brief help
  --compile     try to compile the driver locally
  --download    try to download a prebuilt driver
```

With no options and arguments, it preserves the previous behavior (we should test it deeply).

Notable improvements in messages:
 - All info messages start with `* Message...`
 - Success messages with `* Success: message...`
 - Messages without `*` represent the output of another program or an error
 - Verbosity increased 

/cc @kris-nova 
/cc @leodido 
/cc @fntlnz 

### falco-driver-loader (pre refactoring behavior)

Preflight check:
- if not root, then exit 1
- if not `curl`, then exit 1

#### module

Preflight check:
- if not `lsmod`, then exit 1
- if not `modprobe`, then exit 1
- if not `rmmod`, then exit 1

1. Unload module, if present
2. If module still loaded, then exit 0
3. If UEK host, then
    - do nothing
    *else*
    - if `dkms install` and `insmod` then exit 0
4. if `modprobe` module, then exit 0
5. if unsupported OS, then exit 1 (see `get_target_id`)
6. if `~/.falco/falco_<target_id>_<kernel_release>_<kernel_version>.ko`, then:
    - if `insmod` then exit 0, else exit 1
7. if `curl` module, then
    - if `insmod` then exit 0, else exit 1
    *else*
    - exit 1
8. exit 0

#### bpf 

Preflight check:
- Mount debugfs (continue on error)
- if kernel config not found, exit 1 (see `get_kernel_config`)
- if unsupported OS, then exit 1 (see `get_target_id`)

1. if NOT `~/.falco/falco_<target_id>_<kernel_release>_<kernel_version>.o`, then go to step 6.
    - if COS, MINKUBE, or `BPF_USE_LOCAL_KERNEL_SOURCES` then download kernel source
        - if download failed, then exit 1
        - `customize_kernel_build`
    - make bpf probe to `~/.falco/falco_<target_id>_<kernel_release>_<kernel_version>.o`
2. if NOT `~/.falco/falco_<target_id>_<kernel_release>_<kernel_version>.o`, 
    - `curl` bpf probe (continue on error)
3. if `~/.falco/falco_<target_id>_<kernel_release>_<kernel_version>.o`, then
    - if symlink to `~/.falco/falco-bpf.o` then exit 0, else exit 1
    *else*
    - exit 1
4. exit 0

**N.B.**: **bpf** exists with 0 even though `curl` failed (see step 2.)

```release-note
update(scripts): improve `falco-driver-loader` output messages
new(scripts): options and command-line usage for `falco-driver-loader`
```

### DIFF
--- a/scripts/falco-driver-loader
+++ b/scripts/falco-driver-loader
@@ -68,29 +68,29 @@ cos_version_greater()
 
 get_kernel_config() {
 	if [ -f /proc/config.gz ]; then
-		echo "Found kernel config at /proc/config.gz"
+		echo "* Found kernel config at /proc/config.gz"
 		KERNEL_CONFIG_PATH=/proc/config.gz
 	elif [ -f "/boot/config-${KERNEL_RELEASE}" ]; then
-		echo "Found kernel config at /boot/config-${KERNEL_RELEASE}"
+		echo "* Found kernel config at /boot/config-${KERNEL_RELEASE}"
 		KERNEL_CONFIG_PATH=/boot/config-${KERNEL_RELEASE}
 	elif [ -n "${HOST_ROOT}" ] && [ -f "${HOST_ROOT}/boot/config-${KERNEL_RELEASE}" ]; then
-		echo "Found kernel config at ${HOST_ROOT}/boot/config-${KERNEL_RELEASE}"
+		echo "* Found kernel config at ${HOST_ROOT}/boot/config-${KERNEL_RELEASE}"
 		KERNEL_CONFIG_PATH="${HOST_ROOT}/boot/config-${KERNEL_RELEASE}"
 	elif [ -f "/usr/lib/ostree-boot/config-${KERNEL_RELEASE}" ]; then
-		echo "Found kernel config at /usr/lib/ostree-boot/config-${KERNEL_RELEASE}"
+		echo "* Found kernel config at /usr/lib/ostree-boot/config-${KERNEL_RELEASE}"
 		KERNEL_CONFIG_PATH="/usr/lib/ostree-boot/config-${KERNEL_RELEASE}"
 	elif [ -n "${HOST_ROOT}" ] && [ -f "${HOST_ROOT}/usr/lib/ostree-boot/config-${KERNEL_RELEASE}" ]; then
-		echo "Found kernel config at ${HOST_ROOT}/usr/lib/ostree-boot/config-${KERNEL_RELEASE}"
+		echo "* Found kernel config at ${HOST_ROOT}/usr/lib/ostree-boot/config-${KERNEL_RELEASE}"
 		KERNEL_CONFIG_PATH="${HOST_ROOT}/usr/lib/ostree-boot/config-${KERNEL_RELEASE}"
 	elif [ -f "/lib/modules/${KERNEL_RELEASE}/config" ]; then
 		# this code works both for native host and agent container assuming that
 		# Dockerfile sets up the desired symlink /lib/modules -> $HOST_ROOT/lib/modules
-		echo "Found kernel config at /lib/modules/${KERNEL_RELEASE}/config"
+		echo "* Found kernel config at /lib/modules/${KERNEL_RELEASE}/config"
 		KERNEL_CONFIG_PATH="/lib/modules/${KERNEL_RELEASE}/config"
 	fi
 
 	if [ -z "${KERNEL_CONFIG_PATH}" ]; then
-		echo "Cannot find kernel config"
+		>&2 echo "Cannot find kernel config"
 		exit 1
 	fi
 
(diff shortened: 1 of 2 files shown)


scripts/falco-driver-loader
```diff
@@ -420,18 +480,100 @@ fi
...
+	ENABLE_DOWNLOAD="yes"
+fi
+
 if [ "$(id -u)" != 0 ]; then
-	echo "Installer must be run as root (or with sudo)."
+	>&2 echo "This program must be run as root (or with sudo)"
 	exit 1
 fi
 
 if ! hash curl > /dev/null 2>&1; then
```
leodido, line 566: Can you please add a check for `dkms` when user wants to compile?

leogr: It is already present [here](https://github.com/falcosecurity/falco/pull/1200/files#diff-b1a242527066116a4e7485802c03bbb3R154) (see the `else` branch).
The check will not terminate the execution as like as the previous script version.

scripts/falco-driver-loader
```diff
@@ -140,6 +139,84 @@ get_target_id() {
...
+					echo "* Running dkms build failed, couldn't find ${DKMS_LOG}"
+				fi
+			fi
+		else
+			echo "* Skipping dkms install (dkms not found)"
+		fi
+	fi
+
+	if [ -n "$SKIP_LOAD" ]; then
+		if modprobe "${DRIVER_NAME}" > /dev/null 2>&1; then
```
leogr, line 186: It should be a `--dry-run`

scripts/falco-driver-loader
leogr: Furthermore, we need to continue the execution so the download strategy can happen.
When `dkms` does not exist, users will see `* Skipping dkms install (dkms not found)`

scripts/falco-driver-loader
```diff
@@ -420,18 +480,100 @@ fi
...
+      	ENABLE_COMPILE="yes"
+	    has_opts="true"
+      shift
+      ;;
+    --download)
+	    ENABLE_DOWNLOAD="yes"
+	    has_opts="true"
+      shift
+      ;;
+	--skip-load)
```
leodido, line 539: Indentation

scripts/falco-driver-loader
```diff
@@ -420,18 +480,100 @@ fi
...
+	    has_args="true"
+      shift
+    fi
+      ;;
+    -h|--help)
+      print_usage
+      exit 0
+      ;;
+    --compile)
+      	ENABLE_COMPILE="yes"
```
leodido, line 530: Indentation

scripts/falco-driver-loader
```diff
@@ -177,226 +241,222 @@ load_kernel_module() {
...
+			load_bpf_probe_compile
+		fi
+	fi
 
-		curl -L --create-dirs "${FALCO_DRIVER_CURL_OPTIONS}" -o "${HOME}/.falco/${BPF_PROBE_FILENAME}" "${URL}"
+	if [ -n "$ENABLE_DOWNLOAD" ]; then
+		if [ -f "${HOME}/.falco/${BPF_PROBE_FILENAME}" ]; then
+			echo "* Skipping download, eBPF probe is already present in ${HOME}/.falco/${BPF_PROBE_FILENAME}"
+		else
+			load_bpf_probe_download
```
leodido, line 440: Check the operation was successful so to give hints to the user

leogr:
> The only thing I'm not really sure about is the usage of `check_skip_load` to circuit-break only the load phase of the drivers (not their installation).
> 
> So, implemented in this way, what's the use case of `--skip-load`? Maybe I'm missing something..

My thought when I saw the proposal in the issue was... we might need this for testing purposes only. Ofc, Falco will not work if `--skip-load` is used.

Should we remove it?

leogr:
Hey @fntlnz 
I'm working on tests, so we can add them.

That being said, really important design question: after talking with @leodido I believe we should remove the `--skip-load` option. WDYT?
If you agree on to remove it too, I will remove it ASAP.

leodido:
I think `--skip-load` flag should be removed: it adds more complexity than it removes. 

The `--download` and `--compile` flags are pretty explanatory and they cover the use cases

My 2 cents

scripts/falco-driver-loader
leogr: Fixed here :point_right: https://github.com/falcosecurity/falco/pull/1200/files#diff-b1a242527066116a4e7485802c03bbb3R378

leogr:
Update:
- `--skip-load` has been removed
- do not continue when epbf download fails

leogr:
**IMPORTANT**

I have noticed that:
```
	echo "* Trying to load a system ${DRIVER_NAME} driver, if present"
	if modprobe "${DRIVER_NAME}" > /dev/null 2>&1; then
		echo "* Success: ${DRIVER_NAME} module found and loaded with modprobe"
		exit 0
	fi
```
might load a kernel module that does not fit the installed Falco version. Currently, it happens before trying a local prebuilt and the download strategy, but that should be the last resort.

leogr:
Rebased on top of master to include the fix introduced by https://github.com/falcosecurity/falco/pull/1212

I removed WIP, because it's ok for me, though it's not possible to test every single case.

@fntlnz @leodido So, question: do we want this PR to be merged for the 0.23.0?

leogr:
Rebased on top of master to include lastest fixes.

PR #1213, opened 2020-05-15
rule(macro user_known_set_setuid_or_setgid_bit_conditions): create macro

marier-nico:
**What type of PR is this?**

/kind feature
/kind rule-update

/area rules

This macro will be useful because it will make it possible to filter out
events with a higher degree of granularity than is currently possible
for the `Set Setuid or Setgid bit` rule.

For example, if some application is expected to set the setuid or the
setgid bit under a specific condition, like if it's started with a
specific command, then the `user_known_chmod_applications` list is not
enough because we don't want to filter out _all_ events by this
application, only specific ones. This macro allows that.

A concrete example of this is when using docker-in-docker : the `dockerd` process is triggering the `Set Setuid or Setgid bit` rule and every time it does, `proc.cmdline` is the same. This is expected in our case, but we don't want to just ignore _all_ events from `dockerd` within this rule, just this specific one with this specific `proc.cmdline` value.

None

```release-note
rule(macro user_known_set_setuid_or_setgid_bit_conditions): create macro
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -2514,6 +2514,12 @@
 - list: user_known_chmod_applications
   items: [hyperkube, kubelet]
 
+# This macro should be overridden in user rules as needed. This is useful if a given application
+# should not be ignored alltogether with the user_known_chmod_applications list, but only in
+# specific conditions.
+- macro: user_known_set_setuid_or_setgid_bit_conditions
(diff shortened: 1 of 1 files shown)


Kaizhe:
Thanks!
/lgtm

fntlnz:
reopening to let the CI do its thing

PR #1217, opened 2020-05-18
Unix socket for the gRPC server

### LINKED ISSUE #858: GRPC Output: Also support unix domain sockets
When we add GRPC Support to falco, it would be good to have an option to support connecting to the grpc server via a unix domain socket in addition to a tcp connection.

**Why is this needed**:
This allows for more security for use cases where clients can be colocated with falco and allows for a slightly smoother process when developing with falco, where you might not want to generate tls certificates.

fntlnz:
**What type of PR is this?**

/kind feature

/area engine

Fixes #858

```release-note
new: unix socket for the gRPC server
```


falco.yaml
```diff
@@ -167,21 +167,35 @@ http_output:
...
+# By default, the gRPC server is disabled, with no enabled services (see grpc_output)
+# please comment/uncomment and change accordingly the options below to configure it.
+# Important note: if Falco has any troubles creating the gRPC server
+# this information will be logged, however the main Falco daemon will not be stopped.
+# gRPC server over network with (mandatory) mutual TLS configuration.
+# This gRPC server is secure by default so you need to generate certificates and update their paths here.
 # By default the gRPC server is off.
 # You can configure the address to bind and expose it.
 # By modifying the threadiness configuration you can fine-tune the number of threads (and context) it will use.
+# grpc:
```
mstemm, line 182: I don't have a strong preference, but maybe we should make a unix socket the default?

PR #1222, opened 2020-05-19
Change chread namespace

marier-nico:
**What type of PR is this?**

/kind feature
/kind rule-update

/area rules

**Editing the `exe_running_docker_save` macro**:
While using Falco, I noticed we were getting many events that were virtually identical to those that were previously filtered out by the `exe_running_docker_save` macro, but where the `cmdline` was something like `exe /var/run/docker/netns/cc5c7b9bb110 all false`. I believe this is caused by the use of docker-in-docker. This is removed by the new `or proc.cmdline contains "/var/run/docker"` part of the macro.

**`Change thread namespace` rule edit**:
These application binaries raise events in the `Change thread namespace` rule as part of their normal operation.

Here are more details regarding each binary :

- `protokube` : See [this](https://github.com/kubernetes/kops/tree/master/protokube)
- `dockerd` : The `dockerd` process name is whitelisted already in this
  rule, but not if it is the parent, which will happen if you are doing
  docker-in-docker.
- `tini` : See [this](https://github.com/krallin/tini)
- `aws` : This one I noticed because Falco itself uses the AWS CLI to
  send events to SNS, which was triggering this rule.

None

```release-note
Allow `protokube`, `dockerd`, `tini` and `aws` binaries to change thread namespace.
Edit `exe_running_docker_save` to filter out cmdlines containing `/var/run/docker`.
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -1551,7 +1551,7 @@
     and not proc.name in (user_known_change_thread_namespace_binaries)
     and not proc.name startswith "runc"
     and not proc.cmdline startswith "containerd"
-    and not proc.pname in (sysdigcloud_binaries, hyperkube, kubelet)
+    and not proc.pname in (sysdigcloud_binaries, hyperkube, kubelet, protokube, dockerd, tini, aws)
     and not python_running_sdchecks
     and not java_running_sdjagent
     and not kubelet_running_loopback
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -848,7 +848,8 @@
 - macro: exe_running_docker_save
   condition: >
     proc.name = "exe"
-    and proc.cmdline contains "/var/lib/docker"
+    and (proc.cmdline contains "/var/lib/docker"
+    or proc.cmdline contains "/var/run/docker")
     and proc.pname in (dockerd, docker)
 
 # Ideally we'd have a length check here as well but sysdig


Kaizhe:
@marier-nico thanks for submitting this PR. I like the changes to the rule `Change thread namesapce`. However, you loosen too much in the macro `exe_running_docker_save`. I am not convinced by the changes.

marier-nico:
> @marier-nico thanks for submitting this PR. I like the changes to the rule `Change thread namesapce`. However, you loosen too much in the macro `exe_running_docker_save`. I am not convinced by the changes.

@Kaizhe yeah, I was hoping to discuss that a bit! I had the feeling it would be too loose, but I didn't get much feedback in the Slack channel, so I figured it could be discussed in this PR. The thing is, I'm not really sure what to do with that macro, because there are thousands of events that I'm getting where `proc.pname` is just `<NA>`, is it possible that there is a bug somewhere in Falco? To the best of my knowledge, there wasn't any massive change to our infrastructure that would have spawned those events.

Kaizhe:
> > @marier-nico thanks for submitting this PR. I like the changes to the rule `Change thread namesapce`. However, you loosen too much in the macro `exe_running_docker_save`. I am not convinced by the changes.
> 
> @Kaizhe yeah, I was hoping to discuss that a bit! I had the feeling it would be too loose, but I didn't get much feedback in the Slack channel, so I figured it could be discussed in this PR. The thing is, I'm not really sure what to do with that macro, because there are thousands of events that I'm getting where `proc.pname` is just `<NA>`, is it possible that there is a bug somewhere in Falco? To the best of my knowledge, there wasn't any massive change to our infrastructure that would have spawned those events.

can you describe more about your scenario?

marier-nico:
> can you describe more about your scenario?

@Kaizhe Sure! So it's a bit odd, but I just went in our log stack to find some sample events to show you what types of events I wanted to ignore, and I couldn't find any that had `proc.pname=<NA>`. I'm not sure what happened to those events, because I _definitely_ know I saw a bunch like that, although it might have been an isolated incident.

Given that I can't find those events anymore, I think I'll just remove the commit that makes the macro too permissive! :smile:

Kaizhe:
@leodido test hangs?

fntlnz:
Thanks again @marier-nico !

fntlnz:
If you see an edit in the description, I just edited the release notes to follow [our convention](https://github.com/falcosecurity/falco/blob/master/CONTRIBUTING.md#commit-convention)

PR #1237, opened 2020-05-28
Check keepalive first, and throttle try_pop

scallopboat:
Signed-off-by: Chris Meyer <scallopboat@gmail.com>

**What type of PR is this?**
/kind bug

**Any specific area of the project related to this PR?**
/area engine

**What this PR does / why we need it**:
Reduces the CPU usage of falco when the grpc server is enabled to ~15%.

**Which issue(s) this PR fixes**:
Fixes #1126 

**Special notes for your reviewer**:
Submitted as a potential (short term?) fix.  It'd likely be better to refactor such that a sleep isn't needed by using a local queue as suggested on 05-20-20 call, but wanted to submit this option regardless.

**Does this PR introduce a user-facing change?**:
NONE

```release-note
fix: High CPU usage when grpc server is enabled [#1126]
```

PR #1241, opened 2020-05-29
bidirectional gRPC outputs + fix server streaming gRPC outputs

### LINKED ISSUE #856: Catch and handle gRPC status and logs

Our gRPC server should be able to listen for gRPC runtime statuses and react to them accordingly.

Also the Falco's gRPC server should catch the gRPC logs.

**Why is this needed**:

The gRPC C++ implementation does not uses exception as per Google C++ coding guide. Thus we do not have exceptions to detect misbehaviours. We need to observer the gRPC internal status and react to it.

For example when input certificates for mutual TLS are wrong (eg., empty files) it segfaults and we are not able to correcly handle it.

Similarly we would like to catch the gRPC C++ implementation logs in order to filter and/or expose them with our logger and our logging format.

Refs #822

### LINKED ISSUE #1126: High CPU usage when grpc is enable
**Describe the bug**

When grpc is enabled in falco 0.21.0 there is a huge increase in cpu usage on some pods on kops cluster, k8s 1.16, debian nodes.

grpc disabled
NAME                   CPU(cores)   MEMORY(bytes)
falco-htqvd            7m           16Mi
falco-hzxvx            7m           16Mi
falco-lw9f7            7m           16Mi
falco-nmvqq            7m           16Mi
falco-phz9d            8m           16Mi
falco-scbcr            9m           16Mi
falco-snwqj            8m           16Mi
falco-svswf            7m           16Mi
falco-tgsq8            8m           16Mi

grpc enabled
NAME                   CPU(cores)   MEMORY(bytes)
falco-48zhv            8m           17Mi
falco-8prq5            7m           17Mi
falco-9wtl4            8m           17Mi
falco-ckfl4            1930m        18Mi
falco-fwlzc            1833m        17Mi
falco-gpx6g            7m           17Mi
falco-hrfv8            1825m        18Mi
falco-tv4f2            1004m        17Mi
falco-wx8dx            1003m        17Mi
**How to reproduce it**

Install falco using the stable helm chart with grpc enabled.
`grpc:
   enabled: true
   bind_address: "0.0.0.0:5060"
   threadiness: 8
   private_key: /etc/falco/certs/server.key
   cert_chain: /etc/falco/certs/server.crt
   root_certs: /etc/falco/certs/ca.crt

grpc_output:
      enabled: true`

**Expected behaviour**

CPU usage should be consistent across all nodes, and should be at a reasonable level.

**Screenshots**

**Environment**

k8s 1.16 deployed via kops, standard debian nodes.

- Falco version:
0.21.0
- System info:

AWS
- OS:
`Distributor ID: Debian
Description:    Debian GNU/Linux 9.11 (stretch)
Release:        9.11
Codename:       stretch
`
- Kernel:
4.4.0-1052-aws
- Installation method:
Kubernetes

**Additional context**

### LINKED ISSUE #1240: Bidirectional request/response streaming for the outputs
**Motivation**

As discussed in the past two community calls, the bug reported in https://github.com/falcosecurity/falco/issues/1126 is related to the wrong usage that our gRPC server is making for the `output:subscribe` call where it receives a single request to stream an infinite number of responses.

While that approach is nice on paper, having an infinite loop reading our output queue (which is thread safe and written by multiple threads) causes a very high CPU usage.

**Feature**
To solve that bug and to make our approach on outputs more flexible and controllable by the client we'd like to transform the streaming response service we  have to a bidirectional streaming service.

In this way, every time the client sends a stream of requests as it receives a streaming of responses where the current response is not closed until the queue is empty, then it goes to the next request which waits until the stream of responses is done and so on.

It's measurable that by trying this approach the CPU usage of Falco becomes instantly predictable and stays at very close levels as it is when the gRPC server is not used at all.
**Alternatives**

An alternative approach to making this could be to throttle the calls to try_pop and gate it with a client controlled keepalive. While that approach leads to a more sane CPU usage, it is blocking server side, which is what we ultimately don't want.

**Additional context**

leodido:
**What type of PR is this?**

/kind bug
/kind design
/kind feature

NONE

Long story short:
while debugging and profiling the Falco gRPC server because of #1126, @fntlnz and I identified that the existing gRPC outputs API (server streaming) had an architectural problem causing a high CPU consumption.

The implementation of that API was looping on all the remaining enqueued alerts at every process call to "keepalive" (pun intended) the stream.

An initial fix (announced by me during the community calls, very similar to #1237) of swapping the conditional operands into the `if` and the `while` and profit the short-circuiting did not lead to notable performance improvements in our tests.

_Anyways, more details about this topic have been said and presented during the 2 past community calls_.

Thus, with this PR we:
- [x] remove `keepalive` field from output request schema
- [x] introduce and implement a bi-direction gRPC API (`falco.outputs.service/sub`) to watch Falco alerts 
- [x] fix the high CPU consumption of server streaming gRPC outputs
- [x] re-implement the server streaming gRPC outputs as (`falco.outputs.service/get`)
- [x] reproduce/debug end-of-stream errors

This PR contains **breaking change**s.

Some updates have to be carried on other Falco projects:

- [x] [client-go PR](https://github.com/falcosecurity/client-go/pull/45)
- [x] [client-py issue](https://github.com/falcosecurity/client-py/issues/35)
- [x] [client-rs issue](https://github.com/falcosecurity/client-rs/issues/15)
- [x] [website PR](https://github.com/falcosecurity/falco-website/pull/216)

Fixes #1126
Fixes #1240
Fixes #856

**Special notes for your reviewers**:

Here is a graph showing how the `try_pop` used to pop out Falco alerts does not govern anymore the whole thing.

Notice that this graph is related to the following situation:
- a client flooding the server with continuous and parallel requests (~4MLN calls)
- 13 alerts in 10 seconds of Falco execution

_Click on the image and zoom it_.

profile 1

While the `thread_process` (beacuse of the `while` on the `try_pop`) was occupying ~87.38% before ([ref](https://github.com/falcosecurity/falco/issues/1126#issuecomment-630036463)), now it's "only" occupying 37.74% even with a client deliberately flooding the server with requests.

_Jun 5th 2020 update_:

We did further improvements regarding memory allocation (and CPU) while allocating the response.
The latest change in [this commit](https://github.com/falcosecurity/falco/pull/1241/commits/cd858c02e6dccdb30d714632d048d40c8b12632f) allows us to decrease the `thread_process` impact by 20% (it totals to ~17% now). This is very useful in very intensive scenarios or when the server is flooded by requests.

profile 2

**How to try it out**:

1. Enable gRPC server and gRPC Outputs API (with Unix socket for simplicity).

```diff
diff --git a/falco.yaml b/falco.yaml
index dcd76f7..2051583 100644
--- a/falco.yaml
+++ b/falco.yaml
@@ -189,7 +189,7 @@ http_output:
 
 # gRPC server using an unix socket
 grpc:
-  enabled: false
+  enabled: true
   bind_address: "unix:///var/run/falco.sock"
   threadiness: 8
 
@@ -198,4 +198,4 @@ grpc:
 # By enabling this all the output events will be kept in memory until you read them with a gRPC client.
 # Make sure to have a consumer for them or leave this disabled.
 grpc_output:
-  enabled: false
+  enabled: true
```

2. Run Falco

```console
sudo ./build/userspace/falco/falco -r rules/falco_rules.yaml -c falco.yaml
```

3 (a). Try the **/get** method to obtain a stream of all the Falco alerts in the queue and shutdown.

```console
sudo grpcurl -d '{}' --import-path userspace/falco/ --proto out
puts.proto --plaintext --keepalive-time 500000 --unix /var/run/falco.sock falco.outputs.service.get | jq
```

3 (b). Call the **/sub** method to obtain a stream of all the Falco alerts in the queue and wait online for (eventually) other alerts.

```console
cat > /tmp/requests.txt
```

```console
tail -f /tmp/requests.txt | sudo grpcurl -d @ --import-path userspace/falco/ --proto outputs.proto --plaintext --keepalive-time 500000 --unix /var/run/falco.sock falco.outputs.service.sub | jq
```

Write `{}` (on a new line each time) in the `/tmp/requests.txt` file.

```release-note
BREAKING CHANGE: server streaming gRPC outputs method is now `falco.outputs.service/get`
new: new bi-directional async streaming gRPC outputs (`falco.outputs.service/sub`) 
fix: high CPU usage when using server streaming gRPC outputs
```


fntlnz:
/hold
Adding an hold on this so other Prs can be merged by poiana, feel free to remove once we have two approvals.

userspace/falco/outputs.proto
```diff
@@ -0,0 +1,55 @@
...
+
+package falco.outputs;
+
+option go_package = "github.com/falcosecurity/client-go/pkg/api/outputs";
+
+// This service defines the RPC methods
+// to `request` a stream of output `response`s.
+service service {
+  // Subscribe to a stream of Falco outputs by sending a stream of requests.
+  rpc sub(stream request) returns (stream response);
```
mstemm, line 30: Why does the request have to be a stream? Isn't the big distinction that sub continues forever while get() returns outputs and stops?

userspace/falco/falco_outputs.cpp
```diff
@@ -22,11 +22,10 @@ limitations under the License.
 
 #include "formats.h"
 #include "logger.h"
-#include "falco_output_queue.h"
+#include "falco_outputs_queue.h"
```
kris-nova, line 25: Knowing you @leodido you put a lot of thought into this.

Can you share your thinking behind the change. As I start digging more and more into the code I will be opening up more PRs and would like to understand your thinking before suggesting changes.

userspace/falco/falco_outputs_queue.h
```diff
@@ -16,12 +16,12 @@ limitations under the License.
...
-#include "output.pb.h"
+#include "outputs.pb.h"
 #include "tbb/concurrent_queue.h"
 
 namespace falco
 {
-namespace output
+namespace outputs
 {
 typedef tbb::concurrent_queue<response> response_cq;
```
kris-nova, line 27: It looks like `queue` has some public methods/members. I know this is a nit, and is out of scope for this PR but some doc blocks that reference how we are supposed to use these methods would be useful.

The output queue is going to be very important as Falco gains adoptions so spending a few minutes to write down how others should use them might make sense.

Traditionally I have also made it a point to at the minimum do this in the public members of the header files, but however we decide to do it would be great! 

Feel free to ignore this if we want to bring up a broader coding style/documentation discussion as a community. Just sharing thoughts as I look at the code.

userspace/falco/grpc_context.h
```diff
@@ -36,7 +36,7 @@ class context
 {
 public:
 	context(::grpc::ServerContext* ctx);
-	~context() = default;
+	virtual ~context() = default;
```
kris-nova, line 39: Same here with context, why virtual now? and what does `context` mean to you?

userspace/falco/grpc_request_context.cpp
```diff
@@ -48,40 +48,46 @@ void request_stream_context<falco::output::service, falco::output::request, falc
...
 	// Processing
-	output::response res;
-	(srv->*m_process_func)(*m_stream_ctx, m_req, res); // subscribe()
+	outputs::response res;
+	(srv->*m_process_func)(*m_stream_ctx, m_req, res); // get()
+
+	if(!m_stream_ctx->m_is_running)
+	{
+		m_state = request_context_base::FINISH;
+		m_res_writer->Finish(::grpc::Status::OK, this);
```
kris-nova, line 57: Why is `Finish()` capitalized? I haven't seen many other capital methods in the code.

userspace/falco/grpc_server.cpp
```diff
@@ -211,7 +224,7 @@ void falco::grpc::server::run()
 
 	while(server_impl::is_running())
 	{
-		sleep(1);
+		std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
kris-nova, line 227: I can see why we have a performance increase now :)

userspace/falco/grpc_server_impl.cpp
```diff
@@ -28,29 +29,39 @@ bool falco::grpc::server_impl::is_running()
...
 	}
-	else
+
+	ctx.m_is_running = is_running();
+
+	// Start or continue streaming
+	// m_status == stream_context::STREAMING?
+	// todo(leodido) > set m_stream
+
+	ctx.m_has_more = outputs::queue::get().try_pop(res);
```
kris-nova, line 47: This new queue replaces the `lua_pop()` stuff right? 

Also does the `try` language mean we should expect this to throw an exception?

userspace/falco/grpc_server_impl.cpp
```diff
@@ -28,29 +29,39 @@ bool falco::grpc::server_impl::is_running()
...
-		ctx.m_has_more = !is_running() ? false : req.keepalive();
+		ctx.m_stream = nullptr;
+		return;
 	}
+
+	ctx.m_is_running = is_running();
+
+	// Start or continue streaming
+	// m_status == stream_context::STREAMING?
+	// todo(leodido) > set m_stream
```
kris-nova, line 62: Should we open an issue for this?

userspace/falco/outputs.proto
kris-nova: I think we made that decision when we wrote up the original proposal https://github.com/falcosecurity/falco/pull/1241/files#diff-7399e66b7cefcefcb482bd41d0384dbb

I remember talking about the synchronous nature of the engine, but my memory is very bad as every one can tell you. 

It looks like we have a really nice queue we can use now for outputs that would alleviate the need for us to solve asynchronicity (wow is that really a word?) at the gRPC level. 

Design choice for the Kubernetes use case? I honestly can't remember.

userspace/falco/schema.proto
```diff
@@ -1,3 +1,19 @@
+/*
```
kris-nova, line 1: Thank you <3

userspace/falco/version.proto
```diff
@@ -1,3 +1,19 @@
+/*
```
kris-nova, line 1: Thank you <3

userspace/falco/grpc_server.cpp
```diff
@@ -44,6 +44,15 @@ limitations under the License.
 		c.start(this);                                           \
 	}
 
+#define REGISTER_BIDI(req, res, svc, rpc, impl, num)                          \
+	std::vector<request_bidi_context<svc, req, res>> rpc##_contexts(num); \
+	for(request_bidi_context<svc, req, res> & c : rpc##_contexts)         \
+	{                                                                     \
+		c.m_process_func = &server::impl;                             \
+		c.m_request_func = &svc::AsyncService::Request##rpc;          \
+		c.start(this);                                                \
```
kris-nova, line 53: I am indifferent about using the `this` keyword (personally I like it, but I have only seen it in a few places).

Again - I think we should talk coding style at some point. I know it is a hot topic in other projects.

userspace/falco/grpc_request_context.cpp
leodido: It is a method from the gRPC engine, not defined by us.

Docs: https://grpc.github.io/grpc/cpp/classgrpc__impl_1_1_server_async_writer_interface.html#a6683f8dfe66786b7d8de4466ea0908cd

userspace/falco/grpc_context.h
leodido: The rationale is: `context` is a base class and instances of the derived classes get destructed.

When a base class destructor is not virtual and you have a variable pointing to an object derived from the base class, deleting the derived instance has undefined behaviour. Which can lead to memory leaks.

userspace/falco/grpc_server.cpp
leodido: Which keyword would you use here?

userspace/falco/grpc_server_impl.cpp
leodido: This todo is for future work on https://github.com/falcosecurity/falco/issues/857

userspace/falco/grpc_server.cpp
fntlnz: I can't see how this line can be related to that.

userspace/falco/outputs.proto
mstemm: I see the note in the description, but I didn't see anywhere in the server where a second request message was handled. (Maybe I missed it).

I thought the intent for this method was that you send a single request, and then you're subscribed to all responses, forever. What does a second request do?

fntlnz: @mstemm sorry I just saw your question.

The way it works is that each request has a stream of responses until the engine does not have anything else to send. Then another request is sent in the same stream.

The way it was before was slightly different. Only one request was made and a stream of responses was given which was put on hold in case the engine didn't have events. That hold was a bit problematic because it was a kind of pull mechanism which causes a lot of hits into the shared memory between all the threads. With the new approach we are using a push mechanism instead.

PR #1252, opened 2020-06-09
build: libyaml in bundled deps

fntlnz:

**What type of PR is this?**

/kind feature

/area build

There's no reason we rely on system libyaml when we release since it doesn't add many MB to the build and doesn't require a long time to build either.

Fixes #

```release-note
new: libyaml is now bundled in the release process. Users can now avoid installing libyaml directly when getting Falco from the official release.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -160,12 +160,7 @@ ExternalProject_Add(
   INSTALL_COMMAND "")
 
 # libyaml
-find_library(LIBYAML_LIB NAMES libyaml.so)
-if(LIBYAML_LIB)
-  message(STATUS "Found libyaml: lib: ${LIBYAML_LIB}")
-else()
-  message(FATAL_ERROR "Couldn't find system libyaml")
-endif()
+include(libyaml)
 
 # lyaml
 set(LYAML_SRC "${PROJECT_BINARY_DIR}/lyaml-prefix/src/lyaml/ext/yaml")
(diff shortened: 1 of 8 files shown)


fntlnz:
/hold

This needs more testing with the packages to avoid breaking the release.

fntlnz:
We also probably want to remove libyaml from the builder docker image

leogr:
> We also probably want to remove libyaml from the builder docker image

:+1: 
I can take care of that.

docker/builder/Dockerfile
```diff
@@ -20,7 +20,7 @@ ENV FALCO_VERSION=${FALCO_VERSION}
 
 # build toolchain
 RUN yum -y install centos-release-scl && \
-    INSTALL_PKGS="devtoolset-7-gcc devtoolset-7-gcc-c++ devtoolset-7-toolchain devtoolset-7-libstdc++-devel devtoolset-7-elfutils-libelf-devel llvm-toolset-7 glibc-static autoconf automake libtool createrepo expect git which libcurl-devel zlib-devel ncurses-devel rpm-build libyaml-devel" && \
+    INSTALL_PKGS="devtoolset-7-gcc devtoolset-7-gcc-c++ devtoolset-7-toolchain devtoolset-7-libstdc++-devel devtoolset-7-elfutils-libelf-devel llvm-toolset-7 glibc-static autoconf automake libtool createrepo expect git which libcurl-devel zlib-devel ncurses-devel rpm-build" && \
```
fntlnz, line 23: I think we can revert this one for now and leave libyaml. Users can still use it even if CMake doesn't by default.

leogr: As discussed yesterday, I agree. I will revert it soon so we can unblock this PR.

leogr: Done and rebased

PR #1258, opened 2020-06-11
feat(build): Including the falco-driver-loader into the container image

kris-nova:
**What type of PR is this?**

A small one.

 /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind rule-update

> /kind rule-create

/area build

> /area engine

> /area rules

> /area tests

> /area proposals

Fixes #

I left a big commit message because I was off in terminal land

```
So I might be wrong here, but I believe this is missing.

I tried to build to the `docker/falco/Dockerfile` locally and push to GKE and was
running into some errors getting Falco started.

I checked both the GKE node, as well as the running pod and I couldn't find the script.

So I think we want to include the script directly into the container.

Anyway I was able to now use this Dockerfile to run my local version of Falco in GKE and
load a BPF probe fine.

Other thoughts: Do we want to consider pull the flags from the `falco-driver-loader` script
up to the container image as an ENV or something?

Other thoughts: It looks like the other container images all are based on this one so this
should naturally flow down to the others.

If we don't need this feel free to ignore/close. :)
```

```release-note
Baking `falco-driver-loader` directly into the `falco` image.
```

### DIFF
--- a/docker/falco/Dockerfile
+++ b/docker/falco/Dockerfile
@@ -12,6 +12,8 @@ ENV FALCO_VERSION=${FALCO_VERSION}
 ENV HOST_ROOT /host
 ENV HOME /root
 
+ADD ../../scripts/falco-driver-loader /usr/local/bin/falco-driver-loader
+
 RUN cp /etc/skel/.bashrc /root && cp /etc/skel/.profile /root
 
 RUN apt-get update \


kris-nova:
We might not need this - in which case we can just close it.

But I was trying to use the `falco` docker image to push a local copy of the Falco code to GKE for some debugging and when I built with the `falco` docker image and then pushed the image to my registry I was able to cycle my pods and they all broke.

This fixed that issue, but unsure what we want to do.

PR #1283, opened 2020-06-26
rule(Launch Package Management Process in Container): allow apk manifest

marier-nico:
**What type of PR is this?**

/kind rule-update

/area rules

The `apk manifest` command allows "Displaying checksums for files contained in a given package", according to the Alpine Linux [wiki](https://wiki.alpinelinux.org/wiki/Alpine_Linux_package_management).

This is a very harmless command, so the events that get raised because of that command aren't necessarily relevant. This PR filters out `apk manifest` usage.

None

None

```release-note
rule(Launch Package Management Process in Container): allow apk manifest
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -2333,6 +2333,7 @@
     spawned_process
     and container
     and user.name != "_apt"
+    and not proc.cmdline startswith "apk manifest"
     and package_mgmt_procs
     and not package_mgmt_ancestor_procs
     and not user_known_package_manager_in_container


Kaizhe:
Thanks @marier-nico !

This is a 50/50 change to me. And we do have a place holder for customization:

```
# In a local/user rules file, create a condition that matches legitimate uses
# of a package management process inside a container.
#
# For example:
# - macro: user_known_package_manager_in_container
#   condition: proc.cmdline="dpkg -l"
- macro: user_known_package_manager_in_container
  condition: (never_true)
```

Do you consider customize your falco rule instead of make changes in the upstream?

marier-nico:
Sure thing @Kaizhe ! :smile: 

We have considered doing this in our own rules, however since the command is harmless, we thought others could benefit from this. If it would be better for everyone to filter out `apk manifest` in their `local_rules.yml`, that's no problem, we can just close this PR!

Kaizhe:
I really appreciate the effort,

maybe you can add `proc.cmdline startswith "apk manifest"` in the macro `user_known_package_manager_in_container`

marier-nico:
I'm happy to hear it! That's already done on our end, usually we add these changes to our own rules at the same time we open a PR to Falco.

But really it's not a problem is this isn't suitable for upstream! :smile:

PR #1289, opened 2020-06-29
rule(Write below etc): allow snapd to write its unit files

marier-nico:
**What type of PR is this?**

/kind feature
/kind rule-update

/area rules

In an AWS environment, snap can be used to install the SSM agent, which will write unit files under `etc`. This is the expected behaviour from snap.

See [the AWS docs](https://docs.aws.amazon.com/systems-manager/latest/userguide/agent-install-ubuntu.html) for more information.

None

None

```release-note
rule(Write below etc): allow snapd to write its unit files
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -219,7 +219,7 @@
 # The truncated dpkg-preconfigu is intentional, process names are
 # truncated at the sysdig level.
 - list: package_mgmt_binaries
-  items: [rpm_binaries, deb_binaries, update-alternat, gem, pip, pip3, sane-utils.post, alternatives, chef-client, apk]
+  items: [rpm_binaries, deb_binaries, update-alternat, gem, pip, pip3, sane-utils.post, alternatives, chef-client, apk, snapd]
 
 - macro: package_mgmt_procs
   condition: proc.name in (package_mgmt_binaries)


rules/falco_rules.yaml
```diff
@@ -905,6 +905,9 @@
 - macro: calico_writing_envvars
   condition: (proc.name=start_runit and fd.name startswith "/etc/envvars" and container.image.repository endswith "calico/node")
 
+- macro: snapd_writing_service
+  condition: (proc.name=snapd and (fd.name startswith "/etc/systemd/system/snap-amazon" or fd.name startswith "/etc/systemd/system/snap.amazon-ssm"))
```
leogr, line 909: Maybe we should specify that this condition is amazon specific. I'm not a `snapd` expert, but I assume this macro is not just for snap.

marier-nico: Yeah, you're absolutely right. What do you think about something like `snapd_installing_aws_ssm_agent`?

leogr: Not really sure about the naming.
/cc @Kaizhe 
/cc @leodido

Kaizhe: can you add `snap` in the `package_mgmt_binaries` list which is already in the whitelist for writing below etc.

marier-nico: Oh wow, sorry for missing that! I did it just now :smile:

PR #1303, opened 2020-07-07
feat(userspace/falco): print version at startup

### LINKED ISSUE #1302: Add the Falco version at the very beginning of the log
**Motivation**

Since there's no way to match logs reported by the users with the corresponding Falco version, I'd like Falco to print its version out at the beginning of the log.

It can be advantageous when debugging issues. The version reported by the user might be wrong for several reasons (by mistake, wrong copy/paste, cached container image, etc.)

**Feature**

Print the Falco and driver versions (similar to the output of `--version`) at the very beginning of the output.

**Alternatives**

Simply nothing :)

**Additional context**

IMHO, no other information should be printed out (for example, the gcc version, dependency version, and so on) to avoid mixing build-time metadata with run-time one. The goal here is just to match the log with a specific Falco version, not to provide detailed debugging information. Also, note that the version may contain the commit ID when Falco has not built from a tag matching a release. 

/cc @leodido 
/cc @fntlnz 

/assign

/milestone 0.24.0

leogr:
**What type of PR is this?**

/kind feature

/area engine

See #1302

Fixes #1302

/milestone 0.24.0
/cc @fntlnz 
/cc @leodido 
/cc @kris-nova 

```release-note
new(userspace/falco): print the Falco and driver versions at the very beginning of the output.
```

### DIFF
--- a/userspace/falco/falco.cpp
+++ b/userspace/falco/falco.cpp
@@ -795,12 +795,16 @@ int falco_init(int argc, char **argv)
 			falco_logger::set_time_format_iso_8601(config.m_time_format_iso_8601);
 
 			// log after config init because config determines where logs go
+			falco_logger::log(LOG_INFO, "Falco version " + std::string(FALCO_VERSION) + " (driver version " + std::string(DRIVER_VERSION) + ")\n");
 			falco_logger::log(LOG_INFO, "Falco initialized with configuration file " + conf_filename + "\n");
 		}
 		else
 		{
 			config.init(cmdline_options);
 			falco_logger::set_time_format_iso_8601(config.m_time_format_iso_8601);
+
+			// log after config init because config determines where logs go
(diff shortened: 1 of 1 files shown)


leogr:
> SGTM!
> 
> Anyway, I have a question.
> 
> Why do not stick with the usual two-line version (line 626-627) logs we use for the `--version` flag in order to be more consistent?
> 
> There's a particular reason to print Falco version and driver version this way when Falco runs?

Lines 626-627 uses `printf`, here I use the initialized logger. Moreover, I have found just one-line output briefer.
No other reasons. 
Let me know if you prefer a two-line, and I will change it shortly.

leodido:
I don't have particular preferences honestly.

I was just asking if there was a ratio behind putting both versions on the same line. That's all!

leodido:
Moving this to next release because it hasn't reached 2 approvals

/milestone 0.25.0

fntlnz:
Approved, I think we can get this in now!

PR #1343, opened 2020-08-04
docker-entrypoint: Match /usr/bin/gcc version to kernel /proc/version

antoinedeschenes:
**What type of PR is this?**

/kind bug

/area build

dkms always seems to default to /usr/bin/gcc. The override built in the Dockerfile causes compilation problems with the Ubuntu 20.04 AMI kops is planning to moving to. (We've tested installing gcc-9 to the image, that doesn't seem to fix the problem)

Behavior:
* Match major gcc version with version stated in /proc/version whenever possible
* Use lowest installed version if kernel gcc version not found or lower (supersedes gcc-5 link in Dockerfile)
* Use highest installed gcc version if kernel gcc version higher (fixes problems with ubuntu focal AMI soon to be used by kops)

Fixes #

```release-note
docker-entrypoint: Match closest /usr/bin/gcc major version with version stated in /proc/version for module compilation
```

### DIFF
--- a/docker/driver-loader/docker-entrypoint.sh
+++ b/docker/driver-loader/docker-entrypoint.sh
@@ -25,4 +25,21 @@ do
     ln -s "$i" "/usr/src/$base"
 done
 
+GCC_AVAILABLE_VERSIONS=( $(ls /usr/bin | grep -Po "^gcc-\K[0-9]+") )
+GCC_MIN_VERSION="${GCC_AVAILABLE_VERSIONS[0]}"
+GCC_MAX_VERSION="${GCC_AVAILABLE_VERSIONS[-1]}"
+GCC_KERNEL_VERSION="$(grep -Po "gcc version \K[0-9]+" /proc/version 2>/dev/null)"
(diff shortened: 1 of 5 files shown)


antoinedeschenes:
@leogr absolutely

Now just added a commit adding gcc-7 to the local and falco images, gcc 5, 6 and 8 are currently installed, I figured that this could have happened when the `debian:stable` image switched from stretch to buster.

leogr (commented):
Hey @antoinedeschenes 

during the today community call, we talked about another PR https://github.com/falcosecurity/falco/pull/1408
that solve this problem and cover more use cases. So, once #1408 gets merged, this PR should not needed anymore.

Since you did a great job anyway, before closing this, I'd ask your opinion and the confirmation from you that #1408 solved the same problem this PR tries to fix.

cc'ing @leodido and @fntlnz in case they want to add more context.

Thank you!

antoinedeschenes:
@leogr looks like #1408 solves the same problem

PR #1350, opened 2020-08-12
macro(exe_running_docker_save): add better support for centos

admiral0:
**What type of PR is this?**

/kind bug
/kind rule-update

/area rules

dockerd and docker have "-current" suffix on centos and rhel. This
macro does not match causing false positives on multiple rules
using it

```release-note
macro(exe_running_docker_save): add better support for centos
```


fntlnz:
I edited the release note as this is a user facing change. This is scheduled to go into 0.25.0

fntlnz:
@Kaizhe PTAL

PR #1372, opened 2020-08-28
Fix/address addl rules fps

mstemm:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

/kind rule-update

> /kind rule-create

> /area build

> /area engine

/area rules

> /area tests

> /area proposals

Address several sources of FPs, primarily from GKE environments.

Fixes #

```release-note
Address several sources of FPs, primarily from GKE environments.
```


rules/falco_rules.yaml
```diff
@@ -746,6 +746,13 @@
 - macro: runuser_reading_pam
   condition: (proc.name=runuser and fd.directory=/etc/pam.d)
 
+# CIS Linux Benchmark program
+- macro: linux_bench_reading_etc_shadow
+  condition: ((proc.aname[2]=linux-bench and
```
Kaizhe, line 751: can we use `proc.aname=linux-bench` instead of specifying the index. linux-bench doesn't need to be the grant parent of awk, cut, grep

rules/falco_rules.yaml
```diff
@@ -928,10 +935,12 @@
   items: [sources.list]
 
 - list: repository_directories
-  items: [/etc/apt/sources.list.d, /etc/yum.repos.d]
+  items: [/etc/apt/sources.list.d, /etc/yum.repos.d, /etc/apt]
 
 - macro: access_repositories
-  condition: (fd.filename in (repository_files) or fd.directory in (repository_directories))
+  condition: (fd.directory in (repository_directories) or
```
Kaizhe, line 941: `fd.directory in (repository_directories)` is redundant given the changes of using `pmatch`

rules/k8s_audit_rules.yaml
```diff
@@ -48,6 +48,8 @@
     "minikube", "minikube-user", "kubelet", "kops", "admin", "kube", "kube-proxy", "kube-apiserver-healthcheck",
     "kubernetes-admin",
     vertical_pod_autoscaler_users,
+    cluster-autoscaler
```
Kaizhe, line 51: missing `,`

rules/falco_rules.yaml
mstemm: proc.aname is a bit more expensive, as it traverses the history until it finds an ancestor with the name or runs into init/the root process. We could do that to make it more flexible, but the cost will be higher for the negative case.

rules/falco_rules.yaml
mstemm: I think both halves are important--the first half allows for any files directly below repository_directories. The second looks for any of the files in repository_files, but allows them to be multiple levels below repository_directories.

Maybe this is overkill and we can drop the first half, but they are trying to do different things.

Kaizhe: that's fine, we can keep both.

mstemm:
@fntlnz check out the last commit, it changes the trace files that get downloaded by the tests to have versioning now. I think this is a clean way to handle changes in trace files, as I needed for this PR.

We can also put the trace files in git somewhere but some of them are pretty big.

leogr (changes requested):
Just found a dup entry (see the comment below), otherwise it looks very good to me!!!

Thank you!!!

rules/falco_rules.yaml
```diff
@@ -227,6 +227,7 @@
 - list: deb_binaries
   items: [dpkg, dpkg-preconfigu, dpkg-reconfigur, dpkg-divert, apt, apt-get, aptitude,
     frontend, preinst, add-apt-reposit, apt-auto-remova, apt-key,
+    apt-auto-remova,
```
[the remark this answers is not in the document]
mstemm: Yes thanks! Will fix.

leogr (approved):
LGTM now

Thank you!

Kaizhe (approved):

fntlnz:
@mstemm how are the integration tests file generated? Now that they are versioned I would like to understand how we can get them to a shape where everyone can read a document on how to add new ones.

PR #1376, opened 2020-09-01
Proposal on better exception handling

mstemm:
This proposes adding exceptions as a first class object to falco rules
files.

It adds a new key "exceptions" to rule objects that allows a rule
writer to define tuples of field names that comprise an exception, and a
new top level object "exception" that contains lists of tuples of field
values that define exceptions to rules.

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind rule-update

> /kind rule-create

> /area build

> /area engine

> /area rules

> /area tests

/area proposals

Fixes #

```release-note
NONE
```

### DIFF
--- a/proposals/20200828-structured-exception-handling.md
+++ b/proposals/20200828-structured-exception-handling.md
@@ -147,19 +147,38 @@ To address some of these problems, we will add the notion of Exceptions as top l
    - name: container_writer
      fields: [container.image.repository, fd.directory]
      comps: [=, startswith]
+   - name: proc_filenames
+     fields: [proc.name, fd.name]
+	 comps: [=, in]
+   - name: filenames
+     fields: fd.filename
+     comps: in
 ```
 
-This rule defines two kinds of exceptions: one called proc_writer with a combination of proc.name and fd.directory, and a second called container_writer with a combination of container.image.repository and fd.directory. The specific strings "proc_writer" and "container_writer" are arbitrary strings and don't have a special meaning to the rules file parser. They're only used to link together the list of field names with the list of field values that exist in the exception object.
+This rule defines four kinds of exceptions:
+  * proc_writer: uses a combination of proc.name and fd.directory
+  * container_writer: uses a combination of container.image.repository and fd.directory
+  * proc_filenames: uses a combination of process and list of filenames.
+  * filenames: uses a list of filenames
+
+The specific strings "proc_writer"/"container_writer"/"proc_filenames"/"filenames" are arbitrary strings and don't have a special meaning to the rules file parser. They're only used to link together the list of field names with the list of field values that exist in the exception object.
 
 proc_writer does not have any comps property, so the fields are directly compared to values using the = operator. container_writer does have a comps property, so each field will be compared to the corresponding exception items using the corresponding comparison operator.
 
+proc_filenames uses the in comparison operator, so the corresponding values entry should be a list of filenames.
+
+filenames differs from the others in that it names a single field and single comp operator. This changes how the exception condition snippet is constructed (see below).
+
 Notice that exceptions are defined as a part of the rule. This is important because the author of the rule defines what construes a valid exception to the rule. In this case, an exception can consist of a process and file directory (actor and target), but not a process name only (too broad).
 
-We'll add a new object exception that defines exceptions to a rule:
+Exception values will most commonly be defined in rules with append: true. Here's an example:
 
 ```
-- exception: Write below binary dir
-  items:
+- list: apt_files
+  items: [/bin/ls, /bin/rm]
+
+- rule: Write below binary dir
+  exceptions:
   - name: proc_writer
     values:
     - [apk, /usr/lib/alpine]
@@ -167,32 +186,32 @@ We'll add a new object exception that defines exceptions to a rule:
   - name: container_writer
     values:
     - [docker.io/alpine, /usr/libexec/alpine]
+  - name: proc_filenames
+    values:
+	 - [apt, apt_files]
+	 - [rpm, [/bin/cp, /bin/pwd]]
+  - name: filenames
+    values: [python, go]
 ```
-
-The name of the exception links it to the rule.
 
 A rule exception applies if for a given event, the fields in a rule.exception match all of the values in some exception.item. For example, if a program `apk` writes to a file below `/usr/lib/alpine`, the rule will not trigger, even if the condition is met.
 
-Append will be supported for exception objects. If append: is true, the items in the second definition will be added to the items in the earlier definition. For example, adding:
+Notice that an item in a values list can be a list. This allows building exceptions with operators like "in", "pmatch", etc. that work on a list of items. The item can also be a name of an existing list. If not present surrounding parantheses will be added.
 
-```
-- exception: Write below binary dir
-  items:
-  - container_writer:
-    - [docker.io/golang-alpine, /usr/libexec/go]
-  append: true
-```
-
-Would add a second container_writer exception.
+Finally, note that the structure of the values property differs between the items where fields is a list of fields (proc_writer/container_writer/proc_filenames) and when it is a single field (procs_only). This changes how the condition snippet is constructed.
 
 ### Implementation
 
-Each exception can be thought of as an implicit "and not (field1 cmp1 val1 and field2 cmp2 val2 and...)" appended to the rule's condition. In practice, that's how exceptions will be implemented.
+For exception items where the fields property is a list of field names, each exception can be thought of as an implicit "and not (field1 cmp1 val1 and field2 cmp2 val2 and...)" appended to the rule's condition. For exception items where the fields property is a single field name, the exception can be thought of as an implict "and not field cmp (val1, val2, ...)". In practice, that's how exceptions will be implemented.
 
 When a rule is parsed, the original condition will be wrapped in an extra layer of parentheses and all exception values will be appended to the condition. For example, using the example above, the resulting condition will be:
 
 ```
-(<Write below binary dir condition>) and not ((proc.name = apk and fd.directory = /usr/lib/alpine) or (proc.name = npm and fd.directory = /usr/node/bin) or (container.image.repository = docker.io/alpine and fd.directory startswith /usr/libexec/alpine))
+(<Write below binary dir condition>) and not (
+    (proc.name = apk and fd.directory = /usr/lib/alpine) or (proc.name = npm and fd.directory = /usr/node/bin) or
+	(container.image.repository = docker.io/alpine and fd.directory startswith /usr/libexec/alpine) or
+	(proc.name=apt and fd.name in (apt_files))) or
+	(fd.filename in (python, go))))
 ```
 
 The exceptions are effectively syntatic sugar that allows expressing sets of exceptions in a concise way.
--- a/proposals/20200828-structured-exception-handling.md
+++ b/proposals/20200828-structured-exception-handling.md
@@ -9,7 +9,7 @@ rule should be allowed. For example, The rule Write Below Binary Dir
 has exceptions for specific programs that are known to write below
 these directories as a part of software installation/management:
 
-```
+```yaml
 - rule: Write below binary dir
   desc: an attempt to write to any file below a set of binary directories
   condition: >
@@ -23,7 +23,7 @@ these directories as a part of software installation/management:
 ```
 In most cases, these exceptions are expressed as concatenations to the original rule's condition. For example, looking at the macro package_mgmt_procs:
 
-```
+```yaml
 - macro: package_mgmt_procs
   condition: proc.name in (package_mgmt_binaries)
 ```
@@ -53,7 +53,7 @@ The exceptions all generally follow the same structure--naming a program and a d
 
 An important way to customize rules and macros is to use `append: true` to add to them, or `append: false` to define a new rule/macro, overwriting the original rule/macro. Here's an example from Update Package Repository:
 
-```
+```yaml
 - list: package_mgmt_binaries
   items: [rpm_binaries, deb_binaries, update-alternat, gem, pip, pip3, sane-utils.post, alternatives, chef-client, apk, snapd]
 
@@ -74,7 +74,7 @@ An important way to customize rules and macros is to use `append: true` to add t
 
 If someone wanted to add additional exceptions to this rule, they could add the following to the user_rules file:
 
-```
+```yaml
 - list: package_mgmt_binaries
   items: [puppet]
   append: true
@@ -99,7 +99,7 @@ Although the concepts of macros and lists in condition fields, combined with app
 
 * Appending to conditions can result in incorrect behavior, unless the original condition has its logical operators set up properly with parentheses. For example:
 
-```
+```yaml
 rule: my_rule
 condition: (evt.type=open and (fd.name=/tmp/foo or fd.name=/tmp/bar))
 
@@ -114,7 +114,7 @@ Results in unintended behavior. It will match any fd related event where the nam
 
 * Appends and overrides can get confusing if you try to apply them multiple times. For example:
 
-```
+```yaml
 macro: allowed_files
 condition: fd.name=/tmp/foo
 
@@ -131,7 +131,7 @@ If someone wanted to override the original behavior of allowed_files, they would
 
 To address some of these problems, we will add the notion of Exceptions as top level objects alongside Rules, Macros, and Lists. A rule that supports exceptions must define a new key `exceptions` in the rule. The exceptions key is a list of identifier plus list of tuples of filtercheck fields. Here's an example:
 
-```
+```yaml
 - rule: Write below binary dir
   desc: an attempt to write to any file below a set of binary directories
   condition: >
@@ -173,7 +173,7 @@ Notice that exceptions are defined as a part of the rule. This is important beca
 
 Exception values will most commonly be defined in rules with append: true. Here's an example:
 
-```
+```yaml
 - list: apt_files
   items: [/bin/ls, /bin/rm]
 
@@ -236,7 +236,6 @@ However, there are a few changes we'll have to make to Falco rules file parsing:
 
 * Currently, Falco will reject files containing anything other than rule/macro/list top-level objects. As a result, `exception` objects would be rejected. We'll probably want to make a one-time change to Falco to allow arbitrary top level objects.
 * Similarly, Falco will reject rule objects with exception keys. We'll also probably want to change Falco to allow unknown keys inside rule/macro/list/exception objects.
-
 
 
 
--- a/proposals/20200828-structured-exception-handling.md
+++ b/proposals/20200828-structured-exception-handling.md
@@ -1,0 +1,214 @@
+# Proposal for First Class Structured Exceptions in Falco Rules
+
+## Summary
+
+## Motivation
+
+Almost all Falco Rules have cases where the behavior detected by the
+rule should be allowed. For example, The rule Write Below Binary Dir
+has exceptions for specific programs that are known to write below
+these directories as a part of software installation/management:
+
+```
+- rule: Write below binary dir
+  desc: an attempt to write to any file below a set of binary directories
+  condition: >
+    bin_dir and evt.dir = < and open_write
+    and not package_mgmt_procs
+    and not exe_running_docker_save
+    and not python_running_get_pip
+    and not python_running_ms_oms
+    and not user_known_write_below_binary_dir_activities
+...
+```
+In most cases, these exceptions are expressed as concatenations to the original rule's condition. For example, looking at the macro package_mgmt_procs:
+
+```
+- macro: package_mgmt_procs
+  condition: proc.name in (package_mgmt_binaries)
+```
+
+The result is appending `and not proc.name in (package_mgmt_binaries)` to the condition of the rule.
+
+A more extreme case of this is the write_below_etc macro used by Write below etc rule. It has tens of exceptions:
+
+```
+...
+    and not sed_temporary_file
+    and not exe_running_docker_save
+    and not ansible_running_python
+    and not python_running_denyhosts
+    and not fluentd_writing_conf_files
+    and not user_known_write_etc_conditions
+    and not run_by_centrify
+    and not run_by_adclient
+    and not qualys_writing_conf_files
+    and not git_writing_nssdb
+...
+```
+
+The exceptions all generally follow the same structure--naming a program and a directory prefix below /etc where that program is allowed to write files.
+
+### Using Appends/Overwrites to Customize Rules
+
+An important way to customize rules and macros is to use `append: true` to add to them, or `append: false` to define a new rule/macro, overwriting the original rule/macro. Here's an example from Update Package Repository:
+
+```
+- list: package_mgmt_binaries
+  items: [rpm_binaries, deb_binaries, update-alternat, gem, pip, pip3, sane-utils.post, alternatives, chef-client, apk, snapd]
+
+- macro: package_mgmt_procs
+  condition: proc.name in (package_mgmt_binaries)
+
+- macro: user_known_update_package_registry
+  condition: (never_true)
+
+- rule: Update Package Repository
+  desc: Detect package repositories get updated
+  condition: >
+    ((open_write and access_repositories) or (modify and modify_repositories))
+`    and not package_mgmt_procs
+    and not exe_running_docker_save
+    and not user_known_update_package_registry
+```
+
+If someone wanted to add additional exceptions to this rule, they could add the following to the user_rules file:
+
+```
+- list: package_mgmt_binaries
+  items: [puppet]
+  append: true
+
+- macro: package_mgmt_procs
+  condition: and not proc.pname=chef
+  append: true
+
+- macro: user_known_update_package_registry
+  condition: (and not proc.name in (npm)
+  append: false
+```
+
+This adds an 3 different exceptions:
+* an additional binary to package_mgmt_binaries (because append is true),
+* adds to package_mgmt_procs, adding an exception for programs spawned by chef (because append is true)
+* overrides the macro user_known_update_package_registry to add an exception for npm (because append is false).
+
+### Problems with Appends/Overrides to Define Exceptions
+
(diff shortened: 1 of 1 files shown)


proposals/20200828-structured-exception-handling.md
```diff
@@ -0,0 +1,214 @@
...
+  items: [puppet]
+  append: true
+
+- macro: package_mgmt_procs
+  condition: and not proc.pname=chef
+  append: true
+
+- macro: user_known_update_package_registry
+  condition: (and not proc.name in (npm)
+  append: false
```
Kaizhe, line 88: should this be `append: true` ?

proposals/20200828-structured-exception-handling.md
```diff
@@ -0,0 +1,214 @@
...
+```
+
+This adds an 3 different exceptions:
+* an additional binary to package_mgmt_binaries (because append is true),
+* adds to package_mgmt_procs, adding an exception for programs spawned by chef (because append is true)
+* overrides the macro user_known_update_package_registry to add an exception for npm (because append is false).
```
Kaizhe, line 94: if the purpose is to overwrite, then the condition should probably be `(not proc.name in (npm))`

proposals/20200828-structured-exception-handling.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Advantages
+
+Adding Exception objects as described here has several advantages:
+
+* All rules will implicitly support exceptions. A rule writer doesn't need to define a user_known_xxx macro and add it to the condition.
+* The rule writer has some controls on what defines a valid exception. The rule author knows best what is a good exception, and can define the fields that make up the exception.
+* With this approach, it's much easier to add and manage multiple sets of exceptions from multiple sources. You're just combining lists of tuples of filtercheck field values.
+
+## Backwards compatibility
```
Kaizhe, line 201: If users want to use `exception` object in the rule (or use rules contains exception), they need to upgrade the binary first. I think we should emphasize here.

proposals/20200828-structured-exception-handling.md
```diff
@@ -0,0 +1,214 @@
...
+
+When a rule is parsed, the original condition will be wrapped in an extra layer of parentheses and all exception values will be appended to the condition. For example, using the example above, the resulting condition will be:
+
+```
+(<Write below binary dir condition>) and not ((proc.name=apk and fd.directory=/usr/lib/alpine) or (proc.name=npm and fd.directory=/usr/node/bin) or (container.image.repository=docker.io/alpine and fd.directory=/usr/libexec/alpine))
+```
+
+The exceptions are effectively syntatic sugar that allows expressing sets of exceptions in a concise way.
+
+### Advantages
```
Kaizhe, line 193: I would like to add a section to talk about the limitation here.  The exception approach only supports `=` operator, if users want to use operators like `pmatch`, `contains`, they still need to user the user_known_xxx macros.

proposals/20200828-structured-exception-handling.md
```diff
@@ -0,0 +1,214 @@
...
+```
+
+This rule defines two kinds of exeptions: one called proc_writer with a combination of proc.name and fd.directory, and a second called container_writer with a combination of container.image.repository and fd.directory.
+
+Notice that exceptions are defined as a part of the rule. This is important because the author of the rule defines what construes a valid exception to the rule. In this case, an exception can consist of a process and file directory (actor and target), but not a process name only (too broad).
+
+We'll add a new object exception that defines exceptions to a rule:
+
+```
+- exception: Write below binary dir
```
Kaizhe, line 156: by the definition here, can we remove the `exceptions` field in the rule?

proposals/20200828-structured-exception-handling.md
```diff
@@ -0,0 +1,214 @@
...
+- list: package_mgmt_binaries
+  items: [puppet]
+  append: true
+
+- macro: package_mgmt_procs
+  condition: and not proc.pname=chef
+  append: true
+
+- macro: user_known_update_package_registry
+  condition: (and not proc.name in (npm)
```
leodido, line 87:
```suggestion
  condition: (not proc.name in (npm))
```

proposals/20200828-structured-exception-handling.md
```diff
@@ -0,0 +1,214 @@
...
+```
+rule: my_rule
+condition: (evt.type=open and (fd.name=/tmp/foo or fd.name=/tmp/bar))
+
+rule: my_rule
+condition: or fd.name=/tmp/baz
+append: true
+```
+
+Results in unintended behavior--it will match any fd related event where the name is /tmp/baz, when the intent was probably to add /tmp/baz as an additional opened file.
```
leodido, line 111: Are the two dashes here intended?

proposals/20200828-structured-exception-handling.md
```diff
@@ -0,0 +1,214 @@
...
+
+* All rules will implicitly support exceptions. A rule writer doesn't need to define a user_known_xxx macro and add it to the condition.
+* The rule writer has some controls on what defines a valid exception. The rule author knows best what is a good exception, and can define the fields that make up the exception.
+* With this approach, it's much easier to add and manage multiple sets of exceptions from multiple sources. You're just combining lists of tuples of filtercheck field values.
+
+## Backwards compatibility
+
+This approach does not remove the ability to append to exceptions nor the existing use of user_xxx macros to define exceptions to rules. It only provides an additional way to express exceptions. Hopefully, we can migrate existing exceptions to use this approach, but there isn't any plan to make wholesale rules changes as a part of this.
+
+This approach is for the most part backwards compatible with older falco releases. To implement exceptions, we'll add a preprocessing element to rule parsing. The main falco engine is unchanged.
```
leodido, line 205:
```suggestion
This approach is for the most part backwards compatible with older Falco releases. To implement exceptions, we'll add a preprocessing element to rule parsing. The main Falco engine is unchanged.
```

proposals/20200828-structured-exception-handling.md
```diff
@@ -0,0 +1,214 @@
...
+
+However, there are a few changes we'll have to make to falco rules file parsing:
```
leodido, line 207:
```suggestion
However, there are a few changes we'll have to make to Falco rules file parsing:
```

proposals/20200828-structured-exception-handling.md
```diff
@@ -0,0 +1,214 @@
...
+
+* Currently, falco will reject files with top level `-exception` objects. We'll probably want to make a one-time change to Falco to allow unknown top level objects.
```
leodido, line 209:
```suggestion
* Currently, Falco will reject files with top-level `-exception` objects. We'll probably want to make a one-time change to Falco to allow unknown top-level objects.
```

proposals/20200828-structured-exception-handling.md
```diff
@@ -0,0 +1,214 @@
...
+* Similarly, falco will reject rule objects with exception keys. We'll also probably want to change falco to allow unknown keys inside rule/macro/list/exception objects.
```
leodido, line 210:
```suggestion
* Similarly, Falco will reject rule objects with exception keys. We'll also probably want to change Falco to allow unknown keys inside rule/macro/list/exception objects.
```

leodido (commented):
My second round of review, after the one we had all together on Slack.

Just some nits.

In the meantime, I'm trying to better figure out all the use case combinations about the "backward compatibility" topic.

Anyways, thanks a lot for this Mark!

proposals/20200828-structured-exception-handling.md
leogr: Should it be `- exception` instead of `-exception`?

leogr (commented):
First of all, thank you for this proposal!

It seems good to me, just some very minor issues that @Kaizhe and @leodido already have highlighted (I put a :+1: on their comments).


Other than that, just a question: could this strategy be used to optimize the performance of rule matching at runtime?

proposals/20200828-structured-exception-handling.md
mstemm: Ok, I'll add a sentence.

proposals/20200828-structured-exception-handling.md
mstemm: Ok, I added it a bit earlier in the discussion of exceptions but I call out equality.

proposals/20200828-structured-exception-handling.md
mstemm: I think you need this key to uniquely identify the object. I think you had some related questions, so we can also discuss in person.

proposals/20200828-structured-exception-handling.md
mstemm: Just a hyphen. I can reword it.

proposals/20200828-structured-exception-handling.md
mstemm: Ok, s/falco/Falco/g/ :)

proposals/20200828-structured-exception-handling.md
mstemm: I was just trying to call out the exception object as a top level list item. I can reword it.

mstemm:
About performance, depending on the implementation, it *could* make rule evaluation faster. If we stored the set of field values separate from the condition, we could match against them in parallel. However, @ldegio suggested we actually fold the exception into the condition to keep the rule parsing centralized around condition expressions, so that's what I proposed here.

mstemm:
Okay, I think I addressed all the feedback. Please take another look!

Kaizhe:
Right now, the exception works in two parts:
1. Define exception handlers (e.g. filter names) for each rule
2. 1st class Exception object with:
   - Exception name matches the rule name
   - define values that match the filters (e.g. container.image.repository=`xxx`, fd.name=`yyy`) for each exception handler. 

My main concern is on the usability:
1. If the exception handlers doesn't meet user's use case (e.g. missing some filters), he need to either overwrite the rule with new exception handlers or use the placeholder macros. 

My proposal is similar to [SCC](https://docs.openshift.com/enterprise/3.0/admin_guide/manage_scc.html) in openshift, which is the 1st class object as SecurityContext. Users have full flexibility of what filters can be used as exceptions.

mstemm:
So @Kaizhe 's proposal would define Exceptions like this instead (please let me know if I misinterpreted it):
```
- rule: Write below binary dir
  desc: an attempt to write to any file below a set of binary directories
  condition: >
    bin_dir and evt.dir = < and open_write
    and not package_mgmt_procs
    and not exe_running_docker_save
    and not python_running_get_pip
    and not python_running_ms_oms
    and not user_known_write_below_binary_dir_activities
...
exception: Write below binary dir
  fields:
   - proc_writer: [proc.name, fd.directory]
   - container_writer: [container.image.repository, fd.directory]
  items:
  - container_writer:
    - [docker.io/golang-alpine, /usr/libexec/go]
  append: true
```

The big difference is that the definition of the fields that make up an exception move into the exception object and not the rule, with the rule being unchanged.

I don't think it's a really big difference but I do prefer the version in the original proposal, with exception fields embedded in the rule and exception values provided separately:
* I think it's a good practice to keep the exception fields close to the rule. A rule writer should know best what makes up a good exception to a rule, and can define tuples of fields that define an exception. This can discourage an exception based on something overly broad such as a process or image name in isolation.
* Although we want to encourage rule writers to define exceptions as a tuple of field names, in practice we hope that many rules will not require any exceptions. However, if field names are in a separate object, you always have to create a separate exception object for each new rule and jump between them to find the rule's conditions and valid exceptions.
* Appending/overriding exceptions is very unambiguous when the items are in a separate object. You're always simply appending the tuple values to each named list when append=true, or replacing the lists when append=false. If you have the fields and items together in a single object, what does append mean? Is it legal to add to the list of field name tuples with append=true, or should append only apply to the field values?

I'd like of course to see what other people think. I'm sure we can make it work one way or the other.

Kaizhe:
@mstemm , I am thinking the exception in a different way:
```
exception: Write below binary dir
  excep_1:
    container.image.repository: docker.io/golang-alpine
    fd.name: /usr/libexec/go
  excep_2:
    proc.name: touch
    fd.name: /check
  append: true
```

mstemm:
I see—the field names and values would be combined.

That would make it really hard to express lists of exceptions though, wouldn’t it? You’d have to repeat the field names over and over for each set of field values.

Kaizhe:
> I see—the field names and values would be combined.


> I see—the field names and values would be combined.
> 
> That would make it really hard to express lists of exceptions though, wouldn’t it? You’d have to repeat the field names over and over for each set of field values.

You're right, or

```
exception: Write below binary dir
  excep_1:  # (container.image.repository in <list> and fd.name in <list>)
    container.image.repository: [docker.io/golang-alpine]
    fd.name: [/usr/libexec/go]
  excep_2: # (proc.name in <list> and fd.name in <list>)
    proc.name: [touch]
    fd.name: [/check]
  append: true
```
This is like privilege grant in the role definition in k8s.

mstemm:
We discussed this during the Falco community call on 9/16 and I think there was a slight preference for having the exception fields embedded in the rule. I encouraged people during the call to add their preferences to the comments here.

It also sounds like there was consensus to add this to Falco, so hopefully we can merge this PR and I'll also jump in on the implementation!

fntlnz (commented):
I've been going trough this with @leodido so this comment is from a shared thought:

We analyzed the proposed language addition and we understood that to make a complete exception one would need to write something like:


```yaml
- rule: Write below binary dir
  desc: an attempt to write to any file below a set of binary directories
  condition: >
    bin_dir and evt.dir = < and open_write
    and not package_mgmt_procs
    and not exe_running_docker_save
    and not python_running_get_pip
    and not python_running_ms_oms
    and not user_known_write_below_binary_dir_activities
  exceptions:
   - proc_writer: [proc.name, fd.directory]
   - container_writer: [container.image.repository, fd.directory]

- exception: Write below binary dir
  items:
  - proc_writer:
    - [apk, /usr/lib/alpine]
    - [npm, /usr/node/bin]
  - container_writer:
    - [docker.io/alpine, /usr/libexec/alpine]
```


We unrolled that and we think that if we want to make the exception a rules only construct, we'd prefer them to be 100% colocated to the rule. Something like this:


```yaml
- rule: Write below binary dir
  desc: an attempt to write to any file below a set of binary directories
  condition: >
    bin_dir and evt.dir = < and open_write
    and not package_mgmt_procs
    and not exe_running_docker_save
    and not python_running_get_pip
    and not python_running_ms_oms
    and not user_known_write_below_binary_dir_activities
  exceptions:
   - proc_writer:
      - proc.name: apk
        fd.directory: /usr/lib/alpine
      - proc.name: npm
        fd.directory: /usr/node/bin
   - container_writer:
     - container.image.repository: docker.io/alpine
       fd.directory: /usr/libexec/alpine
```

This completely remove the need of a first class citizen exception construct which could also lead to misunderstandings while using them because one might think that they are reusable given their top level nature.

Ideas? Do you envision any pros/cons of this other approach? We kind tried to reverse engineer your thought for the syntax :)

leogr (commented):
Quoting this [comment](https://github.com/falcosecurity/falco/pull/1376#pullrequestreview-491419222):

> We unrolled that and we think that if we want to make the exception a rules only construct, we'd prefer them to be 100% colocated to the rule. Something like this:
> 
> ```yaml
> - rule: Write below binary dir
>   desc: an attempt to write to any file below a set of binary directories
>   condition: >
>     bin_dir and evt.dir = < and open_write
>     and not package_mgmt_procs
>     and not exe_running_docker_save
>     and not python_running_get_pip
>     and not python_running_ms_oms
>     and not user_known_write_below_binary_dir_activities
>   exceptions:
>       - proc.name: apk
>         fd.directory: /usr/lib/alpine
>       - proc.name: npm
>         fd.directory: /usr/node/bin
>       - container.image.repository: docker.io/alpine
>         fd.directory: /usr/libexec/alpine
> ```
> 
> This completely remove the need of a first class citizen exception construct which could also lead to misunderstandings while using them because one might think that they are reusable given their top level nature.

I agree to make the exception a rule only construct.

However, the approach above will still let users the ability to use `append: true` and extend the exception, right?

For example:
```yaml
- rule: Write below binary dir
  append: true
  exceptions:
       - proc.name: ...
         fd.directory: /path/to/...
```

Though exceptions might not always be colocated to the rule, I believe that's acceptable.

leodido:
> I ended up changing the structure of exceptions a bit, but I think with a couple of small additional changes it will be easier to think of exceptions as properties on rules objects with an easy way to append to rules, as you suggested.
> 
> The new syntax looks like this:
> 
> ```
> - rule: Write below binary dir
>   desc: an attempt to write to any file below a set of binary directories
>   condition: >
>     bin_dir and evt.dir = < and open_write
>     and not package_mgmt_procs
>     and not exe_running_docker_save
>     and not python_running_get_pip
>     and not python_running_ms_oms
>     and not user_known_write_below_binary_dir_activities
>   exceptions:
>    - name: proc_writer
>      fields: [proc.name, fd.directory]
>    - name: container_writer
>      fields: [container.image.repository, fd.directory]
>      comps: [=, startswith]
> 
> - exception: Write below binary dir
>   items:
>   - name: proc_writer
>     values:
>     - [apk, /usr/lib/alpine]
>     - [npm, /usr/node/bin]
>   - name: container_writer
>     values:
>     - [docker.io/alpine, /usr/libexec/alpine]
> ```
> 
> Instead of having object keys be used for exception item names, the exceptions property for a rule just has a list of objects. The name provides the name, and the fields property has the field names.
> 
> Note that I added a comps property that will allow for comparison operators other than =. If not specified, = is assumed.
> 
> Separately, exception objects also have an items property with a list of objects. The values property is a list of lists containing field values.
> 
> So starting with this idea, I think a way to allow exceptions to be colocated with a rule _and_ allow appending to exceptions would be to 1) allow the items for the rule exceptions property to optionally contain a values property with lists of field values:
> 
> ```
> - rule: Write below binary dir
>   desc: an attempt to write to any file below a set of binary directories
>   condition: >
>     bin_dir and evt.dir = < and open_write
>     and not package_mgmt_procs
>     and not exe_running_docker_save
>     and not python_running_get_pip
>     and not python_running_ms_oms
>     and not user_known_write_below_binary_dir_activities
>   exceptions:
>    - name: proc_writer
>      fields: [proc.name, fd.directory]
>      values: 
>         - [apk, /usr/lib/alpine]
>         - [npm, /usr/node/bin]
>    - name: container_writer
>      fields: [container.image.repository, fd.directory]
>      comps: [=, startswith]
> ```
> 
> and also 2) allow appending to a rule's exceptions by specifying append: true, specifying an exceptions property, and providing only name and values:
> 
> ```
> - rule: Write below binary dir
>   exceptions:
>     - name: container_writer
>       values:
>         - [docker.io/alpine, /usr/libexec/alpine]
>    append: true
> ```
> 
> Would that be a reasonable compromise?

It would for me.

I like way more this re-design of the `exception` construct.

mstemm:
Yeah to be clear both 1) and 2) would be supported. They are two halves of the complete syntax.

Ok thanks, I will adjust the PR #1427 to match this.

leogr:
I like this new re-design too!

Just a doubt, what will happen if...
```
- rule: Write below binary dir
  exceptions:
    - name: container_writer
      fields: [something, something] // NOTE: I have changed this when appending
      values:
        - [docker.io/alpine, /usr/libexec/alpine]
   append: true
```

:smile_cat:  ?

mstemm:
Appending to a rule and specifying exception fields wouldn't be allowed, so the rule loading will return an error. The same thing happens if you append to a rule but specify an output.

leogr:
Could you rebase this PR, please?
That should fix the two failing checks.

proposals/20200828-structured-exception-handling.md
```diff
@@ -0,0 +1,242 @@
...
+    and not package_mgmt_procs
+    and not exe_running_docker_save
+    and not python_running_get_pip
+    and not python_running_ms_oms
+    and not user_known_write_below_binary_dir_activities
+...
+```
+In most cases, these exceptions are expressed as concatenations to the original rule's condition. For example, looking at the macro package_mgmt_procs:
+
+```
```
leogr, line 26:
```suggestion
```yaml
```

proposals/20200828-structured-exception-handling.md
```diff
@@ -0,0 +1,242 @@
...
+...
+```
+
+The exceptions all generally follow the same structure--naming a program and a directory prefix below /etc where that program is allowed to write files.
+
+### Using Appends/Overwrites to Customize Rules
+
+An important way to customize rules and macros is to use `append: true` to add to them, or `append: false` to define a new rule/macro, overwriting the original rule/macro. Here's an example from Update Package Repository:
+
+```
```
leogr, line 56:
```suggestion
```yaml
```

proposals/20200828-structured-exception-handling.md
```diff
@@ -0,0 +1,242 @@
...
+  condition: >
+    ((open_write and access_repositories) or (modify and modify_repositories))
+`    and not package_mgmt_procs
+    and not exe_running_docker_save
+    and not user_known_update_package_registry
+```
+
+If someone wanted to add additional exceptions to this rule, they could add the following to the user_rules file:
+
+```
```
leogr, line 77:
```suggestion
```yaml
```

proposals/20200828-structured-exception-handling.md
```diff
@@ -0,0 +1,242 @@
...
+* adds to package_mgmt_procs, adding an exception for programs spawned by chef (because append is true)
+* overrides the macro user_known_update_package_registry to add an exception for npm (because append is false).
+
+### Problems with Appends/Overrides to Define Exceptions
+
+Although the concepts of macros and lists in condition fields, combined with appending to lists/conditions in macros/rules, is very general purpose, it can be unwieldy:
+
+* Appending to conditions can result in incorrect behavior, unless the original condition has its logical operators set up properly with parentheses. For example:
+
+```
```
leogr, line 102:
```suggestion
```yaml
```

proposals/20200828-structured-exception-handling.md
```diff
@@ -0,0 +1,242 @@
...
+append: true
+```
+
+Results in unintended behavior. It will match any fd related event where the name is /tmp/baz, when the intent was probably to add /tmp/baz as an additional opened file.
+
+* A good convention many rules use is to have a clause "and not user_known_xxxx" built into the condition field. However, it's not in all rules and its use is a bit haphazard.
+
+* Appends and overrides can get confusing if you try to apply them multiple times. For example:
+
+```
```
leogr, line 117:
```suggestion
```yaml
```

proposals/20200828-structured-exception-handling.md
```diff
@@ -0,0 +1,242 @@
...
+append: true
+```
+
+If someone wanted to override the original behavior of allowed_files, they would have to use `append: false` in a third definition of allowed_files, but this would result in losing the append: true override.
+
+## Solution: Exceptions as first class objects
+
+To address some of these problems, we will add the notion of Exceptions as top level objects alongside Rules, Macros, and Lists. A rule that supports exceptions must define a new key `exceptions` in the rule. The exceptions key is a list of identifier plus list of tuples of filtercheck fields. Here's an example:
+
+```
```
leogr, line 134:
```suggestion
```yaml
```

proposals/20200828-structured-exception-handling.md
```diff
@@ -0,0 +1,242 @@
...
+
+proc_filenames uses the in comparison operator, so the corresponding values entry should be a list of filenames.
+
+filenames differs from the others in that it names a single field and single comp operator. This changes how the exception condition snippet is constructed (see below).
+
+Notice that exceptions are defined as a part of the rule. This is important because the author of the rule defines what construes a valid exception to the rule. In this case, an exception can consist of a process and file directory (actor and target), but not a process name only (too broad).
+
+Exception values will most commonly be defined in rules with append: true. Here's an example:
+
+```
```
leogr, line 176:
```suggestion
```yaml
```

proposals/20200828-structured-exception-handling.md
```diff
@@ -0,0 +1,242 @@
...
+## Summary
+
+## Motivation
+
+Almost all Falco Rules have cases where the behavior detected by the
+rule should be allowed. For example, The rule Write Below Binary Dir
+has exceptions for specific programs that are known to write below
+these directories as a part of software installation/management:
+
+```
```
leogr, line 12:
```suggestion
```yaml
```

proposals/20200828-structured-exception-handling.md
```diff
@@ -0,0 +1,241 @@
...
+  condition: proc.name in (package_mgmt_binaries)
+
+- macro: user_known_update_package_registry
+  condition: (never_true)
+
+- rule: Update Package Repository
+  desc: Detect package repositories get updated
+  condition: >
+    ((open_write and access_repositories) or (modify and modify_repositories))
+`    and not package_mgmt_procs
```
leogr, line 70:
```suggestion
    and not package_mgmt_procs
```

proposals/20200828-structured-exception-handling.md
```diff
@@ -0,0 +1,241 @@
...
+    values:
+    - [apk, /usr/lib/alpine]
+    - [npm, /usr/node/bin]
+  - name: container_writer
+    values:
+    - [docker.io/alpine, /usr/libexec/alpine]
+  - name: proc_filenames
+    values:
+	 - [apt, apt_files]
+	 - [rpm, [/bin/cp, /bin/pwd]]
```
leogr, line 192:
```suggestion
    - [apt, apt_files]
    - [rpm, [/bin/cp, /bin/pwd]]
```

proposals/20200828-structured-exception-handling.md
```diff
@@ -0,0 +1,241 @@
...
+    and not user_known_write_below_binary_dir_activities
+  exceptions:
+   - name: proc_writer
+     fields: [proc.name, fd.directory]
+   - name: container_writer
+     fields: [container.image.repository, fd.directory]
+     comps: [=, startswith]
+   - name: proc_filenames
+     fields: [proc.name, fd.name]
+	 comps: [=, in]
```
leogr, line 152:
```suggestion
     comps: [=, in]
```

leogr (approved):

fntlnz (approved):

Issue #1277, opened 2020-06-24
Fields by engine version


fntlnz:
**What to document**

While working on #1273 I noticed that the `required_engine_version` field can be a bit confusing to users due to the fact that we don't two key things:

- We don't know how an engine version is correlated to a Falco version (unless you dig into git)
- We don't have a list of what fields are in what engine version

I'd like to have them documented.

leodido:
Agree.

Just for completeness, about the first bullet point, I opened https://github.com/falcosecurity/falco/issues/1269 2 days ago.

Basically, the Falco Engine version behave like the driver version w.r.t. the Falco version. And we need to expose it in an easier way for users.

About point 2: I also agree. In the comment [here](https://github.com/falcosecurity/event-generator/pull/31#pullrequestreview-434756749) I proposed a mapping Falco Engine version <=> fields <=> rules.

PR #1273, opened 2020-06-23
rules: the required_engine_version is now on by default

fntlnz:
Co-Authored-By: Leonardo Di Donato <leodidonato@gmail.com>
Signed-off-by: Lorenzo Fontana <lo@linux.com>

**What type of PR is this?**
/kind feature

> If contributing rules or changes to rules, please make sure to also uncomment one of the following line:

/kind rule-update

**Any specific area of the project related to this PR?**

/area rules

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #1272

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
update: the required_engine_version is now on by default 
rule(Container Drift Detected (open+create)): specify that rule is only enabled with engine 6
rule(Create Disallowed Pod): required_engine_version 5
rule(Create Privileged Pod): required_engine_version 5
rule(Create Sensitive Mount Pod): required_engine_version 5
rule(Create HostNetwork Pod): required_engine_version 5
rule(Pod Created in Kube Namespace): required_engine_version 5
rule(ClusterRole With Wildcard Created): required_engine_version 5
rule(ClusterRole With Write Privileges Created): required_engine_version 5
rule(ClusterRole With Pod Exec Created): required_engine_version 5

```

PR #1381, opened 2020-09-07
update: bump Falco engine version to 7

### LINKED ISSUE #1272: Falco rules file should have `required_engine_version: 6`
**Describe the bug**

Hi, I noticed that https://github.com/falcosecurity/falco/commit/70b9bfe1d6bf4d44e7ad000d726603418ad37342#diff-6acd8aa2e534d1ed83e275bf658d936b added a new rule Container Drift Detected (open+create) that requries falco engine version 6. However, the rules file itself doesn’t have any required_engine_version block. There’s a commented out version here: https://github.com/falcosecurity/falco/blob/master/rules/falco_rules.yaml#L26. 

If users tried to load this rules file with older falco versions, they wouldn't be able to load the file, as the new field `evt.is_open_exec` doesn't work with older falco engine versions.

I'd suggest at least updating the required_engine_version to 6 in the rules file.

**How to reproduce it**

**Expected behaviour**

**Screenshots**

**Environment**

- Falco version:

- System info:

- OS:

- Kernel:

- Installation method:

**Additional context**

### LINKED ISSUE #1380: Falco engine version not updated when renameat2 was added
I noticed that this rules change (https://github.com/falcosecurity/falco/commit/1f1f7c16b66b18ab6c3c36f45b1296c2a52b697a#diff-6acd8aa2e534d1ed83e275bf658d936b) added support for the renameat2 syscall, which was added to sysdig in this commit: https://github.com/draios/sysdig/commit/f0bf26959c6a9bd1856d45ee5c80448201124f08#diff-b17bd28471a61e4b30f3c557d17da87a. I think we should have updated the falco engine version from 6 to 7 when falco was updated  to use this sysdig version, as it's a change to the set of fields that the engine supports.

Also, I think at the time the falco_rules file was updated, we should have modified the required_engine_version field in the rules file to set required_engine_version to 7. Also, it doesn't seem like it was updated when an earlier change (https://github.com/falcosecurity/falco/commit/70b9bfe1d6bf4d44e7ad000d726603418ad37342#diff-6acd8aa2e534d1ed83e275bf658d936b) added support for new modes that hadn't been previously allowed in sysdig.

If you try to read the current rules file with falco 0.24.0, you'll get this error:

```
Thu Sep  3 11:33:26 2020: Runtime error: Error loading rules: /usr/share/falco/lua/sinsp_rule_utils.lua:107: Unknown event/syscall "renameat2" in filter: ((open_write and access_repositories) or (modify and modify_repositories)) and not package_mgmt_procs and not package_mgmt_ancestor_procs and not exe_running_docker_save and not user_known_update_package_registry
```

A preferred error would be the one denoting the required engine version. I set the required_engine_version in the rules file as an example.

```
Thu Sep  3 11:36:07 2020: Runtime error: Rules require engine version 7, but engine version is 6
---
- required_engine_version: 7
```

**Expected behaviour**

**Screenshots**

**Environment**

- Falco version:
Falco version: 0.24.0
Driver version: 85c88952b018fdbce2464222c3303229f5bfcfad
- System info:
{
  "machine": "x86_64",
  "nodename": "ubuntu",
  "release": "4.15.0-15-generic",
  "sysname": "Linux",
  "version": "#16~16.04.1-Ubuntu SMP Thu Apr 5 12:19:23 UTC 2018"
}
- OS:

- Kernel:

- Installation method:

**Additional context**

leogr:
**What type of PR is this?**

/kind bug
/kind feature

/area engine
/area rules

This PR bumps the engine version to the expected number (7), as per https://falco.org/docs/rules/#versioning

Furthermore, it also sets the `required_engine_version` to 7 in the rule files, to show the expected error when the ruleset does not match. Unfortunately, this must be set to the entire rule files (even though it's required just by few rules) since the [`required_engine_version` does not work on rule level](https://github.com/falcosecurity/falco/pull/1273#issuecomment-662406908) (thus the PR #1273 is blocked).

/milestone 0.26.0

Fixes #1380
Fixes #1272

I'd propose closing the https://github.com/falcosecurity/falco/pull/1273 in favor of this since the current implementation does not support `required_engine_version` at rule level.

An alternative of setting the `required_engine_version` to 7 for the whole file is to create a rules file for every version depending on what are the supported fields, as initially proposed in [this comment](https://github.com/falcosecurity/falco/issues/1272#issuecomment-648015596). If we got consensus for this alternative, I will open another PR (since the main purpose of this PR is just to bump the engine version number and not to fix the whole problem).

Finally, this PR intentionally does not try to solve the other problem reported by https://github.com/falcosecurity/falco/issues/1277

/cc @fntlnz 
/cc @leodido
/cc @mstemm 

```release-note
update: bump Falco engine version to 7
update: the required_engine_version is now on by default
```

### DIFF
--- a/userspace/engine/falco_engine_version.h
+++ b/userspace/engine/falco_engine_version.h
@@ -1,5 +1,5 @@
 /*
-Copyright (C) 2019 The Falco Authors.
+Copyright (C) 2020 The Falco Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -16,7 +16,7 @@ limitations under the License.
 
 // The version of rules/filter fields/etc supported by this falco
 // engine.
-#define FALCO_ENGINE_VERSION (6)
+#define FALCO_ENGINE_VERSION (7)
 
 // This is the result of running "falco --list -N | sha256sum" and
 // represents the fields supported by this version of falco. It's used
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2019 The Falco Authors.
+# Copyright (C) 2020 The Falco Authors.
 #
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -23,7 +23,7 @@
 # change to this rules file, we'll uncomment this line and set it to
 # the falco engine version in use at the time.
 #
-#- required_engine_version: 2
+- required_engine_version: 7
 
 # Currently disabled as read/write are ignored syscalls. The nearly
 # similar open_write/open_read check for files being opened for


fntlnz (approved):

userspace/engine/falco_engine_version.h
```diff
@@ -16,7 +16,7 @@ limitations under the License.
 
 // The version of rules/filter fields/etc supported by this falco
 // engine.
-#define FALCO_ENGINE_VERSION (6)
+#define FALCO_ENGINE_VERSION (7)
```
mstemm, line 19: I see that FALCO_FIELDS_CHECKSUM was already updated with the new checksum, great. Generally, whenever that checksum changes, the engine version needs to change too.

mstemm (approved):

PR #1384, opened 2020-09-08
K8s env, best practice, container.info over container.id

epcim:
**What type of PR is this?**

/kind rule-update

/area rules

In my case, I log to ES, I want all informations about running pods available within the alert/audit record. Output field of rule, frequently miss important informations (ie, namespace). container.info macro instead can provide bunch of usefull details if k8s is configured (otherwise it has fallback). 

It is then better, to log all details and have them either available on alerts.  

(Here I miss some clarification, havent found it documented, but it seems to me, that only details about an event, that are part of falcosidekick Outputfields (https://github.com/falcosecurity/falcosidekick/blob/master/outputs/alertmanager.go#L25) are the ones key=value between "()" of the output field of rule. Sorry I am lazy to find it again in src code - it might be bit different, but the truth is that alert does not have all event details.

n/a

racionale here is as I cant get more details in alert, I do have need use flexible %container.info macro to give me all important (namespace, podname, ...)

```release-note
rule(Disallowed SSH Connection): use container.info in output fields
rule(Unexpected outbound connection destination): use container.info in output fields
rule(Unexpected inbound connection source): use container.info in output fields
rule(Modify Shell Configuration File): use container.info in output fields
rule(Read Shell Configuration File): use container.info in output fields
rule(Schedule Cron Jobs): use container.info in output fields
rule(Update Package Repository): use container.info in output fields
rule(Read ssh information): use container.info in output fields
rule(Write below etc): use container.info in output fields
rule(Read sensitive file trusted after startup): use container.info in output fields
rule(Write below rpm database): use container.info in output fields
rule(Modify binary dirs): use container.info in output fields
rule(Mkdir binary dirs): use container.info in output fields
rule(System user interactive): use container.info in output fields
rule(Terminal shell in container): use container.info in output fields
rule(System procs network activity): use container.info in output fields
rule(Program run with disallowed http proxy env): use container.info in output fields
rule(Interpreted procs inbound network activity): use container.info in output fields
rule(Interpreted procs outbound network activity): use container.info in output fields
rule(Unexpected UDP Traffic): use container.info in output fields
rule(User mgmt binaries): use container.info in output fields
rule(Create files below dev): use container.info in output fields
rule(Unexpected K8s NodePort Connection): use container.info in output fields
rule(Netcat Remote Code Execution in Container): use container.info in output fields
rule(Launch Suspicious Network Tool in Container): use container.info in output fields
rule(Launch Suspicious Network Tool on Host): use container.info in output fields
rule(Search Private Keys or Passwords): use container.info in output fields
rule(Clear Log Activities): use container.info in output fields
rule(Remove Bulk Data from Disk): use container.info in output fields
rule(Create Hidden Files or Directories): use container.info in output fields
rule(Launch Remote File Copy Tools in Container): use container.info in output fields
rule(Create Symlink Over Sensitive Files): use container.info in output fields
rule(Detect outbound connections to common miner pool ports): use container.info in output fields
rule(Detect crypto miners using the Stratum protocol): use container.info in output fields
rule(Packet socket created in container): use container.info in output fields
rule(Network Connection outside Local Subnet): use container.info in output fields
rule(Redirect STDOUT/STDIN to Network Connection in Container): use container.info in output fields
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -368,7 +368,7 @@
 - rule: Disallowed SSH Connection
   desc: Detect any new ssh connection to a host other than those in an allowed group of hosts
   condition: (inbound_outbound) and ssh_port and not allowed_ssh_hosts
-  output: Disallowed SSH Connection (command=%proc.cmdline connection=%fd.name user=%user.name user_loginuid=%user.loginuid container_id=%container.id image=%container.image.repository)
+  output: Disallowed SSH Connection (command=%proc.cmdline connection=%fd.name user=%user.name user_loginuid=%user.loginuid %container.info image=%container.image.repository)
   priority: NOTICE
   tags: [network, mitre_remote_service]
 
@@ -399,7 +399,7 @@
     ((fd.sip in (allowed_outbound_destination_ipaddrs)) or
      (fd.snet in (allowed_outbound_destination_networks)) or
      (fd.sip.name in (allowed_outbound_destination_domains)))
-  output: Disallowed outbound connection destination (command=%proc.cmdline connection=%fd.name user=%user.name user_loginuid=%user.loginuid container_id=%container.id image=%container.image.repository)
+  output: Disallowed outbound connection destination (command=%proc.cmdline connection=%fd.name user=%user.name user_loginuid=%user.loginuid %container.info image=%container.image.repository)
   priority: NOTICE
   tags: [network]
 
@@ -422,7 +422,7 @@
     ((fd.cip in (allowed_inbound_source_ipaddrs)) or
      (fd.cnet in (allowed_inbound_source_networks)) or
      (fd.cip.name in (allowed_inbound_source_domains)))
-  output: Disallowed inbound connection source (command=%proc.cmdline connection=%fd.name user=%user.name user_loginuid=%user.loginuid container_id=%container.id image=%container.image.repository)
+  output: Disallowed inbound connection source (command=%proc.cmdline connection=%fd.name user=%user.name user_loginuid=%user.loginuid %container.info image=%container.image.repository)
   priority: NOTICE
   tags: [network]
 
@@ -461,7 +461,7 @@
     and not proc.name in (shell_binaries)
     and not exe_running_docker_save
   output: >
-    a shell configuration file has been modified (user=%user.name user_loginuid=%user.loginuid command=%proc.cmdline pcmdline=%proc.pcmdline file=%fd.name container_id=%container.id image=%container.image.repository)
+    a shell configuration file has been modified (user=%user.name user_loginuid=%user.loginuid command=%proc.cmdline pcmdline=%proc.pcmdline file=%fd.name %container.info image=%container.image.repository)
   priority:
     WARNING
   tags: [file, mitre_persistence]
@@ -483,7 +483,7 @@
      fd.directory in (shell_config_directories)) and
     (not proc.name in (shell_binaries))
   output: >
-    a shell configuration file was read by a non-shell program (user=%user.name user_loginuid=%user.loginuid command=%proc.cmdline file=%fd.name container_id=%container.id image=%container.image.repository)
+    a shell configuration file was read by a non-shell program (user=%user.name user_loginuid=%user.loginuid command=%proc.cmdline file=%fd.name %container.info image=%container.image.repository)
   priority:
     WARNING
   tags: [file, mitre_discovery]
@@ -503,7 +503,7 @@
     not user_known_cron_jobs
   output: >
     Cron jobs were scheduled to run (user=%user.name user_loginuid=%user.loginuid command=%proc.cmdline
-    file=%fd.name container_id=%container.id container_name=%container.name image=%container.image.repository:%container.image.tag)
+    file=%fd.name %container.info container_name=%container.name image=%container.image.repository:%container.image.tag)
   priority:
     NOTICE
   tags: [file, mitre_persistence]
@@ -513,7 +513,7 @@
 # When displaying container information in the output field, use
 # %container.info, without any leading term (file=%fd.name
 # %container.info user=%user.name user_loginuid=%user.loginuid, and not file=%fd.name
-# container=%container.info user=%user.name user_loginuid=%user.loginuid). The output will change
+# %container.info user=%user.name user_loginuid=%user.loginuid). The output will change
 # based on the context and whether or not -pk/-pm/-pc was specified on
 # the command line.
 - macro: container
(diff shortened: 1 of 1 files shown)


epcim:
To change the rules, I have used these few SED cmds..https://github.com/epcim/falco/blob/container-info-dockerfile/docker/falco/Dockerfile#L113 (the targed branch is example, I do have different build pipeline)

leogr (commented):
LGTM

leogr (approved):

leodido (commented):
Thansk @epcim for sending this.

Could you please adjust the release-note block as per [contributing guidelines](https://github.com/falcosecurity/.github/blob/master/CONTRIBUTING.md#rule-type) (it should theoretically list all the rules when you made changes) and rebase this on top of the master?

epcim:
> Thansk @epcim for sending this.
> 
> Could you please adjust the release-note block as per [contributing guidelines](https://github.com/falcosecurity/.github/blob/master/CONTRIBUTING.md#rule-type) (it should theoretically list all the rules when you made changes) and rebase this on top of the master?
> 
> Anyways, I tend to agree with your changes but there's at least a point that we need to take into consideration.
> 
> When `container.info` is not available the output will contain its fallback. Is such fallback something like what we have now (eg., `container_id=<ID>`)?
> Could you provide a fallback example, please?

- ok will rebase
- technically it possibly touch many rules, so release-note this time needs to be "generic statement" but I will list few If catch them.
- as far as I remember, container.info is makro has `container_id=<ID>` as fallback

epcim:
@leodido so I have rebased + updated commit message. 

`Each rule change must be on its own commit...`
For obvious reason I want to avoid split this into series of 37 commits.

Modified rules are:
- Disallowed SSH Connection
- Unexpected outbound connection destination
- Unexpected inbound connection source
- Modify Shell Configuration File
- Read Shell Configuration File
- Schedule Cron Jobs
- Update Package Repository
- Read ssh information
- Write below etc
- Read sensitive file trusted after startup
- Write below rpm database
- Modify binary dirs
- Mkdir binary dirs
- System user interactive
- Terminal shell in container
- System procs network activity
- Program run with disallowed http proxy env
- Interpreted procs inbound network activity
- Interpreted procs outbound network activity
- Unexpected UDP Traffic
- User mgmt binaries
- Create files below dev
- Unexpected K8s NodePort Connection
- Netcat Remote Code Execution in Container
- Launch Suspicious Network Tool in Container
- Launch Suspicious Network Tool on Host
- Search Private Keys or Passwords
- Clear Log Activities
- Remove Bulk Data from Disk
- Create Hidden Files or Directories
- Launch Remote File Copy Tools in Container
- Create Symlink Over Sensitive Files
- Detect outbound connections to common miner pool ports
- Detect crypto miners using the Stratum protocol
- Packet socket created in container
- Network Connection outside Local Subnet
- Redirect STDOUT/STDIN to Network Connection in Container

epcim:
@leodido all ok?

Kaizhe:
I think `image=%container.image.repository` will be redundant if we have `container.info`
I have some concerns over replacing an output field, as it may breaks the consumer code who may parse by `container.id`

leodido:
@epcim I was referring to put them in the release-note block only in this case, not every rule change in its own commit clearly 😊 

I'm okay with the changes. But I also get the @Kaizhe point.

It's definitely something to discuss in the community call!

Thanks for your awesome contribution!

leodido (approved):

fntlnz:
Putting this on hold until we have two positive reviews, feel free to hold cancel once this is ready. I need to unblock the merge queue.

/hold

leodido:
Two questions from @ldegio  (during our Falco Community Calls) today:

- why we can't just retain `container_id=%container.id` field and just append `%container.info` after it?
- what is the typical format of a `%container.info`? does it contains a JSON object, tuples, what?

Because maybe, disregarding eventual container ID duplication, this could be the way to go to also take care of the @Kaizhe's legit concerns.

Kaizhe:
> Two questions from @ldegio (during our Falco Community Calls) today:
> 
> * why we can't just retain `container_id=%container.id` field and just append `%container.info` after it?
> * what is the typical format of a `%container.info`? does it contains a JSON object, tuples, what?
> 
> Because maybe, disregarding eventual container ID duplication, this could be the way to go to also take care of the @Kaizhe's legit concerns.

Appending `container.info` in the output looks good to me

Kaizhe (changes requested):
Please append `container.info` instead of replacing `container.id`

epcim:
> container.info 

`vs container.id vs. id=...`

So do we need any other default than? https://github.com/falcosecurity/falco/blob/3693b16c917d6920adffac03e80d38c1b64317ea/userspace/engine/lua/rule_loader.lua#L694


> I have some concerns over replacing an output field, as it may breaks the consumer code who may parse by `container.id`

As long as `container.info` has fallback to `container.id` and as long as it containse container.id, thats not an issue:

Kaizhe:
> > container.info
> 
> `vs container.id vs. id=...`
> 
> So do we need any other default than?
> 
> https://github.com/falcosecurity/falco/blob/3693b16c917d6920adffac03e80d38c1b64317ea/userspace/engine/lua/rule_loader.lua#L694
> 
> > I have some concerns over replacing an output field, as it may breaks the consumer code who may parse by `container.id`
> 
> As long as `container.info` has fallback to `container.id` and as long as it containse container.id, thats not an issue:

Can you confirm I can use the same json pointer(I don't need to change anything, e.g. pointer name, level) to get the container ID if you use `container.info`?

epcim:
> > > container.info
> > 
> > 
> > `vs container.id vs. id=...`
> > So do we need any other default than?
> > https://github.com/falcosecurity/falco/blob/3693b16c917d6920adffac03e80d38c1b64317ea/userspace/engine/lua/rule_loader.lua#L694
> > 
> > > I have some concerns over replacing an output field, as it may breaks the consumer code who may parse by `container.id`
> > 
> > 
> > As long as `container.info` has fallback to `container.id` and as long as it contains container.id, thats not an issue:
> 
> Can you confirm I can use the same json pointer(I don't need to change anything, e.g. pointer name, level) to get the container ID if you use `container.info`?

Seems like, but somebody else better to confirm. I am not Falco developer, just contributor. Macro `container.info` need some ownership and as you suggested it req. to be backward compatible, future support and notify existing users early if any change in behaviour.

Also `image=%container.image.repository will be redundant` is still not fixed. Having some filter that would uniq. filter keys of `extra_` would be handy.

epcim:
I wont do rebase anymore.On 20 Jan 2021 17:06, poiana <notifications@github.com> wrote:
@epcim: PR needs rebase.

Instructions for interacting with me using PR comments are available here.  If you have questions or suggestions related to my behavior, please file an issue against the kubernetes/test-infra repository.

PR #1419, opened 2020-09-29
feat(falco): Provide a parameter for loading lua files from an alternate path

admiral0:
This will be used by the static build to load lua files from
alternate directories that are not tied to the compile flags

**What type of PR is this?**

/kind feature

/area engine

This allows us to provide tarballs of static builds of falco and use `--alternate-lua-dir $(pwd)/lua` to specify where to look for lua
files.

-

This will improve user experience for users of the static build, because currently that path is hardcoded at compile time.

```release-note
new: CLI flag `--alternate-lua-dir` to load Lua files from arbitrary paths
```

### DIFF
--- a/userspace/falco/falco.cpp
+++ b/userspace/falco/falco.cpp
@@ -86,6 +86,7 @@ static void usage()
 	   " -h, --help                    Print this page\n"
 	   " -c                            Configuration file (default " FALCO_SOURCE_CONF_FILE ", " FALCO_INSTALL_CONF_FILE ")\n"
 	   " -A                            Monitor all events, including those with EF_DROP_SIMPLE_CONS flag.\n"
+	   " --alternate-lua-dir <path>    Specify an alternate path for loading Falco lua files\n"
 	   " -b, --print-base64            Print data buffers in base64.\n"
 	   "                               This is useful for encoding binary data that needs to be used over media designed to.\n"
 	   " --cri <path>                  Path to CRI socket for container metadata.\n"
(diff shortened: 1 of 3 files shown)


userspace/falco/falco.cpp
```diff
@@ -479,7 +480,8 @@ int falco_init(int argc, char **argv)
 
 	static struct option long_options[] =
 	{
-		{"cri", required_argument, 0},
+		{"alternate-lua-dir", required_argument, 0},
```
fntlnz, line 483: please reformat the whole list using clang-format

userspace/falco/falco.cpp
```diff
@@ -86,6 +86,7 @@ static void usage()
 	   " -h, --help                    Print this page\n"
 	   " -c                            Configuration file (default " FALCO_SOURCE_CONF_FILE ", " FALCO_INSTALL_CONF_FILE ")\n"
 	   " -A                            Monitor all events, including those with EF_DROP_SIMPLE_CONS flag.\n"
+	   " --alternate-lua-dir <path>    Specify an alternate path for loading falco lua files\n"
```
fntlnz, line 89:
```suggestion
	   " --alternate-lua-dir <path>    Specify an alternate path for loading Falco lua files\n"
```

fntlnz: Added a suggestion to make Falco upper case. I know that there are other places where it's lower case but we are trying to slowly make everything adhere the branding guidelines.

https://github.com/falcosecurity/falco/tree/master/brand#yes


```
Writing about Falco
Yes

Notice the capitalization of the following terms.

    The Falco Project
    Falco

No

    falco
    the falco project
    the Falco project

```

fntlnz, line 89: This is probably better as `--lua-dir <path>` and in the description we can add `default " FALCO_ENGINE_SOURCE_LUA_DIR"`

fntlnz (commented):
Thanks for bringing this up @admiral0 - I also really needed this for some deployments I was doing this week - I like the general approach but had some comments.


Also please take a look here: https://github.com/falcosecurity/falco/blob/ac84e3e5f3931154c9803998753c2f70690fbaca/userspace/engine/falco_common.cpp#L59
Looks like we are ignoring the `alternate_lua_dir` parameter in falco common.

userspace/falco/falco.cpp
admiral0: it is not actually the lua dir. The main lua dir will still be specified at compile time from installation path. Is `--fallback-lua-dir` better?

fntlnz:
We probably need to add the trailing slash if the user didn't provide it:

This does not work
```
sudo ./userspace/falco/falco -r ../rules/falco_rules.yaml -u --alternate-lua-dir /tmp/lua 
alternate lua dir: /tmp/lua
Tue Sep 29 14:55:27 2020: Runtime error: Could not find Falco Lua entrypoint (tried /usr/share/falco/lua/rule_loader.lua, /tmp/luarule_loader.lua). Exiting.
```

This works

```
sudo ./userspace/falco/falco -r ../rules/falco_rules.yaml -u --alternate-lua-dir /tmp/lua/
alternate lua dir: /tmp/lua/
Tue Sep 29 14:56:19 2020: Falco version 0.26.0-2+ac84e3e (driver version 2aa88dcf6243982697811df4c1b484bcbe9488a2)
Tue Sep 29 14:56:19 2020: Falco initialized with configuration file /home/fntlnz/Projects/falcosecurity/falco/falco.yaml
Tue Sep 29 14:56:19 2020: Loading rules from file ../rules/falco_rules.yaml:
```

leodido (commented):
Thanks Radu for sending this PR!

This PR is in the right direction. Anyways, it is not completed yet IMHO.

For example in `falco_outputs.cpp` there is this line that needs to be updated too.

```cpp
falco_common::init(m_lua_main_filename.c_str(), FALCO_SOURCE_LUA_DIR);
```

Note that the `falco_common::init()` is also called by the ctor of `falco_engine` that you updated.

userspace/falco/falco.cpp
fntlnz: I actually took a closer look at the code, it looks like t he two variables are separate internally so `--alternate-lua-dir` is a good option to me.

leodido:
@admiral0 updated the release-note line, waiting for your new commits :)

userspace/falco/falco.cpp
leodido: I vote for `--alternate-lua-dir`

leogr (commented):
I'd personally prefer to completely remove the need for external files (by embedding them into the binary, as proposed in #1367).

However, I can agree with this approach for now :+1: 

Just my 2 cents :)

userspace/falco/falco.cpp
leodido, line 89:
```suggestion
	   " --alternate-lua-dir <path>    Specify an alternate path for loading Falco lua files\n"
```

admiral0: well, that's what happens when I do not do `git add` .....

leodido (approved):

fntlnz (approved):

PR #1420, opened 2020-09-29
rule(macro user_known_remote_file_copy_tools_in_container_conditions)

mobouchardcoveo:
**What type of PR is this?**

/kind rule-update

/area rules

With this we will be able to create custom condition. We have a special use case with a tool called Rosco.

None

None

```release-note
rule(macro user_known_remote_file_copy_tools_in_container_conditions): Allow custom condition for remote file copy tools
```


mobouchardcoveo:
Will create a new one. with signed-off-by

PR #1422, opened 2020-09-30
rule(macro user_know_remote_file_copy_tools_in_container_conditions)

mobouchardcoveo:
**What type of PR is this?**

/kind rule-update

/area rules

With this we will be able to create custom condition. We have a special use case with a tool called Rosco.

None

None

```release-note
rule(macro user_known_remote_file_copy_tools_in_container_conditions): Allow custom condition for remote file copy tools
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -2719,7 +2719,7 @@
 
 # Users should overwrite this macro to specify conditions under which a
 # Custom condition for use of remote file copy tool in container
-- macro: user_known_remote_file_copy_tools_in_container_conditions
+- macro: user_known_remote_file_copy_activities
   condition: (never_true)
 
 - rule: Launch Remote File Copy Tools in Container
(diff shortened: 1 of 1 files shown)


leodido (commented):
It seems a reasonable macro to add for exclusions.

Not sure though if we want its name to be `user_known_remote_file_copy_tools_in_container_conditions` or just something like `user_known_remote_file_copy_tools`.

I'd wait for @Kaizhe opinion anyway!

Kaizhe:
@mobouchardcoveo I know there some places using `user_known_xxx_conditions`, but his one is too long, let's use `user_known_remote_file_copy_activities`

Kaizhe:
And thank you for making the change! Really apprecaite!

mobouchardcoveo:
Thanks for the comment and yes I think it's better with your name. I just did the change.

Kaizhe (approved):

leodido (approved):

PR #1427, opened 2020-10-03
Add exceptions support

mstemm:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

/kind rule-update

> /kind rule-create

> /area build

/area engine

/area rules

> /area tests

> /area proposals

This adds support for exceptions as rule attributes as described in https://github.com/falcosecurity/falco/pull/1376. Also included are a set of changes to falco_rules.yaml to use exceptions whenever possible.

Fixes #

```release-note
new: Add support for exceptions as rule attributes to provide a compact way to add exceptions to rules.
```


mstemm:
I believe this is ready for review now, with all the changes (including rules changes) in place and all the tests passing.

leogr (commented):
First of all, you did a great job :clap: 

I tried this new feature, and it seems to work very well.

Although it seems good to me, I need just a bit more time to complete the review since this PR is huge :smiley_cat:

Meanwhile, I have left just two comments (see below).

leogr:
Also, using some operators can produce an invalid YAML syntax, for example:
```yaml
comps: !=
```
or
```yaml
comps: [!=]
```

Not sure if that is a real issue anyway since it can be just solved using a duoble-quoted string.

userspace/engine/lua/rule_loader.lua
```diff
@@ -126,6 +126,31 @@ function set_output(output_format, state)
    end
 end
 
+-- This should be keep in sync with parser.lua
+defined_comp_operators = {
+   ["="]=1,
+   ["=="] = 1,
+   ["!"] = 1,
```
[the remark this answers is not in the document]
mstemm: Yes, good catch! Fixed.

userspace/engine/lua/rule_loader.lua
```diff
@@ -592,52 +942,22 @@ function load_rules(sinsp_lua_parser,
 	    sinsp_rule_utils.check_for_ignored_syscalls_events(filter_ast, 'rule', v['rule'])
 	 end
 
-	 evttypes, syscallnums = sinsp_rule_utils.get_evttypes_syscalls(name, filter_ast, v['condition'], warn_evttypes, verbose)
+	 evttypes, syscallnums = sinsp_rule_utils.get_evttypes_syscalls(name, filter_ast, v['compile_condition'], warn_evttypes, verbose)
```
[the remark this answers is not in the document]
mstemm: I do think it's good to keep a verbose flag to suppress this message: https://github.com/falcosecurity/falco/blob/add-exceptions-support/userspace/engine/lua/sinsp_rule_utils.lua#L193. It can be useful when you're debugging, but I don't think you'd want to print it by default.

mstemm:
Okay, I rebased and fixed the one typo using !=. About the use of the operators direct in lists, I agree that the best solution will just be to quote the values. Also, I don't think the not equals style operators will be used that often anyway, as there's already an implied "not" by the use of the exception.

fntlnz (dismissed):

mstemm:
Added a couple of small changes to make it easier to deal with exceptions when processing events.

rules/falco_rules.yaml
```diff
@@ -395,10 +381,20 @@
 - rule: Unexpected outbound connection destination
   desc: Detect any outbound connection to a destination outside of an allowed set of ips, networks, or domain names
   condition: >
-    consider_all_outbound_conns and outbound and not
-    ((fd.sip in (allowed_outbound_destination_ipaddrs)) or
-     (fd.snet in (allowed_outbound_destination_networks)) or
-     (fd.sip.name in (allowed_outbound_destination_domains)))
+    consider_all_outbound_conns and outbound
+  exceptions:
```
leodido, line 385: Mark, just out of curiosity, have you run some benchmarks to evaluate the performance changes?

I mean, expressing "exceptions" with this new first-class citizen is faster or slower than expressing them with the "not" operator in the condition.

test/falco_tests.yaml
```diff
@@ -292,20 +295,10 @@ trace_files: !mux
       - rules/invalid_array_item_not_object.yaml
     trace_file: trace_files/cat_write.scap
 
-  invalid_unexpected object:
```
leodido, line 295: Is this deletion intended?

leodido: I saw that you deleted `test/rules/invalid_unexpected_object.yaml`. Is this test not needed anymore?

userspace/engine/falco_engine.cpp
```diff
@@ -350,33 +328,50 @@ unique_ptr<falco_engine::rule_result> falco_engine::process_k8s_audit_event(json
 	}
 
 	unique_ptr<struct rule_result> res(new rule_result());
+	res->source = "k8s_audit";
```
leodido, line 331: 👍

leodido (commented):
Good job!

Also, the extensive test suite for exceptions is lovely!

Left just some comments, but this looks good to me (as long the current failing tests get fixed).

rules/falco_rules.yaml
leodido: I imagine there's little (negligible) overhead during the parsing phase.

I don't expect performance to decrease at runtime since exceptions get translated...

Am I correct?

mstemm: The exceptions are folded into the condition string at the time the rules are loaded, so I'm not worried about the conversion overhead.

I think the resulting condition is as about as efficient as the current chain of macros are. Each macro effectively was one tuple of exception fields + values.

test/falco_tests.yaml
mstemm: Yes, this test was removed as the rule loading is now tolerant of unknown top level objects. This will allow defining new top level objects (something analogous to rule/exception/macro/list) without interfering with older falco instances that might not know about them.

rules/falco_rules.yaml
leodido: Yeah, exactly what I was supposing. Thanks for confirming it Mark!

test/falco_tests.yaml
leodido: Ok, that sounds like something to mention in the release notes :)

Because users may find ways to use this feature to annotate Falco rules ;)

leodido (approved):
LGTM

But let's keep this on hold until next week (ie., after Falco 0.27.0 is released)

fntlnz (approved):

leogr (approved):

PR #1447, opened 2020-10-15
fix(userspace/engine): free formatters, if any

### LINKED ISSUE #1446: segmentation fault after SIGHUP
**Describe the bug**

Falco can restart itself when a SIGHUP is received (this feature was introduced by #457).

Unfortunately, with the current development version (ie. the current code on the master branch) when a SIGHUP is received then after a while Falco crashes (just before printing out any other alerts).

Note that this problem is not present in the last published version `0.26.1`.

**How to reproduce it**

Build Falco from sources. I used `cmake -DCMAKE_BUILD_TYPE=Release ..`.

Run Falco (eg. `sudo userspace/falco/falco -c /etc/falco/falco.yaml`) and generate some activity (eg. `sudo cat /etc/shadow`).

Send a SIGHUP (eg. `sudo kill -s SIGHUP <pid>`). 

Falco will restart itself as expected, then generate some activity (eg. `sudo cat /etc/shadow`) again, and Falco will crash.

**Expected behaviour**

No segmentation fault.

**Screenshots**

```
Thu Oct 15 18:37:43 2020: Falco version 0.26.1-43+47fa7d5 (driver version 2aa88dcf6243982697811df4c1b484bcbe9488a2)
Thu Oct 15 18:37:43 2020: Falco initialized with configuration file /etc/falco/falco.yaml
Thu Oct 15 18:37:43 2020: Loading rules from file /etc/falco/falco_rules.yaml:
Thu Oct 15 18:37:43 2020: Loading rules from file /etc/falco/falco_rules.local.yaml:
Thu Oct 15 18:37:43 2020: Loading rules from file /etc/falco/k8s_audit_rules.yaml:
Thu Oct 15 18:37:44 2020: Starting internal webserver, listening on port 8765
18:37:48.016285983: Warning Sensitive file opened for reading by non-trusted program (user=root user_loginuid=1000 program=cat command=cat /etc/shadow file=/etc/shadow parent=sudo gparent=zsh ggparent=tmux: gggparent=systemd container_id=host image=<NA>)
Thu Oct 15 18:38:08 2020: SIGHUP received, restarting...
Events detected: 1
Rule counts by severity:
   WARNING: 1
Triggered rules by rule name:
   Read sensitive file untrusted: 1
   - num times actions taken: 0
Thu Oct 15 18:38:08 2020: Falco version 0.26.1-43+47fa7d5 (driver version 2aa88dcf6243982697811df4c1b484bcbe9488a2)
Thu Oct 15 18:38:08 2020: Falco initialized with configuration file /etc/falco/falco.yaml
Thu Oct 15 18:38:08 2020: Loading rules from file /etc/falco/falco_rules.yaml:
Thu Oct 15 18:38:08 2020: Loading rules from file /etc/falco/falco_rules.local.yaml:
Thu Oct 15 18:38:08 2020: Loading rules from file /etc/falco/k8s_audit_rules.yaml:
Thu Oct 15 18:38:09 2020: Starting internal webserver, listening on port 8765
[1]    180475 segmentation fault  sudo userspace/falco/falco -c /etc/falco/falco.yaml
```

**Environment**

- Falco version: ` 0.26.1-43+47fa7d5 (driver version 2aa88dcf6243982697811df4c1b484bcbe9488a2)`

- System info:

- OS: Arch
- Kernel: `5.8.14-arch1-1`

- Installation method: from source

**Additional context**

This problem might be related to https://github.com/falcosecurity/falco/pull/1412

leogr:
**What type of PR is this?**

/kind bug

/area engine

Previously, formatters were freed by LUA code when re-opening outputs.
Since now, outputs are not controlling anymore the falco_formats class (see #1412), we just free formatters only if were already initialized.

That is needed when the engine restarts (see #1446).

By doing so, we also ensure that the correct inspector instance is set to the formatter cache.

Fixes #1446

```release-note
NONE
```

### DIFF
--- a/userspace/engine/formats.cpp
+++ b/userspace/engine/formats.cpp
@@ -1,5 +1,5 @@
 /*
-Copyright (C) 2019 The Falco Authors.
+Copyright (C) 2020 The Falco Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -44,8 +44,9 @@ void falco_formats::init(sinsp *inspector,
 	s_json_include_output_property = json_include_output_property;
 	if(!s_formatters)
 	{
-		s_formatters = new sinsp_evt_formatter_cache(s_inspector);
+		delete(s_formatters);
 	}
+	s_formatters = new sinsp_evt_formatter_cache(s_inspector);
 
 	luaL_openlib(ls, "formats", ll_falco, 0);
 }
--- a/userspace/engine/formats.cpp
+++ b/userspace/engine/formats.cpp
@@ -24,7 +24,7 @@ sinsp *falco_formats::s_inspector = NULL;
 falco_engine *falco_formats::s_engine = NULL;
 bool falco_formats::s_json_output = false;
 bool falco_formats::s_json_include_output_property = true;
-sinsp_evt_formatter_cache *falco_formats::s_formatters = NULL;
+std::unique_ptr<sinsp_evt_formatter_cache> falco_formats::s_formatters = NULL;
 
 const static struct luaL_reg ll_falco[] =
 	{
@@ -42,11 +42,9 @@ void falco_formats::init(sinsp *inspector,
 	s_engine = engine;
 	s_json_output = json_output;
 	s_json_include_output_property = json_include_output_property;
-	if(!s_formatters)
-	{
-		delete(s_formatters);
-	}
-	s_formatters = new sinsp_evt_formatter_cache(s_inspector);
+
+	// todo(leogr): we should have used std::make_unique, but we cannot since it's not C++14
+	s_formatters = std::unique_ptr<sinsp_evt_formatter_cache>(new sinsp_evt_formatter_cache(s_inspector));
 
 	luaL_openlib(ls, "formats", ll_falco, 0);
 }
--- a/userspace/engine/formats.h
+++ b/userspace/engine/formats.h
@@ -53,7 +53,7 @@ public:
 
 	static sinsp *s_inspector;
 	static falco_engine *s_engine;
-	static sinsp_evt_formatter_cache *s_formatters;
+	static std::unique_ptr<sinsp_evt_formatter_cache> s_formatters;
 	static bool s_json_output;
 	static bool s_json_include_output_property;
 };


userspace/engine/formats.cpp
```diff
@@ -44,8 +44,10 @@ void falco_formats::init(sinsp *inspector,
 	s_json_include_output_property = json_include_output_property;
 	if(!s_formatters)
 	{
-		s_formatters = new sinsp_evt_formatter_cache(s_inspector);
+		delete(s_formatters);
+		s_formatters = NULL;
```
leodido, line 48: Why assigning it to `NULL` after the `delete`?

There's the possibility of a double delete?

Also, maybe it's the case to think about a "smart" pointer here. Even to reason if we now need a pointer or not here...

leodido (commented):
Please take a look at my comment :)

userspace/engine/formats.cpp
leogr: It's a WIP, and I was just trying the previous approach:
https://github.com/falcosecurity/falco/blob/0a449afc3ed0efc7d6ddf66fe206d4081ee405ae/userspace/engine/formats.cpp#L115-L123

leogr: And ofc, now there's no need to assign it to `NULL` since two lines below it gets assigned to `new sinsp_evt_formatter_cache(s_inspector)` always (so no way to double delete it):

https://github.com/falcosecurity/falco/blob/c0f3b5acf4b53970549569e8feadbe5710c4d88c/userspace/engine/formats.cpp#L50

Alternatively, an `std::unique_ptr` should work here too.

However, you're probably right: we don't really need a pointer here. I will try that.

Thanks for the hints!

leogr: Fixed.

N.B.
Although the pointer is not really needed, removing it implies having an empty `sinsp_evt_formatter_cache` for a while (when declaring the static member) and later freed at the first `init` call. That can lead to problems if `sinsp_evt_formatter_cache` implementation changed in the future. Thus I opted to keep the pointer (with `std::unique_ptr`).

leogr:
Ready for review.

Thanks, @leodido for the suggestions.

fntlnz (approved):

fntlnz:
We can probably use `make_shared` from cpp+20 here/ I know that we are on C++11 so we might just use the polyfills from abseil cpp that we already have for gRPC .

abseil: https://github.com/abseil/abseil-cpp
grpc importing abseil: https://github.com/falcosecurity/falco/blob/master/cmake/modules/gRPC.cmake#L119

fntlnz:
Putting an hold since this is blocking the merge queue. Feel free to remove once it's approved by two maintainers.

/hold

leodido (approved):

leogr:
/hold cancel

:partying_face:

PR #989, opened 2020-01-06
debian package: use systemd if available

juju4:
**What type of PR is this?**

/kind feature

**Any specific area of the project related to this PR?**

NONE

**What this PR does / why we need it**:

#484

```release-note
* switch to systemd if available in debian package
* hardened systemd settings
```

PR #1448, opened 2020-10-16
feat: migrate Debian/RPM package from init to systemd

### LINKED ISSUE #1265: Migrate Debian packages to Systemd
**Motivation**

Currently the Falco debian package use an outdated feature to manage the service on Debian systems called `init`. This currently hasn't caused any concern because the new feature `systemd` is smart enough to manage the older `init` style configurations. 

Most `systemd` users aren't even aware that `systemd` is actually referencing the old falco `init` scripts.

In issue https://github.com/falcosecurity/falco/issues/640 we have a very strong use case to migrate from `init` to `systemd` so that the unit file can be configured in such a way that it will automatically be restarted.

**Feature**

Can we please change the debian packaging code in [/scripts](https://github.com/falcosecurity/falco/tree/master/scripts) to support `systemd` instead of `init`?

Can we also craft a unit file for `systemd`? Here is what I am running in production currently.

```
[Unit]
Description=Falco Runtime Security

[Service]
ExecStart=/usr/bin/falco --pidfile=/var/run/falco.pid -c /etc/falco/falco.yaml
ExecReload=/bin/kill -HUP $MAINPID
KillMode=process
Restart=always

[Install]
WantedBy=multi-user.target
```

**Alternatives**

For #640 we could also bake in a restart feature into `init` but at this point I don't think we should continue to build on top of `init` any more.

**Additional context**

Note: Now that the #1217 Unix Socket work is making it to production, we can configure Falco in Kubernetes to communicate directly with systemd to get meta information about the service and auto-configure Falco read-only pods to listen on the domain socket on the host :nerd_face:

jenting:
**What type of PR is this?**

/kind feature

/area build

Current Linux Os migrates from `init` to `systemd`, so we update the Falco Debian/RPM packages to `systemd` also.

Fixes #1265

- tested on Ubuntu Focal
  ```
  mkdir -p build
  cd build
  cmake ..
  make package
  ls *.deb | xargs dpkg -i
  /usr/bin/falco-driver-loader
  systemctl status falco.service
  systemctl start falco.service
  systemctl status falco.service
  ```

- tested on CentOS 8 & openSUSE Leap 15.2
  ```
  mkdir -p build
  cd build
  cmake ..
  make package
  ls *.rpm | xargs rpm -i
  /usr/bin/falco-driver-loader
  systemctl status falco.service
  systemctl start falco.service
  systemctl status falco.service
  ```

```release-note
update: Debian/RPM package migrated from init to systemd
BREAKING CHANGE: the init.d service unit is not shipped anymore in deb/rpm packages in favor of a systemd service file
```


leogr (commented):
Thank you for your contribution!

Could we do the same thing for the DEB package too?

jenting:
> Thank you for your contribution!
> 
> Could we do the same thing for the DEB package too?

np! Let me make some code change and test on a Debian-based operating system.

jenting:
Ready to be reviewed :rocket:

scripts/debian/falco.service
```diff
@@ -0,0 +1,29 @@
...
+ProtectKernelTunables=true
+RestrictRealtime=true
+#RestrictAddressFamilies=AF_INET AF_INET6 AF_UNIX
+RestrictAddressFamilies=~AF_PACKET
+SystemCallFilter=~@debug @mount @cpu-emulation @obsolete @privileged
+# FIXME!
+# PANIC: unprotected error in call to Lua API (runtime code generation failed, restricted kernel?)
+# https://www.freelists.org/post/luajit/luajit-crashes-with-grsec-kernel,1
+# MemoryDenyWriteExecute=true
+# PrivateMounts=true
```
leogr, line 26: Could you elaborate more on that?

scripts/rpm/falco.service
```diff
@@ -0,0 +1,29 @@
...
+ProtectKernelTunables=true
+RestrictRealtime=true
+#RestrictAddressFamilies=AF_INET AF_INET6 AF_UNIX
+RestrictAddressFamilies=~AF_PACKET
+SystemCallFilter=~@debug @mount @cpu-emulation @obsolete @privileged
+# FIXME!
+# PANIC: unprotected error in call to Lua API (runtime code generation failed, restricted kernel?)
+# https://www.freelists.org/post/luajit/luajit-crashes-with-grsec-kernel,1
+# MemoryDenyWriteExecute=true
+# PrivateMounts=true
```
leogr, line 26: Same as above

scripts/debian/falco.service
jenting: I'm referencing this PR https://github.com/falcosecurity/falco/pull/989

I'm not sure which part should be preserved or should I directly refer to the issue description https://github.com/falcosecurity/falco/issues/1265#issue-640658962?

leogr: Not a systemd expert. But for sure I'd avoid leaving those comments if we are not sure about that. 

/cc @fntlnz 
/cc @leodido 
/cc @kris-nova 

WDYT?

nibalizer:
I tried the systemd unit file from this on ubuntu and it worked well. 

I've been using the juju4 ansible module which [provides its own systemd unit file](https://github.com/juju4/ansible-falco/blob/master/templates/systemd-falco.service.j2) for several months. The diff is below:


```$ diff pull-request/falco.service ansible-managed/falco.service
0a1,3
> #
> # Ansible managed
> #
3c6
<
---
>
12a16
> PrivateTmp=true
23,27c27,31
< # PANIC: unprotected error in call to Lua API (runtime code generation failed, restricted kernel?)
< # https://www.freelists.org/post/luajit/luajit-crashes-with-grsec-kernel,1
< # MemoryDenyWriteExecute=true
< # PrivateMounts=true
<
---
> #   PANIC: unprotected error in call to Lua API (runtime code generation failed, restricted kernel?)
> #   ?https://www.freelists.org/post/luajit/luajit-crashes-with-grsec-kernel,1
> #MemoryDenyWriteExecute=true
> #PrivateMounts=true
>
```

jenting:
@nibalizer really thanks to the testing and sharing on the falco works well with juju4 provided systemd service file works well for serverl months.

I've updated the PR to remove the comments in the systemd file, PTAL @leogr if you have time.

scripts/debian/falco.service
```diff
@@ -0,0 +1,22 @@
+[Unit]
+Description=Falco: Container Native Runtime Security
+
+[Service]
+Type=simple
+User=root
+ExecStart=/usr/bin/falco --pidfile=/var/run/falco.pid
```
fntlnz, line 7: This is missing the logic to do the modprobe/dkms - if we accept this PR in the next update Falco will not behave as it does now.

fntlnz (changes requested):
Hi! Thanks for your PR and the interest in bringing this up. I have a major comment

scripts/debian/falco.service
jenting: I'll look at this later to recap my memory.

fntlnz: Thank you!

jenting: Correct me if I am wrong.

If I remember correctly, without the PR change, falco performs the modprobe/dkms at the package install and uninstall stage. Therefore, this PR does not break the behavior change.

leogr: Almost correct, but there's more. The current scripts are also calling [modprobe](https://github.com/falcosecurity/falco/blob/master/scripts/debian/falco#L67)/[rmmod](https://github.com/falcosecurity/falco/blob/master/scripts/debian/falco#L99) on the kernel module when starting/stopping the service.

I guess @fntlnz was referring to that. And I agree we should keep the same behavior.

PS
IIRC, the Falco process now tries to `insmod` the kernel module if it does not find it running yet. However, that behavior was implemented as a fallback and we should not rely on that. Moreover, it won't `rmmod` the kernel module. 
Thus, the systemd's script should still responsible for handling the whole kernel module lifecycle.

jenting: Address `/sbin/modprobe falco` and `/sbin/rmmod falco` in systemd service file.

jenting:
This is ready to review.

leogr (approved):
DEB tested with Ubuntu 20.10 :+1: 
RPM tested with CentOS 8 :+1: 

LGTM!!! :clap: 

Thank you @jenting 

/cc @fntlnz 
/cc @leodido

fntlnz (approved):

fntlnz:
I believe it's now time to drop support for init.d based systems now.

Thanks for your patience on this @jenting and for pushing this forward for almost 1.5 years!

I'm signaling this as a breaking change in the release notes.

leogr:
:clap:

PR #1474, opened 2020-11-05
fix(scripts/falco-driver-loader): lsmod usage

### LINKED ISSUE #1468: falco-driver-loader rmmod check/loop inadvertently finds other modules
**Describe the bug**

The falco-driver-loader script just does a basic `lsmod | grep -q "falco"` so will partially match and can inadvertently think that the falco module is loaded when actually an unrelated "falcon_nf_netcontain" kernel module (e.g.,) happened to be loaded on the host

https://github.com/falcosecurity/falco/blob/294804daf436b4d146731322f46a99e2736feee0/scripts/falco-driver-loader#L223-L238

**How to reproduce it**

Attempt to start falco on a host that has a similarly named "falcon" module and notice that the falco-driver-loader will loop attempting to `rmmod falco` when falco is not loaded.

**Expected behaviour**

falco-driver-loader should inspect only the first column of /proc/modules and require the whole search string to match

i.e.,
```sh
while cut -d' ' -f1 </proc/modules | grep -qx "${KMOD_NAME}" && [ $WAIT_TIME -lt "${MAX_RMMOD_WAIT}" ]; do 
```

**Screenshots**

**Environment**

- Falco version:
- System info: Falco version: 0.23.0
Driver version: 96bd9bc560f67742738eb7255aeb4d03046b8045
- OS: Ubuntu bionic
- Kernel: 4.15.0-122-generic
- Installation method: deb

dnwe:
**What type of PR is this?**

/kind bug

/area engine

Attempting to start falco on a host that had a similarly named module
(e.g., "falcon") would cause the falco-driver-loader to loop attempting
to rmmod falco when falco was not loaded.

falco-driver-loader will now inspect only the first column of lsmod
output and require the whole search string to match

Fixes #1468

```release-note
NONE
```

### DIFF
--- a/scripts/falco-driver-loader
+++ b/scripts/falco-driver-loader
@@ -220,7 +220,7 @@ load_kernel_module() {
 	rmmod "${DRIVER_NAME}" 2>/dev/null
 	WAIT_TIME=0
 	KMOD_NAME=$(echo "${DRIVER_NAME}" | tr "-" "_")
-	while lsmod | grep "${KMOD_NAME}" > /dev/null 2>&1 && [ $WAIT_TIME -lt "${MAX_RMMOD_WAIT}" ]; do
+	while lsmod | cut -d' ' -f1 | grep -qx "${KMOD_NAME}" && [ $WAIT_TIME -lt "${MAX_RMMOD_WAIT}" ]; do 
 		if rmmod "${DRIVER_NAME}" 2>/dev/null; then
 			echo "* Unloading ${DRIVER_NAME} module succeeded after ${WAIT_TIME}s"
 			break
@@ -232,7 +232,7 @@ load_kernel_module() {
 		sleep 1
 	done
 
-	if lsmod | grep "${KMOD_NAME}" > /dev/null 2>&1; then
+	if lsmod | cut -d' ' -f1 | grep -qx "${KMOD_NAME}" > /dev/null 2>&1; then
 		echo "* ${DRIVER_NAME} module seems to still be loaded, hoping the best"
 		exit 0
 	fi


scripts/falco-driver-loader
```diff
@@ -220,7 +220,7 @@ load_kernel_module() {
 	rmmod "${DRIVER_NAME}" 2>/dev/null
 	WAIT_TIME=0
 	KMOD_NAME=$(echo "${DRIVER_NAME}" | tr "-" "_")
-	while lsmod | grep "${KMOD_NAME}" > /dev/null 2>&1 && [ $WAIT_TIME -lt "${MAX_RMMOD_WAIT}" ]; do
+	while lsmod | cut -d' ' -f1 | grep -qx "${KMOD_NAME}" && [ $WAIT_TIME -lt "${MAX_RMMOD_WAIT}" ]; do
```
leogr, line 223: Is there any particular reason why  `> /dev/null 2>&1` has been removed?

leogr (commented):
Thank you for your contribution!

We definitively need this fix.
/milestone 0.27.0

Just a question, see below.

leodido (changes requested):
I have some concerns on the portability of `-q` flag.

> ... unlike GNU grep, 7th Edition Unix grep did not conform to POSIX, because it lacked -q and its -s option behaved like GNU grep's -q option. USG-style grep also lacked -q but its -s option behaved like GNU grep. Portable shell scripts should avoid both -q and -s and should redirect standard and error output to /dev/null instead.

From the [manual pages](https://linux.die.net/man/1/grep).

Also, we could think of using `/proc/modules` rather than `lsmod`. WDYT?

scripts/falco-driver-loader
leodido: Probably for this reason:

> Normally, the exit status is 0 if selected lines are found and 1 otherwise. But the exit status is 2 if an error occurred, unless the -q or --quiet or --silent option is used and a selected line is found. 


Ref `man grep`.

dnwe:
@leodido I did originally suggest reading from /proc/modules in my issue https://github.com/falcosecurity/falco/issues/1468 but as it has the advantage of non-root and file reading rather than a spawn, but there seemed to be some concerns and it isn’t necessary to fix the bug so I reverted to lsmod

dnwe:
RE: portability of `grep -q` — whilst this was probably an issue historically, I checked AIX, Busybox, FreeBSD, HP-UX, OpenBSD and Solaris variants of grep, and they all support the expected `-q` behaviour

scripts/falco-driver-loader
leogr: Gotcha tnx

leogr:
> @leodido I did originally suggest reading from /proc/modules in my issue #1468 as it has the advantage of non-root and file reading rather than a process spawn, but there seemed to be some concerns and it isn’t necessary to fix the bug so I reverted to lsmod

Regarding my [comment](https://github.com/falcosecurity/falco/issues/1468#issuecomment-722420607) about `/proc/modules`: although my concern about its availability in a container context, I don't have a strong opinion on that, and I'm not against its usage. Feel free to go with `/proc/modules` if it helps, and we are sure there are no portability issues. 

Thank you again.

leodido:
> RE: portability of `grep -q` — whilst this was probably an issue historically, I checked AIX, Busybox, FreeBSD, HP-UX, OpenBSD and Solaris variants of grep, and they all support the expected `-q` behaviour

I was referring to what man pages say. Not checked personally, by hand. 

So, if the `-q` flag has been included in (non GNU) `grep`s and it's not a portability issue anymore as you saying, the last remaining topic to get this merged is understand if the `/proc/modules` approach as limitations or not.

BTW great job, this is really needed 👏

dnwe:
RE: container runtimes, I checked the only two I have to hand (docker and containerd). In both cases if I start a barebones container with `--cap-drop all` and `--user nobody` I confirmed that I was able to read from /proc/modules file without issue so I think it would probably be fine.

However, there's no great motivation to change it when the `falco-driver-loader` script is already running as root anyway

leodido (approved):

leogr (approved):

PR #1481, opened 2020-11-10
docs: changelog for 0.26.2

leodido:
Co-authored-by: Lorenzo Fontana <lo@linux.com>

**What type of PR is this?**

/kind documentation

NONE

Sync the CHANGELOG with the hotfix 0.26.2 release one.

NONE

This is needed because the 0.26.2 hotfix release is not on the `master` branch, rather it is on the `release/0.26.2` branch because - as discussed in our community calls - we wanted to include only the new prebuilt drivers URL (`DRIVERS_REPO` env variable) in it.

So the CHANGELOG needs to be updated this way.

The alternative would be to wait for 0.27.0 and also add 0.26.2 changes to the CHANGELOG when we release the 0.27.0.

We (me, @fntlnz, and @leogr) chose this path for simplicity.

```release-note
NONE
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Change Log
 
+## v0.26.2
+
+Released on 2020-11-10
+
+### Major Changes
+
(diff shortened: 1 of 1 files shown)


leogr (approved):
That's perfectly fine, IMHO.

fntlnz (approved):

PR #1488, opened 2020-11-16
falco-driver-loader improvements (custom driver name + clean previous drivers functionality)

### LINKED ISSUE #1486: Since the Falco dirver rename, Falco is unable to unload the falco_probe module
**Describe the bug**

In Falco 0.23.0, the Falco driver was renamed from `falco_probe` to `falco`.

https://github.com/falcosecurity/falco/commit/3ec4b5b65272e8c86504edd29cf36023880aec43

That change changed the value of `DRIVER_NAME` which affected the falco-driver-loader script ability to run `rmmod falco_probe`.

Basically:

- The falco-driver-loader script does `rmmod falco 2>/dev/null`
- The `lsmod | cut -d' ' -f1 | grep -qx falco` test returns true because `falco_probe` is still loaded
- The script tries over and over for a minute
- The script ends with the `"* falco module seems to still be loaded, hoping the best"` message

Screen Shot 2020-11-13 at 1 02 43 PM

**How to reproduce it**

- Deploy Falco 0.22.0 on a host
- Validate the `falco_probe` module is loaded with `lsmod | grep falco`
- Upgrade to Falco 0.26.2
- Watch it try to `rmmod falco`

**Expected behaviour**

The https://github.com/falcosecurity/falco/blob/master/scripts/falco-driver-loader script should look for the old version of the Falco module and run `rmmod`.

For example:
```bash
if lsmod | cut -d' ' -f1 | grep -qx "falco_probe"; then
  echo "* Module falco_probe still present"

  if rmmod -v "falco_probe" 2>/dev/null; then
    echo "* Unloading falco_probe module succeeded"
  fi
fi
```

leodido:
**What type of PR is this?**

/kind documentation

/kind feature

NONE

It lets users specify a custom `DRIVER_NAME` getting it from the environment. It can be useful when the user has a Falco kernel-module with a different name (eg., `falco_probe`) from the current default one (ie., `falco`).

This way (`DRIVER_NAME=falco_probe falco-driver-loader module`) it will look for a kernel module (via `lsmod`) named `falco_probe` and remove it (`rmmod`).

Furthermore, the module should also be removed from `dkms`. **WIP** for this reason.

Last but not least, for debugging and clarity purposes, this PR make the `falco-driver-loader` script aware of the Falco version it has been built for.

And outputs it at:
- runtime time
- help/usage time

Fixes #1486 

/cc @leogr 
/cc @fntlnz 

```release-note
new: falco-driver-loader script can get a custom driver name from DRIVER_NAME env variable
new: falco-driver-loader know the Falco version
```


leogr:
Thank you for this PR. We definitely need that :+1: 

Although specifying a custom `DRIVER_NAME` can be useful for various reasons, I believe it's not the best way to clean up old drivers since:
- the script will then try to insmod/modprobe with the given `DRIVER_NAME`
- the script does not have action only to remove the driver

Thus, as discussed privately, I believe it worth adds just a specific cleanup command for the old `falco_probe` that we don't use anymore. I will work on that.

leodido:
Yep, that's the reason this PR is **WIP** @leogr :)

As discussed privately, we probably need to add 3rd action to the script (eg., `falco-driver-loader clean`)

leogr:
> Yep, that's the reason this PR is **WIP** @leogr :)
> 
> As discussed privately, we probably need to add 3rd action to the script (eg., `falco-driver-loader clean`)

That's ok for me too. But, do we still want a simple clean-up for `falco_probe` that runs always?

leodido:
That's up to us to decide.

leogr:
Ok, let's do this step-by-step.

So I can work on the `clean` action since I believe we all agree on that.

leodido:
Yes, we all agree on that. As also discussed privately. :)

Then, we could also think about how to make the upgrade process straightforward even if the driver name changes again.

leogr:
> Yes, we all agree on that. As also discussed privately. :)

Done by https://github.com/falcosecurity/falco/pull/1488/commits/aa892e498c358e4c02d38b4b06c222a37b872431 :)
It's the first draft anyway, so PTAL.

> Then, we could also think about how to make the upgrade process straightforward even if the driver name changes again.
> And, given that a new kernel module name change is very unlikely: if it's worth.

:+1:

leogr (approved):

fntlnz (approved):

Issue #313, opened 2018-01-12
Running Falco on Azure Linux Ubuntu VM: "Unable to load the driver. Exiting." error


alukyan:
Trying to use Falco on Azure Linux Ubuntu VM

```
$ uname -r
4.13.0-1005-azure
```

Installation works OK
```
~$ curl -s https://s3.amazonaws.com/download.draios.com/stable/install-falco | sudo bash
* Detecting operating system
* Installing Sysdig public key
OK
* Installing falco repository
* Installing kernel headers
* Installing falco
Selecting previously unselected package falco.
(Reading database ... 101509 files and directories currently installed.)
Preparing to unpack .../archives/falco_0.8.1_amd64.deb ...
Unpacking falco (0.8.1) ...
Processing triggers for systemd (229-4ubuntu21) ...
Processing triggers for ureadahead (0.100.0-19) ...
Setting up falco (0.8.1) ...
Loading new falco-0.8.1 DKMS files...
First Installation: checking all kernels...
Building only for 4.13.0-1005-azure
Building initial module for 4.13.0-1005-azure
Done.

falco-probe:
Running module version sanity check.
 - Original module
   - No original module exists within this kernel
 - Installation
   - Installing to /lib/modules/4.13.0-1005-azure/updates/dkms/

depmod....

DKMS: install completed.
```

Starting Falco service fails with error:
```
~$ service falco status
● falco.service - LSB: Falco syscall activity monitoring agent
   Loaded: loaded (/etc/init.d/falco; bad; vendor preset: enabled)
   Active: active (exited) since Fri 2018-01-12 22:24:13 UTC; 6min ago
     Docs: man:systemd-sysv-generator(8)
  Process: 127374 ExecStart=/etc/init.d/falco start (code=exited, status=0/SUCCESS)

Jan 12 22:24:12 gwtest6a3000001 falco[127374]: Fri Jan 12 22:24:12 2018: Falco initialized with configuration file /etc/falco/falco.yaml
Jan 12 22:24:13 gwtest6a3000001 falco[127390]: Parsed rules from file /etc/falco/falco_rules.yaml
Jan 12 22:24:13 gwtest6a3000001 falco[127390]: Parsed rules from file /etc/falco/falco_rules.local.yaml
Jan 12 22:24:13 gwtest6a3000001 falco[127374]: Fri Jan 12 22:24:13 2018: Parsed rules from file /etc/falco/falco_rules.yaml
[...]

PR #1492, opened 2020-11-19
fix(docker): missing HOST_ROOT env in "no-driver" docker image

### LINKED ISSUE #1491: Missing container metadata when using falcosecurity/falco-no-driver image
**Describe the bug**

When using the `falcosecurity/falco-no-driver` image, only `container.id` is reported within the output message, and other container-related fields are missing.

**How to reproduce it**

Run the `falcosecurity/falco-no-driver:0.26.2` image so:
```
docker run --rm -i -t --privileged -v /var/run/docker.sock:/host/var/run/docker.sock -v /dev:/host/dev -v /proc:/host/proc:ro -v /boot:/host/boot:ro -v /lib/modules:/host/lib/modules:ro -v /usr:/host/usr:ro -v /etc:/host/etc falcosecurity/falco-no-driver:0.26.2
```

Then in another shell, use the following command to generate an alert:
```
docker run --rm -it alpine
```

Then the Falco output will miss the `image` field:

```
2020-11-19T11:43:09.332827604+0000: Notice A shell was spawned in a container with an attached terminal (user=<NA> user_loginuid=-1 <NA> (id=5d2a4b8d7f06) shell=sh parent=<NA> cmdline=sh terminal=34816 container_id=5d2a4b8d7f06 image=<NA>)
```

**Expected behaviour**

```
2020-11-19T11:46:38.250373343+0000: Notice A shell was spawned in a container with an attached terminal (user=<NA> user_loginuid=-1 happy_thompson (id=25ea176ce5df) shell=sh parent=<NA> cmdline=sh terminal=34816 container_id=25ea176ce5df image=alpine)
```

**Environment**

- Falco version: 0.26.2 (the issue seems to be present also since 0.25.0)
- System info:

- OS: ArchLinux

- Kernel:  .9.8-arch1-1

- Installation method: docker

**Additional context**

The problem is caused by the missing `HOST_ROOT` env var. A quick workaround is to set it via the docker option (i.e. `-e HOST_ROOT=/host`).

The initial of this issue started in Slack. Here the link for reference: https://kubernetes.slack.com/archives/CMWH3EH32/p1604437988412500

Special thanks to @JPLachance that helped me to debug this and confirmed that setting `HOST_ROOT` solved the problem in his use case.

I'm making a PR to definitively fix that soon.

/assign

leogr:
**What type of PR is this?**

/kind bug

/kind cleanup

/area build

This PR solves the problem described in #1491 and makes env vars consistent between the "falco" and the "no-driver" docker images.
It also updates the container labels with the correct docker usage command line.

Fixes #1491

/milestone 0.27.0

```release-note
fix: set `HOST_ROOT=/host` environment variable for the `falcosecurity/falco-no-driver` container image by default
```

### DIFF
--- a/docker/driver-loader/Dockerfile
+++ b/docker/driver-loader/Dockerfile
@@ -3,7 +3,7 @@ FROM falcosecurity/falco:${FALCO_IMAGE_TAG}
 
 LABEL maintainer="cncf-falco-dev@lists.cncf.io"
 
-LABEL usage="docker run -i -t -v /dev:/host/dev -v /proc:/host/proc:ro -v /boot:/host/boot:ro -v /lib/modules:/host/lib/modules:ro -v /usr:/host/usr:ro --name NAME IMAGE"
+LABEL usage="docker run -i -t --privileged -v /root/.falco:/root/.falco -v /proc:/host/proc:ro -v /boot:/host/boot:ro -v /lib/modules:/host/lib/modules:ro -v /usr:/host/usr:ro -v /etc:/host/etc:ro --name NAME IMAGE"
 
 ENV HOST_ROOT /host
 ENV HOME /root
--- a/docker/falco/Dockerfile
+++ b/docker/falco/Dockerfile
@@ -2,7 +2,7 @@ FROM debian:stable
 
 LABEL maintainer="cncf-falco-dev@lists.cncf.io"
 
-LABEL usage="docker run -i -t -v /var/run/docker.sock:/host/var/run/docker.sock -v /dev:/host/dev -v /proc:/host/proc:ro -v /boot:/host/boot:ro -v /lib/modules:/host/lib/modules:ro -v /usr:/host/usr:ro --name NAME IMAGE"
+LABEL usage="docker run -i -t --privileged -v /var/run/docker.sock:/host/var/run/docker.sock -v /dev:/host/dev -v /proc:/host/proc:ro -v /boot:/host/boot:ro -v /lib/modules:/host/lib/modules:ro -v /usr:/host/usr:ro -v /etc:/host/etc --name NAME IMAGE"
 
 ARG FALCO_VERSION=latest
 ARG VERSION_BUCKET=deb
--- a/docker/no-driver/Dockerfile
+++ b/docker/no-driver/Dockerfile
@@ -22,6 +22,9 @@ RUN sed -e 's/time_format_iso_8601: false/time_format_iso_8601: true/' < /falco/
 
 FROM scratch
 
+ENV HOST_ROOT /host
+ENV HOME /root
+
 COPY --from=ubuntu /falco /
 
 CMD ["/usr/bin/falco", "-o", "time_format_iso_8601=true"]
--- a/docker/no-driver/Dockerfile
+++ b/docker/no-driver/Dockerfile
@@ -1,7 +1,5 @@
 FROM ubuntu:18.04 as ubuntu
 
-LABEL maintainer="cncf-falco-dev@lists.cncf.io"
-
 ARG FALCO_VERSION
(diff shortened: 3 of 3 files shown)


fntlnz:
We didn't do the /host env variable thing in that image because with --device you have no way to remap falco devices to /host/dev and thus we were advising to do /var/run/docker.sock on the no-driver image.

If we just merge this PR but don't adjust the wording around that too this will just create more confusion.

docker/driver-loader/Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM falcosecurity/falco:${FALCO_IMAGE_TAG}
 
 LABEL maintainer="cncf-falco-dev@lists.cncf.io"
 
-LABEL usage="docker run -i -t -v /dev:/host/dev -v /proc:/host/proc:ro -v /boot:/host/boot:ro -v /lib/modules:/host/lib/modules:ro -v /usr:/host/usr:ro --name NAME IMAGE"
+LABEL usage="docker run -i -t --privileged -v /root/.falco:/root/.falco -v /proc:/host/proc:ro -v /boot:/host/boot:ro -v /lib/modules:/host/lib/modules:ro -v /usr:/host/usr:ro -v /etc:/host/etc:ro --name NAME IMAGE"
```
fntlnz, line 6: Look at what we do here https://falco.org/docs/running/

fntlnz (commented):
Added a comment

docker/driver-loader/Dockerfile
leogr: Please note that this :point_up: is the `falco-driver-loader` image and I have used the documentation as source, indeed it matches:
```
docker pull falcosecurity/falco-driver-loader:latest
docker run --rm -i -t \
    --privileged \
    -v /root/.falco:/root/.falco \
    -v /proc:/host/proc:ro \
    -v /boot:/host/boot:ro \
    -v /lib/modules:/host/lib/modules:ro \
    -v /usr:/host/usr:ro \
    -v /etc:/host/etc:ro \
    falcosecurity/falco-driver-loader:latest
```
Am I missing something else?

leogr:
> We didn't do the /host env variable thing in that image because with --device you have no way to remap falco devices to /host/dev and thus we were advising to do /var/run/docker.sock on the no-driver image.
> 
> If we just merge this PR but don't adjust the wording around that too this will just create more confusion.
> 
> See: https://falco.org/docs/running/

Good point, I have not considered that.

Anyway I believe is still convenient `HOST_ROOT=/host` by default, and then update the documentation for the "least privileged" in this way:
```
docker pull falcosecurity/falco-no-driver:latest
docker run --rm -i -t \
    -e HOST_ROOT=/ \
    --cap-add SYS_PTRACE --pid=host $(ls /dev/falco* | xargs -I {} echo --device {}) \
    -v /var/run/docker.sock:/var/run/docker.sock \
    falcosecurity/falco-no-driver:latest
```
WDYT?

I had already prepared a PR to update the doc :point_right: https://github.com/falcosecurity/falco-website/pull/313, so if you agree, I can update it again accordingly.

fntlnz:
I agree, if we want to stick with the least privileged wording.

leogr:
> I agree, if we want to stick with the least privileged wording.

:+1: 

Yes, I prefer this option, since having different defaults across images can be more confusing. I will update the doc PR ASAP.

leogr:
Hey @fntlnz 

done, PTAL :point_right: https://github.com/falcosecurity/falco-website/pull/313/files

And thank you for the suggestion! :hugs:

leodido (approved):
I've looked at the #313 PR on the falco-website, and the edits you did to this one.

/lgtm

fntlnz (approved):

PR #1495, opened 2020-11-24
chore(cmake/modules): avoid useless rebuild

deepskyblue86:
Because of https://gitlab.kitware.com/cmake/cmake/-/issues/16419, every
time one compiles, some external projects gets updated causing rebuild.

Set UPDATE_DISCONNECTED on EXTERNALPROJECT_ADD, since we can assume git
tags we're using will not change.

Signed-off-by: deepskyblue86 <angelopuglisi86@gmail.com>

**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind rule-update

> /kind rule-create

cmake

/area build

> /area engine

> /area rules

> /area tests

> /area proposals

Avoid unnecessary rebuild

Fixes #

No

```release-note
NONE
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,15 @@ option(BUILD_WARNINGS_AS_ERRORS "Enable building with -Wextra -Werror flags" OFF
 option(MINIMAL_BUILD "Build a minimal version of Falco, containing only the engine and basic input/output (EXPERIMENTAL)" OFF)
 option(MUSL_OPTIMIZED_BUILD "Enable if you want a musl optimized build" OFF)
 
+# We shouldn't need to set this, see https://gitlab.kitware.com/cmake/cmake/-/issues/16419
+option(EP_UPDATE_DISCONNECTED "ExternalProject update disconnected" OFF)
+if (${EP_UPDATE_DISCONNECTED})
+  set_property(
+    DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+    PROPERTY EP_UPDATE_DISCONNECTED TRUE)
+endif()
+
+
 # Elapsed time
 # set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CMAKE_COMMAND} -E time") # TODO(fntlnz, leodido): add a flag to enable this
 


cmake/modules/OpenSSL.cmake
```diff
@@ -10,6 +10,7 @@
 # "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
 # specific language governing permissions and limitations under the License.
 #
+mark_as_advanced(OPENSSL_BINARY)
```
leodido, line 13: It seems to me this is also in #1496, Angelo

Do you want to mark variables as advanced in this PR or in that one?

Because, depending on what you want to do (two separate PRs, or all in one), we can arrange PRs accordingly :)

cmake/modules/gRPC.cmake
```diff
@@ -121,6 +128,7 @@ else()
     BUILD_BYPRODUCTS ${GRPC_LIB} ${GRPCPP_LIB}
     INSTALL_COMMAND ""
     CONFIGURE_COMMAND ""
+    UPDATE_DISCONNECTED TRUE
```
leodido, line 131: Can we make the value a cache variable rather than hard-coding it?

This way we would give control to developers over whether or not to perform updates.

Another approach to achieve the same goal would be to define `EP_UPDATE_DISCONNECTED` directory options

leodido (commented):
Thanks for submitting this!

Left two comments :)

cmake/modules/OpenSSL.cmake
deepskyblue86: It's here by mistake, I intended it to be in the other PR.

deepskyblue86:
@leodido since `ci/circleci: centos7` was failing and this option is mainly for developers, I switched the default to OFF.

deepskyblue86:
> @leodido since `ci/circleci: centos7` was failing and this option is mainly for developers, I switched the default to OFF.

Now it passed. CMake version maybe? I guess it's fine like that anyway.

fntlnz (commented):
LGTM

We probably want to mention this option in the source documentation page:

markdown: https://github.com/falcosecurity/falco-website/blob/master/content/en/docs/getting-started/source.mdfa

live: https://falco.org/docs/getting-started/source/

fntlnz (approved):

leogr (approved):

PR #1502, opened 2020-12-09
update: reference for the logo

fntlnz:

**What type of PR is this?**

/kind documentation

The CNCF has a single place where to find the various projects artwork, since that is the source of truth I thought that just linking it instead of having it here would be a good idea.

```release-note
NONE
```


leodido (approved):

fntlnz:
/hold
I need to rework this better since there are many unknown places linking to the PNGs here

fntlnz:
/close
for now

PR #1504, opened 2020-12-10
Avoid raw pointers usage (#1360)

### LINKED ISSUE #1360: Improve exception safety with smart pointers
Would you like to [wrap any pointer data members](https://github.com/falcosecurity/falco/blob/e46641d24d74fbe7921039834fdf5b01c4715009/userspace/engine/rules.cpp#L39 "Update candidate: falco_rules class") with the class template “[std::unique_ptr](https://en.wikipedia.org/wiki/Smart_pointer#unique_ptr "Description for the usage of smart pointers")”?

deepskyblue86:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind rule-update

> /kind rule-create

> /area build

/area engine

> /area rules

> /area tests

> /area proposals

Adoption of unique_ptr for better readability and safety

Fixes #1360

```release-note
NONE
```

### DIFF
--- a/userspace/engine/falco_utils.h
+++ b/userspace/engine/falco_utils.h
@@ -26,6 +26,17 @@ limitations under the License.
 
 #pragma once
 
+#if (__cplusplus < 201402L)
+namespace std
+{
+template<typename T, typename... Args>
+std::unique_ptr<T> make_unique(Args &&... args)
+{
+	return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
+}
+}
+#endif
+
 namespace falco
 {
 
--- a/userspace/engine/ruleset.h
+++ b/userspace/engine/ruleset.h
@@ -75,7 +75,10 @@ public:
 
 private:
 
-	struct filter_wrapper {
+	struct filter_wrapper
+	{
+		filter_wrapper(const std::set<uint32_t> &event_tags, std::unique_ptr<gen_event_filter> filter);
+
 		std::unique_ptr<gen_event_filter> filter;
 
 		// Indexes from event tag to enabled/disabled.
--- a/userspace/engine/ruleset.h
+++ b/userspace/engine/ruleset.h
@@ -88,6 +88,8 @@ private:
 	// A group of filters all having the same ruleset
 	class ruleset_filters {
 	public:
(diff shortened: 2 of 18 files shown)


userspace/engine/falco_utils.h
```diff
@@ -26,6 +26,17 @@ limitations under the License.
 
 #pragma once
 
+#if (__cplusplus < 201402L)
```
leodido, line 29: We should already have `absl::make_unique` available if I don't remember it wrong.

This way we could avoid the polyfill here.

Would you please look into it?

leodido (commented):
Thanks for this Angelo!

Left a comment :)

userspace/engine/falco_utils.h
deepskyblue86: I can't find it used anywhere. I do see it as a gRPC submodule, meaning its use would conflict MINIMAL_BUILD.
Do you prefer to use abseil and have it as a separate ExternalProject?

leodido: Right! I was not remembering that it comes as gRPC submodule...

In this case, I'd left your polyfill.

Anyways, if we're talking about what I prefer. Well, I'd prefer to port the whole codebase to C++14 (or greater). 🤣

userspace/engine/ruleset.h
```diff
@@ -106,18 +108,18 @@ class falco_ruleset
...
-		std::vector<std::list<filter_wrapper *> *> m_filter_by_event_tag;
+		std::vector<std::unique_ptr<filter_wrapper_list>> m_filter_by_event_tag;
 
 	};
 
-	std::vector<ruleset_filters *> m_rulesets;
+	std::vector<std::unique_ptr<ruleset_filters>> m_rulesets;
 
 	// Maps from tag to list of filters having that tag.
-	std::map<std::string, std::list<filter_wrapper *>> m_filter_by_event_tag;
```
deepskyblue86, line 113: @mstemm do you recall why we have a list here but a list pointer in `ruleset_filters::m_filter_by_event_tag`?
I don't want to change it in this PR but I'm curious...

Issue #1333, opened 2020-07-29
The outputs throttling mechanism skips notifications without any notice


leogr:
**Motivation**

The current throttling mechanism can be configured within [falco.yaml](https://github.com/falcosecurity/falco/blob/master/falco.yaml) as shown below:

 ```yaml
# A throttling mechanism implemented as a token bucket limits the
# rate of falco notifications. This throttling is controlled by the following configuration
# options:
#  - rate: the number of tokens (i.e. right to send a notification)
#    gained per second. Defaults to 1.
#  - max_burst: the maximum number of tokens outstanding. Defaults to 1000.
#
# With these defaults, falco could send up to 1000 notifications after
# an initial quiet period, and then up to 1 notification per second
# afterward. It would gain the full burst back after 1000 seconds of
# no activity.

outputs:
  rate: 1       
  max_burst: 1000
```

However, when this throttling mechanism takes actions then the user is not informed and AFAIK it's not mentioned anywhere that Falco is just skipping notifications when that happens. This behavior is not related to the very well known "drops" (see #669 for some examples of those).

Furthermore:
- there's no way to disable it
- setting `rate: 0` and `max_burst: 0` just disallow any notifications to be emitted

Finally, it's important to understand that this mechanism affects any output, including the gRPC APIs. Thus, the gRPC consumer must be aware that some events may not arrive. 

**Feature**

Implement a sort of notification or warning message to inform the user that notification could not show up when notification are being rate-limited (as like as we already did with "drops").

Also, add the ability to completely disable the throttling mechanism (ie. no rate-limit).

**Alternatives**

Just document the current behavior to make users aware about it.

**Additional context**

I have discovered this issue when working on https://github.com/falcosecurity/event-generator/pull/36 . Initially, I wrongly thought it was a gRPC problem. Later I got it was just the rate-limit that was affecting all outputs (including gRPC).

Below an example, that shows how notifications are limited up to 1 per second (after an initial burst of 1000 as per the default configuration) though the producer is still generating circa 400 EPS. Please note the timestamps.

[![asciicast](https://asciinema.org/a/350389.svg)](https://asciinema.org/a/350389)

jasondellaluce:
/reopen

/remove-lifecycle rotten

This is still a valid point

PR #1509, opened 2020-12-14
[WIP] Make output rate limiting optional

irozzo-1A:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind rule-update

> /kind rule-create

> /area build

/area engine

> /area rules

> /area tests

> /area proposals

This PR partially address #1333, in particular it allows to disable output throttling.

Fixes #

I profited to refactor a bit the token bucket to:

- Use monotonic clock instead of wall time.
- Having a raw interface where timestamps are provided explicitly and a facade that generates the timestamps automatically, mixing the two does not seem to be a good idea, especially if different types of clocks can be used.

```release-note
NONE
```

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -129,6 +129,7 @@ output_timeout: 2000
 outputs:
   rate: 1
   max_burst: 1000
+  rate_limit_enabled: true
 
 # Where security notifications should go.
 # Multiple outputs can be enabled.
--- a/tests/engine/test_token_bucket.cpp
+++ b/tests/engine/test_token_bucket.cpp
@@ -23,7 +23,7 @@ using namespace Catch::literals;
 
 TEST_CASE("token bucket default ctor", "[token_bucket]")
 {
-	auto tb = std::make_shared<token_bucket>();
+	auto tb = std::make_shared<raw_token_bucket>();
 
 	REQUIRE(tb->get_tokens() == 1);
 
@@ -37,10 +37,16 @@ TEST_CASE("token bucket default ctor", "[token_bucket]")
 	}
 }
 
+TEST_CASE("token bucket default initialization", "[token_bucket]")
+{
+	raw_token_bucket tb;
+	REQUIRE(tb.get_tokens() == 1);
+}
+
 TEST_CASE("token bucket ctor with custom timer", "[token_bucket]")
 {
 	auto t = []() -> uint64_t { return 22; };
-	auto tb = std::make_shared<token_bucket>(t);
+	auto tb = std::make_shared<token_bucket>(1, 1, t);
 
 	REQUIRE(tb->get_tokens() == 1);
 	REQUIRE(tb->get_last_seen() == 22);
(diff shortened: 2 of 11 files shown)


irozzo-1A:
Not finished, but if someone has the time to make a first pass would be great to have early feedback. ;-)

leodido (commented):
Hello Iacopo, thanks for submitting this.

I like the idea of separating the responsibility about which clock to use.
Even if, there's no real use case for it in this codebase. Since we have to use the timestamp coming from sinsp for performance reasons.

ANyways, looking forward to this PR completion!

A nitpick: could you please revert the lines changes that only contain formatting?
We have to work on code conventions, I agree. But that's a task for separate PRs/working groups.
Having here only those changes coherent with this PR will help the maintainers to review it.

Thanks again!

irozzo-1A:
@leodido Thx for your feedback.

> I like the idea of separating the responsibility about which clock to use.
> Even if, there's no real use case for it in this codebase. Since we have to use the timestamp coming from sinsp for performance reasons.

Interesting, let me expand a bit on my reasoning. I saw `token_bucket` is used in two places with different patterns:

https://github.com/falcosecurity/falco/blob/6beb9838d6309dc9d754cae481eaaf38f0a97878/userspace/falco/event_drops.cpp#L90
https://github.com/falcosecurity/falco/blob/6beb9838d6309dc9d754cae481eaaf38f0a97878/userspace/falco/falco_outputs.cpp#L147

In the first use case, the timestamp used in `claim` is provided from outside, and the second is "standard" usage. I thought that from a pure API design standpoint it's better to separate the two for clarity and to reduce the possibility of misusage. Concretely speaking what I did does not bring a lot of benefits :-) at the moment.

Secondly, I noticed that [sinsp_utils::get_current_time_ns](https://github.com/draios/sysdig/blob/8855f4505dc42ad9e328ef6df94c61984cf657e7/userspace/libsinsp/utils.cpp#L920-L926) makes use of [gettimeofday](https://man7.org/linux/man-pages/man2/gettimeofday.2.html), but as we are interested in computing elapsed time would be more convenient to use [clock_gettime](https://man7.org/linux/man-pages/man2/clock_gettime.2.html) with `CLOCK_MONOTONIC ` as it guarantees time is linearly increasing (note that at the moment I used `std::chrono::steady_clock` for convenience). AFAIU your claim is that `gettimeofday` in general is faster, isn't it?
I'll try to benchmark it (just out of curiosity, not for mistrust :smile:).

If you estimate that the benefit is not significant of course I'm open to removing this from the PR as it is not strictly needed/related with the issue :wink:.

> A nitpick: could you please revert the lines changes that only contain formatting?
> We have to work on code conventions, I agree. But that's a task for separate PRs/working groups.
> Having here only those changes coherent with this PR will help the maintainers to review it.

Definitely, I agree with you it was not intended. I'll clean it up.

leodido:
> Interesting, let me expand a bit on my reasoning. I saw `token_bucket` is used in two places with different patterns:
> 
> https://github.com/falcosecurity/falco/blob/6beb9838d6309dc9d754cae481eaaf38f0a97878/userspace/falco/event_drops.cpp#L90
> 
> 
> https://github.com/falcosecurity/falco/blob/6beb9838d6309dc9d754cae481eaaf38f0a97878/userspace/falco/falco_outputs.cpp#L147
> 
> In the first use case, the timestamp used in `claim` is provided from outside, and the second is "standard" usage. I thought that from a pure API design standpoint it's better to separate the two for clarity and to reduce the possibility of misusage. Concretely speaking what I did does not bring a lot of benefits :-) at the moment.

Yup, exactly what I meant. :)
Anyways, in case the problem to solve was "reduce the possibility of misusage", there was also a simpler way to do it. 
However, as said, I like it from an API design standpoint.

> Secondly, I noticed that [sinsp_utils::get_current_time_ns](https://github.com/draios/sysdig/blob/8855f4505dc42ad9e328ef6df94c61984cf657e7/userspace/libsinsp/utils.cpp#L920-L926) makes use of [gettimeofday](https://man7.org/linux/man-pages/man2/gettimeofday.2.html), but as we are interested in computing elapsed time would be more convenient to use [clock_gettime](https://man7.org/linux/man-pages/man2/clock_gettime.2.html) with `CLOCK_MONOTONIC ` as it guarantees time is linearly increasing (note that at the moment I used `std::chrono::steady_clock` for convenience). AFAIU your claim is that `gettimeofday` in general is faster, isn't it?
> I'll try to benchmark it (just out of curiosity, not for mistrust 😄).

To clarify my claim was that our optimization time is better spent somewhere else.

Lemme try to explain.
Last summer, we've benchmarked various timers a bit.
As far as I remember, whether `gettimeofday` is faster/slower than `clock_gettime` largely depends on whether the CPU models you run on have a synchronized TSC.
In fact, in this case, modern Linux kernels use a userspace TSC-based `gettimeofday` (the vdso pulls in an implementation that applies an offset to `rdtsc`'s value, thus computing the time of day without entering the kernel).

Also, `steady_clock` already uses `CLOCK_MONOTONIC` ([ref](https://github.com/gcc-mirror/gcc/blob/3defceaa1a2987fa90296abfbcc85d7e9ad59684/libstdc%2B%2B-v3/src/c%2B%2B11/chrono.cc#L80-L95)) (even if not always).

Finally, while `m_notifications_tb::claim` is called inside every `handle_event`, `m_bucket::claim` is called inside `process_event` when there have been drops.

So, is probably better first quantify those two frequencies and understand which time resolution are we exposed too.

Because we could also use `CLOCK_MONOTIC_COARSE` in case we don't mind losing resolution.

I've not rechecked things that I remember from last summer and that I wrote here (sadly I don't find numbers in my notes), so maybe I remember them completely wrong. In that case, be kind to me :)



> Definitely, I agree with you it was not intended. I'll clean it up.

Yup, please remove the only-formatting changes, if you don't mind.

irozzo-1A:
Thx for the insights @leodido! I'm a bit busy at the moment, but I'll finalize this ;-)

> Also, `steady_clock` already uses `CLOCK_MONOTONIC` ([ref](https://github.com/gcc-mirror/gcc/blob/3defceaa1a2987fa90296abfbcc85d7e9ad59684/libstdc%2B%2B-v3/src/c%2B%2B11/chrono.cc#L80-L95)) (even if not always).

That's what I expected, but I did not verify ;-)

irozzo-1A:
@leodido I removed the formatting-related changes and put the token bucket refactoring in a dedicated commit so that I can easily get rid of it if you don't need/want it.
I won't have much time to work on this in the near future. If you need some changes and you cannot wait for me feel free to close this and take ownership (you can reuse part of the code I did of course).

japui:
WIP] Tornar a limitação da taxa de saída opcional

jasondellaluce:
Might be superseded by https://github.com/falcosecurity/falco/pull/2080, which removes the output rate limiter entirely.

irozzo-1A:
> Might be superseded by #2080, which removes the output rate limiter entirely.

Ok, l close it. Thanks for letting me know @jasondellaluce 😉

PR #1530, opened 2021-01-19
new(proposals): libraries donation

leodido:
Co-authored-by: Lorenzo Fontana <lo@linux.com>

**What type of PR is this?**

/kind documentation

/area proposals

Donate:
- libsinsp
- libscap
- the kernel module driver
- the eBPF driver sources

by moving them to the Falco project.

NONE

```release-note
docs(proposals): libraries and drivers donation
```

### DIFF
--- a/proposals/2021019-libraries-donation.md
+++ b/proposals/2021019-libraries-donation.md
@@ -1,0 +1,167 @@
+# OSS Libraries Donation Plan
+
+## Summary
+
+Sysdig Inc. intends to donate **libsinsp**, **libscap**, the **kernel module driver** and the **eBPF driver sources** by moving them to the Falco project.
+
+This means that some parts of the [draios/sysdig](https://github.com/draios/sysdig) repository will be moved to a new GitHub repository called [falcosecurity/libs](https://github.com/falcosecurity/libs).
+
+This plan aims to describe and clarify the terms and goals to get the donation done.
+
+## Motivation
+
+There are two main OSS projects using the libraries and drivers that we are aware of:
+
+- [sysdig](https://github.com/draios/sysdig)  the command line tool
+- [Falco](https:/github.com/falcosecurity/falco), the CNCF project.
+
+Since the Falco project is a heavy user of the libraries, a lot more than the sysdig cli tool, Sysdig (the company) decided to donate the libraries and the driver to the Falco community.
+
+Sysdig (the command line tool) will continue to use the libraries now provided by the Falco community underneath.
+
+This change is win-win for both parties because of the following reasons:
+
+- The Falco community owns the source code of the three most important parts of the software it distributes.
+  - Right now it is "only" an engine on top of the libraries. This **donation** helps in making the scope of the Falco project broader. Having the majority of the source code under an **open governance** in the same organization gives the Falco project more contribution opportunities, helps it in **evolving independently** and makes the whole Falco community a strong owner of the processes and decision making regarding those crucial parts.
+
+- Given the previous point, Sysdig (the command line tool) will benefit from the now **extended contributors base**
+
+- Sysdig (the company) can now focus on the user experience and user space features
+
+- **Contributions** to the libraries and drivers will be **easier** to spread across the Falco community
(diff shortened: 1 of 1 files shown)


deepskyblue86:
How do you think to proceed with their versioning? I read everything will be in falcosecurity/libs, do you want to release everything with the same version?

leodido:
@deepskyblue86 at the moment Falco uses a commit hash to reference the libs. That hash is also the "driver version".

See https://github.com/falcosecurity/falco/blob/2883df5808084985d94fefb1b97463900740ab57/cmake/modules/sysdig.cmake#L33

This mechanism proved to work pretty well (for sure, better than the previous one).

**Anyway**, this proposal is about the donation of the libraries (and the drivers' sources) to the falcosecurity organization. That's the goal here. :)

Changes to their versioning scheme, like versioning them separately, need to be eventually proposed and discussed in a follow-up proposal. Because such changes would impact a lot of areas (build system, release system, prebuilt drivers machinery, etc.).

deepskyblue86:
Sorry @leodido, but I raised it because I read the following:
> 8. Distribute the `libsinsp.so` library and headers as an artifact (rpm, deb, tar.gz) following the falcosecurity current process
> 9. Distribute the `libsinsp.a` library and headers as an artifact (rpm, deb, tar.gz) following the falcosecurity current process
> 10. Creation of the CI scripts using the Falco CI and Falco Infra
> 11. The CI scripts will need to publish the artifacts in the current falcosecurity artifacts repository
> 12. Artifacts will be pushed for every tag (release) and for every master merge (development release)
> ... 
> ### Falco
> 2. When distributing the deb and rpm, libscap and libsinsp will need to be install dependencies and not anymore compiled into Falco

so I assumed it was to be discussed now.

leodido:
In fact, we are discussing it, Angelo, no worries :)

What I was trying to say is: the falcosecurity current process (at the moment) relies on that hash, which is the same for all the 3 (libscap, libsinp, drivers) underlying pieces.

Changing the versioning scheme would require to also change the current falcosecurity processes for building and distributing software. Which rather is not the scope of this proposal.

deepskyblue86:
Discussed privately with @leodido. I was just raising the issue of having a semver for the libraries and not just the git commit hash, due to deb/rpm package management. Of course, it can just be something like x.y.z.<githash> if needed.

We also discussed having just a single package with all these components — like falcosecurity-libs — for release and maintenance convenience, given that there's already the idea to have them with the same version.

leogr:
> In this document we just propose to donate those libraries by keeping the versioning, building, releasing etc as much as similar to the existing one.

:+1:  this proposal explains "the full plan", and I really love that.

Initial steps to start the migration are:

> 1. Extract libsinsp from `draios/sysdig/userspace/libsinsp` (keeping the commit history) into [falcosecurity/libs](https://github.com/falcosecurity/libs)
> 
> 2. The migration comes first, then we can do additional PRs for the points below so that we do only one thing at a time and keep the history linear
>
> 3. Keep the same code, refactorings will need to be done in subsequent PRs and approved separately

And I totally agree with that. Doing so will allow us to start using the migrated libraries almost immediately. Since I wanted to try, I already did a PoC (will share soon :smile_cat: ).

Then we can proceed with other refinements in a step-by-step fashion.

Btw, if the main question here is "Would you like to accept the donation?" - as I believe - my answer is: a strong YES :grinning:

leogr (approved):
I strongly agree with this proposal, so I'm approving. :partying_face: 

I have also created a sort of PoC (see below) to demonstrate that the very first steps of this migration can be done without any disruption for our users.

- https://github.com/leogr/falco-libs It is an example of how the repo falcosecurity/libs will look like (btw, I believe it's better to import the git repo, not forking it)
- https://github.com/leogr/falco/tree/update/lib-migration This branch represents the changes to be applied to Falco to make it point to the above-mentioned libs

N.B. The two links above are just examples that I wanted to share. They might not exactly represent the final work.

fntlnz (approved):

PR #1538, opened 2021-01-26
Allow unknown sources

mstemm:
**What type of PR is this?**

In #1427 rule processing was already relaxed a bit to allow rules to have new properties without failures. This further relaxes processing to completely ignore rules that have a source other than "syscall" or "k8s_audit"., for example:
```
- rule: Rule with unknown source
  condition: some unknown filteer
  output: some unknown output
  priority: INFO
```

Without this change, falco will try to read rules with unknown sources and match the filtercheck fields against known supported sets for k8s audit/syscall.

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind rule-update

> /kind rule-create

> /area build

/area engine

> /area rules

> /area tests

> /area proposals

Fixes #

```release-note
Silently allow and ignore rules where the source is something other than "syscall" or "k8s_audit" (or blank, which defaults to "syscall").
```

### DIFF
--- a/test/falco_tests.yaml
+++ b/test/falco_tests.yaml
@@ -1277,3 +1277,10 @@ trace_files: !mux
     trace_file: trace_files/cat_write.scap
     stdout_contains: "2016-08-04T16:17:57.882054739\\+0000: Warning An open was seen"
     stderr_contains: "^\\d\\d\\d\\d-\\d\\d-\\d\\dT\\d\\d:\\d\\d:\\d\\d\\+0000"
+
+  unknown_source:
+    detect: True
+    detect_level: WARNING
+    rules_file:
+      - rules/unknown_source.yaml
+    trace_file: trace_files/cat_write.scap
--- a/test/rules/unknown_source.yaml
+++ b/test/rules/unknown_source.yaml
@@ -1,0 +1,31 @@
+#
+# Copyright (C) 2021 The Falco Authors.
+#
+#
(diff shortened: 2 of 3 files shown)


leogr (commented):
Please note that the `if append then` is evaluated after the check to skip unknown source. So, it seems to me that by doing so, the `append` property will not be considered for sources other than `syscall` and `k8s_audit`. 

 If that was intended, then I don't understand the reason.

mstemm:
The idea is to entirely skip rules where the source is something other than syscall/k8s_audit. So that would also mean skipping appending rules to other rules, since the appended rule would be one that falco should skip anyway. I think I got that part right--which if append then were you looking at?

leogr:
> The idea is to entirely skip rules where the source is something other than syscall/k8s_audit. So that would also mean skipping appending rules to other rules, since the appended rule would be one that falco should skip anyway. I think I got that part right--which if append then were you looking at?

Ok, your intent is clear to me now, thanks for the explanation. 

Anyway, another point: it is not exactly clear to me (as a user) what's the purpose to silently ignore those rules. It could lead to errors that can be hard to discover. So, what's the use case, could you provide an example?

mstemm:
I discussed more with @leogr and I'm going to close this for now. Maybe we'll bring it back when falco starts handling a wider variety of event sources.

PR #1546, opened 2021-01-30
falco: add healthz endpoint

cpanato:
**What type of PR is this?**
/kind feature

Add an initial implementation to have a `heatlhz` endpoint, this is useful when running the workload in kubernetes so we can add liveness/read probes to do the checks and fail when is not responding.

In the future, we can have some deep checks if needed.

This PR is the initial to fix this issue: https://github.com/falcosecurity/charts/issues/53

tried my best with c++ 😄 

/assign @fntlnz  @leodido 

Fixes #

```release-note
falco: add healthz endpoint
```

### DIFF
--- a/userspace/falco/configuration.cpp
+++ b/userspace/falco/configuration.cpp
@@ -34,6 +34,7 @@ falco_configuration::falco_configuration():
 	m_webserver_enabled(false),
 	m_webserver_listen_port(8765),
 	m_webserver_k8s_audit_endpoint("/k8s-audit"),
+	m_webserver_k8s_healthz_endpoint("/healthz"),
 	m_webserver_ssl_enabled(false),
 	m_config(NULL)
 {
@@ -193,6 +194,7 @@ void falco_configuration::init(string conf_filename, list<string> &cmdline_optio
 	m_webserver_enabled = m_config->get_scalar<bool>("webserver", "enabled", false);
 	m_webserver_listen_port = m_config->get_scalar<uint32_t>("webserver", "listen_port", 8765);
 	m_webserver_k8s_audit_endpoint = m_config->get_scalar<string>("webserver", "k8s_audit_endpoint", "/k8s-audit");
+	m_webserver_k8s_healthz_endpoint = m_config->get_scalar<string>("webserver", "k8s_healthz_endpoint", "/healthz");
 	m_webserver_ssl_enabled = m_config->get_scalar<bool>("webserver", "ssl_enabled", false);
 	m_webserver_ssl_certificate = m_config->get_scalar<string>("webserver", "ssl_certificate", "/etc/falco/falco.pem");
 
--- a/userspace/falco/webserver.cpp
+++ b/userspace/falco/webserver.cpp
@@ -34,6 +34,15 @@ k8s_audit_handler::~k8s_audit_handler()
 {
 }
 
+bool k8s_healthz_handler::handleGet(CivetServer *server, struct mg_connection *conn)
+{
+	const std::string status_body = "{\"status\": \"ok\"}";
+	mg_send_http_ok(conn, "application/json", status_body.size());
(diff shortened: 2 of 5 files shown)


userspace/falco/webserver.cpp
```diff
@@ -34,6 +34,23 @@ k8s_audit_handler::~k8s_audit_handler()
...
+{
+}
+
+k8s_healthz_handler::~k8s_healthz_handler()
+{
+}
+
+bool k8s_healthz_handler::handleGet(CivetServer *server, struct mg_connection *conn)
+{
+	std::string status_body = "{\"status\": \"ok\"}";
```
fntlnz, line 47:
```suggestion
	const std::string status_body = "{\"status\": \"ok\"}";
```

fntlnz (commented):
Good job @cpanato !

Just a little nit then LGTM

userspace/falco/webserver.cpp
cpanato: thanks @fntlnz for your review, did the change and also update https://github.com/falcosecurity/falco/pull/1546/files#diff-759cf16077c5a03cbf4e9b6dd8af34208252447adb0f3f894e5aadb6f218335dR168 to match your feedback here 😄 

PTAL

userspace/falco/configuration.cpp
```diff
@@ -193,6 +194,7 @@ void falco_configuration::init(string conf_filename, list<string> &cmdline_optio
 	m_webserver_enabled = m_config->get_scalar<bool>("webserver", "enabled", false);
 	m_webserver_listen_port = m_config->get_scalar<uint32_t>("webserver", "listen_port", 8765);
 	m_webserver_k8s_audit_endpoint = m_config->get_scalar<string>("webserver", "k8s_audit_endpoint", "/k8s-audit");
+	m_webserver_k8s_healthz_endpoint = m_config->get_scalar<string>("webserver", "k8s_healthz_endpoint", "/healthz");
```
leogr, line 197: It should be added to `falco.yaml` too, so folks can find this option easily.

leogr (commented):
Overall it seems good to me, but I'm not sure about one thing.

We assume the (already implemented) Falco web server is enabled, so `/heathz` endpoint works. But this assumption can be false under some circumstances since the web server can easily be disabled by the config (see https://github.com/falcosecurity/falco/blob/master/falco.yaml#L156 ).
It might also be confusing as the only purpose of the webserver has been to ingest K8s Auditing log events until now.

So, IMHO we should either:
- implement the `/healthz` endpoint on another web server instance (btw, I don't like this idea)
OR
- document very well the new purposes of the webserver (making it clear that it's not for k8s events only, my vote goes for this option)

PS
well done, @cpanato ! We really missed this feature, thank you! :clap:

cpanato:
@leogr @fntlnz Agree with documentation the endpoint, please let me know what y'all prefer then I will update the PR accordingly

userspace/falco/webserver.cpp
```diff
@@ -34,6 +34,23 @@ k8s_audit_handler::~k8s_audit_handler()
 {
 }
 
+k8s_healthz_handler::k8s_healthz_handler()
```
leodido, line 37: Would define in webserver.h

leogr:
> @leogr @fntlnz Agree with documentation the endpoint, please let me know what y'all prefer then I will update the PR accordingly

Hey @cpanato 

I think that just adding the new configuration option to [falco.yaml](https://github.com/falcosecurity/falco/blob/master/falco.yaml#L145-L162) (with the usage description in the comments) would be the right way to document it.

cpanato:
@leogr PTAL

cpanato:
ping 😄  @leogr @fntlnz

leogr (approved):
LGTM

leodido (approved):

PR #1569, opened 2021-03-08
add flag to drop events via unset_eventmask in libsinsp

zemek:
**What type of PR is this?**
/kind feature

/area engine

We would like to drop certain noisy events that we don't care about at the kernel level to improve performance. It looks like the capability to do this already exists in the libraries/driver. This adds a new flag `--ignore-event <event>`  which can be called multiple times, for example `--ignore-event open --ignore-event close`.

Fixes #

```release-note
new: flag "--ignore-event <event_name>" which can be used to drop unwanted events in kernelspace
```

### DIFF
--- a/userspace/falco/falco.cpp
+++ b/userspace/falco/falco.cpp
@@ -191,6 +191,7 @@ static void display_fatal_err(const string &msg)
 
 // Splitting into key=value or key.subkey=value will be handled by configuration class.
 std::list<string> cmdline_options;
+std::set<string> ignored_event_names_cmdline;
 
 #ifndef MINIMAL_BUILD
 // Read a jsonl file containing k8s audit events and pass each to the engine.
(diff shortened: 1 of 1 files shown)


leodido (commented):
LGTM.

/hold
to postpone this for the next release Falco 0.28.1

FedeDP:
Hi @zemek !  
Thanks for this great PR. Unfortunately i think it is a very tricky choice to let users be able to manually disable some events, for a couple of reasons:
* first of all, not all syscalls are heavyweight
* some syscalls are **required** for sinsp metadata enrichment

Given all of the above, we don't really want to let users manually disable syscalls; instead a better solution is to port current _simpleconsumer_ mode to be kernel side; this is what i've done in https://github.com/falcosecurity/libs/pull/115.
With a kernel side simpleconsumer mode we get all the benefits of dropping heavy syscalls, while we can be sure that sinsp enrichment is untouched, new containers are correctly fetched, and so on.

What do you think?

PR #1578, opened 2021-03-11
Use url-safe characters in falco version

mstemm:
In some cases, you might want to host falco packages in a way where
they're directly accessible via http. The '+' character that separates
the version and the git hash ends up breaking naive solutions that don't
properly url-escape the package name before doing the http fetch.

Of course, clients can properly url-escape, but switching to a tilde is
url safe and I think still preserves the idea of separating the version
and hash.

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind rule-update

> /kind rule-create

/area build

> /area engine

> /area rules

> /area tests

> /area proposals

Make the falco version a bit more url-friendly by using tilde as the version/hash separator instead of plus.

Fixes #

```release-note
Make the falco version a bit more url-friendly by using tilde as the version/hash separator instead of plus.
```

### DIFF
--- a/cmake/modules/GetFalcoVersion.cmake
+++ b/cmake/modules/GetFalcoVersion.cmake
@@ -27,7 +27,7 @@ if(NOT FALCO_VERSION)
       set(FALCO_VERSION "0.0.0")
     endif()
     # Format FALCO_VERSION to be semver with prerelease and build part
-    string(REPLACE "-g" "+" FALCO_VERSION "${FALCO_VERSION}")
+    string(REPLACE "-g" "~" FALCO_VERSION "${FALCO_VERSION}")
   else()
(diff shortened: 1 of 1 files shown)


leogr (commented):
We cannot change this because it's a SemVer spec:
https://semver.org/#spec-item-10

Also, I don't believe this is an issue that should be solved here since it's a client issue
(btw the client should always encode URL properly).

leodido (commented):
I see your point Mark but this is what the SemVer 2.0 specification demands for the build metadata part (see point 10 of the spec).



Anyways moving this to 
/milestone 0.28.1
for future discussion

PR #1599, opened 2021-04-02
make the falco driver loader downloads first, then fallback to compilation

leodido:
**What type of PR is this?**

/kind feature

ALL (since we don't have a scripts area).

This PR inverts the logic of the `falco-driver-loader` script.

The new logic (expand 👇 ) is:

<details><summary>kernel module case (default one)</summary>

1. Unloading the Falco module (`falco.ko`) if present
2. If the Falco module is present but the script was unable to remove it, exit (and try to use it, notice that the driver version is not checked)
3. Try to add a Falco module from the system modules directory - `/lib/modules/$(uname -r)` - exit in case operation succeeds (notice that the driver version is not checked)
4. Look for a Falco module - eg., `falco_{distro}_{kernel_release}_{kernel_version}.ko` - in `${HOME}/.falco/` - insert it if present (notice that the driver version is not checked)
5. Try to find a suitable prebuilt kernel module on https://download.falco.org
6. Try to compile the kernel module on the fly if not found a prebuilt one 

</details>

<details><summary>eBPF probe case</summary>

1. Look for a Falco eBPF probe - eg., `falco_{distro}_{kernel_release}_{kernel_version}.o` - in `${HOME}/.falco/` - go to 4 (notice that the driver version is not checked)
2. Try to find a suitable prebuilt eBPF probe on https://download.falco.org
3. Try to compile the eBPF probe on the fly if not found a prebuilt one
4. Symlink `${HOME}/.falco/falco_{distro}_{kernel_release}_{kernel_version}.o` to `${HOME/.falco/falco-bpf.o`

</details>

Fixes #

The reason for this PR is to reduce the number of cases in which the Falco users need to wait for a Falco container to build a driver on the fly.

This will also speed up the boot of containerized Falco (assuming a prebuilt exists for the host machine, we now have ~4K prebuilt Falco drivers at https://download.falco.org).

The only cases in which https://download.falco.org will not be hit by a request will be:

- a Falco eBPF probe already exists in `${HOME}/.falco/falco_{distro}_{kernel_release}_{kernel_version}.o`
- a Falco kernel module already exists in `${HOME}/.falco/falco_{distro}_{kernel_release}_{kernel_version}.ko`
- a Falco kernel module is already inserted (and has not been removed)
- a Falco kernel module is already present in the modules directory

In this way, the Falco community could also have more real statistics about the Falco usage and usage patterns by looking at the logs of https://download.falco.org.

Notes that this PR also moves out from the `load_bpf_probe` function the target detection logic in its place (`get_target_id`) for `minikube` and `COS`.

```release-note
update: falco-driver-loader tries to download a Falco driver before then compiling it on the fly for the host
BREAKING CHANGE: SKIP_MODULE_LOAD env variable no more disables the driver loading (use SKIP_DRIVER_LOADER env variable)
```


leodido:
Ready for review even if **WIP**

leogr (commented):
Overall, it seems very good to me :clap: 

I have just found an issue that should easily be fixable (see the comment below about the need of mounting `debugfs`).

Other comments are just small thing, not really needed :)

scripts/falco-driver-loader
```diff
@@ -216,54 +217,56 @@ load_kernel_module() {
 		exit 1
 	fi
 
-	echo "* Unloading ${DRIVER_NAME} module, if present"
+	echo "* Unloading the Falco module, if present"
```
fntlnz, line 220: I would leave DRIVER_NAME here, we can discuss this change in another PR since it's not the scope of this one, plus i don't believe this change is useful as it prevents users from reusing this scripts for using multiple Falco instances on the same host.

fntlnz: Same for all the other places where DRIVER_NAME was replaced

scripts/falco-driver-loader
```diff
@@ -314,14 +317,14 @@ load_bpf_probe_compile() {
...
-		sed -i "s/LOCALVERSION=\"\"/LOCALVERSION=\"${KERNEL_EXTRA_VERSION}\"/" .config
+			sed -i "s/LOCALVERSION=\"\"/LOCALVERSION=\"${KERNEL_EXTRA_VERSION}\"/" .config
 		fi
 		make olddefconfig > /dev/null
 		make modules_prepare > /dev/null
 	}
 
-	if [ -n "${COS}" ]; then
-		echo "* COS detected (build ${BUILD_ID}), using cos kernel headers"
+	if [ "${TARGET_ID}" == "cos" ]; then
```
fntlnz, line 326: This type of stuff goes to another PR to be reviewed separately, has nothing to do with inverting the order of download vs compile.

scripts/falco-driver-loader
```diff
@@ -278,31 +281,31 @@ clean_kernel_module() {
...
-		echo "* No ${DRIVER_NAME} module found in dkms"
+		echo "* There is no Falco module in dkms"
 		return
 	fi
 	for CURRENT_VER in ${DRIVER_VERSIONS}; do
 		if dkms remove -m "${DRIVER_NAME}" -v "${CURRENT_VER}" --all 2>/dev/null; then
-			echo "* Removing ${DRIVER_NAME}/${CURRENT_VER} succeeded"
-		else 
-			echo "* Removing ${DRIVER_NAME}/${CURRENT_VER} failed"
+			echo "* Removing the Falco module (version ${CURRENT_VER}) succeeded"
```
fntlnz, line 306: I liked the old way this was printed, it was way more clear printing the actual name of the removed dkms module

scripts/falco-driver-loader
```diff
@@ -352,7 +355,8 @@ load_bpf_probe_compile() {
 		}
 	fi
 
-	if [ -n "${MINIKUBE}" ]; then
+	if [ "${TARGET_ID}" == "minikube" ]; then
```
fntlnz, line 358: Again this has nothing to do with the topic of the PR

scripts/falco-driver-loader
```diff
@@ -426,70 +432,40 @@ load_bpf_probe_download() {
...
-
-	echo "* Mounting debugfs"
-
-	if [ ! -d /sys/kernel/debug/tracing ]; then
-		mount -t debugfs nodev /sys/kernel/debug
-	fi
-
-	get_kernel_config
-
-	if [ -n "${HOST_ROOT}" ] && [ -f "${HOST_ROOT}/etc/os-release" ]; then
```
fntlnz, line 444: I don't think this has anything to do with the inversion of download vs compile here

scripts/falco-driver-loader
```diff
@@ -506,13 +482,14 @@ print_usage() {
 	echo "Options:"
 	echo "  --help         show brief help"
 	echo "  --clean        try to remove an already present driver installation"
-	echo "  --compile      try to compile the driver locally"
-	echo "  --download     try to download a prebuilt driver"
+	echo "  --compile      try to compile the driver locally (default on)"
```
fntlnz, line 485:
```suggestion
	echo "  --compile      try to compile the driver locally (default true)"
```

scripts/falco-driver-loader
```diff
@@ -506,13 +482,14 @@ print_usage() {
...
+	echo "  --download     try to download a prebuilt driver (default on)"
```
fntlnz, line 486:
```suggestion
	echo "  --download     try to download a prebuilt driver (default true)"
```

fntlnz (changes requested):

scripts/falco-driver-loader
```diff
@@ -426,70 +432,40 @@ load_bpf_probe_download() {
...
 	fi
 }
 
 load_bpf_probe() {
-
-	echo "* Mounting debugfs"
-
-	if [ ! -d /sys/kernel/debug/tracing ]; then
-		mount -t debugfs nodev /sys/kernel/debug
-	fi
```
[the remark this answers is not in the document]
leodido: Right, my fault.

scripts/falco-driver-loader
fntlnz: I agree on the guidelines in general but this is the name of the module rather than the name of the project and the module is called `falco` rather than `Falco` now

scripts/falco-driver-loader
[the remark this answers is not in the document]
fntlnz: Thanks Leo, I agree we need to do this but I would prefer to do it in a second moment if you agree

scripts/falco-driver-loader
leodido: Yup, no problem at all. I'll revert ASAP

scripts/falco-driver-loader
```diff
@@ -140,33 +142,33 @@ get_target_id() {
...
 	if [[ $(uname -r) == *uek* ]]; then
-		echo "* Skipping dkms install for UEK host"
+		>&2 echo "Skipping because the dkms install always fail (on UEK hosts)"
 		return
 	fi
 
-	if ! hash dkms &>/dev/null; then
-		echo "* Skipping dkms install (dkms not found)"
+	if ! hash dkms >/dev/null 2>&1; then
+		>&2 echo "This program requires dkms"
```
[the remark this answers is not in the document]
leodido: It seems always required to me if the script reaches this point.

scripts/falco-driver-loader
```diff
@@ -181,22 +183,21 @@ load_kernel_module_compile() {
...
+	echo "* Trying to download a prebuilt Falco module from ${URL}"
 	if curl -L --create-dirs "${FALCO_DRIVER_CURL_OPTIONS}" -o "${HOME}/.falco/${FALCO_KERNEL_MODULE_FILENAME}" "${URL}"; then
 		echo "* Download succeeded"
-		insmod "${HOME}/.falco/${FALCO_KERNEL_MODULE_FILENAME}" && echo "* Success: ${DRIVER_NAME} module loaded"
+		insmod "${HOME}/.falco/${FALCO_KERNEL_MODULE_FILENAME}" && echo "* Success: Falco module found and inserted"
 		exit $?
 	else
-		>&2 echo "Download failed, consider compiling your own ${DRIVER_NAME} module and loading it or getting in touch with the Falco community"
-		exit 1
+		>&2 echo "Unable to find a prebuilt Falco module"
```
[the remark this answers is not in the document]
leodido: This was already considered an error, and I tend to continue to consider it an error. Except that now, it's recoverable if the script is later able to compile the kernel module on the fly.

scripts/falco-driver-loader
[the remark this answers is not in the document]
leodido: We double-checked that it's correct to consider it an error like I did.

Resolving

scripts/falco-driver-loader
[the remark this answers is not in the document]
leodido: The key point anyways is the return below. Rather than the previous exit 1.

If the script was called with `--download` flag only this needs to be the definitive error.
Otherwise, an attempt to recover from this error (by compiling the kernel module) will be done soon after.

scripts/falco-driver-loader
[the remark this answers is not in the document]
leodido: Ideally, the correct place for this mount would be in Falco binary when the `FALCO_BPF_PROBE` env variable is present.

I'll revert it back for now, anyways.

scripts/falco-driver-loader
leodido: Thanks for the suggestions. I'm gonna use "yes" which is the default value when both flags are not passed. It's also the value that gets assigned to `ENABLE_COMPILE` when `--compile` is rather passed in by the user (same for `--download`).

leogr (approved):
LGTM! :+1:

fntlnz (approved):

Issue #1653, opened 2021-05-14
Cut 0.28.2 release


maxgio92:
## Scheduled to happen: > 2021-05-16, < 2021-06-09
This is a patch release. We can make the release happen as the features are merged.

## Release Steps
Process is described in this document: https://github.com/falcosecurity/falco/blob/master/RELEASE.md

## Release Blocking Features

- #1552
- [test-infra #398](https://github.com/falcosecurity/test-infra/pull/398)

## Action Items

- [x]  Update release documentation (#1652)
- [x] Update DBG configurations ([test-infra #398](https://github.com/falcosecurity/test-infra/pull/398))
- [x] Prebuilt drivers publish ([Jobs status](https://prow.falco.org/?repo=falcosecurity%2Ftest-infra&type=postsubmit&job=build-drivers-*))
- [x] Pre: Release notes
- [x] Pre: Milestones
- [x] Pre: Changelog update (#1661)
- [ ] Tag
- [ ] GH Release
- [x] Meeting-notes ([community #138](https://github.com/falcosecurity/community/pull/138))
- [ ] Blog ([falco-website #442](https://github.com/falcosecurity/falco-website/pull/442))
- [ ] Send email to cncf-falco-dev mailing list
- [ ] Announce in Falco slack channel

## Open Questions
/milestone 0.28.2
/assign @maxgio92 @leogr @zuc 
/kind documentation

leogr:
Update:

The last PR for cutting 0.28.2 is ready to be reviewed :point_down:
https://github.com/falcosecurity/falco/pull/1552

TODO: Since the driver version changed, we should also update the Driver Build Grid configuration in test-infra

maxgio92:
Thanks @leogr. Updated the checklist in [this issue](https://github.com/falcosecurity/falco/issues/1653)'s description.

maxgio92:
I'm going to prepare DBG configurations

zuc:
hello 👋

maxgio92:
Hello @zuc!
As discussed @zuc would want to be the release shadow
/assign @zuc

maxgio92:
Update:

Drivers have been built with success, not published yet ([test-infra #398](https://github.com/falcosecurity/test-infra/pull/398)).

maxgio92:
Update:

Release blocking feature #398 gots merged

maxgio92:
Update:
[...]

PR #1662, opened 2021-06-01
build(cmake/modules): avoid gRCP files to be installed into packages

leogr:
**What type of PR is this?**

/kind bug

/area build

This patch is to be intended as a temporary workaround.
The root issue needs to be addressed in `falcosecurity/libs`.
A definite solution will be implemented later.

Fixes #

This is a hotfix to unblock the current release. 
It addresses the issue described in https://github.com/falcosecurity/falco/issues/1653#issuecomment-850438225

/milestone 0.28.2

```release-note
NONE
```

### DIFF
--- a/cmake/modules/falcosecurity-libs-repo/patch/grpc.patch
+++ b/cmake/modules/falcosecurity-libs-repo/patch/grpc.patch
@@ -1,0 +1,12 @@
+diff --git a/cmake/modules/grpc.cmake b/cmake/modules/grpc.cmake
+index ffa4e6c4..346939b9 100644
+--- a/cmake/modules/grpc.cmake
++++ b/cmake/modules/grpc.cmake
+@@ -125,6 +125,7 @@ else()
+ 				-DZLIB_ROOT:STRING=${ZLIB_SRC}
+ 			BUILD_IN_SOURCE 1
+ 			BUILD_BYPRODUCTS ${GRPC_LIB} ${GRPCPP_LIB} ${GPR_LIB} ${GRPC_LIBRARIES}
++			INSTALL_COMMAND DESTDIR=/ make install
+ 		)
+ 	endif()
+ endif()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v0.28.2
 
-Released on 2021-05-27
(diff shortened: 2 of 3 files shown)


cmake/modules/falcosecurity-libs-repo/patch/grpc.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/cmake/modules/grpc.cmake b/cmake/modules/grpc.cmake
+index ffa4e6c4..346939b9 100644
+--- a/cmake/modules/grpc.cmake
++++ b/cmake/modules/grpc.cmake
+@@ -125,6 +125,7 @@ else()
+ 				-DZLIB_ROOT:STRING=${ZLIB_SRC}
+ 			BUILD_IN_SOURCE 1
+ 			BUILD_BYPRODUCTS ${GRPC_LIB} ${GRPCPP_LIB} ${GPR_LIB} ${GRPC_LIBRARIES}
++			INSTALL_COMMAND DESTDIR=/ make install
```
leogr, line 9: N.B. 
the resulting installation directory is computed by `DESTDIR` + `CMAKE_INSTALL_PREFIX`.
Basically, it forces gRPC installation files to be always installed into the path specified by `CMAKE_INSTALL_PREFIX` that's set to `${GRPC_SRC}/target` (see [here](https://github.com/falcosecurity/libs/blob/master/cmake/modules/grpc.cmake#L94)).

See also:
- https://cmake.org/cmake/help/latest/envvar/DESTDIR.html
https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html

leodido (commented):
I see your point in patching this here to complete the hotfix release of Falco 0.28 line.

But, wouldn't be simpler and more straightforward to fix the thing where it really belongs to (`falcosecurity/libs`) and cancel this hotfix release since Falco 0.29 is due in 7 days as per milestones?

Yes, a new driver version would be required and 13ec67e (that in this case never made part of an official Falco version) removed... But to me, it would sound all more correct.

Just my 2 cents

leogr:
As we all agree on releasing 0.29.0 directly, I'm closing this and moving the fix to `libs` directly.
:+1:

PR #1667, opened 2021-06-05
Add customizable metadata fetching params

zuc:
Signed-off-by: Michele Zuccala <michele@zuccala.com>

**What type of PR is this?**

/kind feature

Not sure about the appropriate `/area` label here, could you please advise?

This PR adds support for customizable metadata fetching params that are going to be introduced with https://github.com/falcosecurity/libs/pull/40.

WIP because I'm planning to set the correct libs version and checksum once https://github.com/falcosecurity/libs/pull/40 gets merged.

For now it's set to the feature branch name on https://github.com/falcosecurity/libs in order for this PR to be easily tested.

```release-note
new(userspace/falco): add customizable metadata fetching params
```

### DIFF
--- a/userspace/falco/configuration.cpp
+++ b/userspace/falco/configuration.cpp
@@ -252,6 +252,10 @@ void falco_configuration::init(string conf_filename, list<string> &cmdline_optio
 	{
 		throw logic_error("Error reading config file(" + m_config_file + "): the maximum consecutive timeouts without an event must be an unsigned integer > 0");
 	}
+
+	m_metadata_download_max_mb = m_config->get_scalar<uint32_t>("metadata_download", "max_mb", 100);
+	m_metadata_download_chunk_wait_us = m_config->get_scalar<uint32_t>("metadata_download", "chunk_wait_us", 1000);
+	m_metadata_download_watch_freq_sec = m_config->get_scalar<uint32_t>("metadata_download", "watch_freq_sec", 1);
 }
 
 void falco_configuration::read_rules_file_directory(const string &path, list<string> &rules_filenames)
--- a/userspace/falco/configuration.cpp
+++ b/userspace/falco/configuration.cpp
@@ -254,8 +254,17 @@ void falco_configuration::init(string conf_filename, list<string> &cmdline_optio
 	}
 
 	m_metadata_download_max_mb = m_config->get_scalar<uint32_t>("metadata_download", "max_mb", 100);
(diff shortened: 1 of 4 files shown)


zuc:
Note for the reviewer(s): I chose to keep `m_metadata_download_max_mb` expressed in megabytes for user-friendliness reasons, and to convert the value in bytes only when calling sinsp. Happy to discuss if you have a different point of view here!

zuc:
Since falcosecurity/libs#40 got merged, this is now ready for review.
Also, driver version has been bumped to https://github.com/falcosecurity/libs/commit/c415d577b79bef3b95028b1030ce42a5ec3c5c32.

userspace/falco/configuration.cpp
```diff
@@ -252,6 +252,10 @@ void falco_configuration::init(string conf_filename, list<string> &cmdline_optio
 	{
 		throw logic_error("Error reading config file(" + m_config_file + "): the maximum consecutive timeouts without an event must be an unsigned integer > 0");
 	}
+
+	m_metadata_download_max_mb = m_config->get_scalar<uint32_t>("metadata_download", "max_mb", 100);
```
leodido, line 256: Can we please cap these values (both lower and upper bounds)?

leodido (dismissed):
I believe that checking the config values (coming from the user) against limits (especially upper ones in this case) is necessary.

Example: a user configuring Falco to have `max_mb` equal to eg. `20000`.

Since we know that this approach of fetching the container orchestrator metadata has an impact on memory (and Falco performances), I think it's sane to provide safe boundaries for them.

zuc:
@leodido I added an upper boundary for the maximum download size, and a check to force frequency seconds to be higher than 0.

I don't think it would make much sense to cap the remaining params/boundaries given their meaning, but happy to discuss.

mstemm (dismissed):

leogr:
/hold
to let other PRs get merged in

leogr (dismissed):

userspace/falco/configuration.cpp
leogr: It has been fixed here 
https://github.com/falcosecurity/falco/pull/1667/files#diff-9bd683a06f316f8dab7df7c882f9d1857df11de6ae476356e5f2f86068891157R257

leogr (approved):
/approve

leodido (approved):

PR #1697, opened 2021-07-19
Systemd unit: Do not send falco alerts to /var/log/messages by default

### LINKED ISSUE #1673: falco logs contaminating /var/log/messages
**Describe the bug**
falco writing logs in /var/log/messages and contaminating syslogs.
After having changed the configuration in `/etc/falco/falco.yaml` by disabling `syslog_output`:
```yaml
syslog_output:  
  enabled: false
```
Falco seems to ignore this configuration and continues to write to `/var/log/messages`

**How to reproduce it**
1. add falco repositories to systems `curl -s -o /etc/yum.repos.d/falcosecurity.repo https://falco.org/repo/falcosecurity-rpm.repo`
2. install falco i.e. deb/rpm `yum install falco` (or `apt install falco`)
3. check the '/var/log/messages`

**Expected behaviour**
After disabling `syslog_output` falco should not write logs into /var/log/messages

**Environment**

- Falco version: 0.28.2
- System info:

 ```json
{
  "machine": "x86_64",
  "nodename": "demo-milan.acsia.io",
  "release": "3.10.0-1160.11.1.el7.x86_64",
  "sysname": "Linux",
  "version": "#1 SMP Fri Dec 18 16:34:56 UTC 2020"
}
```
- Cloud provider or hardware configuration: AWS
- OS: CentOS/Debian/Ubuntu/Suse
- Kernel:3.10.0-1160.11.1.el7.x86_64
- Installation method: RPM/DEB from repo

This creates 2 particular issues:  
1. contaminating /var/log/messages (compliance)
2. generating big volume of logs in messages (syslog)

### PR DESCRIPTION
**Pull request by chirabino**
Signed-off-by: Domenico Chirabino <chirabino@protonmail.com>

**What type of PR is this?**
/kind bug

/area engine

**What this PR does**:
This PR set `StandardOutput` to `null` in the systemd unit configuration `falco.service`.

**Why we need it**
After the migration from initd to systemd (#1448), when falco is installed via systemd it writes to `/var/log/messages` by default.

The problem is not present when manually running Falco (without using systemd).

This is due to the default value in the systemd configuration: `DefaultStandardOutput=journal` (which falco inherits), so that Falco standard output is redirected to `/var/log/messages` (see https://manpages.debian.org/wheezy/systemd/systemd.conf.5.en.html).

At the time of writing a workaround is to disable `stdout_output`:
```yaml
stdout_output:
  enabled: false
```

I think that falco should not write to system logs by default. Moreover, asking users to disable `stdout_output` (enabled by default) may be misleading and confusing (ie. I want `stdout_output` enabled so I can start falco manually for debugging or for testing rules)

Fixes #1673

Thanks for your time :)
/cc @leogr 

```release-note
fix(scripts): correct standard output redirection in systemd config
```

### DIFF
--- a/scripts/rpm/falco.service
+++ b/scripts/rpm/falco.service
@@ -19,6 +19,7 @@ ProtectSystem=full
 ProtectKernelTunables=true
 RestrictRealtime=true
 RestrictAddressFamilies=~AF_PACKET
+StandardOutput=null
 
 [Install]
 WantedBy=multi-user.target


leogr (approved):

mstemm:
I can kind of agree that falco logs shouldn't go to /var/log/messages by default. But with this change, where are falco alerts logged when run under systemd?

leogr:
> I can kind of agree that falco logs shouldn't go to /var/log/messages by default. But with this change, where are falco alerts logged when run under systemd?

Short answer: alerts will be still sent to syslog since the `syslog_output` is enabled by default :point_down: 
https://github.com/falcosecurity/falco/blob/7c98d0047cb4e926eb0147bc6142dac9e41c0e12/falco.yaml#L154-L155

TL;DR
Actually, this issue is trying to fix a problem introduced in 0.28.0 by https://github.com/falcosecurity/falco/pull/1448 and does not prevent alerts from being sent to syslog.
Before that version, Falco was shipped with the `initd` configuration. `syslog_output` was the way to deliver alerts. Moreover, standard output was discarded by `initd`.
Starting from 0.28.0, when Falco migrated to `systemd`, the problem was introduced. Basically,  `systemd` redirects (by default) the Falco's standard output to `journal` too (or to other destinations, depending the system configuration), while the `syslog_output` is still enabled. The final result is that we are getting duplicated log lines in `/var/log/messages` (you can find an example in [this comment](https://github.com/falcosecurity/falco/issues/1673#issuecomment-866806974)).
Luckily, the `systemd`'s behavior can be configured (see `StandardOutput=` in https://manpages.debian.org/wheezy/systemd/systemd.exec.5.en.html).
Indeed, this fix just tells `systemd` not to copy the Falco standard output to `/var/log/messages`, avoiding duplicated log lines. Thus, the expected result is to restore the original Falco behavior (pre 0.28.0).

Note: It's been a long time since I've experienced this problem, so that I may have forgotten something. Also, I may be missing something, so make sure my reasons are correct :smile_cat:

mstemm:
Sounds good, as long as there's some output channel still enabled.

mstemm (approved):

FedeDP:
Sorry for necrobumping; does anyone remember why the patch was only applied to rpm service, and not debian too?

leogr:
> Sorry for necrobumping; does anyone remember why the patch was only applied to rpm service, and not debian too? @leogr @chirabino

IIRC correctly, either the Debian default was different (so the issue was not present there) or we just forget it.

Issue #1647, opened 2021-05-07
Expose event source and rule tag(s) in json output


sboschman:
**Motivation**

Forwarding alerts from Falco to Falcosidekick / Falcosidekick-ui would benefit from being able to reason about event source and rule tag(s) in Falcosidekick. Consider the following use cases:
* enrich output fields with Falcosidekick before sending to a "central" Falcosidekick / Falcosidekick-ui with info about the "origin", i.e. host or k8s cluster. Based on the event source Falcosidekick can add the appropiate "origin", e..g a host field is added for syscall events, and a cluster id is added for k8s audit events
* use Falcosidekick outputs based on tags and not only according to priority (brought up by @Issif on falco slack)
* allow to add a routing mechanism based on tags, e.g. sent alerts to different Slack channels based on the alert tag(s) (brought up by @Issif on falco slack)

**Feature**

Add event source (syscall / k8s audit event) and the tag list to the json output (especially the http output used by Falcosidekick)

**Alternatives**

Running many Falco + Falcosidekick instances for each routing scenario. E.g. one Falco+Falcosidekick for only syscall events and one Falco+Falcosidekick for only k8s audit events to be able to 'label' the generated alerts with additional output fields using Falcosidekick. As both will be running as systemd service on the same host it feels a bit hacky.

Issue #1713, opened 2021-08-26
grpc: add rule tags to outputs service


jasondellaluce:
**Motivation**

Currently, the `outputs` grpc service does not support [rule tags](https://falco.org/docs/rules/#tags). Tags can be useful to grpc clients, such as [falco-exporter](https://github.com/falcosecurity/falco-exporter).

**Feature**

The `outputs` grpc service should have a field for rule tags inside the `response` message.

**Additional context**

This has been a `todo` of the grpc API for a while, as visible [here](https://github.com/falcosecurity/falco/blob/f28688551cbf44f12a5f28ab519c0b9f96e46282/userspace/falco/outputs.proto#L53). Also, this issue blocks https://github.com/falcosecurity/falco-exporter/issues/53.

jasondellaluce:
Fixed by: https://github.com/falcosecurity/falco/pull/1714

PR #1714, opened 2021-08-26
new(outputs): expose rule tags and event source in gRPC and json outputs

jasondellaluce:
**What type of PR is this?**

/kind feature

/area engine

1. Adds rule tags as a new field to the gRPC `outputs` service.
2. Adds rule tags and event source fields to the json outputs.
3. Updates the test suite to check that rule tags and event source are correctly retrievable through the gRPC api and through json outputs.

Fixes https://github.com/falcosecurity/falco/issues/1647, https://github.com/falcosecurity/falco/issues/1713

```release-note
new(outputs): expose rule tags and event source in gRPC and json outputs
```


leogr (commented):
Just found a spacing issue (see the comment below), otherwise LGTM! 
Btw, great improvements, thank you! :+1: 

Note for reviewers: since the Lua code has been changed, to avoid a headache, make sure your local Falco's built is pointing to the correct Lua files :smiley_cat:

leogr:
cc @Issif

leogr (approved):
/approve

mstemm (approved):

leodido:
Late to the party...

Have you folks considered making the tags configurable?

The reason they were not there was also that, as far as I remember, the `tags` are not a mandatory part (while `output` is)  of the Falco rule definitions. 

Furthermore, a test covering such a scenario (no tags defined in the rule) would have been helpful to understand what's the resulting json in that case (empty array? other? valid json?).

userspace/falco/outputs.proto
```diff
@@ -50,6 +50,5 @@ message response {
   string output = 5;
   map<string, string> output_fields = 6;
   string hostname = 7;
-  // TODO(leodido,fntlnz): tags not supported yet, keeping it for reference.
-  // repeated string tags = 8;
+  repeated string tags = 8;
```
leodido, line 53: Maybe this should have been `optional` in the protobuf def?

userspace/falco/outputs_grpc.cpp
```diff
@@ -64,5 +64,9 @@ void falco::outputs::output_grpc::output(const message *msg)
 	auto host = grpc_res.mutable_hostname();
 	*host = m_hostname;
 
+	// tags
+	auto tags = grpc_res.mutable_tags();
+	*tags = {msg->tags.begin(), msg->tags.end()};
```
leodido, line 69: Let's check what we end up with when a Falco rule does **not** have tags defined.

Do we obtain `...,"tags":[],...` or not? I'd prefer a clean JSON, but that's my personal preference only :)

leodido (commented):
Great job Jason!

Left two comments

Issif:
I'm really of this PR, when falco 0.30.0 is out, I'll update the format of json in Falcosidekick either 👍 .

userspace/falco/outputs.proto
leogr: AFAIK, a `repeated` field is inherently `optional`.

Moreover, the protobuf compiler does not accept both modifiers simultaneously, for example:

```
optional repeated string tags = 8;
```

will produce a syntax error:

```
optional repeated string tags = 8;
```

Unless I'm missing something, we cannot add `optional` (and we don't need it).

userspace/falco/outputs_grpc.cpp
leogr: I don't have any preferences. Anyway, I agree with @leodido that a test covering such a scenario (especially for the JSON output format) would be really helpful to clarify the expected result.

userspace/falco/outputs.proto
jasondellaluce: I endorse the comment of @leogr. From my understanding, the behavior of `repeated` is quite similar to `optional` already.

userspace/falco/outputs_grpc.cpp
jasondellaluce: Thank you for your review! I opened a PR to address these concerns, so we can eventually move this conversation there: https://github.com/falcosecurity/falco/pull/1733

Issue #1723, opened 2021-09-07
test: output_strictly_contains check leads to non-deterministic test failures


jasondellaluce:
**Which jobs are failing**:
`tests/integration-static`

**Which test(s) are failing**:
`stdout_output_strict`, `stdout_output_json_strict`

**Test link**:
https://github.com/falcosecurity/falco/blob/2a8c0e8bb7ba5ea2d4fa99913257fc32955d718d/test/falco_tests.yaml#L675
https://github.com/falcosecurity/falco/blob/2a8c0e8bb7ba5ea2d4fa99913257fc32955d718d/test/falco_tests.yaml#L686

**Reason for failure**:
The `output_strictly_contains` check ensures that the expected output (read from a file) is contained in the actual Falco's output by using simple string inclusion comparison (`if expected not in actual: self.fail(...)`). As such, the test is _fragile_ and can fail if even a single character in the actual output breaks the pattern of the expected string.

Since merging https://github.com/falcosecurity/falco/pull/1451, alerts are written on the outputs by an additional worker thread. While running integration tests, it may happen that the main thread terminates _before_ the last alert has been emitted by the working thread. As such, Falco's termination message `Events detected: ... Rule counts by severity: ...` may be written on stdout before the last alert.

In this specific scenario, `output_strictly_contains` fails because the termination message appears in the middle of the output, thus not matching the expected string. This only happens while checking the stdout output. This is not deterministically reproducible, and happens in execution environments with poor computation resources.

**Anything else we need to know**:
The `tests/integration-static` job occasionally fails on CircleCI. Failures related to this issue have an error message like the following:
```
Output 'stdout' does not strictly contains the expected content 'output_files/single_rule_with_cat_write.json'
```

An example of this failure is visible in one of the currently open PRs: https://app.circleci.com/pipelines/github/Spartan-65/falco/3/workflows/44b81732-5123-47f5-9642-d56708092bd2/jobs/37/parallel-runs/0/steps/0-103

The log lines containing the corrupted output, and the error, are reported below:
```
04:21:35 DEBUG| [stdout] {"output":"2016-08-04T16:17:57.881781397+0000: Warning An open was seen (command=cat /dev/null)","priority":"Warning","rule":"open_from_cat","time":"2016-08-04T16:17:57.881781397Z", "output_fields": {"evt.time.iso8601":1470327477881781397,"proc.cmdline":"cat /dev/null"}}
[...]

PR #1724, opened 2021-09-08
fix(test): avoid output_strictly_contains failures

jasondellaluce:
Signed-off-by: Jason Dellaluce <jasondellaluce@gmail.com>

**What type of PR is this?**
/kind failing-test

/area tests

This fixes non-deterministic test failures related to the `output_strictly_contains` check.

`output_strictly_contains` now ensures that the actual output contains _every line_ of the expected output. This offers less guarantees than checking that the expected output is perfectly contained in the actual one, but is more fault tolerant and does not alter the semantics of our tests.

https://github.com/falcosecurity/falco/issues/1723

```release-note
NONE
```

### DIFF
--- a/test/falco_test.py
+++ b/test/falco_test.py
@@ -409,10 +409,12 @@ class FalcoTest(Test):
             else:
                 actual = open(output['actual']).read()
 
-            if expected not in actual:
-                self.fail("Output '{}' does not strictly contains the expected content '{}'".format(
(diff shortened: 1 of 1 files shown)


leogr (approved):
/approve

mstemm (approved):
Thank you for tracking down the problem! This fix seems like a reasonable trade-off.

PR #1733, opened 2021-09-23
update(outputs): add configuration option for tags in json outputs

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind feature

/area engine

/area tests

Since the recent merge of https://github.com/falcosecurity/falco/pull/1714, `tags` and `source` fields are now part of both the grpc and json outputs. However, the semantic of the json output is not clear in some corner cases, such as when no tags are defined in a rule. Also, users can't configure whether to include `tags` in the json outputs or not, which instead is possible for the `output` field. Moreover, no test cases are yet present for asserting the expected outcome of these corner cases. These issues were reported in https://github.com/falcosecurity/falco/pull/1714#issuecomment-924800821 by @leodido (thank you!!)

As such, this PR contributes with the following:

- Adding the `json_include_tags_property` configuration option, so that users can decide whether to include `tags` in the json outputs or not. By default this is set to `true` to not introduce breaking changes for clients that currently expect the field to be present.
- Clarifying the expected outcome in the corner cases arising when a rule has no defined tags.
- Adding test cases asserting the expected outcomes of the corner cases, and updating the test suite to support the new configuration option.

**Additional Context**:
The table below describes the expected outcomes in the json outputs by using the new configuration option.
|   |  ` json_include_tags_property = true`  |  ` json_include_tags_property = false` |
|---|---|---|
| **Rule with no tags**  |  **(1)** json has a field `tags=[]` |  **(2)** json has no `tags` field  |
| **Rule with tags**  |  **(3)** json has a field `tags=[...]`  |  **(4)** json has no `tags` field  |

These cases are covered by the following tests:

- `json_output_empty_tags_property`: Case **(1)**
- `json_output_no_tags_property`: Cases **(2)** and **(4)**
- `stdout_output_json_strict`: Case **(3)**

```release-note
update(outputs): add configuration option for tags in json outputs
```

### DIFF
--- a/test/falco_tests.yaml
+++ b/test/falco_tests.yaml
@@ -1111,6 +1111,25 @@ trace_files: !mux
     trace_file: trace_files/cat_write.scap
     stdout_contains: "^(?!.*Warning An open of /dev/null was seen.*)"
 
+  json_output_no_tags_property:
+    json_output: True
+    json_include_tags_property: False
+    detect: True
+    detect_level: WARNING
+    rules_file:
+      - rules/rule_append.yaml
+    trace_file: trace_files/cat_write.scap
+    stdout_contains: "^(?!.*\"tags\":[ ]*\\[.*\\],.*)"
+
+  json_output_empty_tags_property:
+    json_output: True
+    detect: True
+    detect_level: WARNING
+    rules_file:
+      - rules/rule_append.yaml
+    trace_file: trace_files/cat_write.scap
+    stdout_contains: "^(.*\"tags\":[ ]*\\[\\],.*)"
+
   in_operator_netmasks:
     detect: True
     detect_level: INFO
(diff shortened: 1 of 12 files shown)


test/falco_tests.yaml
```diff
@@ -1111,6 +1111,25 @@ trace_files: !mux
...
+
+  json_output_empty_tags_property:
+    json_output: True
+    detect: True
+    detect_level: WARNING
+    rules_file:
+      - rules/rule_append.yaml
+    trace_file: trace_files/cat_write.scap
+    stdout_contains: "^(.*\"tags\":[ ]*\\[\\],.*)"
+
```
leogr, line 1132: :+1:

leogr (approved):
LGTM.
Great improvement :+1:

leodido (approved):
Thank you Jason for implementing my feedback. 🤗 

/LGTM

PR #1738, opened 2021-09-28
update(proposal): clarify that old drivers are not removed anymore

leodido:

**What type of PR is this?**

/kind cleanup

/area proposals

Clarify that the `drivers/cleanup` job we needed when we were using Bintray to store the Falco drivers is not needed anymore. And that, its related proposal has been superseded by a newer one (already implemented and in place during the last year).

NONE

/milestone 0.30.0

```release-note
docs: clarify that previous Falco drivers will remain available at https://download.falco.org and no automated cleanup is run anymore
```

### DIFF
--- a/proposals/20200901-artifacts-cleanup.md
+++ b/proposals/20200901-artifacts-cleanup.md
@@ -2,6 +2,8 @@
 
 This document reflects when and how we clean up the Falco artifacts from their storage location.
 
+**Superseeded by**: [drivers-storage-s3 proposal](https://github.com/falcosecurity/falco/blob/master/proposals/20201025-drivers-storage-s3.md).
+
 ## Motivation
 
 The [bintray](https://bintray.com/falcosecurity) open-source plan offers 10GB free space for storing artifacts.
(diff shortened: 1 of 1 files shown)


leogr (approved):
Thanks for the clarification.

kris-nova:
Can you please add more detail about what the expectation is to a user?

Will Falco break? If so when should we expect it to break after this change? What will we expect Falco to do if/when it breaks? Does Falco now have a log line that will reference this change that is printed to `stdout` or `stderr` once the break has occured?

I do not have any issues with the change, however I do not think this does a good job at explaining what might go wrong after this change takes place. Maybe I am missing this documentation somewhere else? My main concern is that people will be affected by this, and we will see GitHub issues like "Falco unable to find driver on <os> and <kernel>" and it will take one of us a large amount of time to connect these new issues back to this change. 

We should do what we can to make our end-users as self sufficient as possible, so that they do not need a maintainer to help them understand why Falco stopped working.

leodido:
> Can you please add more detail about what the expectation is to a user?
> 
> Will Falco break? If so when should we expect it to break after this change? What will we expect Falco to do if/when it breaks? Does Falco now have a log line that will reference this change that is printed to `stdout` or `stderr` once the break has occured?
> 
> I do not have any issues with the change, however I do not think this does a good job at explaining what might go wrong after this change takes place. Maybe I am missing this documentation somewhere else? My main concern is that people will be affected by this, and we will see GitHub issues like "Falco unable to find driver on {os} and {kernel}" and it will take one of us a large amount of time to connect these new issues back to this change.
> 
> We should do what we can to make our end-users as self sufficient as possible, so that they do not need a maintainer to help them understand why Falco stopped working.

These edits just reflects changes in place for all the past year, Kris.

In fact, it is just a documentation change about the fact that we store drivers at https://download.falco.org and we do not need anymore to delete the previous Falco drivers (like we had to do in the past) because we are free from storage limits and constraints.

There is nothing that can go wrong since this just updates an almost 2 years old proposal for reflecting the new process that has been in place for 6 releases already.

leogr:
> Can you please add more detail about what the expectation is to a user?

Nothing will change. This is just a further clarification of what is already in place.
 
> Will Falco break? If so when should we expect it to break after this change? What will we expect Falco to do if/when it breaks? Does Falco now have a log line that will reference this change that is printed to `stdout` or `stderr` once the break has occurred?

No, Falco will not break, because nothing is changing here.

> I do not have any issues with the change, however I do not think this does a good job at explaining what might go wrong after this change takes place. Maybe I am missing this documentation somewhere else? My main concern is that people will be affected by this, and we will see GitHub issues like "Falco unable to find driver on {os} and {kernel}" and it will take one of us a large amount of time to connect these new issues back to this change.

This PR clarifies more that we are keeping all previous prebuilts on S3, available for download at https://download.falco.org/. So we should not see that kind of issues :) 

> We should do what we can to make our end-users as self sufficient as possible, so that they do not need a maintainer to help them understand why Falco stopped working.

That's exactly what we did when decided to do NOT delete drivers from S3 anymore. So, again I just want to ensure you that this PR it's just a documentation change to clarify that we are already going to the right direction :)

mstemm (approved):

fntlnz (approved):

PR #1768, opened 2021-11-02
Allow append of new exceptions to rules

sai-arigeli:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> If contributing rules or changes to rules, please make sure to also uncomment one of the following line:

> /kind rule-update

> /kind rule-create

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area build

> /area engine

/area rules

> /area tests

> /area proposals

**What this PR does / why we need it**:
PR introduces appending of new fully valid exception objects to existing rules. 
**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #1775, opened 2021-11-08
fix(userspace/engine): supporting enabled-only overwritten rules

### LINKED ISSUE #1510: Cannot disable rule with append and enabled
**Describe the bug**
Cannot disable rule using documented way, e.g:
``` yaml
- rule: Change thread namespace
  append: true
  enabled: false
```
logs:
```
DKMS: install completed.
* falco module installed in dkms, trying to insmod
* Success: falco module found and loaded in dkms
Mon Dec 14 20:19:49 2020: Falco version 0.26.2 (driver version 2aa88dcf6243982697811df4c1b484bcbe9488a2)
Mon Dec 14 20:19:49 2020: Falco initialized with configuration file /etc/falco/falco.yaml
Mon Dec 14 20:19:49 2020: Loading rules from file /etc/falco/falco_rules.yaml:
Mon Dec 14 20:19:50 2020: Loading rules from file /etc/falco/falco_rules.local.yaml:
Mon Dec 14 20:19:51 2020: Loading rules from file /etc/falco/rules.d/custom.local.yaml:
Mon Dec 14 20:19:51 2020: Runtime error: Rule must have property condition
---
- rule: Change thread namespace
  append: true
  enabled: false
---. Exiting.
```
where the custom rules are located in custom.local.yaml and are properly loaded after the falco_rules.yaml file.

It will however work if using the following:
``` yaml
- rule: Change thread namespace
  append: true
  condition: and (never_true)
```
suggesting that the appending is probably working properly and file is loaded in proper order.

**How to reproduce it**
Deploy falco using falco chart with the following values
``` yaml
customRules:
  custom.local.yaml: |-
    - rule: Change thread namespace
      append: true     
      enabled: false
```

**Expected behaviour**

Disables the falco rule

**Environment**

- Falco version:
0.26.2
- System info:
```
Mon Dec 14 20:38:02 2020: Falco version 0.26.2 (driver version 2aa88dcf6243982697811df4c1b484bcbe9488a2)
Mon Dec 14 20:38:02 2020: Falco initialized with configuration file /etc/falco/falco.yaml
Mon Dec 14 20:38:02 2020: Loading rules from file /etc/falco/falco_rules.yaml:
Mon Dec 14 20:38:03 2020: Loading rules from file /etc/falco/falco_rules.local.yaml:
Mon Dec 14 20:38:04 2020: Loading rules from file /etc/falco/rules.d/custom.local.yaml:
{
  "machine": "x86_64",
  "nodename": "falco-swwzp",
  "release": "5.4.0-1032-azure",
  "sysname": "Linux",
  "version": "#33~18.04.1-Ubuntu SMP Tue Nov 17 11:40:52 UTC 2020"
}

```
Azure AKS

### LINKED ISSUE #1537: Disabling a rule by using "append" does not work as expected
**Describe the bug**

Disabling a rule should be possible by just appending `enabled: false` as per the [official documentation](https://falco.org/docs/rules/#via-custom-rule-definition).

Unfortunately, it does not work.

**How to reproduce it**

Add a file under `/etc/falco/rules.d/custom-rules.yaml` with the following content:
```yaml
- rule: Unexpected UDP Traffic
  append: true
  enabled: false
```
Finally, run falco, and you get:

```
Tue Jan 26 15:24:39 2021: Falco version 0.27.0 (driver version 5c0b863ddade7a45568c0ac97d037422c9efb750)
Tue Jan 26 15:24:39 2021: Falco initialized with configuration file /home/leogr/code/github.com/falcosecurity/falco/falco.yaml
Tue Jan 26 15:24:39 2021: Loading rules from file /etc/falco/falco_rules.yaml:
Tue Jan 26 15:24:39 2021: Loading rules from file /etc/falco/falco_rules.local.yaml:
Tue Jan 26 15:24:39 2021: Loading rules from file /etc/falco/k8s_audit_rules.yaml:
Tue Jan 26 15:24:39 2021: Loading rules from file /etc/falco/rules.d/custom-rules.yaml:
Tue Jan 26 15:24:39 2021: Runtime error: Rule must have property condition
---
- rule: Unexpected UDP Traffic
  append: true
  enabled: false
---. Exiting.
```
N.B: rule files were loaded in the correct order (that's important to ensure that `append` works).

**Expected behaviour**

No error and the rule being disabled as suggested by the official documentation [here](https://falco.org/docs/rules/#via-custom-rule-definition).

**Environment**

- Falco version: 0.27.0 (previous versions seem to be affected too)

- System info:

- OS:

- Kernel:

- Installation method:

**Additional context**

Disabling a rule seemed to work, as discussed in the past (see https://github.com/falcosecurity/falco/issues/1469 ).

However, as noticed by a user in the slack channel (original discussion [here](https://kubernetes.slack.com/archives/CMWH3EH32/p1611648695038300)) it doesn't work. 

**Considerations**
Firstly, the rule parser implements a [too strict condition on the existence of the `condition` property](https://github.com/falcosecurity/falco/blob/0.27.0/userspace/engine/lua/rule_loader.lua#L396-L403) that actually disallows this possibility. Secondly, after a quick look, it seems to me that implementation of the `enabled` flag would not work in appending mode even after removing the check on the `condition` property. Further investigation is needed.

**Workaround**
It is still possible to disable a rule using the `never_true` macro, for example:
```yaml
- rule: Unexpected UDP Traffic
  condition: and (never_true)
  append: true
```

jasondellaluce:
Signed-off-by: Jason Dellaluce <jasondellaluce@gmail.com>
Co-authored-by: Leonardo Grasso <me@leonardograsso.com>

**What type of PR is this?**

/kind bug

/area engine

This enables overwriting rules by specifying their name and the `enabled` field only, so that users can easily enable/disable already-defined rules. The expected behavior is to be able to overwrite rules like the following (example from @leogr):
```
- rule: A rule enabled by default
  enabled: false
```

The following issues currently prevent this from being supported:
- Every rule is forced to either specify a `condition` or an `exception` field. However, this conceptually applies only to rules having `append` set as `true`.
- Overwritten rules are entirely substituted to their already-defined counterpart. This is correct, however it does not support changing the `enabled` flag only, which is a feature demanded by the community.

Proof of this ambiguity can be seen in the current version of `falco-website`, which reports a non-working example: https://github.com/falcosecurity/falco-website/blob/4a7b5eb82027da5dadd50dc688c7ebd55c24bd9e/content/en/docs/rules/_index.md?plain=1#L263

Fixes #1537 
Fixes #1510

See https://github.com/falcosecurity/falco/issues/1537 for additional context.

```release-note
fix(userspace/engine): supporting enabled-only overwritten rules
```

### DIFF
--- a/test/falco_tests.yaml
+++ b/test/falco_tests.yaml
@@ -354,7 +354,7 @@ trace_files: !mux
     exit_status: 1
     stdout_is: |+
       1 errors:
-      Rule must have exceptions or condition property
+      Appended rule must have exceptions or condition property
       ---
       - rule: no condition rule
         append: true
--- a/test/falco_tests.yaml
+++ b/test/falco_tests.yaml
@@ -634,7 +634,13 @@ trace_files: !mux
     rules_file:
       - rules/single_rule_enabled_flag.yaml
     trace_file: trace_files/cat_write.scap
-
+  
+  disabled_rule_using_false_enabled_flag_only:
+    detect: False
+    rules_file:
+      - rules/disabled_rule_using_enabled_flag_only.yaml
+    trace_file: trace_files/cat_write.scap
+  
   disabled_and_enabled_rules_1:
     exit_status: 1
     stderr_contains: "Runtime error: You can not specify both disabled .-D/-T. and enabled .-t. rules. Exiting."
--- a/test/falco_tests.yaml
+++ b/test/falco_tests.yaml
@@ -641,6 +641,14 @@ trace_files: !mux
       - rules/disabled_rule_using_enabled_flag_only.yaml
     trace_file: trace_files/cat_write.scap
   
+  enabled_rule_using_false_enabled_flag_only:
+    detect: True
+    detect_level: WARNING
+    rules_file:
+      - rules/enabled_rule_using_enabled_flag_only.yaml
+    trace_file: trace_files/cat_write.scap
+    stdout_contains: "Warning An open was seen"
+
   disabled_and_enabled_rules_1:
     exit_status: 1
     stderr_contains: "Runtime error: You can not specify both disabled .-D/-T. and enabled .-t. rules. Exiting."
--- a/test/rules/disabled_rule_using_enabled_flag_only.yaml
+++ b/test/rules/disabled_rule_using_enabled_flag_only.yaml
@@ -1,0 +1,24 @@
+#
+# Copyright (C) 2021 The Falco Authors.
+#
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+- rule: open_from_cat
+  desc: A process named cat does an open
+  condition: evt.type=open and proc.name=cat
+  output: "An open was seen"
+  priority: WARNING
+
+- rule: open_from_cat
+  enabled: false
--- a/test/rules/enabled_rule_using_enabled_flag_only.yaml
+++ b/test/rules/enabled_rule_using_enabled_flag_only.yaml
@@ -1,0 +1,25 @@
+#
+# Copyright (C) 2021 The Falco Authors.
+#
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+- rule: open_from_cat
+  desc: A process named cat does an open
+  condition: evt.type=open and proc.name=cat
+  output: "An open was seen"
+  priority: WARNING
+  enabled: false
+
+- rule: open_from_cat
+  enabled: true
--- a/test/rules/invalid_append_rule_without_condition.yaml
+++ b/test/rules/invalid_append_rule_without_condition.yaml
@@ -1,2 +1,24 @@
+#
+# Copyright (C) 2021 The Falco Authors.
+#
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+- rule: no condition rule
+  desc: simpe rule 
+  condition: evt.type=open
+  output: simple output
+  priority: WARNING
+
 - rule: no condition rule
   append: true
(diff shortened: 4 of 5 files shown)


jasondellaluce:
The test `invalid_append_rule_without_condition` asserted that rules with `append` set as `true` contained a `condition` field. However, according to our documentation, this is not a requirement. This is explicitly stated here 👇🏼 
https://falco.org/docs/rules/#via-custom-rule-definition

Accordingly, this PR removes the `invalid_append_rule_without_condition` test as it fails with the latest updates. Also, rules can now specify omit the `condition` and `exception` fields if they have `append` as `true`.

@mstemm @leogr

mstemm:
About exceptions, the docs may not state it, but you're not supposed to be able to append exceptions at the moment, only conditions. @sai-arigeli is working on a separate PR to change that: https://github.com/falcosecurity/falco/pull/1768.

And about enabled, I'm not sure we want disable a rule by appending to it with a different enabled value. Is there a compelling use case for this? Why wouldn't redefining the original rule work?

leogr:
Hey @mstemm 

> And about enabled, I'm not sure we want disable a rule by appending to it with a different enabled value. Is there a compelling use case for this? Why wouldn't redefining the original rule work?

Currently, it works neither :point_down: 

_Redefine_
```
Wed Nov 10 08:26:22 2021: Runtime error: Could not load rules file /etc/falco/falco_rules.local.yaml: 1 errors:                                                                                           
Rule must have exceptions or condition property                                                                                                                                                           
---                                                                                                                                                                                                       
- rule: Unexpected inbound connection source                                                                                                                                                              
  enabled: false                                                                                                                                                                                          
---                                            
. Exiting.
```

_Append_
```
Wed Nov 10 08:27:23 2021: Runtime error: Could not load rules file /etc/falco/falco_rules.local.yaml: 1 errors:
Rule must have exceptions or condition property
---
- rule: Unexpected inbound connection source
  append: true
  enabled: false
---
. Exiting.
```

I guess the reason why @jasondellaluce tried to fix the "append" case is that the official documentation advice it: https://falco.org/docs/rules/#via-custom-rule-definition 
(that's currently broken; indeed, we had to add a note with an alternative)

I would correct both cases. In my opinion, `enabled: false` should work both in redefining and appending cases since they are semantically equivalent.

In general, `enabled` needs to be fixed because `enabled` is currently not really usable.

jasondellaluce:
I don't think this strictly overlaps with #1768, as this is not correlated to rule exceptions. In that perspective, I only weakened the requirement of needing a condition or an exception in appended rules, in order to allow users to append rules defining only the `enabled` field:
```
- rule: Unexpected inbound connection source
  append: true
  enabled: false
```
Users raised many issues about this (such as #1510 and #1537), since we indicate this as feasible in our doc (as @leogr said).

> I would correct both cases. In my opinion, enabled: false should work both in redefining and appending cases since they are semantically equivalent.

I agree with this. Some additional work is required for this, so I'm gonna wait until we are on the same page on this discussion before pushing some additional commit.

leogr:
After having discussed this privately, I'm ok with fixing only the redefinition case.

Basically, what this PR should do is:

1. Allow to disable a rule by redefining it:
```
- rule: A rule enabled by default
  enabled: false
```

2. Allow to enable a rule by redefining it:
```
- rule: A rule disabled by default
  enabled: true
```

The `append: true` behavior will continue to be as-is now.
Consequently, we will have to update [this piece of documentation](https://falco.org/docs/rules/#via-custom-rule-definition) too.


@mstemm do you agree?

mstemm:
Currently the rule loading code doesn't handle cases where append is false, and the only thing in the second rule definition is an enabled flag. Currently, if a rule/macro has append false, everything in the prior rule is discarded and only the properties in the new rule are used instead.

So this would introduce a special case where a rule with append false and *only* an enabled flag changes the enabled value from the original rule, but otherwise everything in the original rule is used instead?

I guess this is okay, but to me this seems like a problem where the docs are wrong, but we're changing the code to match the docs rather than fixing the docs.

leogr:
> I guess this is okay, but to me this seems like a problem where the docs are wrong, but we're changing the code to match the docs rather than fixing the docs.

@mstemm The real point here is that users are asking to disable/enable a rule **without modifying the default ruleset file** (that is a very common use case) and we should provide a way to do that. The docs might be wrong in the how, but they are perfectly correct w.r.t. the need. For these reasons, my last proposal suggests fixing both the code and the docs.

jasondellaluce:
I just refactored the PR trying to implement @leogr's proposal. For clarity, I also added two more integration tests that assert the expected behavior wit `enabled`-only rules. Also, the `invalid_append_rule_without_condition` test has been restored and adapted to the new changes. Let me know what you think.

leogr (approved):
LGTM now

/approve

mstemm (approved):

mstemm:
Ok, this seems like a reasonable way to handle this use case. Approved!

Issue #1654, opened 2021-05-14
Build failure when specifying version


shane-lawrence:
**Describe the bug**

I am unable to build falco when I specify a custom version with `-DFALCO_VERSION`. cmake runs successfully but make fails. If I don't specify FALCO_VERSION then the build is successful.

**How to reproduce it**

```
git clone https://github.com/falcosecurity/falco.git
cd falco/
ls
git checkout 0.28.1
mkdir build
cd build
cmake -DBUILD_BPF=Off -DBUILD_DRIVER=Off -DCMAKE_BUILD_TYPE=Release -DUSE_BUNDLED_DEPS=On -DFALCO_VERSION=0.28.1-shane1 ..
make -j4 falco
```

**Expected behaviour**

The build should complete successfully.

**Screenshots**
```
[ 96%] Building CXX object userspace/falco/CMakeFiles/falco.dir/grpc_request_context.cpp.o
/media/ramdisk/falco/userspace/falco/grpc_server_impl.cpp: In member function ‘void falco::grpc::server_impl::version(const falco::grpc::context&, const falco::version::request&, falco::version::response&)’:
/media/ramdisk/falco/userspace/falco/grpc_server_impl.cpp:82:35: error: no matching function for call to ‘falco::version::response::set_major()’
  res.set_major(FALCO_VERSION_MAJOR);
                                   ^
In file included from /media/ramdisk/falco/build/userspace/falco/version.grpc.pb.h:23,
                 from /media/ramdisk/falco/userspace/falco/grpc_server_impl.h:21,
                 from /media/ramdisk/falco/userspace/falco/grpc_server_impl.cpp:19:
/media/ramdisk/falco/build/userspace/falco/version.pb.h:532:13: note: candidate: ‘void falco::version::response::set_major(google::protobuf::uint32)’
 inline void response::set_major(::PROTOBUF_NAMESPACE_ID::uint32 value) {
             ^~~~~~~~
/media/ramdisk/falco/build/userspace/falco/version.pb.h:532:13: note:   candidate expects 1 argument, 0 provided
/media/ramdisk/falco/userspace/falco/grpc_server_impl.cpp:83:35: error: no matching function for call to ‘falco::version::response::set_minor()’
  res.set_minor(FALCO_VERSION_MINOR);
                                   ^
In file included from /media/ramdisk/falco/build/userspace/falco/version.grpc.pb.h:23,
[...]

PR #1778, opened 2021-11-09
build: always populate partial version variables

dnwe:
**What type of PR is this?**

/kind bug

/area build

When FALCO_VERSION was provided via a CMake variable, the build would
eventually fail because the partial version variables hadn't been
populated. Move the creation of those outside the check of FALCO_VERSION
being set so they also happen when that is provided too.

Fixes: #1654

```release-note
NONE
```


dnwe:
@jasondellaluce sure. I originally put it just as Contributes-to as I was waiting for the OP to confirm it worked for them too. Have updated to Fixes now I’ve had another confirm though.

leogr (approved):
LGTM, thank you!

/approve

mstemm (approved):

PR #1792, opened 2021-11-17
support arbitrary-depth nested values in YAML configuration

jasondellaluce:
**What type of PR is this?**

/kind design

/kind feature

/area engine

/area tests

Falco loads its configuration from a YAML file. However, we currently support fields with a maximum depth of 2. Fields nested after the second level are not reachable programmatically. This is a limiter for future features and contributions. At the same time, the current code only loads the configuration from file, which makes it impossible to write unit tests for the configuration loader.

This PR introduces a representation-agnostic notation to access configuration fields, independently from their level of nesting and from their underlying representation (YAML or others). Accordingly, the `yaml_configuration` class has been slightly refactored to address this changes.

Unit tests have been implement to assert that everything works as expected.

The newly proposed syntax for accessing config fields resembles the way nested objects are accessed in JSON. Consider the following YAML file:
```yaml
base:
  id: 1
  subvalue:
    subvalue2:
      name: 'test'
  list:
    - elem1
    - elem2
```
Each field can be accessed as:
```
base.id
base.subvalue.subvalue2.name
base.list[0]
base.list[1]
```

```release-note
update(userspace/falco): support arbitrary-depth nested values in YAML configuration
```

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -20,7 +20,7 @@ else()
   set(FALCO_TESTS_SOURCES test_base.cpp engine/test_token_bucket.cpp engine/test_rulesets.cpp engine/test_falco_utils.cpp falco/test_webserver.cpp)
 endif()
 
-set(FALCO_TESTED_LIBRARIES falco_engine)
+set(FALCO_TESTED_LIBRARIES falco_engine yaml-cpp)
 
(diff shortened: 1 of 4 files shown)


leogr (approved):
/approve

whoa! :clap:

mstemm (approved):
Looks like a valuable change. It's very similar to json pointers, but for yaml!

PR #1793, opened 2021-11-18
Allow overwriting existing macros with conditions defined after them

### LINKED ISSUE #706: Validation Error when using a macro as definition of an overwritten macro.
**What happened / Replication Steps**:

Given a macro that is defined in the default ruleset and for the purposes of being concrete let's use [`Clear Log Activities`](https://github.com/falcosecurity/falco/blob/dev/rules/falco_rules.yaml#L2341-L2342):

```
- macro: allowed_clear_log_files
  condition: (never_true)
```

1. Now, if you overwrite this macro in [`falco_rules.local.yaml`](https://github.com/falcosecurity/falco/blob/dev/rules/falco_rules.local.yaml), for example:

```
- macro: allowed_clear_log_files
  condition: >
    (container.name = "I am testing something")
```

When I run my validation script: 
```
falco -c /etc/falco/falco.yml.erb               \
--validate /etc/falco/falco_rules.yml.erb        \                           
--validate /etc/falco/falco_macros.local.yml.erb  \                
```

Everything succeeds 🎉

2. However, if I overwrite this macro in [`falco_rules.local.yaml`](https://github.com/falcosecurity/falco/blob/dev/rules/falco_rules.local.yaml) using another macro, for example:

```
- macro: allowed_clear_log_files
  condition: >
    consider_all_chmods
``` 

I get the following error message: 
`Runtime error: Error loading rules: /usr/share/falco/lua/compiler.lua:65: Undefined macro 'consider_all_chmods' used in filter.. Exiting.`

(It is important to note that `consider_all_chmods` is in defined in the default ruleset, so I am not sure why it is failing here.

**Anything else we need to know?**:

**Environment**:
- Falco version (use `falco --version`): falco 0.14.0
- System info 
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools (e.g. in kubernetes, rpm, deb, from source):
- Others:

wdoekes:
Thanks to @fntlnz for providing a minimal working example.

Now allowed:

    - macro: something_happens
      condition: (evt.num=0)

    - macro: something_else
      condition: (evt.num=1)

    # This -- usually seen in a local file -- was previously not
    # allowed, because during resolution of the /first/ something_happens,
    # the /then/ unknown something_else would be encountered.
    - macro: something_happens
      condition: (evt.num=2) or (something_else)

/kind bug

/area rules

Fixes #706

```release-note
fix: allow overwriting existing macros with conditions defined after them
```

--

P.S. Unrelated to this PR. When I opened a PR, I got these tips:

1. If this is your first time, please read our contributor guidelines in the github.com/falcosecurity/.github/blob/master/CONTRIBUTING.md file and learn how to compile Falco from source falco.org/docs/source

If I wanted tips on how to compile Falco, I think I wouldn't be at the pull request stage. So that's weird. Also, the mentioned URL gives a 404.

The tip that _isn't_ mentioned, but is found in the CONTRIBUTING docs, is the `-s` signed-off that is apparently mandatory. Maybe put that here.

### DIFF
--- a/userspace/engine/lua/rule_loader.lua
+++ b/userspace/engine/lua/rule_loader.lua
@@ -811,6 +811,46 @@ function build_exception_condition_string_single_field(eitem, exfields)
 
 end
 
+-- Process macros. Broken off from the big load_rules() function for readability.
+-- Tries loading all the macros, pushing back any failed ones to a next try.
+-- Stops and fails first when a run fails without completing any resolution.
+-- Returns: an error or nil on success
+local function load_rule_macros(s)
+   -- Initialize macros to resolve.
+   local idx
+   local macros_to_try = {}
+   for idx = 1, #s.ordered_macro_names do
+      macros_to_try[idx] = s.ordered_macro_names[idx]
+   end
+
+   -- v and ast are used when reporting failure.
+   local status, v, ast
+   local did_anything = true
+   repeat
+      local macros_to_retry = {}
+      did_anything = false
+      for idx = 1, #macros_to_try do
+	 local name = macros_to_try[idx]
+	 v = s.macros_by_name[name]
+	 status, ast = compiler.compile_macro(v['condition'], s.macros, s.lists)
+	 if status == false then
+	    macros_to_retry[#macros_to_retry + 1] = name
+	 else
+	    s.macros[v['macro']] = {["ast"] = ast.filter.value, ["used"] = false}
+	    did_anything = true
+	 end
+      end
+      macros_to_try = macros_to_retry
+   until #macros_to_try == 0 or did_anything == false
+
+   -- Work left to do?
+   if #macros_to_try ~= 0 then
+      return build_error_with_context(v['context'], ast)
+   end
+
+   return nil
+end
+
 -- Returns:
 -- - Load Result: bool
 -- - required engine version. will be nil when load result is false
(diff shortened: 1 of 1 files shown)


jasondellaluce:
This is a great step forward. I haven't tested this, but my only concern is that this would also allow cases like this to be accepted:
```
- macro: something_happens
  condition: (something_else)

- macro: something_else
  condition: (evt.num=1)
```
Which kind brokes the ordering semantics. Did you test this specific case?

leogr:
Hey @wdoekes 

Thank you for this PR! 
I've just updated the PR description since we need the release-note block. Also, I added the release note text, since this PR is actually fixing an issue.

wdoekes:
> This is a great step forward. I haven't tested this, but my only concern is that this would also allow cases like this to be accepted:

Yes. It accepts that case. I haven't tested it, but that's what the code (also) implies.

Is there a reason why we would not want that?

If there is, then the fix would've been to simply insert updated macros at the end of ordered_macros when loading them from yaml. Apparently there is a reason why you want to keep some kind of ordering. But I'm in the dark about what reasons you have. Keeping the original order _whilst_ allowing the bugfix, but not your suggested case, seems to me like trying to fit a round peg through a square hole.

leogr:
Although I like this fix, I believe we have to rethink the whole sematic a bit. So, I'm moving it to the next milestone
/milestone 0.32.0

jasondellaluce:
Hey @wdoekes, I was finally able to work on a refactoring of the rule loader 👉🏼 https://github.com/falcosecurity/falco/pull/1966
Right now, all that is done is porting the existing loader logic from Lua to C++, but some bugs have been fixed in the meanwhile, including the one you're tackling here I think. It would be great if you could take a look!

leogr:
I'm closing this PR in favor of #1966 (that's already merged).
/close

wdoekes:
> I was finally able to work on a refactoring of the rule loader 👉🏼 https://github.com/falcosecurity/falco/pull/1966

Okay. I was finally able to test it. It works :+1:

PR #1800, opened 2021-11-19
fix(userspace/falco): accept 'Content-Type' header that contains "application/json", but it is not strictly equal to it

### LINKED ISSUE #1702: Forwarding audit logs to falco from an AKS cluster using sysdiglabs/aks-audit-log
**Bug**

When trying to forward AKS audit logs to a falco instance using the setup from [sysdiglabs/aks-audit-log](https://github.com/sysdiglabs/aks-audit-log/blob/master/docs/readme-dev.md) the logs indicate all POST requests fail with a 400 code.

`**Error sending POST, retry 1, result: [BadRequest] System.Net.Http.HttpConnectionResponseContent`

**Problem**

The issue is that the falco webserver only accepts requests with the header `application/json` while the audit log forwarder sends requests with the header `application/json; charset=utf-8`. [See webserver.cpp](https://github.com/falcosecurity/falco/blob/46425b392cc949b78d754c71875ad9a9c5121106/userspace/falco/webserver.cpp#L151)

The webserver.cpp file should be modified to accept encoding values in the request headers.

Related issue : https://github.com/sysdiglabs/aks-audit-log/issues/4

FedeDP:
Signed-off-by: Federico Di Pierro <nierro92@gmail.com>

**What type of PR is this?**

/kind bug

/area engine

Do not strictly require "application/json" content-type string in k8s_audit_handler::handlePost(); instead, accept any Content-Type that has "application/json" in it.

Fixes #1702 

```release-note
fix(userspace/falco): accept 'Content-Type' header that contains "application/json", but it is not strictly equal to it
```

### DIFF
--- a/userspace/falco/webserver.cpp
+++ b/userspace/falco/webserver.cpp
@@ -150,7 +150,7 @@ bool k8s_audit_handler::handlePost(CivetServer *server, struct mg_connection *co
 	// Ensure that the content-type is application/json
 	const char *ct = server->getHeader(conn, string("Content-Type"));
 
-	if(ct == NULL || string(ct) != "application/json")
+	if(ct == NULL || strstr(ct, "application/json") == NULL)
 	{
 		mg_send_http_error(conn, 400, "Wrong Content Type");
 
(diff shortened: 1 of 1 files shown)


leogr (approved):
LGTM

/approve

mstemm (approved):
I believe this will only detect prefixes, right? So I would suggest updating the release note but if a prefix check is sufficient, then lgtm.

PR #1810, opened 2021-11-29
Bypass research issues

sberkovich:
**What type of PR is this?**

/kind rule-update
/kind rule-create

/area rules

This PR is a result of the bypass research published [~~here~~](https://github.com/blackberry/Falco-bypasses) (for now kept private). The research introduced several new bypass techniques and expanded on several already-known bypasses. 

This PR touches five existing rules and introduces one new rule.

This PR does not cover the bypasses found in the original research fully. Rather it follows a low-hanging fruit approach and focuses on syntactical fixes that do not require structural changes to the rule engine or kernel probes.

```release-note
rule(Create Hardlink Over Sensitive Files): new rule to detect hard links created over sensitive files
rule(Detect crypto miners using the Stratum protocol): add `stratum2+tcp` and `stratum+ssl` protocols detection
rule(Sudo Potential Privilege Escalation): correct special case for the CVE-2021-3156 exploit
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -2820,7 +2820,7 @@
 
 - rule: Detect crypto miners using the Stratum protocol
   desc: Miners typically specify the mining pool to connect to with a URI that begins with 'stratum+tcp'
-  condition: spawned_process and proc.cmdline contains "stratum+tcp"
+  condition: spawned_process and (proc.cmdline contains "stratum+tcp" or proc.cmdline contains "stratum2+tcp" or proc.cmdline contains "stratum+ssl" or proc.cmdline contains "stratum2+ssl")
   output: Possible miner running (command=%proc.cmdline container=%container.info image=%container.image.repository)
   priority: CRITICAL
   tags: [process, mitre_execution]
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -68,6 +68,9 @@
 - macro: create_symlink
   condition: evt.type in (symlink, symlinkat) and evt.dir=<
 
+- macro: create_hardlink
+  condition: evt.type in (link, linkat) and evt.dir=<
+
 - macro: chmod
   condition: (evt.type in (chmod, fchmod, fchmodat) and evt.dir=<)
 
@@ -2696,7 +2699,17 @@
     (evt.arg.target in (sensitive_file_names) or evt.arg.target in (sensitive_directory_names))
   output: >
     Symlinks created over sensitive files (user=%user.name user_loginuid=%user.loginuid command=%proc.cmdline target=%evt.arg.target linkpath=%evt.arg.linkpath parent_process=%proc.pname)
-  priority: NOTICE
+  priority: WARNING
+  tags: [file, mitre_exfiltration]
+
+- rule: Create Hardlink Over Sensitive Files
+  desc: Detect hardlink created over sensitive files
+  condition: >
+    create_hardlink and
+    (evt.arg.oldpath in (sensitive_file_names))
+  output: >
+    Hardlinks created over sensitive files (user=%user.name user_loginuid=%user.loginuid command=%proc.cmdline target=%evt.arg.oldpath linkpath=%evt.arg.newpath parent_process=%proc.pname)
+  priority: WARNING
   tags: [file, mitre_exfiltration]
 
 - list: miner_ports
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -3039,7 +3039,7 @@
 # A privilege escalation to root through heap-based buffer overflow
 - rule: Sudo Potential Privilege Escalation
   desc: Privilege escalation vulnerability affecting sudo (<= 1.9.5p2). Executing sudo using sudoedit -s or sudoedit -i command with command-line argument that ends with a single backslash character from an unprivileged user it's possible to elevate the user privileges to root.
-  condition: spawned_process and user.uid != 0 and proc.name=sudoedit and (proc.args contains -s or proc.args contains -i) and (proc.args contains "\ " or proc.args endswith \)
+  condition: spawned_process and user.uid != 0 and (proc.name=sudoedit or proc.name = sudo) and (proc.args contains -s or proc.args contains -i or proc.args contains --login) and (proc.args contains "\ " or proc.args endswith \)
   output: "Detect Sudo Privilege Escalation Exploit (CVE-2021-3156) (user=%user.name parent=%proc.pname cmdline=%proc.cmdline %container.info)"
   priority: CRITICAL
   tags: [filesystem, mitre_privilege_escalation]


sberkovich:
After the conversation with @leogr removing the commit that extended the sensitive directories macro. This leaves the symlinks creation rule open for bypass (see original research), but may result in too many FPs as circleci tests showed.

Note: reconsider expanding sensitive directory macro in the future to "contains" /etc/ and /root/

rules/falco_rules.yaml
```diff
@@ -2948,7 +2958,7 @@
 
 - rule: Redirect STDOUT/STDIN to Network Connection in Container
   desc: Detect redirecting stdout/stdin to network connection in container (potential reverse shell).
-  condition: evt.type=dup and evt.dir=> and container and fd.num in (0, 1, 2) and fd.type in ("ipv4", "ipv6") and not user_known_stand_streams_redirect_activities
+  condition: evt.type in ("dup", "dup2", "dup3") and evt.dir=> and container and fd.num in (0, 1, 2) and fd.type in ("ipv4", "ipv6") and not user_known_stand_streams_redirect_activities
```
Kaizhe, line 2961: `dup2` and `dup3` are not supported yet, please refer to [here](https://github.com/falcosecurity/libs/blob/3cc178e06086c126fa539a9d82b3c616796b56a9/driver/event_table.c#L139)

rules/falco_rules.yaml
```diff
@@ -3038,7 +3048,7 @@
 # A privilege escalation to root through heap-based buffer overflow
 - rule: Sudo Potential Privilege Escalation
   desc: Privilege escalation vulnerability affecting sudo (<= 1.9.5p2). Executing sudo using sudoedit -s or sudoedit -i command with command-line argument that ends with a single backslash character from an unprivileged user it's possible to elevate the user privileges to root.
-  condition: spawned_process and user.uid != 0 and proc.name=sudoedit and (proc.args contains -s or proc.args contains -i) and (proc.args contains "\ " or proc.args endswith \)
+  condition: spawned_process and user.uid != 0 and (proc.name=sudoedit or proc.name = sudo) and (proc.args contains -s or proc.args contains -i or proc.args contains --login) and (proc.args contains "\ " or proc.args endswith \)
```
Kaizhe, line 3051: can you elaborate more why `sudo` also works?

rules/falco_rules.yaml
```diff
@@ -2708,7 +2708,17 @@
     (evt.arg.target in (sensitive_file_names) or evt.arg.target in (sensitive_directory_names))
   output: >
     Symlinks created over sensitive files (user=%user.name user_loginuid=%user.loginuid command=%proc.cmdline target=%evt.arg.target linkpath=%evt.arg.linkpath parent_process=%proc.pname)
-  priority: NOTICE
+  priority: WARNING
+  tags: [file, mitre_exfiltration]
+
+- rule: Create Hardlink Over Sensitive Files
```
Kaizhe, line 2714: I believe we did notice quite a lot of false positives, that's why we keep the priority `NOTICE`

Kaizhe (commented):
Thanks you for your contribution, really appreciate it! I do have some other thoughts about the new rules you suggested. Would you please share your research with the reviewers so that we could better understand the idea behind this PR. 

Thanks,

Kaizhe

sberkovich:
> Thanks you for your contribution, really appreciate it! I do have some other thoughts about the new rules you suggested. Would you please share your research with the reviewers so that we could better understand the idea behind this PR.
> 
> Thanks,
> 
> Kaizhe

My apologies Kaizhe, I assumed you've had the access. Invitation has been sent. Mark should've already had access.

rules/falco_rules.yaml
Andreagit97: Actually, we support `dup2` and `dup3` syscalls although I admit that the question is quite confusing. In a nutshell, these two syscalls are mapped to the simple `dup` event, so using a filter like `evt.type=dup` will also capture dup2 and dup3 calls. `dup2` and `dup3` are just a wrapper for the `dup` syscall, so when you call a `dup2` indeed a `dup` syscall is launched. This is the reason why we could catch all these 3 syscalls with just one event. This does not rule out that in the future we may also add events for `dup2` and `dup3` to make things clearer. To get a pointer to the code you can take a look at this [issue](https://github.com/falcosecurity/libs/issues/150) which dealt with a similar question.

leogr:
Hey @sberkovich 

Could you rebase, please?

rules/falco_rules.yaml
```diff
@@ -1910,11 +1913,8 @@
 # doesn't allow exclusions of a full pattern, only single characters.
 - macro: sensitive_mount
   condition: (container.mount.dest[/proc*] != "N/A" or
-              container.mount.dest[/var/run/docker.sock] != "N/A" or
-              container.mount.dest[/var/run/crio/crio.sock] != "N/A" or
               container.mount.dest[/run/containerd/containerd.sock] != "N/A" or
-              container.mount.dest[/var/lib/kubelet] != "N/A" or
-              container.mount.dest[/var/lib/kubelet/pki] != "N/A" or
+              container.mount.dest[/var*] != "N/A" or
```
Kaizhe, line 1917: can you please test with scenario of pod using service account? I bet it will trigger the alert with your changes?

leogr: Hey @Kaizhe 

are there any other blockers for you? I would like to include this PR in 0.31.0. wdyt?

Kaizhe: My concern is this is going to trigger a lot of false positives.

rules/falco_rules.yaml
sberkovich: That's explained in the "A special case of "Sudo Potential Privilege Escalation"" section of the writeup.

rules/falco_rules.yaml
sberkovich: Thats the main idea behind "Bypass rules via symlink creation" section of the writeup - its too easy to use a symlink bypass for WARNING rules, whereas the symlink creation by itself is only a NOTICE. Priority consistency from the attack detection perspective.

rules/falco_rules.yaml
leogr: Hey @sberkovich 

as discussed privately, this solution :point_right: https://github.com/falcosecurity/falco/pull/1810/commits/c28fd3817c628378c017ad2e174aabae287dbd40#diff-ea7e90b2fb13b5aa2c8e6799c1ea20612e8bd29a13c61dfb3653f03f0235436eR1926 seems interesting. 

I'm not aware if there are other legit exceptions, but the two included in `allowed_mount` seem ok to me.
@Kaizhe wdyt?

sberkovich:
Guys, I'd love to get it into 0.31. Any more feedback? I'll commit as much time as needed.

rules/falco_rules.yaml
Kaizhe: Yep, I am fine with this one!

rules/falco_rules.yaml
Kaizhe: It would be hard to iterate all the scenarios like secret, configmap etc. and `/var` directory is supposed to mean variant when Unix/Linux introduce it.

rules/falco_rules.yaml
Kaizhe: It would be hard to iterate all the scenarios like secret, configmap etc. and /var directory is supposed to mean variant when Unix/Linux introduce it.

rules/falco_rules.yaml
Kaizhe: okay, I am sort of okay with this one.

leogr (approved):
/approve

Kaizhe (approved):
/lgtm

Issue #1233, opened 2020-05-21
"BPF doesn't have JIT enabled" reported incorrectly when using docker


leogr:
**Describe the bug**

Even if JIT is enabled for BPF, both `falcosecurity/falco` and `facosecurity/falco-driver-loader` are reporting the following message when run with `-e FALCO_BPF_PROBE=""`:
```
******************************************************************
** BPF doesn't have JIT enabled, performance might be degraded. **
** Please ensure to run on a kernel with CONFIG_BPF_JIT on.     **
******************************************************************
```

**How to reproduce it**

In a system with a kernel >= 4.15  with CONFIG_BPF_JIT on and docker installed, run:
```shell
docker run --rm -i -t \
    --privileged \
    -v /dev:/host/dev \
    -v /proc:/host/proc:ro \
    -v /boot:/host/boot:ro \
    -v /lib/modules:/host/lib/modules:ro \
    -v /usr:/host/usr:ro \
    -v /etc:/host/etc:ro \
    -e FALCO_BPF_PROBE="" \
    falcosecurity/falco:0.23.0
```

**Expected behaviour**

No "BPF doesn't have JIT enabled, performance might be degraded" banner.

**Screenshots**

**Environment**

- Falco version: `0.23.0`
- System info:

- Cloud provider or hardware configuration:
- OS:

- Kernel: `Linux ubuntu-xenial 4.15.0-101-generic #102~16.04.1-Ubuntu SMP Mon May 11 11:38:16 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux`
- Installation method: `docker`

**Additional context**

Because the view of `/proc` inside the container is different, [this check](https://github.com/falcosecurity/falco/blob/d1c9aae8810ee0d21f3bd073dd19ff23288a5d08/scripts/falco-driver-loader#L431) is not working as expected when it runs inside a container. Note that prefixing the path with `${HOST}` will not solve the issue anyway.

leogr:
IIRC, we suppressed the banner. https://github.com/falcosecurity/falco/pull/1811 attempted to re-introduce it, but I had some concern :point_down: 
https://github.com/falcosecurity/falco/pull/1811#pullrequestreview-837474605

cc @LucaGuerra @FedeDP

leogr:
cc @therealbobo @LucaGuerra

therealbobo:
[...]

PR #1811, opened 2021-11-29
fix(scripts): add eBPF jit warning

LucaGuerra:
Signed-off-by: Luca Guerra <luca@guerra.sh>

**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind rule-update

> /kind rule-create

> /area build

/area engine

> /area rules

> /area tests

> /area proposals

When using the eBPF probe we should double check if eBPF JIT is enabled. If not, the system could exhibit much poorer performance than expected. We did document this behavior in the past but it's also good to have a warning printed on the logs to make it easier to troubleshoot this kind of issues.

Fixes #

```release-note

```

### DIFF
--- a/scripts/falco-driver-loader
+++ b/scripts/falco-driver-loader
@@ -471,6 +471,15 @@ load_bpf_probe() {
 	if [ -f "${HOME}/.falco/${BPF_PROBE_FILENAME}" ]; then
 		echo "* eBPF probe located in ${HOME}/.falco/${BPF_PROBE_FILENAME}"
 
+		if [ ! -f /proc/sys/net/core/bpf_jit_enable ] || [ $(cat /proc/sys/net/core/bpf_jit_enable) -eq 0 ]; then
+			echo "**********************************************************"
+			echo "** BPF doesn't have JIT enabled, performance might be   **"
+			echo "** degraded. Please ensure to run on a kernel with      **"
+			echo "** CONFIG_BPF_JIT enabled and/or use --net=host if      **"
+			echo "** running inside a container.                          **"
+			echo "**********************************************************"
+		fi
+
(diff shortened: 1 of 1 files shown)


FedeDP (approved):

leogr (commented):
Hey @LucaGuerra 

We had a warning like this already implemented. IIRC, at some point, it was removed because we had issues like https://github.com/falcosecurity/falco/issues/1233

I believe it worth investigating a bit more, and making sure that reading from `/proc` works as expected under all circumstances (eg, https://falco.org/docs/getting-started/running/#docker).
Otherwise, we will risk giving misleading information to the users.

LucaGuerra:
Thanks for the context @leogr ! I will investigate if we need additional checks on:
* whether the container has enough capability to read that value
* whether you need host network / host UTS namespace in order to do so

In that case, this warning should only be shown if the container is running with the correct privileges, and silent otherwise.

japui:
correção (scripts): adicione aviso de jit eBPF #

scripts/falco-driver-loader
```diff
@@ -471,6 +471,15 @@ load_bpf_probe() {
 	if [ -f "${HOME}/.falco/${BPF_PROBE_FILENAME}" ]; then
 		echo "* eBPF probe located in ${HOME}/.falco/${BPF_PROBE_FILENAME}"
 
+		if [ ! -f /proc/sys/net/core/bpf_jit_enable ] || [ $(cat /proc/sys/net/core/bpf_jit_enable) -eq 0 ]; then
```
FedeDP, line 474: Perhaps we should use `HOST_ROOT/proc` here?

PR #1825, opened 2021-12-16
Fix false-positive of sssd updating /etc/krb5.keytab

mac-chaffee:
Signed-off-by: Mac Chaffee <me@macchaffee.com>

**What type of PR is this?**
/kind bug
/kind rule-update

/area rules

SSSD is a popular tool for getting Linux machines to interoperate with ActiveDirectory. Part of that process involves invoking `adcli` to update /etc/krb5.keytab periodically: https://github.com/SSSD/sssd/blob/a664e9ce08ca6c0f9eb2e260b25463eea9c7829b/src/providers/ad/ad_machine_pw_renewal.c#L30

Looks like we already have exceptions for some tools that update this file (`centrify_writing_krb`), but that exception doesn't include `adcli`

n/a

Here is an example false positive alert:
```
"16:34:45.928110723: Error File below /etc opened for writing (user=root user_loginuid=-1 command=adcli update --domain=ad.renci.org --host-fqdn=k8s-node08.k8s.renci.org --computer-password-lifetime=30 --domain-controller=dc1.ad.renci.org parent=sssd_be pcmdline=sssd_be --domain ad.renci.org --uid 0 --gid 0 --logger=files file=/etc/krb5.keytab program=adcli gparent=sssd ggparent=systemd gggparent=<NA> container_id=host image=<NA>) k8s.ns=<NA> k8s.pod=<NA> container=host k8s.ns=<NA> k8s.pod=<NA> container=host k8s.ns=<NA> k8s.pod=<NA> container=host"
```

No.

```release-note
Fixed a false-positive alert that was being generated when SSSD updates /etc/krb5.keytab
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -772,6 +772,9 @@
 - macro: centrify_writing_krb
   condition: (proc.name in (adjoin,addns) and fd.name startswith /etc/krb5)
 
+- macro: sssd_writing_krb
+  condition: (proc.name=adcli and proc.aname[2]=sssd and fd.name startswith /etc/krb5)
+
 - macro: cockpit_writing_conf
   condition: >
     ((proc.pname=cockpit-kube-la or proc.aname[2]=cockpit-kube-la)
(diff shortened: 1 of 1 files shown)


leogr (approved):
LGTM.

@Kaizhe PTAL :pray:

Kaizhe (approved):

PR #1827, opened 2021-12-20
Logrotation

### LINKED ISSUE #1808: Propose log rotation in parameters for falco events stored in falco file output
**Motivation**

In my perception of Falco configuration, file_output consists in Falco log file. This file is never emptied.
We should implement a system  as kubernetes does for [audit-log](https://kubernetes.io/docs/tasks/debug-application-cluster/audit/):
```yaml
--audit-log-maxage     # defined the maximum number of days to retain old audit log files
--audit-log-maxbackup  # defines the maximum number of audit log files to retain
--audit-log-maxsize    # defines the maximum size in megabytes of the audit log file before it gets rotated
```
**Feature**

Having a mechanism as logrotate or a cron job to manage Falco `file_output` to avoid having an infinite size.

andreabonanno:
**What type of PR is this?**

/kind feature

/area engine

The aim of this PR is to implement a simple log rotation mechanism specifically for logs written via `file_output` by Falco.
As stated in issue #1808, the current behavior is to never empty/rotate these files.

Two new config parameters can be applied to `file_output` in the Falco config file:

- `log_maxage: days` represents the number of days worth of output that are kept in the configured `file_output` file, as is written by Falco. The file is emptied after _days_ days, and the old content is "rotated" (moved to a different file). Default is `log_maxage: 0`, which means that the configured `file_output` is never emptied or rotated.
- `log_maxbackup: n` represents the number of old logs (already rotated) to be kept in the same directory as the configured `file_output`. If _n_ logs are already present and a new one needs to be rotated, Falco will remove the oldest log file, following a FIFO policy. Rotated files follow a _filename_timestamp_ naming convention, with _filename_ being the original configuration of `file_output` and _timestamp_ a representation of year/month/day/hour/min/sec when the file was rotated.

Fixes #1808 implementing the proposed feature.

Original proposal in #1808 mentions also a log rotation policy based on the size of the current `file_output` file.
I chose to not implement this third parameter/policy, as in my opinion it is highly unlikely that the size of the file will be a relevant factor for log rotation, considering that the contents of `file_output` should already be considered exceptional events (alerts).

The proposed implementation keeps track of the rotated logs and timestamps internally, with a minimum number of interaction with the host's filesystem. This is to prevent the logrotate feature from introducing too many additional syscalls by itself. A consequence of this choice is that when Falco is started, already present log files from previous executions are not recognized and thus ignored (they will not be part of the set of `log_maxbackup` logs). 

```release-note
NONE
```

### DIFF
--- a/userspace/falco/outputs_file.h
+++ b/userspace/falco/outputs_file.h
@@ -28,7 +28,6 @@ namespace outputs
 
 class output_file : public abstract_output
 {
-	output_file();
 
 	void output(const message *msg);
 
@@ -38,6 +37,8 @@ class output_file : public abstract_output
 
 	void logrotate();
 
+public:
+	output_file();
 private:
 	void open_file();
 
--- a/userspace/falco/outputs_file.h
+++ b/userspace/falco/outputs_file.h
@@ -28,6 +28,8 @@ namespace outputs
 
 class output_file : public abstract_output
 {
+	output_file();
+
 	void output(const message *msg);
 
 	void cleanup();
@@ -40,8 +42,8 @@ private:
 	void open_file();
 
 	std::ofstream m_outfile;
-	std::queue<string> rotating_queue;
-	std::time_t lastlog = time(nullptr);
+	std::queue<string> m_rotating_queue;
+	std::time_t m_lastlog;
 	const unsigned int m_secs_day = 86400;
 };
 
--- a/userspace/falco/outputs_file.h
+++ b/userspace/falco/outputs_file.h
@@ -19,6 +19,7 @@ limitations under the License.
 #include "outputs.h"
 #include <iostream>
 #include <fstream>
+#include <ctime>
 
 namespace falco
 {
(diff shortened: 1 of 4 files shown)


mathildeHermet:
Hey ! Glad to see you share my point of interest about this issue. When this came into my mind, I was thinking to the built-in system of logrotation made by linux and configured through a file like '/etc/logrotate.d/falco'

userspace/falco/outputs_file.h
```diff
@@ -33,10 +34,15 @@ class output_file : public abstract_output
 
 	void reopen();
 
+	void logrotate();
+
 private:
 	void open_file();
 
 	std::ofstream m_outfile;
+	std::queue<string> rotating_queue;
```
FedeDP, line 43: Better naming it `m_rotating_queue`, i think!

userspace/falco/outputs_file.h
```diff
@@ -33,10 +34,15 @@ class output_file : public abstract_output
...
+	std::time_t lastlog = time(nullptr);
```
FedeDP, line 44: I'd initialize this in the class ctor ;)

FedeDP (commented):
Apart from minor comments, it looks good!

japui:
Fazer uma rotação necessária de log.

PR #1829, opened 2021-12-22
fix: set http output contenttype to text/plain when json output is disabled

### LINKED ISSUE #1824: `http_output` always uses "Content-Type: application/json" even if json formatting is disabled
**Describe the bug**

`Content-Type: application/json` is always sent :point_down: 

https://github.com/falcosecurity/falco/blob/f035829ca2881b70a4d13e22abb5f8eb828d3a5f/userspace/falco/outputs_http.cpp#L31

regardless of the value of json formatting option :point_down: 

https://github.com/falcosecurity/falco/blob/f035829ca2881b70a4d13e22abb5f8eb828d3a5f/falco.yaml#L68

**How to reproduce it**

**Expected behaviour**

Perhaps, `text/plain` when `json_output: false` ?

**Screenshots**

**Environment**

- Falco version:

- System info:

- OS:

- Kernel:

- Installation method:

**Additional context**

FedeDP:
**What type of PR is this?**

/kind bug

/area engine

The PR sets the http output Content-type header to a more meaningful value (text/plain) when json_output is disabled.

Fixes #1824 

```release-note
fix: set http output contenttype to text/plain when json output is disabled
```

### DIFF
--- a/userspace/falco/outputs.h
+++ b/userspace/falco/outputs.h
@@ -63,11 +63,12 @@ class abstract_output
 public:
 	virtual ~abstract_output() {}
 
-	void init(config oc, bool buffered, std::string hostname)
+	void init(config oc, bool buffered, std::string hostname, bool json_output)
 	{
 		m_oc = oc;
 		m_buffered = buffered;
 		m_hostname = hostname;
+		m_json_output = json_output;
 	}
 
 	// Return the output's name as per its configuration.
@@ -89,6 +90,7 @@ protected:
 	config m_oc;
 	bool m_buffered;
 	std::string m_hostname;
+	bool m_json_output;
 };
 
 } // namespace outputs
--- a/userspace/falco/falco_outputs.cpp
+++ b/userspace/falco/falco_outputs.cpp
@@ -135,7 +135,7 @@ void falco_outputs::add_output(falco::outputs::config oc)
 		throw falco_exception("Output not supported: " + oc.name);
 	}
 
-	oo->init(oc, m_buffered, m_hostname);
+	oo->init(oc, m_buffered, m_hostname, m_json_output);
 	m_outputs.push_back(oo);
 }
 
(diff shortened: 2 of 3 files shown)


userspace/falco/outputs.h
```diff
@@ -51,6 +51,7 @@ struct message
 	std::string source;
 	map<std::string, std::string> fields;
 	std::set<std::string> tags;
+	gen_event_formatter::output_format format;
```
leogr, line 54: Could we avoid passing the formatter in the message?

This approach cannot work with `internal` messages, that do not have a formatter :point_down: 
https://github.com/falcosecurity/falco/blob/226d1fb7284319b5dfb7cbed50281ca11cfda7ad/userspace/falco/falco_outputs.cpp#L199-L213

Moreover, the JSON formatting option is a global setting (not a per-message setting) and the message already contains the formatted text. So I believe the formatter does not really belong to the message.

I would rather pass the `m_json_output`'s value to each output via the `init` function :point_down: 
https://github.com/falcosecurity/falco/blob/226d1fb7284319b5dfb7cbed50281ca11cfda7ad/userspace/falco/falco_outputs.cpp#L138

leogr (commented):
Hey @FedeDP 

Thank you for this PR! We really need to fix that.
However, we shouldn't pass the formatter to the output channels. Moreover, if I understood correctly, your implementation won't work with internal messages (because they don't get a formatter). See my comment below.

I propose instead to propagate `m_json_ouput` to the output channels during init.

userspace/falco/outputs.h
FedeDP: Much simpler, you're right! Thanks ;)

leogr (approved):
/approve

jasondellaluce (approved):
Thanks @FedeDP! ❤️ 

/approve

PR #1832, opened 2022-01-01
fix(macro modify_shell_history, truncate_shell_history): avoid false positives from .zsh_history.new and .LOCK files

m4wh6k:
Signed-off-by: m4wh6k <m4wh6k@users.noreply.github.com>

/kind bug
/kind rule-update
/area rules

The default macros modify_shell_history and truncate_shell_history match on .zsh_history.new and .zsh_history.LOCK files, which are temporary files created while using zsh. This causes rules using these macros to produce false positives. This PR changes the macros to use `endswith` rather than `contains`, which makes it work in the expected way.

n/a

No

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -2574,15 +2574,15 @@
   condition: >
     (modify and (
       evt.arg.name contains "bash_history" or
-      evt.arg.name contains "zsh_history" or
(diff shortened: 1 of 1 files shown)


Kaizhe (approved):

leogr (approved):
/approve

PR #1834, opened 2022-01-02
replace ".." concatenation with table.concat

VadimZy:
When testing exceptions with a large number of auto-generated values we faced an issue with Falco validator memory starvation. That was due to the code performing string concatenation with ".." operator.  Due to the string immutability, each of such operations creates a new copy of the result string keeping the old one in the memory till the GC is run, and the GS won't run at least until the function returns. The above generates a snowball of allocated and unused memory. Rough estimates could give at least an N^2 growth rate. In our tests a ~1MB rules YAML required between 1-2 GB of memory to be parsed.

**What type of PR is this?**

>/kind bug

> /kind design

> /area engine

fix memory mismanagement when converting exceptions

fixes memory mismanagement when converting exception

Fixes #

fixes memory mismanagement when converting exception

NONE

original code formating is not consistent 3/4 spaces per indent.
indentation changed to 4 spaces. 
please use ignore whitespace when reviewing. 

```release-note
NONE
```

### DIFF
--- a/userspace/engine/lua/rule_loader.lua
+++ b/userspace/engine/lua/rule_loader.lua
@@ -293,19 +293,19 @@ function split_lines(rules_content)
 end
 
 function get_orig_yaml_obj(rules_lines, row)
-   local ret = ""
+    idx = row
+    local t = {}
+    while (idx <= #rules_lines) do
+        t[#t + 1] = rules_lines[idx]
+        idx = idx + 1
 
-   idx = row
-   while (idx <= #rules_lines) do
-      ret = ret..rules_lines[idx].."\n"
-      idx = idx + 1
-
-      if idx > #rules_lines or rules_lines[idx] == "" or string.sub(rules_lines[idx], 1, 1) == '-' then
-	 break
-      end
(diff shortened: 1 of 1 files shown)


mstemm (approved):
@VadimZy is going to add some comments to the PR that provide more context on why these changes result in more efficient management of exceptions.

But the changes look good to me and thanks for finding the problem and providing a fix!

jasondellaluce:
@VadimZy Thanks for the contribution, this is definitely an issue and the contribution looks promising. Can you [sign-off](https://github.com/falcosecurity/.github/blob/master/CONTRIBUTING.md#developer-certificate-of-origin) your commits? That's a requirement as for our governance.

leodido (commented):
LGTM

VadimZy:
just pushed the message for both commits but one still mentioned as not signed off.

VadimZy:
I see it either/or have dco or tests passed

mstemm (approved):

leogr (approved):
/approve

PR #1840, opened 2022-01-04
update: support non-enumerable event sources in gRPC outputs service

### LINKED ISSUE #1826: wip: schema.proto does not consider custom data source provided by plugins
**Describe the bug**

`schema.proto` that's required by the gRPC implementation enums only predefined data sources (ie. `syscall`, `k8s_audit`, `internal`):

https://github.com/falcosecurity/falco/blob/f035829ca2881b70a4d13e22abb5f8eb828d3a5f/userspace/falco/schema.proto#L51-L63

Consequently, the gRPC output enforces the consistency of `source` field and throws an exception in case it does not match the enum set:
https://github.com/falcosecurity/falco/blob/f035829ca2881b70a4d13e22abb5f8eb828d3a5f/userspace/falco/outputs_grpc.cpp#L37-L42

That, unfortunately, does not allow to set the value returned by a source plugin (which defines its own data source, eg. `cloudtrail`).

**How to reproduce it**

Try to use a source plugin with the gRPC output enabled. The implementation should throw an exception on the first alert notification.

**Expected behaviour**

No exception thrown. The gRPC implementation should return the correct data source.

**Screenshots**

**Environment**

This applies only to the not yet released (at the time of writing) build of Falco including the plugin implementation.

**Additional context**

jasondellaluce:
**What type of PR is this?**

/kind design

/kind feature

/area engine

See https://github.com/falcosecurity/falco/issues/1826 for details about the issue.

This deprecates the `source` field in `outputs.proto`, and adds a `PLUGIN` entry in the source enum. The goal is to maintain backward compatibility for clients relying on old protobufs versions. Accordingly, a new string field for the event source is added. This is required by plugin system, as the source name is determined by the source plugin loaded and it's not enumerable anymore.

Fixes #1826 

```release-note
update: support non-enumerable event sources in gRPC outputs service
```

### DIFF
--- a/userspace/falco/outputs.proto
+++ b/userspace/falco/outputs.proto
@@ -45,10 +45,11 @@ message request {
 message response {
   google.protobuf.Timestamp time = 1;
   falco.schema.priority priority = 2;
-  falco.schema.source source = 3;
+  falco.schema.source source_deprecated = 3 [deprecated=true];
   string rule = 4;
   string output = 5;
   map<string, string> output_fields = 6;
   string hostname = 7;
   repeated string tags = 8;
+  string source = 9;
 }
--- a/userspace/falco/outputs_grpc.cpp
+++ b/userspace/falco/outputs_grpc.cpp
@@ -21,6 +21,20 @@ limitations under the License.
 #include "formats.h"
 #include "banned.h" // This raises a compilation error when certain functions are used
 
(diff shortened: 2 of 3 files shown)


leogr (approved):
/approve

mstemm (approved):

Issue #1692, opened 2021-07-15
Falco Hot Reloader Proposal


developer-guy:
**Motivation**

The Falco project has a [built-in hot-reload feature](https://falco.org/docs/getting-started/running/#hot-reload), so what is our concern to do that kind of project? Simple, the Falco project can reload itself when it receives a SIGHUP signal, so think of that scenario in a Kubernetes environment; the Falco uses a [ConfigMap](https://github.com/falcosecurity/charts/blob/master/falco/templates/configmap.yaml), which includes configuration and rule files Falco can use. We want to add support to monitor these files, detect changes against them, and send SIGHUP signals to Falco. To do that, we need to know a couple of things:

- How can access PID of the Falco process within the same Pod?
- Are ConfigMaps mounted inside the Pod reloading itself against updates?

**How can access PID of the Falco process within the same Pod?**
We know that Pod is a concept that can include one or more containers in it, and we are generally using Pods with one container inside it. However, sometimes we might need an extra container to assist the main container in Pod. So, how containers in the Pod can communicate with each other? There are multiple ways of doing this, such as Shared Volume, IPC, localhost net interface, etc. In our case, we will share the process namespace between containers in a Pod; with that, we will now access the Falco container PID from a different container.

To get more detail about the concept, please follow the [link](https://kubernetes.io/docs/tasks/configure-pod-container/share-process-namespace/).

**Are ConfigMaps mounted inside the Pod reloading itself against updates ?**
There is a really well-explained section about this question in the Kubernetes documentation, and it says:

> When a ConfigMap currently consumed in a volume is updated, projected keys are eventually updated. The kubelet checks whether the mounted ConfigMap is fresh on every periodic sync. However, the kubelet uses its local cache for getting the current value of the ConfigMap. The type of the cache is configurable using the ConfigMapAndSecretChangeDetectionStrategy field in the KubeletConfiguration struct. A ConfigMap can be either propagated by watch (default), TTL-based or redirecting all requests directly to the API server. As a result, the total delay from the moment when the ConfigMap is updated to the moment when new keys are projected to the Pod can be as long as the kubelet sync period + cache propagation delay, where the cache propagation delay depends on the chosen cache type (it equals to watch propagation delay, TTL of cache, or zero correspondingly). ConfigMaps consumed as environment variables are not updated automatically and require a pod restart.

[...]

PR #1849, opened 2022-01-11
Add grpc rules loading

mstemm:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind rule-update

> /kind rule-create

/area build

/area engine

> /area rules

> /area tests

> /area proposals

Fixes #

```release-note
Add the ability to reload rules files on the fly via the embedded grpc server.
```


FedeDP (commented):
The idea seems pretty interesting to me @mstemm, i quite like it; can't say nothing more given that i most probably lack some of the underlying context.

Its it that hard to just replace the ruleset without replacing the entire engine, eg adding some way to "reset" engine state and call load_rules_file() for each new file again? 
I don't know how feasible that is though.

mstemm:
> Is it that hard to just replace the ruleset without replacing the entire engine, eg adding some way to "reset" engine state and call load_rules_file() for each new file again?
I don't know how feasible that is though.

I actually started by going down that route, thinking this would be a trivial change. The big problem with that approach is that we want to to do the rules reload via the grpc server, and the grpc server runs in a separate thread than the one processing events.

So you'd have to add some kind of synchronization between the threads while the reload is in progress to avoid race conditions. But this results in potentially stalling event processing while the reload is happening. We've noticed in some commercial uses of the falco engine that this can lead to dropped syscall events in high volume environments with complex rules files.

Although I did get something to compile, I'm not sure it's going to ultimately be better than the current approach of signaling falco with SIGHUP to simply restart everything. I'm going to clean this up a bit and see what shape it's in.

It would also have the benefit of cleaning up the currently gigantic init_falco function, which has grown to ~1200 (!!) lines and is currently kind of a mess with a mix of command line validation, opening plugins, initializing the engine, loading rules, handling one-off paths like --list/--validate/etc.

FedeDP:
> It would also have the benefit of cleaning up the currently gigantic init_falco function, which has grown to ~1200 (!!) lines and is currently kind of a mess with a mix of command line validation, opening plugins, initializing the engine, loading rules, handling one-off paths like --list/--validate/etc.

+1 +1 +1 +1 +1 from me :) That alone would be superb!

I understand the issue better now, thanks! 
Just a question: isn't the `tbb::concurrent_queue<std::shared_ptr<falco_engine>> m_pending_falco_engine;` just another locking mechanism by itself too? I mean, the concurrent_queue will surely use some sort of locking in it. I am quite curious about its performance and its impact on events drop too.

leogr:
> > It would also have the benefit of cleaning up the currently gigantic init_falco function, which has grown to ~1200 (!!) lines and is currently kind of a mess with a mix of command line validation, opening plugins, initializing the engine, loading rules, handling one-off paths like --list/--validate/etc.
> 
> +1 +1 +1 +1 +1 from me :) That alone would be superb!
> 
> I understand the issue better now, thanks! Just a question: isn't the `tbb::concurrent_queue<std::shared_ptr<falco_engine>> m_pending_falco_engine;` just another locking mechanism by itself too? I mean, the concurrent_queue will surely use some sort of locking in it. I am quite curious about its performance and its impact on events drop too.

For the record, in the past, we did an attempt to load rules on the fly using an external lib. The issue we faced is similar. The most performant way we found is in this commit :point_down: (never merged)
https://github.com/falcosecurity/falco/commit/f8c10f6c2791f902804c4739dc0a5b4c995030e5

The full code is here :point_down: 
https://github.com/falcosecurity/falco/commits/libhawk-rules

FedeDP:
Very cool @leogr , i like it

Kaizhe:
@mstemm ,if falco is running as daemonset, does it mean I will have to update each falco pod through grpc server as well as updating the configmap where the rule resides?

leogr:
> @mstemm ,if falco is running as daemonset, does it mean I will have to update each falco pod through grpc server as well as updating the configmap where the rule resides?

I think so. That may be not the best. Maybe, we have to rethink a bit the UX of this feature. Any idea?

Kaizhe:
> > @mstemm ,if falco is running as daemonset, does it mean I will have to update each falco pod through grpc server as well as updating the configmap where the rule resides?
> 
> I think so. That may be not the best. Maybe, we have to rethink a bit the UX of this feature. Any idea?

I prefer to modify the rule once (via the configmap), and the running Falco pods detects the mounted rule updated, then auto-reload. May a golang thread to monitor the rule file changes and then send request to the grpc server?

leogr:
> > > @mstemm ,if falco is running as daemonset, does it mean I will have to update each falco pod through grpc server as well as updating the configmap where the rule resides?
> > 
> > 
> > I think so. That may be not the best. Maybe, we have to rethink a bit the UX of this feature. Any idea?
> 
> I prefer to modify the rule once (via the configmap), and the running Falco pods detects the mounted rule updated, then auto-reload. May a golang thread to monitor the rule file changes and then send request to the grpc server?

Hey @Kaizhe 

There's always a proposal to do something similar (with the signal instead of the gRPC API) :point_right: https://github.com/falcosecurity/falco/issues/1692 

Anyway, I believe we can watch for rule files changes directly in Falco without needing a gRPC API. It would be straightforward.

Furthermore, I'm thinking that we should differentiate between rules hot reload and config hot reload. 
The config hot reload requires stopping the capture for a while (no detection for a few seconds) and then restarting itself.
On the other hand, I think rules hot reload could be implemented without fully reloading Falco.
(cc @jasondellaluce since I know he's playing with filter checks and perhaps the rule loader too).

Issif:
> > @mstemm ,if falco is running as daemonset, does it mean I will have to update each falco pod through grpc server as well as updating the configmap where the rule resides?
> 
> I think so. That may be not the best. Maybe, we have to rethink a bit the UX of this feature. Any idea?

I totally agree, with current implementation, Falco is the gRPC server, it means you need to know all IP of all your instances. The task is even harder with Falco running inside a k8s cluster (a node can be spawned and a new Falco pod booted while you're already connected to other nodes, and you miss it).

This is not the topic here, but I think we should consider to reverse the situation, Falco should also work as a gRPC client (to falco-exporter or falcosidekick).

For the hot reload of rules, I think a classic `fsnotify` may work in most of the situation, just by telling Falco to follow direct files and `rules.d` folder.

japui:
Adicionar regras grpc

PR #1850, opened 2022-01-12
[feature] Ability to set user-agent variable for http_output

### LINKED ISSUE #1831: Ability to customize falco's http_output for logstash http input plugin.
**Motivation**

I attempting to ship falco events via http_output to logstash's http input. I am missing a field that would identify the incoming events as produced by falco. http_user_agent is also empty.

**Feature**

An ability to add additional field to http payload (or a fixed field) that would help identify the incoming data in logstash. e.g. 

```
"source" => "falco"
```

**Alternatives**

Custom script that would wrap the log message with additional fields. But ideally http_output ought to be a little bit customizable.

**Additional context**

Example of data that comes into logstash http input via falco's http_output, for reference  (as seen in logstash) : 

``` 
{ 
   "output_fields" => {
                          "fd.name" => "172.17.255.86:45264->10.117.196.70:22",
                       "proc.pname" => "sshd",
                     "proc.cmdline" => "sshd",
                        "user.name" => "root",
                           "fd.sip" => "10.117.196.70",
                 "evt.time.iso8601" => 1640956206668757005,
                     "container.id" => "confes02_prod",
                           "fd.num" => 0,
                    "user.loginuid" => 1000,
                   "container.name" => "confes02_prod",
       "container.image.repository" => nil,
                          "fd.type" => "ipv4",
                        "proc.name" => "sshd",
                         "proc.tty" => 0
   },
          "output" => "2021-12-31T13:10:06.668757005+0000: Warning Redirect stdout/stdin to network connection (user=root user_loginuid=1000 confes02_prod (id=confes02_prod) process=sshd parent=sshd cmdline=sshd terminal=0 container_id
=confes02_prod image=<NA> fd.name=172.17.255.86:45264->10.117.196.70:22 fd.num=0 fd.type=ipv4 fd.sip=10.117.196.70)",
         "headers" => {
           "http_accept" => "*/*",
          "content_type" => "application/json",
          "http_version" => "HTTP/1.1",
          "request_path" => "/",
        "request_method" => "POST",
             "http_host" => "172.20.4.82:8080",
        "content_length" => "870",
       "http_user_agent" => nil
   },
      "@timestamp" => 2021-12-31T13:10:06.743Z,
            "host" => "172.17.3.37",
        "@version" => "1",
            "rule" => "Redirect STDOUT/STDIN to Network Connection in Container",
            "time" => "2021-12-31T13:10:06.668757005Z",
          "source" => "syscall",
        "priority" => "Warning",
            "tags" => []
}
```

yoshi314:
/kind feature

/area engine

This allows to identify falco's datastream sent via http_output on the receiving end.

In my particular case that is logstash with its http listener. 

Since I have multiple processes submitting via http from the same machine, I need a way to distinguish falco's events from other incoming data streams for further parsing and processing.

Fixes #1831 

A config entry of user_agent: "some string" is now supported under http_output. If not provided, the value defaults to "falcosecurity/falco".

Setting it to "" ought to restore previous behaviour (that field appears to be present in http headers, but with blank value).

```release-note
Add ability to set User-Agent http header when sending http output. Provide default value of 'falcosecurit/falco'.
```

### DIFF
--- a/userspace/falco/outputs_http.cpp
+++ b/userspace/falco/outputs_http.cpp
@@ -34,10 +34,13 @@ void falco::outputs::output_http::output(const message *msg)
 		} else {
 			slist1 = curl_slist_append(slist1, "Content-Type: text/plain");
 		}
+        slist1 = curl_slist_append(slist1, m_oc.options["user_agent"].c_str());
+
 		curl_easy_setopt(curl, CURLOPT_HTTPHEADER, slist1);
 		curl_easy_setopt(curl, CURLOPT_URL, m_oc.options["url"].c_str());
 		curl_easy_setopt(curl, CURLOPT_POSTFIELDS, msg->msg.c_str());
 		curl_easy_setopt(curl, CURLOPT_POSTFIELDSIZE, -1L);
+
 
 		res = curl_easy_perform(curl);
 
--- a/userspace/falco/configuration.cpp
+++ b/userspace/falco/configuration.cpp
@@ -148,6 +148,10 @@ void falco_configuration::init(string conf_filename, list<string> &cmdline_optio
 		}
 		http_output.options["url"] = url;
 
+		string user_agent;
+		user_agent = m_config->get_scalar<string>("http_output.user_agent","falcosecurity/falco");
+		http_output.options["user_agent"] = user_agent;
+
 		m_outputs.push_back(http_output);
 	}
 


userspace/falco/outputs_http.cpp
```diff
@@ -34,11 +34,14 @@ void falco::outputs::output_http::output(const message *msg)
 		} else {
 			slist1 = curl_slist_append(slist1, "Content-Type: text/plain");
 		}
+        slist1 = curl_slist_append(slist1, m_oc.options["user_agent"].c_str());
```
FedeDP, line 37: Spacing :)

FedeDP:
Hi! Thank you for opening this PR! 
This is very useful IMO!
Note that all your commits must be signed with dco, see https://github.com/falcosecurity/.github/blob/master/CONTRIBUTING.md#developer-certificate-of-origin.
Moreover, commits message must obey to conventional commits: https://github.com/falcosecurity/.github/blob/master/CONTRIBUTING.md#commit-convention

It would be great if you could force-push with these changes in place!

FedeDP (commented):
Apart from a single (small) comment, LGTM!
Thanks!

jasondellaluce:
On top of @FedeDP's comments, I would also add an `http_output.user_agent` in `falco.yaml` too for better visibility.

FedeDP:
Hi!
It is just as simple as rebasing interactively and add the "Signed-off-by ..." bit to each commit message :)

yoshi314:
I hope that fixed it. Also, what exactly is  that remark with spacing about?  A case of tabs/spaces or something else?

jasondellaluce:
Yeah, it looks like it worked (You can see your sign-off on the commit message).

The spacing issue seems to be that `outputs_http.cpp` uses tabs for indentation, whereas the new changes use spacing. If you can fix that, and also add an entry for the new config field in `falco.yaml`, I would proceed and approve!

yoshi314:
It appears vim can be deceiving with indentation. Should be fixed now.

userspace/falco/outputs_http.cpp
```diff
@@ -34,11 +34,14 @@ void falco::outputs::output_http::output(const message *msg)
 		} else {
 			slist1 = curl_slist_append(slist1, "Content-Type: text/plain");
 		}
+		slist1 = curl_slist_append(slist1, m_oc.options["user_agent"].c_str());
```
jasondellaluce, line 37: By testing it locally, I was not able to get the user agent header (but maybe it was me). Either way, not setting the `user-agent: ` prefix properly would result in an invalid header I think. Would it be better to use this instead?
```
curl_easy_setopt(curl, CURLOPT_USERAGENT, m_oc.options["user_agent"].c_str());
```

jasondellaluce (changes requested):

userspace/falco/outputs_http.cpp
yoshi314: i'll have to retest this in my setup. I was pretty sure that worked on my case, when checking logs on remote end.

yoshi314: Oh, i think i missd the "User-Agent:" string from the header, since i used a predefined string before.

jasondellaluce:
@yoshi314 new changes look good. Please sign-off the last commit and we should be good to go!

jasondellaluce (approved):
/approve

Thanks for your contribution! 🥳

leogr (approved):
LGTM

@yoshi314 thank you for this PR!

/approve

PR #1852, opened 2022-01-12
new(configuration): support defining plugin init config as a YAML

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind design

/kind feature

/area engine

This PR cleans up the plugin configuration reading, and supports init configs defined in YAML format. With this, users can define the `init_config` fields either as a string, or as a YAML map. In the first case, the string is opaque and passed as-is to the plugin framework of `libs`. In the second case, the YAML is parsed and converted to a JSON string. This is useful for two purposes:
- Coupled with the feature of https://github.com/falcosecurity/libs/pull/175, this allows defining structured configuration data more easily, along with automatic validation and error checking
- Defining/overriding plugin init configs with Falco's `-o` cli parameter

The YAML data is converted in JSON format for simplicity and because that is the only schema-based data format we support as for https://github.com/falcosecurity/libs/pull/175. This can easily be extended in the future to more formats if required, as the plugin-provided schema dictates the format expected by a given plugin.

This also includes a minor fix to the changes introduced in #1792, and removes the `init_config_file` configuration field.

```release-note
new(configuration): support defining plugin init config as a YAML
```

### DIFF
--- a/userspace/falco/configuration.h
+++ b/userspace/falco/configuration.h
@@ -168,6 +168,10 @@ private:
 					int nodeIdx = std::stoi(key.substr(i + 1, close_param_idx - i - 1));
 					ret.reset(ret[nodeIdx]);
 					i = close_param_idx;
+					if (i < key.size() - 1 && key[i + 1] == '.')
+					{
+						i++;
+					}
 				}
 			}
 		}
--- a/userspace/falco/configuration.h
+++ b/userspace/falco/configuration.h
@@ -333,60 +333,8 @@ namespace YAML {
 	template<>
 	struct convert<falco_configuration::plugin_config> {
 
-		static bool read_file_from_key(const Node &node, const std::string &prefix, std::string &value)
-		{
-			std::string key = prefix;
-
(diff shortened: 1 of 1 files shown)


userspace/falco/configuration.h
```diff
@@ -277,48 +281,60 @@ class falco_configuration
...
+			double double_val;
+			bool bool_val;
+			std::string str_val;
+			nlohmann::json sub{};
+
+			switch (node.Type()) {
+				case YAML::NodeType::Map:
+					for (auto &&it: node)
+					{
+						YAML::convert<nlohmann::json>::decode(it.second, sub);
```
mstemm, line 297: Not that any legitimate use would do this, but we might want to cap the recursion at some big but fixed number of levels.

mstemm (approved):
I don't feel that strongly about the recursion thing, so I'll approve just to skip some back and forth. But go for it if you'd like and I'll re-approve after.

leogr (approved):
/approve

PR #1866, opened 2022-01-25
Rule Update - Adds npm support to pkg mgt processes

rileydakota:
Adds `npm` to `package_mgmt_binaries` for detection of "living off the land" style attacks that utilize NPM to pull down additional tooling/libraries

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

/kind rule-update

> /area build

> /area engine

/area rules

> /area tests

> /area proposals

Adds `npm` to the list of `package_management_binaries`. NodeJS is one of the most popular development languages in the world - and the `node` base image comes packaged with npm. Like other package managers - npm installing packages in an already built container is a best practice violation at best, and a potential IoC at worst (attackers could potentially use it to pull down additional tooling/libraries without tripping the existing package management process alerts).

Fixes #

Yes

```release-note
rule(list package_mgmt_binaries): `npm` added
rule(Launch Package Management Process in Container): support for detecting `npm` usage
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -222,7 +222,7 @@
 # The truncated dpkg-preconfigu is intentional, process names are
 # truncated at the falcosecurity-libs level.
 - list: package_mgmt_binaries
-  items: [rpm_binaries, deb_binaries, update-alternat, gem, pip, pip3, sane-utils.post, alternatives, chef-client, apk, snapd]
+  items: [rpm_binaries, deb_binaries, update-alternat, gem, npm, pip, pip3, sane-utils.post, alternatives, chef-client, apk, snapd]
 
 - macro: package_mgmt_procs
   condition: proc.name in (package_mgmt_binaries)


jasondellaluce:
Hi @rileydakota! Thanks for your contribution. Can you [sign-off](https://github.com/falcosecurity/.github/blob/master/CONTRIBUTING.md#developer-certificate-of-origin) your commit?

leogr (approved):
/approve

jasondellaluce (approved):
/approve

micheas:
This was approved in error and generates false positives over 99% of the time.

`npm start` is the standard way of starting a node js application.

micheas:
Unless it can be rewritten to exclude npm start it should be backed out

micheas:
@rileydakota this is awful. 99% false positives in the real world. 

```
Error Package management process launched in container (user=<NA> user_loginuid=-1 command=npm node /usr/local/bin/npm start container_id=7215a7d30edc container_name=<NA> image=<NA>:<NA>) k8s.ns=<NA> k8s.pod=<NA> container=7215a7d30edc
```

Is the most common log message. While the intention is good it resulted in all of the package management processes being ignored resulting in a net negative.

rileydakota:
potentially could break it into a separate rule and rely on `proc.args` to only flag on instances of `npm install` which was the original intention. The current structure (having all package manager names grouped and then a macro that checks if `proc.name` is present wouldn't suffice however.

PR #1874, opened 2022-01-26
Escape double-quotes in aws cloudtrail rule

mstemm:
The rule Delete Bucket Public Access Block has a predicate
`json.value[/requestParameters/publicAccessBlock]=""` to match
an event snippet like this:

```
			"requestParameters": {
				"publicAccessBlock": "",
```

The cloudtrail plugin properly returns `""` for this field, but the
yaml representation was a literal back-to-back quote, which gets
interpreted by the yaml parser to be an empty string.

Escaping the back-to-back quote fixes the ambiguity.

This is related to https://github.com/falcosecurity/plugins/issues/56. There were actually 3 fixes in plugins/libs/falco, so I won't directly link it here.

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

/kind rule-update

> /kind rule-create

> /area build

> /area engine

/area rules

> /area tests

> /area proposals

Fixes #

```release-note
NONE
```

### DIFF
--- a/rules/aws_cloudtrail_rules.yaml
+++ b/rules/aws_cloudtrail_rules.yaml
@@ -335,7 +335,7 @@
   desc: Detect deleting blocking public access to bucket.
   condition:
     ct.name="PutBucketPublicAccessBlock" and not ct.error exists and
-    json.value[/requestParameters/publicAccessBlock]="" and
+    json.value[/requestParameters/publicAccessBlock]='""' and
       (json.value[/requestParameters/PublicAccessBlockConfiguration/RestrictPublicBuckets]=false or
       json.value[/requestParameters/PublicAccessBlockConfiguration/BlockPublicPolicy]=false or
       json.value[/requestParameters/PublicAccessBlockConfiguration/BlockPublicAcls]=false or


leogr (changes requested):
This solution will not address all cases.

The current implementation of the `json` plugin always [returns](https://github.com/falcosecurity/plugins/blob/f6b32ee7625ea95fe0739c21cbc05dda69d3ffc7/plugins/json/json.go#L133) the literal string as in JSON syntax (ie. all strings will be always double-quoted, see [here](https://github.com/valyala/fastjson/blob/6dae91c8e11a7fa6a257a550b75cba53ab81693e/parser.go#L596)). 

So the bug in the current implementation is that all strings contain unnecessary double quotes. Not only the empty string case is affected. Also, all other string comparisons are affected. For example, also this condition :point_down: won't be triggered.
https://github.com/falcosecurity/falco/blob/6a42f4a1333275dd43f844b326dae60316575a17/rules/aws_cloudtrail_rules.yaml#L52
(because the extracted string is `"Success"` and not `Success` as one would expect).

To fix the problem extracted strings should not contain any double-quote. Thus it should be fixed in the extraction phase and not here. 

We (@jasondellaluce, @FedeDP, and I) are addressing the problem directly in the `json` plugin.

leogr:
Probably superseded by https://github.com/falcosecurity/plugins/pull/59

mstemm:
Ok, I'll close this then.

PR #1875, opened 2022-01-27
fix(userspace/engine): actually make m_filter_all_event_types useful by properly using it as fallback when no filter event types is provided

FedeDP:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind rule-update

> /kind rule-create

/area build

/area engine

> /area rules

> /area tests

> /area proposals

Experimenting together with @jasondellaluce and @leogr  a possible solution for https://github.com/falcosecurity/plugins/issues/57

The issue was that `m_filter_all_event_types` was actually unused: for `syscalls` event source, we always have a `m_filter_by_event_type` value for the filter, event if `evt.type` is missing in the condition: in fact when it is missing, a default set of "all events" is used.  

For k8s audit logs the bug is not appearing because its filter check provides a default `evttypes()` impl that basically means "force-use any event type", even when negated.

Plugins filter checks instead is returning `PPME_PLUGINEVENT_E` as only possible event associated; but, when rule condition gets negated, there is no other set of event type possible and thus the filter check gets added to `m_filter_all_event_types`.

Unfortunately, the old check in `ruleset_filters::run`:
```
if(evt->get_type() >= m_filter_by_event_type.size())
{
	return false;
}
```
prevented it to actually run the filterchecks stored in `m_filter_all_event_types`, de facto making it useless.

Fixes #

Edit by @leogr: we also bumped the json plugin version to include a fix for https://github.com/falcosecurity/plugins/issues/56

```release-note
NONE
```

### DIFF
--- a/cmake/modules/plugins.cmake
+++ b/cmake/modules/plugins.cmake
@@ -27,8 +27,8 @@ install(FILES "${PROJECT_BINARY_DIR}/cloudtrail-plugin-prefix/src/cloudtrail-plu
 
 ExternalProject_Add(
   json-plugin
-  URL "https://download.falco.org/plugins/stable/json-0.2.1-${PLUGINS_SYSTEM_NAME}-${CMAKE_HOST_SYSTEM_PROCESSOR}.tar.gz"
-  URL_HASH "SHA256=14d1cf4c3c651af0daec7a45162ef91172d6f0baba787f0eff0227b3cf2ca39c"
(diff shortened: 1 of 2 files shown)


leogr:
/close

leogr (approved):
/approve

PR #1877, opened 2022-01-28
rule(falco_rules): Polkit Local Privilege Escalation Vulnerability (CVE-2021-4034)

darryk10:
Polkit pkexec Local Privilege Escalation Vulnerability (CVE-2021-4034)

Co-authored-by: javery-sysdig <jason.avery@sysdig.com>

Signed-off-by: darryk10 <stefano.chierici@sysdig.com>

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind rule-update

/kind rule-create

> /area build

> /area engine

/area rules

> /area tests

> /area proposals

This rule detects an attempt to exploit a privilege escalation vulnerability in Polkit's pkexec. By running specially crafted code, a local user can leverage this flaw to gain root privileges on a compromised system. Reference: https://seclists.org/oss-sec/2022/q1/80

Fixes #

```release-note
rule(Polkit Local Privilege Escalation Vulnerability): new rule created to detect CVE-2021-4034
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -3125,9 +3125,9 @@
 - rule: Polkit Local Privilege Escalation Vulnerability (CVE-2021-4034)
   desc: "This rule detects an attempt to exploit a privilege escalation vulnerability in Polkit's pkexec. By running specially crafted code, a local user can leverage this flaw to gain root privileges on a compromised system"
   condition:
-    spawned_process and user.uid != 0 and proc.name=pkexec and proc.args = '' and proc.env icontains 'GCONV_PATH'
+    spawned_process and user.uid != 0 and proc.name=pkexec and proc.args = ''
   output:
-    "Detect Polkit pkexec Local Privilege Escalation Exploit (CVE-2021-4034) (user=%user.loginname uid=%user.loginuid command=%proc.cmdline access_args=%evt.args)"
+    "Detect Polkit pkexec Local Privilege Escalation Exploit (CVE-2021-4034) (user=%user.loginname uid=%user.loginuid command=%proc.cmdline args=%evt.args)"
   priority: CRITICAL
   tags: [process, mitre_privilege_escalation]
 
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -3120,6 +3120,17 @@
   priority: NOTICE
   tags: [network, process, mitre_command_and_control]
 
+# This rule helps detect CVE-2021-4034:
+# A privilege escalation to root through memory corruption
+- rule: Polkit Local Privilege Escalation Vulnerability (CVE-2021-4034)
+  desc: "This rule detects an attempt to exploit a privilege escalation vulnerability in Polkit's pkexec. By running specially crafted code, a local user can leverage this flaw to gain root privileges on a compromised system"
+  condition:
+    spawned_process and user.uid != 0 and proc.name=pkexec and proc.args = '' and proc.env icontains 'GCONV_PATH'
+  output:
+    "Detect Polkit pkexec Local Privilege Escalation Exploit (CVE-2021-4034) (user=%user.loginname uid=%user.loginuid command=%proc.cmdline access_args=%evt.args)"
+  priority: CRITICAL
+  tags: [process, mitre_privilege_escalation]
+
 # Application rules have moved to application_rules.yaml. Please look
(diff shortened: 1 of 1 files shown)


rules/falco_rules.yaml
```diff
@@ -3120,6 +3120,17 @@
   priority: NOTICE
   tags: [network, process, mitre_command_and_control]
 
+# This rule helps detect CVE-2021-4034:
+# A privilege escalation to root through memory corruption
+- rule: Polkit Local Privilege Escalation Vulnerability (CVE-2021-4034)
+  desc: "This rule detects an attempt to exploit a privilege escalation vulnerability in Polkit's pkexec. By running specially crafted code, a local user can leverage this flaw to gain root privileges on a compromised system"
+  condition:
+    spawned_process and user.uid != 0 and proc.name=pkexec and proc.args = '' and proc.env icontains 'GCONV_PATH'
```
jhoos, line 3128: Why `GCONV_PATH` and not just `PATH`? The way the original vulnerability was described (https://blog.qualys.com/vulnerabilities-threat-research/2022/01/25/pwnkit-local-privilege-escalation-vulnerability-discovered-in-polkits-pkexec-cve-2021-4034) it was a manipulation of `PATH`. I wonder if we even need to check `proc.env` at all; it seems that any invocation of `pkexec` with an empty set of args would be suspect, no? Or am I misunderstanding what this all means to falco?

darryk10: Hey @jhoos, thanks for the feedback. 
I added this condition due to the exploitability and GCONV_PATH can force iconv_open() to read another configuration file. Ee-introducing GCONV_PATH into pkexec's environment, and to execute our own shared library it's possible to get root. and exploit CVE-2021-4034.
Reference: https://www.qualys.com/2022/01/25/cve-2021-4034/pwnkit.txt

Kaizhe: Jason, I tend to agree with you here. The exploit relies on a malformed environment variables which is hard to detect during the process execution (no baseline could compare). And execute `pkexe` may be a normal case which may give you bunch of options (the same behavior as `-h` or `--help`). If what I assumed before is correct, I don't see there is an easy to write a rule to detect such exploit.

darryk10: Thanks @Kaizhe for your feedback. I changed the condition removing the proc.env check.
Thanks again.

Kaizhe (approved):

rules/falco_rules.yaml
```diff
@@ -3120,6 +3120,17 @@
...
   tags: [network, process, mitre_command_and_control]
 
+# This rule helps detect CVE-2021-4034:
+# A privilege escalation to root through memory corruption
+- rule: Polkit Local Privilege Escalation Vulnerability (CVE-2021-4034)
+  desc: "This rule detects an attempt to exploit a privilege escalation vulnerability in Polkit's pkexec. By running specially crafted code, a local user can leverage this flaw to gain root privileges on a compromised system"
+  condition:
+    spawned_process and user.uid != 0 and proc.name=pkexec and proc.args = ''
+  output:
+    "Detect Polkit pkexec Local Privilege Escalation Exploit (CVE-2021-4034) (user=%user.loginname uid=%user.loginuid command=%proc.cmdline args=%evt.args)"
```
leogr, line 3130:
```suggestion
    "Detect Polkit pkexec Local Privilege Escalation Exploit (CVE-2021-4034) (user=%user.loginname uid=%user.loginuid command=%proc.cmdline args=%proc.args)"
```

I guess you wanted to print the proc's arguments, right?

leogr (commented):
I've found an issue in the output, otherwise SGMT.

I'm also a bit concerned about false positives since this rule triggers just every time `pkexec` is called without args.

rules/falco_rules.yaml
anthony-alessi-acquia: I'm a bit confused whether 'args=%proc.args' would output anything useful. Wouldn't it just be an empty string? Should we try to print out the environment variable that is being leveraged?

Kaizhe:
> I've found an issue in the output, otherwise SGMT.
> 
> I'm also a bit concerned about false positives since this rule triggers just every time `pkexec` is called without args.

According to the fix, the `pkexe` will return error if runs without argument.

leogr (approved):
/approve

PR #1878, opened 2022-01-28
fix: force-set evt.type for plugin source events

FedeDP:
**What type of PR is this?**

/kind bug

/area engine

/area tests

This completes the changes of https://github.com/falcosecurity/falco/pull/1875, by introducing protective logic on for the `evttypes()` method for the edge case of sources coming from plugins and `k8s_audit`. This prevents rules written from those sources to be unwillingly excluded from the rule index, which caused them to never match on any event.

This also bumps the plugin versions to the latest released versions, both in the CMake dependencies an in the `aws_cloudtrail_rules.yaml` ruleset.

Finally, semver check for required plugin versions has been added.

Fixes #

```release-note
NONE
```

### DIFF
--- a/cmake/modules/plugins.cmake
+++ b/cmake/modules/plugins.cmake
@@ -17,8 +17,8 @@ string(TOLOWER ${CMAKE_HOST_SYSTEM_NAME} PLUGINS_SYSTEM_NAME)
 
 ExternalProject_Add(
   cloudtrail-plugin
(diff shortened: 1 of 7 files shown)


leogr:
Hoping this is the last set of last-minute fixes for this release

/milestone 0.31.0

Huge thanks to @FedeDP and @jasondellaluce :hugs:

leogr (approved):
/approve

jasondellaluce (approved):
/approve

Issue #1884, opened 2022-02-01
Informational Rules Loaded When priority: notice


mike-stewart:
**Describe the bug**

On falco 0.31.0 I am seeing Informational rules being loaded when the priority in the config is set to only load Notice and above.

**How to reproduce it**

I'm running falco.yaml with:
```
priority: notice
```

with no overrides for info-level rules. However, informational rules are still being loaded, and are alerting at the Notice level.

**Expected behaviour**

It is expected that when `priority: notice`, only Notice, Warning, Error, Critical, Alert, and Emergency rules will be loaded and fired.

**Screenshots**
<img width="430" alt="Screen Shot 2022-02-01 at 5 57 39 PM" src="https://user-images.githubusercontent.com/2521245/152058280-50a57a26-5fa3-46cf-908e-377d2be5e2e6.png">

Note that this rule is supposed to be INFO level: https://github.com/falcosecurity/falco/blob/f86423db76ce79917ae30499e17726e942a678b6/rules/falco_rules.yaml#L1886-L1894

**Environment**

- Falco version: 0.31.0

- System info:

- Cloud provider or hardware configuration:
- OS:

- Kernel: 

- Installation method: Kubernetes

**Additional context**

I'm wondering if this code may be the problem: https://github.com/falcosecurity/falco/blob/2f82a9baa193825337049e4ac69793f3074d6110/userspace/engine/lua/rule_loader.lua#L61-L66

Note that Notice and Informational are both set to level 5.

I can open a PR from https://github.com/falcosecurity/falco/compare/master...mike-stewart:patch-2 if that would be helpful.

jasondellaluce:
Great catch @mike-stewart ! Thank you for noticing, would you mind opening a PR?

PR #1885, opened 2022-02-02
Fix for informational rules loaded at notice level

mike-stewart:
**What type of PR is this?**

/kind bug

On falco 0.31.0 I am seeing Informational rules being loaded when the priority in the config is set to only load Notice and above. 

Should fix falcosecurity/falco#1884.

I've not yet had the opportunity to test this resolves the bug, but I'm opening this to get feedback.

```release-note
fix for rules at the informational level being loaded at the notice level.
```

### DIFF
--- a/userspace/engine/lua/rule_loader.lua
+++ b/userspace/engine/lua/rule_loader.lua
@@ -59,10 +59,10 @@ end
 
 -- Permissive for case and for common abbreviations.
 priorities = {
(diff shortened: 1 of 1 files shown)


jasondellaluce (approved):
Thank you for catching this!

/approve

leogr (approved):
/approve

PR #1886, opened 2022-02-03
Add cmdline opts library/initial falco application object

mstemm:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind rule-update

> /kind rule-create

> /area build

/area engine

> /area rules

> /area tests

> /area proposals

Fixes #

```release-note
Replace direct getopt_long() cmdline option parsing with third-party cxxopts library.
```

### DIFF
--- a/cmake/modules/cxxopts.cmake
+++ b/cmake/modules/cxxopts.cmake
@@ -1,0 +1,23 @@
+#
+# Copyright (C) 2022 The Falco Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+# the License. You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+#
+
+set(CXXOPTS_SRC "${PROJECT_BINARY_DIR}/cxxopts-prefix/src/cxxopts/")
+set(CXXOPTS_INCLUDE_DIR "${CXXOPTS_SRC}/include")
+
+ExternalProject_Add(
+        cxxopts
+        URL "https://github.com/jarro2783/cxxopts/archive/refs/tags/v3.0.0.tar.gz"
+        URL_HASH "SHA256=36f41fa2a46b3c1466613b63f3fa73dc24d912bc90d667147f1e43215a8c6d00"
+	CONFIGURE_COMMAND ""
+	BUILD_COMMAND ""
+	INSTALL_COMMAND "")
--- a/userspace/falco/app_cmdline_options.cpp
+++ b/userspace/falco/app_cmdline_options.cpp
@@ -1,0 +1,209 @@
+/*
+Copyright (C) 2022 The Falco Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "app_cmdline_options.h"
+#include "configuration.h"
+#include "config_falco.h"
+
+#include <fstream>
+
+namespace falco {
+namespace app {
+
+// Most bool member variables do not need to be set explicitly, as
+// they are bound to command line options that have default
+// values. However, a few options can be ifdef'd out so explicitly
+// initialize their linked variables.
+cmdline_options::cmdline_options()
+	: event_buffer_format(sinsp_evt::PF_NORMAL),
+	  list_plugins(false),
+	  m_cmdline_opts("falco", "Falco - Cloud Native Runtime Security")
+{
+	define();
+}
+
+cmdline_options::~cmdline_options()
+{
+}
+
+bool cmdline_options::parse(int argc, char **argv, std::string &errstr)
+{
+	try {
+		m_cmdline_parsed = m_cmdline_opts.parse(argc, argv);
+	}
+	catch (std::exception &e)
+	{
+		errstr = e.what();
+		return false;
+	}
+
+	// Some options require additional processing/validation
+	std::ifstream conf_stream;
+	if (!conf_filename.empty())
+	{
+		conf_stream.open(conf_filename);
+		if (!conf_stream.is_open())
+		{
+			errstr = std::string("Could not find configuration file at ") + conf_filename;
+			return false;
+		}
+	}
+	else
+	{
+#ifndef BUILD_TYPE_RELEASE
+		conf_stream.open(FALCO_SOURCE_CONF_FILE);
+		if (conf_stream.is_open())
+		{
+			conf_filename = FALCO_SOURCE_CONF_FILE;
+		}
+		else
+#endif
+		{
+			conf_stream.open(FALCO_INSTALL_CONF_FILE);
+			if (conf_stream.is_open())
+			{
+				conf_filename = FALCO_INSTALL_CONF_FILE;
+			}
+			else
+			{
+				// Note we do not return false here. Although there is
+				// no valid config file, some ways of running falco
+				// (e.g. --help, --list) do not need a config file.
+				//
+				// Later, when it comes time to read a config file, if
+				// the filename is empty we exit with an error.
+				conf_filename = "";
+			}
+		}
+	}
+
+	if(m_cmdline_parsed.count("b") > 0)
+	{
+		event_buffer_format = sinsp_evt::PF_BASE64;
+	}
+
+	// Expand any paths provided via -r and fill in rules_filenames
+	if(m_cmdline_parsed.count("r") > 0)
+	{
+		for(auto &path : m_cmdline_parsed["r"].as<std::vector<std::string>>())
+		{
+			falco_configuration::read_rules_file_directory(path, rules_filenames);
+		}
+	}
+
+	// Convert the vectors of enabled/disabled tags into sets to match falco engine API
+	if(m_cmdline_parsed.count("T") > 0)
+	{
+		for(auto &tag : m_cmdline_parsed["T"].as<std::vector<std::string>>())
+		{
+			disabled_rule_tags.insert(tag);
+		}
+	}
+
+	if(m_cmdline_parsed.count("t") > 0)
+	{
+		for(auto &tag : m_cmdline_parsed["t"].as<std::vector<std::string>>())
+		{
+			enabled_rule_tags.insert(tag);
+		}
+	}
+
+	// Some combinations of arguments are not allowed.
+
+	// You can't both disable and enable rules
+	if((disabled_rule_substrings.size() + disabled_rule_tags.size() > 0) &&
+	   enabled_rule_tags.size() > 0)
+	{
+		errstr = std::string("You can not specify both disabled (-D/-T) and enabled (-t) rules");
+		return false;
+	}
+
+	if (daemon && pidfilename == "") {
+		errstr = std::string("If -d is provided, a pid file must also be provided");
+		return false;
+	}
+
+	list_fields = m_cmdline_parsed.count("list") > 0 ? true : false;
+
+	return true;
+}
+
+std::string cmdline_options::usage()
+{
+	return m_cmdline_opts.help();
+}
+
+void cmdline_options::define()
+{
+	m_cmdline_opts.add_options()
+		("h,help",                        "Print this page", cxxopts::value(help)->default_value("false"))
+#ifdef BUILD_TYPE_RELEASE
+		("c",                             "Configuration file. If not specified uses " FALCO_INSTALL_CONF_FILE ".", cxxopts::value(conf_filename), "<path>")
+#else
+		("c",                             "Configuration file. If not specified tries " FALCO_SOURCE_CONF_FILE ", " FALCO_INSTALL_CONF_FILE ".", cxxopts::value(conf_filename), "<path>")
+#endif
+		("A",                             "Monitor all events, including those with EF_DROP_SIMPLE_CONS flag.", cxxopts::value(all_events)->default_value("false"))
+		("b,print-base64",                "Print data buffers in base64. This is useful for encoding binary data that needs to be used over media designed to.")
+		("cri",                           "Path to CRI socket for container metadata. Use the specified socket to fetch data from a CRI-compatible runtime. If not specified, uses libs default.", cxxopts::value(cri_socket_path), "<path>")
+		("d,daemon",                      "Run as a daemon.", cxxopts::value(daemon)->default_value("false"))
+		("disable-cri-async",             "Disable asynchronous CRI metadata fetching. This is useful to let the input event wait for the container metadata fetch to finish before moving forward. Async fetching, in some environments leads to empty fields for container metadata when the fetch is not fast enough to be completed asynchronously. This can have a performance penalty on your environment depending on the number of containers and the frequency at which they are created/started/stopped.", cxxopts::value(disable_cri_async)->default_value("false"))
+		("disable-source",                "Disable a specific event source. Available event sources are: syscall, k8s_audit, or any source from a configured source plugin. It can be passed multiple times. Can not disable all event sources.", cxxopts::value(disable_sources), "<event_source>")
+		("D",                             "Disable any rules with names having the substring <substring>. Can be specified multiple times. Can not be specified with -t.", cxxopts::value(disabled_rule_substrings), "<substring>")
+		("e",                             "Read the events from <events_file> (in .scap format for sinsp events, or jsonl for k8s audit events) instead of tapping into live.", cxxopts::value(trace_filename), "<events_file>")
+		("i",                             "Print all events that are ignored by default (i.e. without the -A flag) and exit.", cxxopts::value(print_ignored_events)->default_value("false"))
+#ifndef MINIMAL_BUILD
+		("k,k8s-api",                     "Enable Kubernetes support by connecting to the API server specified as argument. E.g. \"http://admin:password@127.0.0.1:8080\". The API server can also be specified via the environment variable FALCO_K8S_API.", cxxopts::value(k8s_api), "<url>")
+		("K,k8s-api-cert",                "Use the provided files names to authenticate user and (optionally) verify the K8S API server identity. Each entry must specify full (absolute, or relative to the current directory) path to the respective file. Private key password is optional (needed only if key is password protected). CA certificate is optional. For all files, only PEM file format is supported. Specifying CA certificate only is obsoleted - when single entry is provided for this option, it will be interpreted as the name of a file containing bearer token. Note that the format of this command-line option prohibits use of files whose names contain ':' or '#' characters in the file name.", cxxopts::value(k8s_api_cert), "(<bt_file> | <cert_file>:<key_file[#password]>[:<ca_cert_file>])")
+		("k8s-node",                      "The node name will be used as a filter when requesting metadata of pods to the API server. Usually, it should be set to the current node on which Falco is running. If empty, no filter is set, which may have a performance penalty on large clusters.", cxxopts::value(k8s_node_name), "<node_name>")
+#endif
+		("L",                             "Show the name and description of all rules and exit.", cxxopts::value(describe_all_rules)->default_value("false"))
+		("l",                             "Show the name and description of the rule with name <rule> and exit.", cxxopts::value(describe_rule), "<rule>")
+		("list",                          "List all defined fields. If <source> is provided, only list those fields for the source <source>. Current values for <source> are \"syscall\", \"k8s_audit\", or any source from a configured source plugin.", cxxopts::value(list_source_fields)->implicit_value(""), "<source>")
+#ifndef MUSL_OPTIMIZED
+		("list-plugins",                  "Print info on all loaded plugins and exit.", cxxopts::value(list_plugins)->default_value("false"))
+#endif
+#ifndef MINIMAL_BUILD
+		("m,mesos-api",                   "Enable Mesos support by connecting to the API server specified as argument. E.g. \"http://admin:password@127.0.0.1:5050\". Marathon url is optional and defaults to Mesos address, port 8080. The API servers can also be specified via the environment variable FALCO_MESOS_API.", cxxopts::value(mesos_api), "<url[,marathon_url]>")
+#endif
+		("M",                             "Stop collecting after <num_seconds> reached.", cxxopts::value(duration_to_tot)->default_value("0"), "<num_seconds>")
+		("N",                             "When used with --list/--list-source, only print field names.", cxxopts::value(names_only)->default_value("false"))
+		("o,option",                      "Set the value of option <opt> to <val>. Overrides values in configuration file. <opt> can be identified using its location in configuration file using dot notation. Elements which are entries of lists can be accessed via square brackets [].\n    E.g. base.id = val\n         base.subvalue.subvalue2 = val\n         base.list[1]=val", cxxopts::value(cmdline_config_options), "<opt>=<val>")
+		("p,print",                       "Add additional information to each falco notification's output.", cxxopts::value(print_additional), "<output_format>")
+		("pcontainer",                    "Add additional information to each falco notification's output, using a container-friendly format. Specifying -pc/-pcontainer will change the interpretation of %%container.info in rule output fields.", cxxopts::value(print_container)->default_value("false"))
+		("pc",                            "Short for -pcontainer", cxxopts::value(print_container)->default_value("false"))
+		("pkubernetes",                   "Add additional information to each falco notification's output, using a kubernetes-friendly format. Additionally, specifying -pk/-pkubernetes will change the interpretation of %%container.info in rule output fields.", cxxopts::value(print_kubernetes)->default_value("false"))
+		("pk",                            "Short for -pkubernetes", cxxopts::value(print_kubernetes)->default_value("false"))
+		("pmesos",                        "Add additional information to each falco notification's output, using a mesos-friendly format. Additionally, specifying -pm/-pmesos will change the interpretation of %%container.info in rule output fields.", cxxopts::value(print_mesos)->default_value("false"))
+		("pm",                            "Short for -pmesos", cxxopts::value(print_mesos)->default_value("false"))
+		("P,pidfile",                     "When run as a daemon, write pid to specified file", cxxopts::value(pidfilename)->default_value("/var/run/falco.pid"), "<pid_file>")
+		("r",                             "Rules file/directory (defaults to value set in configuration file, or /etc/falco_rules.yaml). Can be specified multiple times to read from multiple files/directories.", cxxopts::value<std::vector<std::string>>(), "<rules_file>")
+		("s",                             "If specified, append statistics related to Falco's reading/processing of events to this file (only useful in live mode).", cxxopts::value(stats_filename), "<stats_file>")
+		("stats-interval",                "When using -s <stats_file>, write statistics every <msec> ms. This uses signals, so don't recommend intervals below 200 ms. Defaults to 5000 (5 seconds).", cxxopts::value(stats_interval)->default_value("5000"),  "<msec>")
+		("S,snaplen",                     "Capture the first <len> bytes of each I/O buffer. By default, the first 80 bytes are captured. Use this option with caution, it can generate huge trace files.", cxxopts::value(snaplen)->default_value("0"), "<len>")
+		("support",                       "Print support information including version, rules files used, etc. and exit.", cxxopts::value(print_support)->default_value("false"))
+		("T",                             "Disable any rules with a tag=<tag>. Can be specified multiple times. Can not be specified with -t", cxxopts::value<std::vector<std::string>>(), "<tag>")
+		("t",                             "Only run those rules with a tag=<tag>. Can be specified multiple times. Can not be specified with -T/-D.", cxxopts::value<std::vector<std::string>>(), "<tag>")
+		("U,unbuffered",                  "Turn off output buffering to configured outputs. This causes every single line emitted by falco to be flushed which generates higher CPU usage but is useful when piping those outputs into another process or into a script.", cxxopts::value(unbuffered_outputs)->default_value("false"))
+		("u,userspace",                   "Parse events from userspace. To be used in conjunction with the ptrace(2) based driver (pdig)", cxxopts::value(userspace)->default_value("false"))
+		("V,validate",                    "Read the contents of the specified rules(s) file and exit. Can be specified multiple times to validate multiple files.", cxxopts::value(validate_rules_filenames), "<rules_file>")
+		("v",                             "Verbose output.", cxxopts::value(verbose)->default_value("false"))
+		("version",                       "Print version number.", cxxopts::value(print_version_info)->default_value("false"));
+
+	m_cmdline_opts.set_width(140);
+}
+
+}; // namespace app
+}; // namespace falco
--- a/userspace/falco/app_cmdline_options.h
+++ b/userspace/falco/app_cmdline_options.h
@@ -1,0 +1,88 @@
+/*
+Copyright (C) 2022 The Falco Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#pragma once
+
+#include <event.h>
+
+#include <cxxopts.hpp>
+
+#include <string>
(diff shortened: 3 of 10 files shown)


cmake/modules/cxxopts.cmake
```diff
@@ -0,0 +1,23 @@
+#
+# Copyright (C) 2021 The Falco Authors.
```
leogr, line 2:
```suggestion
# Copyright (C) 2022 The Falco Authors.
```

userspace/engine/falco_engine.cpp
```diff
@@ -91,7 +91,7 @@ void falco_engine::list_fields(std::string &source, bool verbose, bool names_onl
 	// applicable to multiple event sources.
 	for(auto &it : m_filter_factories)
 	{
-		if(source != "" && source != it.first)
+		if(source != "all" && source != it.first)
```
leogr, line 94: Honestly, I don't like the idea to introduce a kind of textual constant (i.e., `all`). I would prefer the empty string as before.

userspace/falco/application.cpp
```diff
@@ -0,0 +1,61 @@
...
+	{
+		return false;
+	}
+
+	m_initialized = true;
+	return true;
+}
+
+}; // namespace app
+}; // namespace falco {
```
leogr, line 61:
```suggestion
}; // namespace falco
```

userspace/falco/application.h
```diff
@@ -0,0 +1,53 @@
...
+	bool init(int argc, char **argv, std::string &errstr);
+
+private:
+
+	cmdline_options m_cmdline_options;
+	bool m_initialized;
+};
+
+}; // namespace app {
+}; // namespace falco {
```
leogr, line 53:
```suggestion
}; // namespace falco
```

leogr (commented):
I took a quick look. I like the idea, but I don't know `cxxopts`, so I need some time to play with it a bit :) 

Meanwhile, I have left some (not really important) suggestions.

userspace/falco/application.h
```diff
@@ -0,0 +1,53 @@
...
+
+// For now, it is only responsible for command line options.
+#pragma once
+
+#include "app_cmdline_options.h"
+
+#include <string>
+
+namespace falco {
+namespace app {
```
jasondellaluce, line 35: This is a matter of taste, but if there's not specific need for this namespace pattern I would prefer having a `falco::application` and `falco::cmdline_options` instead, which is less verbose. And since we are here, I think it would be worthy to be consistent with it for the falco engine and other parts of the code too. This would lead to a better naming consistency and less verbose code in each part of the codebase. By that, I mean some renamings like:
- `falco_engine` -> `falco::engine`
- `falco_formats` -> `falco::formats`
- `falco_outputs` -> `falco::outputs`
- ...

This might need be addressed in future PRs, but let's have a discussion about this!

userspace/falco/falco.cpp
```diff
@@ -564,257 +450,49 @@ int falco_init(int argc, char **argv)
...
 		// If required, set the snaplen
 		//
-		if(snaplen != 0)
+		if(app.copts().snaplen != 0)
 		{
-			inspector->set_snaplen(snaplen);
+			inspector->set_snaplen(app.copts().snaplen);
 		}
 
 		if(print_ignored_events)
```
jasondellaluce, line 498: Where is this initialized, and why is it not part of `app.copts()`?

userspace/falco/application.h
```diff
@@ -0,0 +1,53 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// The falco "app" will eventually replace the monolithic code in
+// falco.cpp. We expect it will be responsible for the following:
+//  - Parsing/validating command line options
+//  - Parsing/validing falco config
+//  - Initialize prerequsites (inspector, falco engine, webserver, etc)
```
jasondellaluce, line 21: Always for testing/modularity purposes, I think I would advocate to do dependency injection here (if possible) and provide the code dependencies from outside.

userspace/falco/application.h
```diff
@@ -0,0 +1,53 @@
...
+//  - Loading plugins
```
jasondellaluce, line 22: Why not having a `plugin_manager` class for this, in future PRs?

userspace/falco/falco.cpp
```diff
@@ -838,73 +517,30 @@ int falco_init(int argc, char **argv)
...
-		if (conf_filename.size())
-		{
-			conf_stream.open(conf_filename);
-			if (!conf_stream.is_open())
-			{
-				throw std::runtime_error("Could not find configuration file at " + conf_filename);
-			}
-		}
-		else
+		// XXX/mstemm technically this isn't right, you could disable syscall *and* k8s_audit and configure a plugin.
```
jasondellaluce, line 525: True. Maybe we might move this check after the plugin loading logic, and re-populate this array there if a source plugin is loaded. Besides, this is a reminder for ourselves that we might want to re-factor this logic once `k8s_audit` gets moved out of Falco to become a plugin. From that moment on, I think it would not make sense to have an `enabled_sources` array and a `disable source` option, as Falco would support one of syscalls and plugin source with mutual exclusion.

userspace/falco/falco.cpp
```diff
@@ -1028,7 +664,7 @@ int falco_init(int argc, char **argv)
 
 		std::list<sinsp_plugin::info> infos = sinsp_plugin::plugin_infos(inspector);
 
-		if(list_plugins)
+		if(app.copts().list_plugins)
```
jasondellaluce, line 667: As for the comment above, this would be a good fit for a `plugin_manager` class in the future (maybe in future PRs).

jasondellaluce (commented):
I left you some comments. I like the direction this is going to!

userspace/falco/app_cmdline_options.h
```diff
@@ -0,0 +1,84 @@
...
+	bool print_support;
+	std::set<std::string> disabled_rule_tags;
+	std::set<std::string> enabled_rule_tags;
+	bool unbuffered_outputs;
+	bool userspace;
+	std::vector<std::string> validate_rules_filenames;
+	bool verbose;
+	bool print_version_info;
+
+	bool parse(int argc, char **argv, std::string &errstr);
```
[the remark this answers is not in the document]
LucaGuerra: I'm happy to see these changes that allow testing and I believe that the biggest advantage of splitting up large functions is to be able to test.
I have a question: what would the test use case be for mocking `parse()`, `usage()` (and possibly `get_option()`) in this case?
From what I see the `parse()` method acts as a "constructor" of sorts that fills the data in the struct based on CLI strings. I'd expect to either exercise the function itself (so, providing a bunch of strings and verifying that the resulting options are right) or to exercise something else that uses the options in which case I would either construct a new `cmdline_options` struct and then sending the command line strings or individually setting the options I need. In these cases I wouldn't mock `parse()` as I could use the regular function.

userspace/falco/falco.cpp
leogr: Btw, I believe we should port the current `k8s_audit` impl. to a plugin __before__ merging this PR. 
That would solve several micro issues like the above one.

userspace/falco/app_cmdline_options.h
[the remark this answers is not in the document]
jasondellaluce: That's a good point. My concern was being able to test `falco::app::application`, and trying to remove unhandy constraints from this class on which it depends. But you're right, in this case is quite simple and maybe mocking would be an over-engineered choice!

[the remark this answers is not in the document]
LucaGuerra: > My concern was being able to test `falco::app::application`, and trying to remove unhandy constraints
A noble concern indeed! 👍 

> maybe mocking would be an over-engineered choice
Experience tells us that mocks tend to break over time and I think the less mocks we have the more of our code we can test!

cmake/modules/cxxopts.cmake
mstemm: Thanks, fixed.

userspace/engine/falco_engine.cpp
mstemm: Sure, I was on the fence. Switched it back.

userspace/falco/application.cpp
mstemm: Ok, I changed all of them to not have a trailing { for the sake of consistency.

userspace/falco/application.h
mstemm: See above, updated all.

userspace/falco/app_cmdline_options.cpp
```diff
@@ -0,0 +1,190 @@
...
+#include "config_falco.h"
+
+#include <fstream>
+
+namespace falco {
+namespace app {
+
+cmdline_options::cmdline_options()
+	: event_buffer_format(sinsp_evt::PF_NORMAL),
+	  m_cmdline_opts("Falco", "Container/Cloud-friendly runtime security tool")
```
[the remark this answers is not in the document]
mstemm: I changed it to "Cloud Native Runtime Security" to match the Slogan on the branding page. (I preferred the -friendly part as falco works fine in non cloud environments, but I'll keep things consistent).

userspace/falco/app_cmdline_options.h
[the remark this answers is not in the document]
mstemm: Ok, no changes here then.

userspace/falco/application.h
```diff
@@ -0,0 +1,53 @@
...
+namespace falco {
+namespace app {
+
+class application {
+public:
+
+	application();
+	virtual ~application();
+
+	cmdline_options &copts();
```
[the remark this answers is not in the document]
mstemm: I wanted something short, but I guess options() isn't that much longer, and as we add more to the application object this won't be used in falco proper, so sure.

userspace/falco/application.h
mstemm: I actually like having the application-related objects (application and cmdline_options for now, more to come later) in their own namespace. For example the engine probably *won't* end up below the application-related namespace because it's not really responsible for setting up and managing falco program behavior. I'm not sure formats and outputs would either, although that is less clear to me--it might be a new object that manages formatting/outputs and wraps the falco::formats/falco::outputs objects.

I did make it `falco::app` to keep it concise.

userspace/falco/falco.cpp
mstemm: Most of the options have a `->default_value(xxx)` call which sets the default value for the option. Exceptions are for options that are bound to strings that can't be used without the option being specified (e.g. --cri <path> to specify an alternate cri socket, --disable-source to disable a source), and --list_plugins, which may be commented out for MUSL_OPTIMIZED builds.

The default for snaplen is 0 (https://github.com/falcosecurity/falco/blob/add-cmdline-opts-library/userspace/falco/app_cmdline_options.cpp#L181) which matches what it was set to for falco before this PR (https://github.com/falcosecurity/falco/blob/8d9dd4440fa9e28911342d6e668f62062b263fff/userspace/falco/falco.cpp#L532)

userspace/falco/application.h
mstemm: Sure, when we get to that we can find a good way to organize it.

userspace/falco/application.h
mstemm: Yeah I think we will have this. Possibly moved to libs.

userspace/falco/falco.cpp
mstemm: Yeah, I thought a good time to revisit this would be when we make initialization an application object instead of just dumped in main. I don't think it's very likely that somebody would want to run falco with plugins and also disable all other sources.

I really don't agree about waiting for k8s_audit being a plugin first, though. There are a few other things we will need to figure out to move k8s_audit to a plugin:
1) Are we going to keep just the webhook-based approach, leaving most of the configuration/setup outside of falco? Or should we have separate k8s_audit plugins for distros like eks, gke, etc that have ways to read and transform proprietary logs into audit events (https://github.com/sysdiglabs/stackdriver-webhook-bridge, https://github.com/sysdiglabs/ekscloudwatch). There may even be a way to do this in a distro-independent way by hooking in at the admission controller.
2) We will need to figure out how to run falco with multiple sets of source plugins running at once, or with a combination of syscalls and plugin events at the same time. Remember that currently, if a source plugin is configured, other events are disabled.
3) We need to decide whether to update k8s_audit_rules.yaml to be more plugin friendly, or to strive for 100% compatibility with the existing rules file, even if we have to change the plugin framework. It *might* be the case that the json plugin can handle the jevt.xxx fields currently used in k8s_audit_rules.yaml, but we'd want to do some testing to make sure.
4) The event model and fields support used for k8s_audit events is very slightly different than sinsp events. For example, for jevt events, a field can actually return a list of items rather than a single item. For example, `ka.req.pod.containers.privileged` returns a list of boolean values for all containers in a pod. I don't believe the sinsp filtercheck model supports this right now, so we'll need to add it.

To be clear, I'm 100% for converting k8s audit support into a plugin so we can get rid of the parallel implementation of events in userspace/engine/json_evt.cpp. A plugin-based approach will be way cleaner and better. I'm just pointing out that it's a lot of work and I wouldn't want to hold this up to wait for it.

userspace/falco/falco.cpp
mstemm: Agreed!

mstemm:
One public-visible change is that --list, which used to support an optional argument, has been broken into two options. --list now always displays all fields, and a new option --list-source displays fields for a specific source. This is because the cxxopts library has a bug (https://github.com/jarro2783/cxxopts/issues/104, https://github.com/jarro2783/cxxopts/issues/211) that does not parse implicit options correctly when specified as "--list <source>".

userspace/falco/falco.cpp
leogr: I believe the k8s_audit plugin will be ready very soon. Moreover, I guess the changeset to clean up the old k8s implementation will be quicker to review than this PR.

Also, to reply to your points:

1. We will likely provide the webhook-based approach as a plugin, along with other plugins, one for each distro (eks, gke, etc..). @Issif, @jasondellaluce, and I have done some experiments, and we will propose something soon, I hope :smile_cat:  Moreover, I'm pretty sure the admission controller approach cannot work to replace the current functionality (since not all the information available in the audit log is also provided to the admission controller. One notable example is the API response object.
2. Totally agree with that. But I guess that's something we cannot fix soon, isn't it?
3. I would like to keep `k8s_audit_rules.yaml` as is now (at least, I would not introduce any backward-incompatible change). It seems to me that ~100% is possible, but I agree we need to do some testing.
4. I'm aware of that issue, and fixing that should be prioritized.


That being said, my advice to wait is not really a blocker, though I still think it would be better.

test/falco_tests.yaml
```diff
@@ -651,7 +651,7 @@ trace_files: !mux
 
   disabled_and_enabled_rules_1:
     exit_status: 1
-    stderr_contains: "Runtime error: You can not specify both disabled .-D/-T. and enabled .-t. rules. Exiting."
+    stderr_contains: "Runtime error: Could not initialize: You can not specify both disabled .-D/-T. and enabled .-t. rules. Exiting."
```
jasondellaluce, line 654: Personally, I don't see a great reason to change the error output now with a prefix. This is not strictly relevant for this contribution, and also causes a user-facing change. I would advocate to leave the errors as is, and consider changing the format in the future.

userspace/falco/app_cmdline_options.cpp
```diff
@@ -0,0 +1,196 @@
...
+		("disable-source",                "Disable a specific event source. Available event sources are: syscall, k8s_audit, or any source from a configured source plugin. It can be passed multiple times. Can not disable all event sources.", cxxopts::value(disable_sources), "<event_source>")
+		("D",                             "Disable any rules with names having the substring <substring>. Can be specified multiple times. Can not be specified with -t.", cxxopts::value(disabled_rule_substrings), "<substring>")
+		("e",                             "Read the events from <events_file> (in .scap format for sinsp events, or jsonl for k8s audit events) instead of tapping into live.", cxxopts::value(trace_filename), "<events_file>")
```
jasondellaluce, line 152: Is `jsonl` a typo?

userspace/falco/falco.cpp
jasondellaluce: I'm not sure I'm understanding well, how this is related to `print_ignored_events`?

userspace/falco/falco.cpp
jasondellaluce: I just opened a PR in libs trying to fix point #4 👉🏼 https://github.com/falcosecurity/libs/pull/200

I generally agree with @leogr, but in this case I would accept merging this before moving k8s-audit, and think about this discussion later.

userspace/falco/falco.cpp
mstemm: Oh sorry--I thought you meant snaplen. I missed this, it used to be the `-i` option. I'll add it.

test/falco_tests.yaml
mstemm: The reason for the "Could not initialize" is that there's an extra layer of functions now, specifically the application object and init. I can remove any prefix from that layer, and just pass lower level errors as-is.

userspace/falco/app_cmdline_options.cpp
mstemm: No, jsonl is a "real" format: https://jsonlines.org/. It's just a collection of json objects, each on their own line, separated by newlines.

leogr (changes requested):
Hey @mstemm,

I really like this refactoring and the way the code is organized! 

I have found some easy-to-fix issues, see my suggestions below.

Most of them are in the spirit of not changing the current UX, since I see this PR as a refactoring operation only.

Furthermore, regarding your comment :point_down:, 

> One public-visible change is that --list, which used to support an optional argument, has been broken into two options. --list now always displays all fields, and a new option --list-source displays fields for a specific source. This is because the cxxopts library has a bug ([jarro2783/cxxopts#104](https://github.com/jarro2783/cxxopts/issues/104), [jarro2783/cxxopts#211](https://github.com/jarro2783/cxxopts/issues/211)) that does not parse implicit options correctly when specified as "--list ".

Not sure if that is a bug. It seems to be the expected behavior. 

As per cxxopts docs, when `--list` has an implicit value, one must use `=` to override the implicit value. For example, assuming the implicit value is an empty string, you will get:
- `--list` :point_right: an empty string
- `--list something` :point_right: still an empty string (note something is considered another argument)
- `--list=syscall` :point_right: `syscall`

Perhaps by coincidence (or by design) the current Falco 0.31.0 (the released version) has exactly the same behavior described above (which is the same of cxxopts when using implicit values).

Try yourself `falco --list=k8s_audit` vs. `falco --list k8s_audit`. The first will work return ka fields only. The second will return all fields.

Here is a small patch (on top of your PR) that emulates the behavior of the previous implementation. 

```diff
diff --git a/userspace/falco/app_cmdline_options.cpp b/userspace/falco/app_cmdline_options.cpp
index 2e83a758..9c7fe6e5 100644
--- a/userspace/falco/app_cmdline_options.cpp
+++ b/userspace/falco/app_cmdline_options.cpp
@@ -129,6 +129,8 @@ bool cmdline_options::parse(int argc, char **argv, std::string &errstr)
 		return false;
 	}
 
+	list_fields = m_cmdline_parsed.count("list") > 0 ? true : false;
+
 	return true;
 }
 
@@ -158,8 +160,7 @@ void cmdline_options::define()
 #endif
 		("L",                             "Show the name and description of all rules and exit.", cxxopts::value(describe_all_rules)->default_value("false"))
 		("l",                             "Show the name and description of the rule with name <rule> and exit.", cxxopts::value(describe_rule), "<rule>")
-		("list",                          "List all defined fields.", cxxopts::value(list_all_fields))
-		("list-source",                   "List all defined fields for the provided source. Current values for <source> are \"syscall\", \"k8s_audit\", or any source from a configured source plugin", cxxopts::value(list_source_fields)->default_value("none"), "<source>")
+		("list",                          "List all defined fields. If <source> is provided, only list those fields for the source <source>. Current values for <source> are \"syscall\", \"k8s_audit\".", cxxopts::value(list_source_fields)->implicit_value(""), "<source>")
 #ifndef MUSL_OPTIMIZED
 		("list-plugins",                  "Print info on all loaded plugins and exit.", cxxopts::value(list_plugins)->default_value("false"))
 #endif
diff --git a/userspace/falco/app_cmdline_options.h b/userspace/falco/app_cmdline_options.h
index e28b83dc..37aa954f 100644
--- a/userspace/falco/app_cmdline_options.h
+++ b/userspace/falco/app_cmdline_options.h
@@ -49,7 +49,7 @@ public:
 	bool describe_all_rules;
 	std::string describe_rule;
 	bool print_ignored_events;
-	bool list_all_fields;
+	bool list_fields;
 	std::string list_source_fields;
 	bool list_plugins;
 	std::string mesos_api;
diff --git a/userspace/falco/falco.cpp b/userspace/falco/falco.cpp
index abacc13e..09a3e495 100644
--- a/userspace/falco/falco.cpp
+++ b/userspace/falco/falco.cpp
@@ -690,14 +690,8 @@ int falco_init(int argc, char **argv)
 			return EXIT_SUCCESS;
 		}
 
-		if(app.options().list_all_fields)
-		{
-			std::string all_sources = "";
-			list_source_fields(engine, app.options().verbose, app.options().names_only, all_sources);
-			return EXIT_SUCCESS;
-		}
 
-		if(app.options().list_source_fields != "none")
+		if(app.options().list_fields) 
 		{
 			list_source_fields(engine, app.options().verbose, app.options().names_only, app.options().list_source_fields);
 			return EXIT_SUCCESS;

```
N.B.: I have tried it a bit, it seems that it perfectly emulates the behavior of the latest released version of Falco. Let me know if it works for you too!!!

mstemm:
> Furthermore, regarding your comment 👇
> 
> > One public-visible change is that --list, which used to support an optional argument, has been broken into two options. --list now always displays all fields, and a new option --list-source displays fields for a specific source. This is because the cxxopts library has a bug ([jarro2783/cxxopts#104](https://github.com/jarro2783/cxxopts/issues/104), [jarro2783/cxxopts#211](https://github.com/jarro2783/cxxopts/issues/211)) that does not parse implicit options correctly when specified as "--list ".
....
> N.B.: I have tried it a bit, it seems that it perfectly emulates the behavior of the latest released version of Falco. Let me know if it works for you too!!!

Thanks for doing the extra debugging and the patch! I went back to a single option, where `--list=<source>` only works with equals.

userspace/falco/app_cmdline_options.cpp
```diff
@@ -0,0 +1,197 @@
...
+		("t",                             "Only run those rules with a tag=<tag>. Can be specified multiple times. Can not be specified with -T/-D.", cxxopts::value<std::vector<std::string>>(), "<tag>")
+		("U,unbuffered",                   "Turn off output buffering to configured outputs. This causes every single line emitted by falco to be flushed which generates higher CPU usage but is useful when piping those outputs into another process or into a script.", cxxopts::value(unbuffered_outputs)->default_value("false"))
+		("u,userspace",                   "Parse events from userspace. To be used in conjunction with the ptrace(2) based driver (pdig)", cxxopts::value(userspace)->default_value("false"))
+		("V,validate",                    "Read the contents of the specified rules(s) file and exit. Can be specified multiple times to validate multiple files.", cxxopts::value(validate_rules_filenames), "<rules_file>")
+		("v",                              "Verbose output.", cxxopts::value(verbose)->default_value("false"))
```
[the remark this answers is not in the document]
mstemm: Ok fixed all these space typos.

userspace/falco/app_cmdline_options.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+void cmdline_options::define()
+{
+	m_cmdline_opts.add_options()
+		("h,help",                        "Print this page", cxxopts::value(help)->default_value("false"))
+		("c",                             "Configuration file. If not specified tries " FALCO_SOURCE_CONF_FILE ", " FALCO_INSTALL_CONF_FILE, cxxopts::value(conf_filename), "<path>")
+		("A",                             "Monitor all events, including those with EF_DROP_SIMPLE_CONS flag.", cxxopts::value(all_events)->default_value("false"))
+		("b,print-base64",                "Print data buffers in base64. This is useful for encoding binary data that needs to be used over media designed to.")
+		("cri",                           "Path to CRI socket for container metadata. Use the specified socket to fetch data from a CRI-compatible runtime. If not specified, uses libs default.", cxxopts::value(cri_socket_path), "<path>")
+		("d,daemon",                       "Run as a daemon.", cxxopts::value(daemon)->default_value("false"))
```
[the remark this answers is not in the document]
mstemm: Ok fixed all these space typos.

userspace/falco/app_cmdline_options.cpp
```diff
@@ -0,0 +1,197 @@
...
+		("S,snaplen",                     "Capture the first <len> bytes of each I/O buffer. By default, the first 80 bytes are captured. Use this option with caution, it can generate huge trace files.", cxxopts::value(snaplen)->default_value("0"), "<len>")
+		("support",                       "Print support information including version, rules files used, etc. and exit.", cxxopts::value(print_support)->default_value("false"))
+		("T",                             "Disable any rules with a tag=<tag>. Can be specified multiple times. Can not be specified with -t", cxxopts::value<std::vector<std::string>>(), "<tag>")
+		("t",                             "Only run those rules with a tag=<tag>. Can be specified multiple times. Can not be specified with -T/-D.", cxxopts::value<std::vector<std::string>>(), "<tag>")
+		("U,unbuffered",                   "Turn off output buffering to configured outputs. This causes every single line emitted by falco to be flushed which generates higher CPU usage but is useful when piping those outputs into another process or into a script.", cxxopts::value(unbuffered_outputs)->default_value("false"))
```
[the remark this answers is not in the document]
mstemm: Ok fixed all these space typos.

userspace/falco/app_cmdline_options.cpp
```diff
@@ -0,0 +1,197 @@
...
+namespace app {
+
+// Most bool member variables do not need to be set explicitly, as
+// they are bound to command line options that have default
+// values. However, a few options can be ifdef'd out so explicitly
+// initialize their linked variables.
+cmdline_options::cmdline_options()
+	: event_buffer_format(sinsp_evt::PF_NORMAL),
+	  list_plugins(false),
+	  m_cmdline_opts("Falco", "Cloud Native Runtime Security")
```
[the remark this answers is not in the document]
mstemm: Sure, that summary sounds better. Updated.

userspace/falco/app_cmdline_options.cpp
```diff
@@ -0,0 +1,197 @@
...
+			{
+				conf_filename = FALCO_INSTALL_CONF_FILE;
+			}
+			else
+			{
+				errstr = std::string("You must create a config file at ")  + FALCO_SOURCE_CONF_FILE + ", " + FALCO_INSTALL_CONF_FILE + " or by passing -c";
+				return false;
+			}
+		}
+	}
```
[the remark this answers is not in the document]
mstemm: Thanks for pointing this out, I brought over the behavior that changes --help output and the order of trying files to app_cmdline_options.cpp.

userspace/falco/falco.cpp
```diff
@@ -564,268 +449,61 @@ int falco_init(int argc, char **argv)
...
-				{
-					if(optarg != NULL)
-					{
-						disable_sources.insert(optarg);
-					}
-				}
-				break;
+		if(!successful)
+		{
+			throw falco_exception(errstr);
```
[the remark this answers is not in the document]
mstemm: Sure, I went back to simply printing to stderr and returning from falco_init.

userspace/falco/app_cmdline_options.cpp
```diff
@@ -0,0 +1,197 @@
...
+#ifndef MUSL_OPTIMIZED
+		("list-plugins",                  "Print info on all loaded plugins and exit.", cxxopts::value(list_plugins)->default_value("false"))
+#endif
+#ifndef MINIMAL_BUILD
+		("m,mesos-api",                   "Enable Mesos support by connecting to the API server specified as argument. E.g. \"http://admin:password@127.0.0.1:5050\". Marathon url is optional and defaults to Mesos address, port 8080. The API servers can also be specified via the environment variable FALCO_MESOS_API.", cxxopts::value(mesos_api), "<url[,marathon_url]>")
+#endif
+		("M",                             "Stop collecting after <num_seconds> reached.", cxxopts::value(duration_to_tot)->default_value("0"), "<num_seconds>")
+		("N",                             "When used with --list/--list-source, only print field names.", cxxopts::value(names_only)->default_value("false"))
+		("o,option",                      "Set the value of option <key> to <val>. Overrides values in configuration file. <key> can be a two-part <key>.<subkey>", cxxopts::value(cmdline_config_options), "(<key>=<val>|<key.subkey>=<val>)")
```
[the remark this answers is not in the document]
mstemm: Sure, moved the updated description over.

userspace/falco/configuration.cpp
```diff
@@ -402,6 +402,11 @@ void falco_configuration::set_cmdline_option(const string &opt)
 {
 	pair<string, string> keyval;
 
+	if(opt.empty())
+	{
+		return;
+	}
+
```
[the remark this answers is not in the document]
mstemm: I guess I ran into it when doing early testing, but I double-checked and we can keep this behavior without causing regression test failures. I removed this change.

userspace/falco/app_cmdline_options.cpp
```diff
@@ -0,0 +1,209 @@
...
+		for(auto &tag : m_cmdline_parsed["t"].as<std::vector<std::string>>())
+		{
+			enabled_rule_tags.insert(tag);
+		}
+	}
+
+	// Some combinations of arguments are not allowed.
+
+	// You can't both disable and enable rules
+	if((disabled_rule_substrings.size() + disabled_rule_tags.size() > 0) &&
```
jasondellaluce, line 127: Maybe I'm missing something, but you seem to initialize `disable_rule_substrings` differently. The first is initialized as a single value with `cxxopts::value(disabled_rule_substrings)` even though it's a vector, whereas the second is initialized manually value-by-value [in this method at line 108](https://github.com/falcosecurity/falco/pull/1886/files#diff-e8ce29b2b289869d198bfb7de910426be6c5c226e4ee26cdc9f42eed9c68d787R108). Is there a reason for this difference, or is `disabled_rule_substrings` supposed to be inizialized manually too?

userspace/falco/falco.cpp
```diff
@@ -1472,66 +1095,67 @@ int falco_init(int argc, char **argv)
...
 			}
-			inspector->init_k8s_client(k8s_api, k8s_api_cert, k8s_node_name, verbose);
-			k8s_api = 0;
-			k8s_api_cert = 0;
+			inspector->init_k8s_client(k8s_api_ptr, k8s_api_cert_ptr, k8s_node_name_ptr, app.options().verbose);
 		}
 		else if(char* k8s_api_env = getenv("FALCO_K8S_API"))
 		{
-			if(k8s_api_env != NULL)
+			// Create string pointers for some config vars
```
jasondellaluce, line 1118: At first glance, this looks like the exact same [block at line 1100](https://github.com/falcosecurity/falco/pull/1886/files#diff-67ae49cc953e68396a677916d8dce080ef900deede8f39489ae611dce4b3cab2R1100). If that's so, it would be great to handle the if-else `k8s_api` string init logic before, and then have a single block for initializing the k8s client.

userspace/falco/falco.cpp
```diff
@@ -1438,7 +1061,7 @@ int falco_init(int argc, char **argv)
 			catch(sinsp_exception &e)
 			{
 				// If syscall input source is enabled and not through userspace instrumentation
-				if (!disable_syscall && !userspace)
+				if (enabled_sources.find(syscall_source) == enabled_sources.end() && !app.options().userspace)
```
jasondellaluce, line 1064: I may be wrong at this, but does not finding a source in the `enabled_sources` mean that the source is disabled? Shouldn't this be `!=` instead of `==`?

jasondellaluce (commented):
Thanks for addressing all the comments, this looks much more solid to me! Left you some comments/concerns. I'm not super-strong about those, but I wanted to have some feedback from you on those points.

userspace/falco/app_cmdline_options.cpp
mstemm: cxxopts allows you to bind an option to a vector (https://github.com/jarro2783/cxxopts#stdvectort-values). In this case, specifying the option multiple times appends to the vector. That's the approach taken for disabled_rule_substrings.

disabled_rule_tags was handled differently mostly to leave the falco engine interface unchanged. `falco_engine::enable_rule_by_tag(const std::set<std::string> &tags, ...)` already takes a std::set as its first argument. However, cxxopts only allows binding to vectors. So the instance variable is a set, but after parsing args the values in the vector for the option `T/t` are added to the set.

We can change the falco interface to use a vector for the method `falco_engine::enable_rule_by_tag()`, in which case we could treat disabled_rule_substrings and disabled_rule_tags the same.

userspace/falco/falco.cpp
mstemm: Sure, the old blocks were nearly identical too, but I can reorganize this to avoid the duplicate code.

userspace/falco/falco.cpp
mstemm: You're right, this is backwards. I got the logic right earlier in line 1043 but not here. Fixed.

userspace/falco/app_cmdline_options.cpp
jasondellaluce: There's no need for now I think, my only concern was that everything worked as expected. I personally like having tags as a set, since that's kind of a good data structure for them.

leogr: I agree we can improve it later. Not a blocker for this PR, IMO.

leogr (approved):
LGTM now.

Thank you, Mark!

/approve

jasondellaluce (approved):
/approve

🥳

PR #1900, opened 2022-02-14
fix(userspace/falco): applies FALCO_INSTALL_CONF_FILE as the default …

### LINKED ISSUE #1897: Unnecessary default value for `-c`
**Describe the bug**

The command-line option `-c` has two default values:

```shell
-c                            Configuration file (default /source/falco/falco.yaml, /etc/falco/falco.yaml)
```

Although it's not really a bug, I think `/source/falco/falco.yaml` is misleading. Furthermore, usually users do not expect that `/source/falco/falco.yaml` takes precedence over `/etc/falco/falco.yaml`

Thus, I would like to do not use `/source/falco/falco.yaml` as a default value.

**How to reproduce it**

Store a configuration file at `/source/falco/falco.yaml`, then run `falco` (without `-c` option).

**Expected behaviour**

If `-c` is not specified, `falco` should look for `/etc/falco/falco.yaml` only.

**Environment**

- Falco version: 0.31.0
- System info:

- OS:

- Kernel:

- Installation method:

**Additional context**

andreabonanno:
…config.

Signed-off-by: Andrea Bonanno <andrea@bonanno.cloud>

**What type of PR is this?**

/kind bug

> /area build

> /area engine

> /area rules

> /area tests

> /area proposals

Makes Falco defaults to the config file in` /etc/falco/falco.yaml` when the `-c` option is not used, instead of evaluating `/source/falco/falco.yaml` as the config source with highest priority. 

Fixes #1897 

```release-note
NONE
```

### DIFF
--- a/userspace/falco/falco.cpp
+++ b/userspace/falco/falco.cpp
@@ -92,7 +92,11 @@ static void usage()
 	   "Usage: falco [options]\n\n"
 	   "Options:\n"
 	   " -h, --help                    Print this page\n"
-	   " -c                            Configuration file (default " FALCO_SOURCE_CONF_FILE ", " FALCO_INSTALL_CONF_FILE ")\n"
+#ifdef BUILD_TYPE_RELEASE
+	   " -c                            Configuration file (default " FALCO_INSTALL_CONF_FILE ")\n"
+#else
+           " -c                            Configuration file (default " FALCO_SOURCE_CONF_FILE ", " FALCO_INSTALL_CONF_FILE ")\n"
+#endif
 	   " -A                            Monitor all events, including those with EF_DROP_SIMPLE_CONS flag.\n"
 	   " -b, --print-base64            Print data buffers in base64.\n"
 	   "                               This is useful for encoding binary data that needs to be used over media designed to.\n"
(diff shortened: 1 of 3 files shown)


leogr (commented):
Thank you!

Your solution works, but I'd rather keep the old behavior when `-DCMAKE_BUILD_TYPE=Debug` and use **only** the installation path when `-DCMAKE_BUILD_TYPE=Release`.

Wdyt? :)

leogr (approved):
LGTM

Thank you!

/approve

userspace/falco/falco.cpp
```diff
@@ -873,21 +877,28 @@ int falco_init(int argc, char **argv)
 		}
 		else
 		{
-			conf_stream.open(FALCO_SOURCE_CONF_FILE);
+
+#ifdef BUILD_TYPE_RELEASE
```
jasondellaluce, line 881: From my understanding:
- If `BUILD_TYPE_RELEASE` is defined, we first try to load from `FALCO_INSTALL_CONF_FILE`, and then from `FALCO_SOURCE_CONF_FILE`
- If `BUILD_TYPE_RELEASE` is not defined, we only try to load from `FALCO_SOURCE_CONF_FILE`

Which is not the semantic we seem to describe in the `-help` section. Am I missing something?

andreabonanno: No worries!

leogr: Thank you :pray:

jasondellaluce (approved):

PR #1901, opened 2022-02-15
update(script): makes user able to pass custom option to driver-loade…

### LINKED ISSUE #1896: Create env variable to add options when loading the driver
**Motivation**

I have a usecase of Falco where I have a workflow that compiles Falco kernel module and upload it in Artifactory. 
When I download it, I have to pass credentials in the header of the request so I need to add options in the curl command in the falco-driver-loader script.

**Feature**

I would like an environment variable to pass options to the curl command that download Falco kernel module

**Alternatives**

Add an environment variable like FALCO_DRIVER_CURL_CUSTOM_OPTIONS because FALCO_DRIVER_CURL_OPTIONS already exists

**Additional context**

andreabonanno:
…r curl command

Signed-off-by: Andrea Bonanno <andrea@bonanno.cloud>

**What type of PR is this?**

/kind feature

Makes possible for the user to pass additional custom options to the `curl` command invoked by `falco-driver-loader`, appending them to the ones usually used.

Fixes #1896 

```release-note
NONE
```

### DIFF
--- a/scripts/falco-driver-loader
+++ b/scripts/falco-driver-loader
@@ -626,6 +626,7 @@ print_usage() {
 	echo "  DRIVERS_REPO             specify different URL(s) where to look for prebuilt Falco drivers (comma separated)"
 	echo "  DRIVER_NAME              specify a different name for the driver"
 	echo "  DRIVER_INSECURE_DOWNLOAD whether you want to allow insecure downloads or not"
+ 	echo "  ADD_CURL_OPTIONS         specify additional options to be passed to curl command used to download Falco drivers"
 	echo ""
 	echo "Versions:"
 	echo "  Falco version  ${FALCO_VERSION}"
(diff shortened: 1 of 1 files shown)


tspearconquest:
As a user with security concerns in mind, I would like to see that this does not allow passing arbitrary curl options, and instead would prefer if the environment variable was targeted toward the specific use case you have in mind.

japui:
update(script): torna o usuário capaz de passar a opção personalizada para o driver-loade

tspearconquest:
Hi @leogr will there be a release before this becomes rotten?

leogr:
> Hi @leogr will there be a release before this becomes rotten?

I apologize I had lost track of this. 
I believe we can merge it, but perhaps we should fix the env vat naming. 

cc @jasondellaluce 

/remove-lifecycle stale

leogr (approved):

Andreagit97 (approved):
/approve

PR #1916, opened 2022-02-24
new: add a new drop category `n_drops_scratch_map`

Andreagit97:

**What type of PR is this?**

/kind feature

/area engine

This PR allows computing the delta also for `scratch_map` drops. 
The `scratch_map` drops are a new drop category introduced [here](https://github.com/falcosecurity/libs/blob/0765e5de629259b71303b39ca9ce4c778faad14f/driver/ppm_events_public.h#L1675). 
These drops are related only to the bpf probe: a `scratch_map` drop happens when an event is too large and cannot be saved into the `frame_scratch_map` before being sent to userspace. The idea here is to clearly separate drops due to a full buffer from those caused by events that are too large. For more information take a look at this PR :point_down:  

https://github.com/falcosecurity/libs/pull/224

```release-note
new: add a new drop category `n_drops_scratch_map`
```


jasondellaluce (approved):
/approve

Thank you @Andreagit97 !!

PR #1925, opened 2022-03-04
Add app actions

mstemm:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind rule-update

> /kind rule-create

> /area build

/area engine

> /area rules

> /area tests

> /area proposals

Split the currently monolithic falco_init into smaller "actions", managed by the falco application's action manager. 

Fixes #

```release-note
Break up the monolithic falco_init into smaller manageable pieces.
```


mstemm:
If we want to use a third-party library for managing actions, instead of what I wrote, we could try some of these:
* https://github.com/taskflow/taskflow (needs c++17)
* https://github.com/tillenius/superglue
* https://github.com/Amanieu/asyncplusplus
* https://github.com/actor-framework/actor-framework
* https://github.com/STEllAR-GROUP/hpx

These all seem to be more for parallel programming where actions also declare what they read and write. Since we don't really need parallel programming for this startup phase of falco_init(), I thought it was kind of overkill.

mstemm:
Although the large number of changed lines looks daunting, remember that a lot of it is moving code around from the monolithic falco_init() into individual files for each of the actions. Here's a rough breakdown of the changed lines:

As pretext, note that the bulk of falco_init was split into 22 separate "actions", with each action represented as its own derived class and in its own file.

4,359 lines added
-1235 lines moved from falco_init()
-227 lines for the base action class definition and implementation. This is split into two files to allow for running unit tests without pulling in too much other falco code
-80 lines for the "init action" class definition and implementation that is the base class of actions run before event processing.
-80 lines for the "run action" class definition and implementation that is the base class of actions run during event processing.
-123 lines to modify the application object to add the 22 actions to the action manager and have the action manager run
-230 lines for the "action manager" class definition and implementation
-38 lines for the single header file that includes all the 22 action header files
-387 lines for actions unit tests
-660 lines for copyright notices for the 44 action files (22 header files, 22 code files)
-1320 lines of "class boilerplate" to define the derived class and implement its basic methods across the 22 files

So if we're concerned about making compact changes, we can switch things to easier-to-declare things like lambdas rather than derived classes. But honestly I'd rather keep things as-is and keep the overheads from class definitions and being split into individual files.

mstemm:
I'm working on a version of this PR where it uses std::function instead of derived classes and the code in each action file has preserved git history. So adding WIP label back for now. There's no need to review this yet.

leogr:
Closed in favour of https://github.com/falcosecurity/falco/pull/1953
/close

PR #1926, opened 2022-03-07
fix(cli): restore behavior of -pc -pm -pk CLI options

LucaGuerra:
Signed-off-by: Luca Guerra <luca@guerra.sh>

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind rule-update

> /kind rule-create

> /area build

/area engine

> /area rules

> /area tests

> /area proposals

Falco is often executed with `-pc` or `-pk` to display additional container or kubernetes metadata. The CLI option handling has been recently refactored, and those options are now handled differently using cxxopts as they're now expected with a double dash. After some testing, I noticed that:
* The behavior of cxxopts when all of `-p` with argument and boolean `--pcontainer` and `--pc` is at least a little weird, since they don't appear to work as-is
* Widely used examples, containers and helm charts rely on the use of a literal `-pc`/`-pk` option with a single dash, which after recent modifications would output literal `k`s or `c`s which are hard to interpret when looking at the output

This PR fixes this by restoring the original behavior, maintaining compatibility with CLIs currently used and removing ambiguity in parsing.

Fixes #

```release-note
NONE
```


mstemm (approved):

jasondellaluce (approved):
/approve

PR #1928, opened 2022-03-08
falco_engine::process_event performance enhancement 

deepskyblue86:
**What type of PR is this?**
`falco_engine::process_event` gets called for every inspector event.
Profiling showed that std::map::find takes about 10% of `falco_engine::process_event`, and that can easily be improved by accessing the source by index.
By leveraging `multi_index` we can still access by source name in all other (non critical performance-wise) methods.

## How I tested it
https://gist.github.com/deepskyblue86/91ca71de9b464e5dcd87deac1b911a6d
## Results
| impl        | IR (just item access) | detail                                              |
|:------------|----------------------:|:----------------------------------------------------|
| original    |                  1450 | std::map::find                                      |
| multi_index |                   320 | boost::multi_index::detail::random_access_index::at |
| vector      |                   150 | std::vector::at                                     |

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind rule-update

> /kind rule-create

> /area build

/area engine

> /area rules

> /area tests

> /area proposals

falco_engine::process_event performance enhancement 

Fixes #

```release-note
faster Falco engine event processing
```


jasondellaluce:
Thanks Angelo, this is a great contribution and I think we definitely need to optimize these code paths. 

The only concern that I have towards this proposal, is that I don't see a compelling use case to include Boost components as new dependencies. From what I saw, the only thing we use it for is a `multi_index` on rulesets. However, it seems like we just really want to use numeric-based indexing all over the place, with the only exception being the `falco_engine::source_index` method.

My opinion is that `falco_engine::source_index` is not used in any hot path. I saw it being only used in https://github.com/falcosecurity/falco/pull/1928/files#diff-759cf16077c5a03cbf4e9b6dd8af34208252447adb0f3f894e5aadb6f218335dR86, which:

1. The source index can potentially be retrieved at initialization time and then be passed to `k8s_audit_handler::accept_data` as a function parameter
2. Will probably be removed in the short term once K8S Audit gets re-implemented as a plugin

If that's the case, we could just use a `std::vector` to index rulesets with an integer key, avoid including Boost, and eventually have a simpler and performant code

Let me know what you think. 
cc @leogr

deepskyblue86:
I opted for multi_index because I assumed I had to keep the string-based access.
If we can go with just a vector, that would be the best-performing option.
BTW, I forgot to add my pref analysis. I'll add that in the description right away.

deepskyblue86:
@jasondellaluce there are also the following:
- `evttypes_for_ruleset`
- `add_filter`
- `is_source_valid`
Since I had no clue about them I followed the easy path :sweat_smile:

leogr (commented):
Hey Angelo,

First, really thanks for this PR! :star_struck: 

Anyway, I second @jasondellaluce on using numeric indexes and on avoiding using boost. Since all data sources are known at initialization time (including loaded plugins), we can build up a local numeric-only index and use it while running. So I'd go for the vector way, that's also the best-performing option.

Other than that I can just say this optimization looks fantastic :+1:

deepskyblue86:
Closing this in favor of #1944.

PR #1938, opened 2022-03-13
Allow to whitelist shell config modifiers

claudio-vellage:
**What type of PR is this?**

/kind bug
/kind feature

/kind rule-update

/area rules

This PR allows to whitelist specific containers to modify the shell config files.

When running on GKE with `COS_CONTAINERD` images, I get frequent messages about modified bash files, e.g.:

```
rule
Modify Shell Configuration File

container.id
host

fd.name
/var/lib/containerd/tmpmounts/containerd-mount584469562/root/.bashrc
```

Because this is a managed cluster, I can't influence what the nodes/host is doing and what processes might modify the bash config, so as a user, I'd like to be able to whitelist this false positive.

I can open an issue if required, but I think it's only a minor change.

```release-note
rule(macro user_known_shell_config_modifiers): allow to allowlist shell config modifiers
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -443,6 +443,9 @@
 - list: shell_config_directories
   items: [/etc/zsh]
 
+- macro: user_known_shell_config_modifiers
+  condition: (never_true)
+
 - rule: Modify Shell Configuration File
   desc: Detect attempt to modify shell configuration files
   condition: >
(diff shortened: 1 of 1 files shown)


leogr (approved):
LGTM

/approve

claudio-vellage:
@leodido @mstemm @Kaizhe is someone able to approve this or do you have an ETA when this can be rolled out? If this will take a while, I will use a local copy of the chart. Currently on GKE (on preemptible nodes), I get the following warnings on our staging environment where we've migrated to `COS_CONTAINERD`:

image

It's crucial to get this merged, because [GKE 1.24 and later only support node images that use the containerd runtime](https://cloud.google.com/kubernetes-engine/docs/how-to/migrate-containerd#overview). GKE 1.24 [will be released to the stable channel on 2022-04-19](https://cloud.google.com/kubernetes-engine/docs/release-schedule#schedule_for_release_channels).

While there is still some time left, we prefer to migrate early so catch potential bugs early and don't get surprised last-minute. I hope I could clarify the importance of this PR a bit and thank you in advance for your feedback.

Kaizhe (approved):

PR #1950, opened 2022-03-20
chore(falco_scripts): Update `falco-driver-loader` cleaning phase

Andreagit97:

**What type of PR is this?**

/kind cleanup

/kind feature

/area build

This PR slightly modifies the behavior of the `falco-driver-loader` script. Now the cleaning phase enforces the deletion of the kernel module if it is still loaded. If the script is not able to detach the module, it waits until the user removes it manually. The user can do this by following some specific instructions that will be provided in the documentation... This is necessary to avoid incompatibility issues with driver versions, obtaining errors like the following:

```
Runtime error: Kernel module does not support PPM_IOCTL_GET_API_VERSION. Exiting.
```

This happens with Falco 0.31.1 because we introduced a mechanism to detect incompatible driver versions.

This new cleaning phase is called every time we try to load a new kernel module typing `falco-driver-loader` or explicitly calling it with `falco-driver-loader --clean`.

Since both `.deb` and `.rpm` packages have to perform the same cleaning phase when uninstalled now they call directly `falco-driver-loader --clean`.

```release-note
NONE
```

### DIFF
--- a/scripts/debian/postinst.in
+++ b/scripts/debian/postinst.in
@@ -1,6 +1,6 @@
 #!/bin/sh
 #
-# Copyright (C) 2021 The Falco Authors.
+# Copyright (C) 2022 The Falco Authors.
 #
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/scripts/debian/postrm.in
+++ b/scripts/debian/postrm.in
@@ -1,6 +1,6 @@
 #!/bin/sh
 #
-# Copyright (C) 2021 The Falco Authors.
+# Copyright (C) 2022 The Falco Authors.
 #
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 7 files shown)


Andreagit97:
The `dkms status` output changed from version `v2.8.6` (https://github.com/dell/dkms/commit/f83b758b6fb8ca67b1ab65df9e3d2a1e994eb483):

Before `v2.8.6` `dkms status` prints:

`module, version, kernel, architecture` 

After: 

`module/version, kernel, architecture`

Before `v2.8.6` it seems that the format never changed. The last commit I considered has more than 15 years and it keeps the same format, so we should be pretty safe...

The last commit I checked: https://github.com/dell/dkms/blob/3abe63b850b4714df34972ae4486f0c0689022f0/dkms#L1470

I addressed this different output with the following line, for more info look directly at the code in this PR:

```bash
DRIVER_VERSIONS=$(dkms status -m "${KMOD_NAME}" | tr -d "," | tr "/" " " | cut -d' ' -f2)
```

jasondellaluce (approved):
/approve

PR #1952, opened 2022-03-22
refactor!: remove K8S audit logs from Falco

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind design

/area build

/area engine

This is part of the ongoing effort of porting the current K8S Audit Logs support implementation out of Falco and turn it into a Falcosecurity PR. The goals of this PR are to:
- Remove the current audit log implementation from Falco
- Adapt all integration tests to use new K8S Audit plugin
- Adapt the K8S default ruleset to use the new K8S Audit plugin
- Drop the civetweb dependency
- Re-implement the `healthz` endpoint with a lightweight HTTP server

Fixes #

This new plugin-based K8S Audit implementation introduces the following **breaking changes**:
- The `falco.yaml` piece of configuration responsible of K8S Audit is now located under the `plugins` section
- List field types now have explicit list typing: a field can either extract a single value or a list of values. For list values, extracting one value means extracting a list containing only one value, which is not the same thing as extracting regular single values
- List field types now only support the `in` and `intersects` operators. For example, checks such as `ka.req.role.rules.verbs contains create` would be rejected and would need to be changed in the equivalent `ka.req.role.rules.verbs intersects (create)`
- Failed field value extraction should be checked with the `exists` operator, and not by comparing with the `<NA>` string.
The `<NA>` string literal is not returned anymore, neither in single-valued fields nor in list fields. For example, in the past field existence was occasionally checked with expressions like `ka.target.subresource != <NA>`, which would now inherently never be true, because if the field is absent the string comparison fails by default. Instead, prefer using the analoguous `ka.target.subresource exists`, which explicitly checks for missing values.
- The `/healtz` endpoint cannot bind to the same port of the K8S Audit Log endpoint (e.g. `/k8s-audit`), due to the fact that they are now managed by two different webservers (one in Falco, one in the plugin)
- Syscalls and K8S Audit event sources cannot be active at the same time anymore. This is caused by the current restriction of having 1 event source active per time in the libs, so that this may change in the future

Blocking PRs:
- [x] https://github.com/falcosecurity/plugins/pull/79
- [x] https://github.com/falcosecurity/falco/pull/1976
- [x] https://github.com/falcosecurity/plugins/pull/90

```release-note
refactor!: remove K8S audit logs from Falco
```


jasondellaluce:
This is currently blocked by the changes of https://github.com/falcosecurity/falco/pull/1976.

userspace/engine/falco_engine_version.h
```diff
@@ -21,4 +21,4 @@ limitations under the License.
 // This is the result of running "falco --list -N | sha256sum" and
 // represents the fields supported by this version of Falco. It's used
 // at build time to detect a changed set of fields.
-#define FALCO_FIELDS_CHECKSUM "77c4c549181b8aac1b9698c0101ac61acb5b2faede84a2c4fecb34834c6de2b9"
+#define FALCO_FIELDS_CHECKSUM "a557747a209f2d16e90a3324d84d56c02cf54d000b6e3ee44598413f19885fcc"
```
jasondellaluce, line 24: This changes because the `ka.*` and `jevt.*` are not listed anymore with the default `falco.yaml` configuration.

userspace/falco/webserver.cpp
```diff
@@ -14,258 +14,100 @@ See the License for the specific language governing permissions and
...
+        throw falco_exception("invalid webserver configuration");
+    }
+
+    std::atomic<bool> failed;
+    failed.store(false, std::memory_order_relaxed);
+    m_server_thread = std::thread([this, listen_port, &failed]
+    {
+        try
+        {
+            this->m_server->listen("localhost", listen_port);
```
leogr, line 69:
```suggestion
            this->m_server->listen("0.0.0.0", listen_port);
```
I guess it should be bound to `0.0.0.0` and not just `127.0.0.1`.

leogr (commented):
Just left a comment, otherwise SGTM! :star_struck:

userspace/falco/webserver.cpp
leogr: PS
It would be better to let the user choose, but this PR is already enough big.

leogr (approved):
:boom:

userspace/engine/json_evt.h
```diff
@@ -48,7 +48,7 @@ class json_event : public gen_event
 	inline uint16_t get_type() const
 	{
 		// All k8s audit events have the single tag "1". - see falco_engine::process_k8s_audit_event
-		return 1;
+		return ppm_event_type::PPME_PLUGINEVENT_E;
```
mstemm, line 51: Can you update the comment as well?

userspace/falco/app_actions/init_falco_engine.cpp
```diff
@@ -56,35 +56,30 @@ application::run_result application::init_falco_engine()
...
-		k8s_audit_formatter_factory->set_output_format(gen_event_formatter::OF_JSON);
 	}
 
 	for(const auto &src : m_options.disable_sources)
 	{
 		m_state->enabled_sources.erase(src);
 	}
 
-	// XXX/mstemm technically this isn't right, you could disable syscall *and* k8s_audit and configure a plugin.
+	// todo(jasondellaluce,leogr): change this once we attain multiple active source
```
mstemm, line 79: This will increase the need to support multiple event sources within one falco process. k8s and syscalls together used to be supported together obviously, and that isn't possible atm.

CMakeLists.txt
```diff
@@ -148,8 +148,8 @@ if(NOT MINIMAL_BUILD)
   # libcurl
   include(curl)
 
-  # civetweb
-  include(civetweb)
+  # cpp-httlib
```
mstemm, line 151: Just curious, why the switch to cpp-httplib?

mstemm (changes requested):
Just a couple of questions and one suggested change. Thanks for making this great change!

userspace/falco/app_actions/init_falco_engine.cpp
jasondellaluce: Definitely. I left this comment as a remainder for when we'll work on supporting multiple sources.

mstemm (approved):
I'll approve it to just save a round-trip on monday. @jasondellaluce promised to update the comment and I'll check back for answers on the other questions.

Thanks again!

CMakeLists.txt
jasondellaluce: Civetweb caused lots of compilation problems in the past, and was in general a troublesome dependency for us in the past. This new http server is header-only and pretty lightweight, which better serves out purposes for now since we only implement `/healthz`.

mstemm:
(whoops, didn't see that @leogr had also approved). Well I'll check back for the answers and it would be nice to clean up the comment in a later PR. Not the end of the world as that code is pretty retired anyway.

PR #1955, opened 2022-03-23
rules: Out of the box ruleset for OKTA Falco Plugin

darryk10:
Creating out of the box rules for the new OKTA Faclo plugins
Opened PR: https://github.com/falcosecurity/plugins/pull/81
Signed-off-by: darryk10 <stefano.chierici@sysdig.com>

**What type of PR is this?**
This PR introduce the out of the box ruleset for the new OKTA Falco plugin created.
Reference https://github.com/falcosecurity/plugins/pull/81

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind rule-update

/kind rule-create

> /area build

> /area engine

/area rules

> /area tests

> /area proposals

This PR add the out of the box ruleset for the new OKTA Falco plugin

Fixes #

Depends on opened PR https://github.com/falcosecurity/plugins/pull/81

```release-note
WIP: Out of the box ruleset for OKTA Falco Plugin
```

### DIFF
--- a/rules/okta_rules.yaml
+++ b/rules/okta_rules.yaml
@@ -1,0 +1,170 @@
+#Example Rule on login in to OKTA. Disabled by default since it might be noisy
+#- rule: User logged in to OKTA
+#  desc: Detect the user login in to OKTA
+#  condition: okta.evt.type = "user.session.start" 
+#  output: "A user has logged in toOKTA (user=%okta.actor.name, ip=%okta.client.ip)"
+#       priority: NOTICE
+#  source: okta
+#  tags: [okta]
+
+- rule: User Changing password in to OKTA
+  desc: Detect a user change password in OKTA
+  condition: okta.evt.type = "user.account.update_password"
+  output: "A user has changed password from OKTA (user=%okta.actor.name, ip=%okta.client.ip)"
+  priority: NOTICE
+  source: okta
+  tags: [okta]
+
+- rule: Creating a new OKTA user account
+  desc: Detect a new OKTA user account created in the OKTA environment
+  condition: okta.evt.type = "user.lifecycle.create"
+  output: "A new OKTA user account created (user=%okta.actor.name, target user=%okta.target.user.name)"
+  priority: NOTICE
+  source: okta
+  tags: [okta]
+
+- rule: User accessing app via single sign on OKTA
+  desc: Detect a user accessing an app via OKTA
+  condition: okta.evt.type = "user.authentication.sso"
+  output: "A user has accessed and app using OKTA (user=%okta.actor.name, app=%okta.app)"
+  priority: NOTICE
+  source: okta
+  tags: [okta]
+
+- rule: User has been locked out in OKTA
+  desc: Detect a user who has been locked out in OKTA
+  condition: okta.evt.type = "user.account.lock"
+  output: "A user has been locked out in OKTA (user=%okta.actor.name, ip=%okta.client.ip)"
+  priority: NOTICE
+  source: okta
+  tags: [okta]
+
+- rule: User has been moved from suspended status in OKTA.
+  desc: Detect a user who has been moved from suspended status in OKTA
+  condition: okta.evt.type = "user.lifecycle.unsuspend"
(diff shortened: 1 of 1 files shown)


rules/okta_rules.yaml
```diff
@@ -0,0 +1,162 @@
...
+  condition: okta.evt.type = "user.lifecycle.create"
+  output: "A new OKTA user account created (user=%okta.actor.name, target user=%okta.target.user.name)"
+  priority: NOTICE
+  source: okta
+  tags: [okta]
+
+- rule: User accessing app via single sign on OKTA
+  desc: Detect a user accessing an app via OKTA
+  condition: okta.evt.type = "user.authentication.sso"
+  output: "A user has accessed and app using OKTA (user=%okta.actor.name, app=%okta.app)"
```
Issif, line 29:
```suggestion
  output: "A user has accessed an app using OKTA (user=%okta.actor.name, app=%okta.app)"
```

Issif (changes requested):

jasondellaluce:
I see the last two commits have a `Co-authored-by: ...` but not a `Signed-off-by: ...`, which is breaking DCO. My guess is that this happened through a GitHub PR suggestion. I fell with this trick lots of times, please remember of always putting your `Signed-off-by: ...` before committing a suggestion! In this case you need to add it to those commits in order to unblock the DCO. Thank you! 😄

rules/okta_rules.yaml
```diff
@@ -0,0 +1,170 @@
+#Example Rule on login in to OKTA. Disabled by default since it might be noisy
+#- rule: User logged in to OKTA
+#  desc: Detect the user login in to OKTA
+#  condition: okta.evt.type = "user.session.start" 
+#  output: "A user has logged in toOKTA (user=%okta.actor.name, ip=%okta.client.ip)"
+#       priority: NOTICE
+#  source: okta
+#  tags: [okta]
+
+- rule: User Changing password in to OKTA
```
mstemm, line 10: I think overall the rules do a good job at showing up what info the plugin provides, but we should decide whether the rules are supposed to be opinionated and look for suspicious/alarming events, or rather provide a changelog of information where a user would do additional analysis on the changelog to find suspicious/alarming events.

The distinction is important because I've seen users, for example, just blindly enable all the k8s audit rules "to provide the best coverage", and then they're drowning in falco alerts that denote all changes to their k8s cluster. And then they blame falco for all the "noise".

I think we should make this distinction at least using rule tags and comments.

darryk10: thanks for your review.
All the rules due to the types of logs and the type of Okta service (IAM) are basically audit rules. 
The response for each rule might change user by user based on what they think it's important for their environment. Hard to fine a one fit all behaviour.
For instance here is the suggestion from OKTA on what it should be monitored from the security prospective: https://github.com/OktaSecurityLabs/CheatSheets/blob/master/SecurityEvents.md
Some of the one proposes are even more noisy and I haven't included that already (login and log out logs).
I can add an "audit" tag on some of them hard to find which one isn't audit due to the log source.

darryk10: @mstemm I disabled by default some rules leaving the most important on security side active. 
The all noisy ones should be disabled by default to reduce the noise as you suggested. 
Thanks for the feedback and let me know your thoughts

jasondellaluce (approved):
/approve

Thank you @darryk10, I think this is fantastic!

Just leaving a note for us in the future. I think adding rulesets for plugins that are not bundled into Falco is not a scalable pattern. What we would like to have in the future is probably a distribution strategy that would improve the UX and allow users to pull plugins & rulesets from a registry/index. I think this is something we'll want to prioritize in the future! But for sake of simplicity, I'm ok to keep this here!

mstemm (approved):
Thanks for disabling the audit-type rules. I agree with @jasondellaluce that we'll need to define a process where we decide which rules/plugins are shipped with falco by default, and which are generally available but a user might need to build the plugin themself or add the rules for that plugin's event source by themself.

I'm okay for now with adding the okta rules, because it's a great use case for plugins.

PR #1956, opened 2022-03-23
Delete or rename shell history: .ash_history

### LINKED ISSUE #1605: Delete or rename shell history: .ash_history
**Describe the bug**

The "Delete or rename shell history" rule detects changes to bash, zsh, and fish history files, but does not detect changes to `.ash_history`. The ash_history file is common for alpine containers, so it might be a good one to include.

https://github.com/falcosecurity/falco/blob/master/rules/falco_rules.yaml#L2612-L2623

**How to reproduce it**

Delete or truncate the .ash_history file in an alpine container.

**Expected behaviour**

Falco would alert for the "Delete or rename shell history" rule.

**Environment**

- Falco version: 0.27.0

- System info:

- OS: Alpine container running on Amazon Linux nodes

- Kernel:

- Installation method: Falco running as a daemonset in Kubernetes

**Additional context**

bdashrad:
**What type of PR is this?**

/kind rule-update

/area rules

Includes .ash_history in `rule: Delete or rename shell history`

Fixes #1605

```release-note
rule(macro truncate_shell_history): include .ash_history
rule(macro modify_shell_history): include .ash_history
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -2593,7 +2593,8 @@
       fd.name contains "bash_history" or
       fd.name endswith "zsh_history" or
       fd.name contains "fish_read_history" or
-      fd.name endswith "fish_history") and evt.arg.flags contains "O_TRUNC")
+      fd.name endswith "fish_history" or
+      fd.name contains "ash_history") and evt.arg.flags contains "O_TRUNC")
 
 - macro: var_lib_docker_filepath
   condition: (evt.arg.name startswith /var/lib/docker or fd.name startswith /var/lib/docker)
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -2575,14 +2575,17 @@
       evt.arg.name endswith "zsh_history" or
       evt.arg.name contains "fish_read_history" or
       evt.arg.name endswith "fish_history" or
+      evt.arg.name contains "ash_history" or
       evt.arg.oldpath contains "bash_history" or
       evt.arg.oldpath endswith "zsh_history" or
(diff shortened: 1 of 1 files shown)


leogr:
/close

rules/falco_rules.yaml
```diff
@@ -2575,22 +2575,26 @@
       evt.arg.name endswith "zsh_history" or
       evt.arg.name contains "fish_read_history" or
       evt.arg.name endswith "fish_history" or
+      evt.arg.name contains "ash_history" or
       evt.arg.oldpath contains "bash_history" or
       evt.arg.oldpath endswith "zsh_history" or
       evt.arg.oldpath contains "fish_read_history" or
       evt.arg.oldpath endswith "fish_history" or
+      evt.arg.oldpath contains "ash_history" or
```
darryk10, line 2583: Would be possible to use endswith operator instead of contains?
Thanks

leogr: Good idea @darryk10 :+1:

Kaizhe: If we introduce `ash_history` (doesn't matter whether using operator `contains` or `endswith`), we can ignore the `bash_history` then. A proper comment will be useful here saying both `bash_history` and `ash_history` will be covered.

bdashrad: Seems like a reasonable approach to me. Do we still want to use `contains` or do we want to change it to `endswith`. I don't have a preference.

leogr:
@Kaizhe ping :hugs:

Kaizhe (approved):

Issue #1922, opened 2022-03-02
Memory usage doubled after 1.16.3 -> 1.17.2 upgrade


mac-chaffee:
**Describe the bug**

After upgrading Falco from 1.16.3 to 1.17.2, I noticed that RAM usage for the falco daemonset roughly doubled from ~100MB to ~200MB per pod. I'm mainly filing this bug report to document this since it's not a serious bug unless you have custom RAM limits set, and the pods start killing oomkilled.

If there was some new dependency added in 1.17 that might be eating extra RAM, that might explain it. Just want to make sure I haven't misconfigured falco in some way that makes it use up more RAM.

**How to reproduce it**

Upgrade from 1.16.3 to 1.17.2

**Expected behaviour**

Consistent RAM usage between versions.

**Screenshots**

Top graph is pre-upgrade, bottom graph is post-upgrade.
<img width="1736" alt="image" src="https://user-images.githubusercontent.com/7581860/156425379-a48a0dc1-638c-40d9-b520-c49543e18f8a.png">

**Environment**

- Falco version: 1.17.2
- System info:
```
{
  "machine": "x86_64",
  "nodename": "monitoring-falco-td4qg",
  "release": "4.18.0-348.12.2.el8_5.x86_64",
  "sysname": "Linux",
  "version": "#1 SMP Mon Jan 17 07:06:06 EST 2022"
}
```
- Cloud provider or hardware configuration: Bare-metal, v1.21
- OS: RHEL 8.5
- Kernel: `Linux k8s-node01 4.18.0-348.12.2.el8_5.x86_64 #1 SMP Mon Jan 17 07:06:06 EST 2022 x86_64 x86_64 x86_64 GNU/Linux`
- Installation method: Kubernetes was installed with kubespray. Falco was installed with Helm.

**Additional context**

This cluster is medium-sized: 18 nodes, 500 pods. That might affect things.

leogr:
@mac-chaffee 

Are you referring to the Falco Helm Chart version? https://github.com/falcosecurity/charts/releases

mac-chaffee:
Oh yes, looks like I've mixed up the versions of the helm chart and falco itself.

PhilipSchmid:
[...]

Issue #1836, opened 2022-01-03
Unable to "override" k8s audit rules


tspearconquest:
**Describe the bug**
We recently setup k8s audit logging in our dev cluster on AKS and configured the audit log forwarder to forward events to falco.
Upon loading the rules, we saw some outputs that we immediately wanted to customize the rules for, so we proceeded to follow the same strategy we use for the primary falco rules, where we have the default rules from your team and then an override configuration which redefines the same rules with our changes.

So we did this for the k8s audit rules and restarted Falco, but the pod does not start. Instead it runs out of memory.

This happens even if we copy a default rule verbatim into the override config file.

Install falco helm chart with audit logging rules. Must use build 0.31.0 or higher
Ensure to enable the web server.
Ensure to configure reasonable resource limits on the pod.

```
          limits:
            cpu: 700m
            memory: 512Mi
          requests:
            cpu: 100m
            memory: 128Mi
```

Add a new audit logging configmap manifest to override the default k8s audit logging rules, similar to the one used to override the falco rules for syscall and network events, and be sure to add the config file to the falco configmap.
Copy in one of the k8s audit logging rules or a macro or a list to the new configmap, deploy the configmaps and redeploy the pods. You do not need to change the rule, macro, or list in any way to trigger the issue.

In case the loading order matters, we load the standard falco rules, then their overrides, then the k8s audit rules and then finally the k8s audit overrides. We did not test another loading order.

Upon trying to start the pods, falco will be oomkilled. Upon commenting out the new config file in the falco config, falco will start properly again.

Falco should start up and override the default k8s audit rules the same way it does for syscall and network rules

**Environment**

- Falco version: 0.31.0

- Cloud provider or hardware configuration: AKS
- OS: Ubuntu 18.04

- Installation method: Kubernetes, customized from the helm chart templates and using an image built from the above falco dev release.

jasondellaluce:
Hey @tspearconquest, I was wondering if you tried to increase the memory resource limit, and if the issue still occurs even in that case.

tspearconquest:
Hi @jasondellaluce Can you suggest a better limit? It has been running fine with this limit for some time so it's not clear what we should adjust that limit to. The helm chart uses 1024Mi, however that seems excessive for Falco by itself to me.

jasondellaluce:
[...]

PR #1965, opened 2022-04-06
refactor(userspace/engine): reduce memory usage for resolving evttypes

jasondellaluce:
**What type of PR is this?**

/kind bug

/kind cleanup

/area engine

/area tests

In https://github.com/falcosecurity/libs/pull/74 we added a feature in libsinsp for searching the event types for a given filtercheck. The set of event types is used internally by the Falco rule engine to index each rule filter by event type to improve performance at runtime. Before this, we used to calculate the set of event types right from Falco.

In the implementation of https://github.com/falcosecurity/libs/pull/74, filterchecks maintain the set of evttypes as part of the internal class state. Since this information is used only at initialization time, maintaining it at runtime causes a memory abuse that ended up in Falco 0.31+ using way more memory than needed, with the limit increasing depending on how many rules are loaded in the engine.

This PR fixes this by re-introducing the evttype search back into Falco, this time leveraging the new AST definitions introduced in https://github.com/falcosecurity/libs/pull/217. We do so by implementing a simple AST visitor class that populates the set of evttypes for a given filter condition. This also deprecates the changes of https://github.com/falcosecurity/libs/pull/74, that can be removed after this PR gets merged. At the same time, we also introduce the `falco_filter` definition inside the rule engine, which simplifies the signatures of many method that used to have 5+ arguments.

The unit tests related to https://github.com/falcosecurity/libs/pull/74 have been ported from libsinp to Falco and adapted to the Catch2 framework.

Here's some data about Falco's memory usage. The following was measured at runtime by after launching Falco by simply loading the `falco_rules.yaml` default ruleset:
- Falco (master): **~118.09 MBs**
- Falco (with these changes):  **~96.01 MBs**
- Falco (with these changes and with the code of falcosecurity/libs#74 removed from sinsp): **~62.28MB**

Not sure this is a direct fix, but this can help for the following issues that reported Falco consuming more memory at runtime:
- https://github.com/falcosecurity/falco/issues/1922
- https://github.com/falcosecurity/falco/issues/1836

Of the 753 LOC added in this PR, 620+ are simple renamings and unit tests ported from libsinsp. The real contribution is just [filter_evttype_resolver.cpp](https://github.com/falcosecurity/falco/compare/refactor/evttypes#diff-4bf293decba287ea22ece28fb1bd8728717ae07a3465df60d0bf04c46d612636).

```release-note
refactor(userspace/engine): reduce memory usage for resolving evttypes
```


tests/engine/test_filter_evttype_resolver.cpp
```diff
@@ -0,0 +1,237 @@
+/*
```
mstemm, line 1: So libsinsp/test/evttype_filter.ut.cpp will be removed after this is merged?

userspace/engine/ruleset.cpp
```diff
@@ -145,25 +141,17 @@ void falco_ruleset::ruleset_filters::evttypes_for_ruleset(std::set<uint16_t> &ev
...
-		evttypes.insert(fevttypes.begin(), fevttypes.end());
+		evttypes.insert(filter->evttypes.begin(), filter->evttypes.end());
 	}
 }
 
-void falco_ruleset::add(string &source,
-			string &name,
-			set<string> &tags,
-			std::shared_ptr<gen_event_filter> filter)
+void falco_ruleset::add(falco_filter& filter)
```
mstemm, line 150: Why not pass a shared_ptr<falco_filter> instead? It's still shared_ptrs within ruleset.cpp and the first thing add() does is make a copy into a shared pointer.

userspace/engine/ruleset.h
```diff
@@ -72,26 +70,7 @@ class falco_ruleset
...
-		std::string source;
-		std::string name;
-		std::set<std::string> tags;
-		std::shared_ptr<gen_event_filter> filter;
-		std::set<uint16_t> evttypes()
-		{
-			// todo(jasondellaluce,leogr): temp workaround, remove when fixed in libs
-			if(source == "syscall" || source == "k8s_audit")
-			{
-				return filter->evttypes();
```
mstemm, line 86: So this is the key change that reduces memory usage right? Not calling gen_event_filter::evttypes() at all, which prevents the set from being populated.

I assume a follow-on PR will remove all the evttypes() code from libs, right?

mstemm (commented):
Thanks for making this change! One comment and 2 questions.

tests/engine/test_filter_evttype_resolver.cpp
jasondellaluce: Exactly, we have one in WIP to be merged after this one: https://github.com/falcosecurity/libs/pull/280

userspace/engine/ruleset.h
jasondellaluce: See the comment above about the cleanup PR in libs. However, the key solution here is to **not** store the event types set in each filtercheck. The sets used for the calculations are deallocated at each function call.

jasondellaluce:
I reverted the refactorings of the `add_filter` method signatures. I think that will require deeper attention and must be part of other future PRs.

userspace/engine/falco_engine_version.h
```diff
@@ -21,4 +21,4 @@ limitations under the License.
 // This is the result of running "falco --list -N | sha256sum" and
 // represents the fields supported by this version of Falco. It's used
 // at build time to detect a changed set of fields.
-#define FALCO_FIELDS_CHECKSUM "4de812495f8529ac20bda2b9774462b15911a51df293d59fe9ccb6b922fdeb9d"
+#define FALCO_FIELDS_CHECKSUM "77c4c549181b8aac1b9698c0101ac61acb5b2faede84a2c4fecb34834c6de2b9"
```
mstemm, line 24: Did this change due to the libs change, and some fields were added to libs?

mstemm (changes requested):
Just the question about the fields checksum changing now.

userspace/engine/falco_engine_version.h
jasondellaluce: You're right. I bumped the engine version to 12!

leogr (approved):
LGTM!

Thank you

/approve

mstemm (changes requested):

cmake/modules/plugins.cmake
```diff
@@ -19,8 +19,8 @@ string(TOLOWER ${CMAKE_HOST_SYSTEM_NAME} PLUGINS_SYSTEM_NAME)
 # released with a 1.0.0 required plugin api version
 ExternalProject_Add(
   cloudtrail-plugin
-  URL "https://download.falco.org/plugins/dev/cloudtrail-0.2.5-0.2.5-3%2B3068d86-${PLUGINS_SYSTEM_NAME}-${CMAKE_HOST_SYSTEM_PROCESSOR}.tar.gz"
```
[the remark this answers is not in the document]
mstemm: I guess it was because libs updated, which changed the API slightly, and that required a new plugin. So sounds good.

mstemm (approved):

PR #1970, opened 2022-04-13
refactor(userspace/engine): decoupling ruleset reading, parsing, and compilation steps

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind design

/kind feature

/area engine

This is PR is **not supposed to bring any breaking changes**. The goal of this is to de-couple the logic related to ruleset file/content loading in the Falco Engine. With this, YAML reading and rule parsing/compilation are operated by the two distinct classes `rule_reader` and `rule_loader`. This not only makes the code more clean, but will also allow us to swap YAML library in the future, or even support alternative methods of loading rulesets.

On top of that, this also adds the optional feature of splitting the rule loading and rule compilation process. In this way, we can load many rulesets file without re-compiling & clearing the same rules in the engine multiple times, which saves time and memory. Plus, this allows clearing the engine internal state definitions when they become unused.

```release-note
refactor(userspace/engine): decoupling ruleset reading, parsing, and compilation steps
```


userspace/engine/rule_loader.h
```diff
@@ -33,110 +33,207 @@ class falco_engine;
...
+		falco_common::priority_type priority;
+		bool enabled;
+		bool warn_evttypes;
+		bool skip_if_unknown_filter;
+	};
+
+	/*!
+		\brief Erases all the internal state and definitions
+	*/
+	virtual void clear();
```
mstemm, line 178: Why are a bunch of methods virtual? I checked and I didn't see any derived class from rule_loader. Did I miss it?

userspace/engine/rule_loader.h
```diff
@@ -33,110 +33,207 @@ class falco_engine;
...
+		the previous definition gets overwritten
+	*/
+	virtual void define(context& ctx, engine_version_info& info);
+	virtual void define(context& ctx, plugin_version_info& info);
+	virtual void define(context& ctx, list_info& info);
+	virtual void define(context& ctx, macro_info& info);
+	virtual void define(context& ctx, rule_info& info);
+
+	/*!
+		\brief Appends an info block to an existing one. An exception
```
mstemm, line 207: We are definitely not consistent about error reporting, but I'd like to make error reporting more consistent, and get away from using exceptions, if possible. I hope that follow-on PRs can provide a richer "error" struct that points to the rule/macro/list that had the error + an error string + a filename etc. I think it's hard to put that kind of info in an exception what().

I know you already used exceptions a lot in the previous iteration of this, and exceptions are still used a lot in rule_reader, so I'm happy to do a separate pass if we can get consensus about how to return these kinds of errors.

userspace/engine/rule_reader.cpp
```diff
@@ -0,0 +1,325 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "rule_reader.h"
+
+#define THROW(cond, err)    { if (cond) { throw falco_exception(err); } }
+
+static rule_loader::mark yaml_get_mark(
```
mstemm, line 21: Don't feel strongly about this, but why the static functions as compared to private methods?

mstemm (changes requested):
This already is looking like a much cleaner way to interact with rules files!

userspace/engine/rule_loader.h
jasondellaluce: I've put them as virtual so that it can be easier to write unit tests by mocking the loader. IIRC we use FakeIt here, which needs virtual methods to create a Mock.

test/falco_tests.yaml
```diff
@@ -1169,7 +1169,7 @@ trace_files: !mux
   skip_unknown_error:
     exit_status: 1
     stderr_contains: |+
-      Could not load rules file.*skip_unknown_error.yaml: 1 errors:
+      Could not compile rules in the engine: 1 errors:
```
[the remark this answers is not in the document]
jasondellaluce: I hear your point and I think you're right. I definitely think we should collect loading info/errors/warnings more accurately, but I agree this may be too much for this single PR.

userspace/engine/falco_engine.h
```diff
@@ -59,15 +59,35 @@ class falco_engine
...
-	void load_rules_file(const std::string &rules_filename, bool verbose, bool all_events);
-	void load_rules(const std::string &rules_content, bool verbose, bool all_events);
+	// If compile is set to true, the defined rules will be compiled and stored
+	// in the engine in a single step. If set to false, this would just load
+	// the rule definitions, and the compilation step can be run separately
+	// with the compile_rules() method. This can be used as a performance
+	// optimization when loading multiple rule files/contents to avoid
+	// re-compiling the same definitions at each iteration. If compilation
+	// is done in a single step, the engine will not clear the internal rule
+	// definitions (see 'clear' parameter of compile_rules()).
```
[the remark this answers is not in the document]
jasondellaluce: I think you're right on this. I'll change the last few commits accordingly.

userspace/engine/rule_loader.h
```diff
@@ -33,110 +33,207 @@ class falco_engine;
...
+			content += m.content;
+		}
+	};
 
 	/*!
-		\brief Erases the internal states and all the loaded rules
+		\brief Contains the info required to load rule definitions
 	*/
-	virtual void clear();
+	struct context
```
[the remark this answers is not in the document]
jasondellaluce: Totally agree on this point. I like the name `context` for the _infos because 1) it respects the old naming convention and 2) I think it represents the contained info properly, and we may also expand that info in the future to improve error handling. I would propose to rename `rule_loader::context` to `rule_loader::configuration`, which I think better explains what the variables inside are for.

userspace/engine/rule_loader.h
jasondellaluce: I think this is a broader topic and we should go for it in a second pass, but I agree that having a well-defined error reporting mechanism would be super helpful. First ideas on top of my mind:
- We can create an "error manager" class to which errors and warnings are reported (instead of having the two vectors) and which decides when an error is blocking/fatal by also gathering as much metadata as possible
- We can create a "falco rule loader exception" that extends `falco_exception` but contains all the info you mentioned

I would personally lean towards option 1 because it allows adding more logic in the error reporting tasks. Plus, exceptions are always blocking in the rule loading logic, and are not suitable for warnings. If we had a class for this, we can potentially make it the single source of truth for loading metadata too, so that we don't need to store all the 'context' information in each rule_info/macro_info/list_info.

userspace/engine/rule_reader.cpp
jasondellaluce: This was simply a matter of not over-complicating the class definition by adding helper methods that are just used internally. If you are strong on this point, I can turn them into private methods.

userspace/engine/rule_loader.h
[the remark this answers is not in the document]
mstemm: yep sounds great.

userspace/engine/falco_engine.h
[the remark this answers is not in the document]
mstemm: Yeah, logging is something that isn't quite there, especially for the engine, which can be embedded. Sounds good to wait for the compile step too.

userspace/engine/rule_loader.h
mstemm: Ah ok, sounds good.

userspace/engine/rule_loader.h
mstemm: Sure, we can work on a design and then fix it later.

userspace/engine/rule_reader.cpp
mstemm: No, it's okay. That's more of a style thing and I can work with either, as long as the static functions don't end up passing too many args that should be instance variables. That's not the case here.

mstemm (approved):

leogr (approved):
LGTM

/approve

PR #1976, opened 2022-04-21
refactor!: deprecate PSP regression tests and warn for unsafe usage of <NA> in k8s audit filters

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind design

/kind feature

/area engine

/area tests

This follows up the discussion in https://kubernetes.slack.com/archives/CMWH3EH32/p1650475403605479, the upcoming deprecation of PSPs in K8S (see https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/), and the recent dropping of the official Falco documentation for PSPs (see: https://github.com/falcosecurity/falco-website/pull/572).

**This PR starts the deprecation process of PSP from Falco** by r**emoving the legacy regression tests** that were present in our test suite. This is also an **enabler for** the project of **porting the K8S Audit** support to a plugin implementation (see: https://github.com/falcosecurity/falco/pull/1952), since the PSP regression tests used unsafe rule filters based on top of notorious `<NA>`-related bugs supported by the current K8S Audit log Falco implementation.

Accordingly, this PR also **adds a warning to the rule loader named `unsafe-na-check`**, that advises the users when a rule relies on unsafe comparisons with `<NA>`. For instance, cases such as `ka.req.pod.volumes.volume_type in (flexVolume,"<NA>")` (taken right from the removed PSP test cases), would produce the warning below:
```
Rule <rule_name>: warning (unsafe-na-check):
    comparing a field value with <NA> is unsafe 
    and can lead to unpredictable behavior of the 
    rule. Consider using the 'exists' operator instead. 
```
The condition above can safely be rewritten as: `(not ka.req.pod.volumes.volume_type exists) or ka.req.pod.volumes.volume_type in (flexVolume)`.

The **warning is non-blocking** and is only applied to K8S Audit-specific fields (`ka.*` and `jevt.*`), so that there is **no effective breaking change**. It's also worth mentioning that no rules like to one above is present in the current officially supported k8s audit ruleset. We may considering extending the warning to every field and-or making it blocking in the future, once we reach proper coherence with `<NA>` values in sinsp filterchecks too.

Breaking changes will be eventually introduced in the the plugin-based K8S Audit implementation proposed in #1952, that will fix the way `<NA>`s are extracted from `ka.*` fields and evaluated at runtime.

```release-note
refactor!: deprecate PSP regression tests and warn for unsafe usage of <NA> in k8s audit filters
```


mstemm (approved):

leogr (approved):
:boom: 

:clap: 

/approve

PR #1979, opened 2022-04-21
build: components opt-in mechanism for packages

leogr:
**What type of PR is this?**

/kind bug

/area build

This PR allows for selectively declaring which components should be included in packages, using [CPACK_INSTALL_CMAKE_PROJECTS](https://cmake.org/cmake/help/latest/module/CPack.html#variable:CPACK_INSTALL_CMAKE_PROJECTS).

It fixes the issue involuntarily introduced by https://github.com/falcosecurity/libs/pull/247. All libs targets are now installing their artifacts in the installation location, so those files were caught by `cpack` and included in Falco packages. However, Falco does not need all those libraries and headers, since those dependencies are already bundled into the Falco executable. The new opt-in mechanism introduced in the Falco building system allows overcoming this issue.

All recent `publish/docker-dev` CI jobs were failing for that reason. For example, see 
https://app.circleci.com/pipelines/github/falcosecurity/falco/2433/workflows/2d4d55c9-5528-4a0c-8fee-5b4ff168c484/jobs/22412
Note that the .deb installation failed because the package was trying to reinstall openssl files.
```
dpkg: error processing archive /var/cache/apt/archives/falco_0.31.1-81+8dd4bea_amd64.deb (--unpack):
 trying to overwrite '/usr/include/openssl/aes.h', which is also in package libssl-dev:amd64 1.1.1n-0+deb10u1
dpkg-deb: error: paste subprocess was killed by signal (Broken pipe)
Errors were encountered while processing:
 /var/cache/apt/archives/falco_0.31.1-81+8dd4bea_amd64.deb
```

Fixes #

Another small issue was present and has been fixed by [this](https://github.com/falcosecurity/falco/commit/06b7d8b18e5686dc97baa03865908fccd3647419). The `LIBS_PACKAGE_NAME` is not propagated in Falco, so all files weren't installed under the `falcosecurity` folder.
Although that's irrelevant for packages (because we are not including those files anymore), it still applies for `make install`.

/milestone 0.32.0

```release-note
NONE
```


mstemm (approved):

jasondellaluce (approved):
/approve

Great fix Leo! 😍

PR #1980, opened 2022-04-21
build: correct conffiles for DEB packages

leogr:
**What type of PR is this?**

/kind cleanup

/area build

Users are not supposed to locally edit ruleset files. They should use local variants only (for example `falco_rules.local.yaml`).
For this reason, I believe we should not include them in `conffiles`.

PS
If you want to know more about *conffiles* :point_right: https://raphaelhertzog.com/2010/09/21/debian-conffile-configuration-file-managed-by-dpkg/

Fixes #

```release-note
NONE
```


jasondellaluce (approved):
/approve

mstemm (approved):

PR #1999, opened 2022-05-13
fix: split filterchecks per source-idx

### LINKED ISSUE #1978: plugins: rule matching fails with trace files if a source plugin is loaded
**Describe the bug**

Configure Falco to load one or more plugin with source capabilities, such as in the following `falco.yaml` example:
```
plugins:
  - name: cloudtrail
    library_path: libcloudtrail.so
    init_config: ""
    open_params: ""
  - name: json
    library_path: libjson.so
    init_config: ""

load_plugins: [cloudtrail]
```
Then, launch Falco with a trace file in CAPTURE mode (`-e` option, used extensively in our regression tests). The capture file will be loaded and read successfully, but no rule will get triggered in any case.

**Explanation**

Since Falco implemented the support to the Plugin System, it started dividing rulesets by event source inside the engine. This happens by assigning a numeric index to each source, which is then use to process a given event (see: https://github.com/falcosecurity/falco/blob/b080d20525aaeaf2c9daa6a27ed290b281902c30/userspace/falco/falco.cpp#L277). This hightlights the fact that the Falco codebase is currently not ready to host the concept of "multiple source" active at the same time. So far, we have the hard limit of having only one source plugin active per time, and either syscalls or plugins activated exclusively.

However, this limitation is bypassed when reading trace files. In CAPTURE mode, Falco loads all the plugins by using only their extraction capabilities, and reads the events one by one without having any expert knowledge about their source. This causes the event to be processed, potentially, by the wrong ruleset. 

The root cause is that there is no way for Falco to know the source of an event when it gets received with `sinsp::next()`.

**Expected behaviour**

Loading one or more plugins should not interfere in any way with the ability to reproduce and process capture files.

FedeDP:
**What type of PR is this?**

/kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind rule-update

> /kind rule-create

> /area build

/area engine

> /area rules

/area tests

> /area proposals

Fixes #1978 

```release-note
NONE
```


leogr (approved):

jasondellaluce (approved):
/approve

PR #2009, opened 2022-05-23
update(userspace/engine): introduce new `check_plugin_requirements` API

Andreagit97:

**What type of PR is this?**

/kind cleanup

/kind feature

/area engine

This PR introduces a new `check_plugin_requirements` API  instead of the old `is_plugin_compatible` to check if a list of plugins satisfies all the version requirements.

```release-note
NONE
```

### DIFF
--- a/userspace/falco/app_actions/load_rules_files.cpp
+++ b/userspace/falco/app_actions/load_rules_files.cpp
@@ -121,7 +121,7 @@ application::run_result application::load_rules_files()
 	std::vector<falco_engine::plugin_version_requirement> plugin_reqs;
 	for (const auto &plugin : m_state->inspector->get_plugin_manager()->plugins())
  	{
-		 falco_engine::plugin_version_requirement req;
+		falco_engine::plugin_version_requirement req;
 		req.name = plugin->name();
 		req.version = plugin->plugin_version().as_string();
 		plugin_reqs.push_back(req);
--- a/userspace/falco/app_actions/load_rules_files.cpp
+++ b/userspace/falco/app_actions/load_rules_files.cpp
@@ -117,16 +117,21 @@ application::run_result application::load_rules_files()
 	}
 
 	// Ensure that all plugins are compatible with the loaded set of rules
-	for(const auto &plugin : m_state->inspector->get_plugin_manager()->plugins())
+	std::string plugin_vers_err = "";
+	std::vector<falco_engine::plugin_version_requirement> plugin_reqs;
+	for (const auto &plugin : m_state->inspector->get_plugin_manager()->plugins())
+ 	{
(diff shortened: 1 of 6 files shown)


userspace/falco/app_actions/load_rules_files.cpp
```diff
@@ -117,16 +117,21 @@ application::run_result application::load_rules_files()
 	}
 
 	// Ensure that all plugins are compatible with the loaded set of rules
-	for(const auto &plugin : m_state->inspector->get_plugin_manager()->plugins())
+	std::string plugin_vers_err = "";
+	std::vector<falco_engine::plugin_version_requirement> plugin_reqs;
+	for (const auto &plugin : m_state->inspector->get_plugin_manager()->plugins())
+ 	{
+		 falco_engine::plugin_version_requirement req;
```
jasondellaluce, line 124:
```suggestion
		falco_engine::plugin_version_requirement req;
```

jasondellaluce (commented):
This is great! Just left a minor indentation suggestion.

jasondellaluce:
I think this fix deserves some additional context. The current plugin/rules loading works as follows:
- The Falco Engine loads ruleset files and collects plugin version requirements (e.g. `required_plugin_versions`)
- Falco loads all the plugins depending on its configuration
- For each loaded plugin, Falco checks if it is compatible with the version requirements in the engine

This logic has the simple flow that Falco is not able to find out if the loaded rules require a certain plugin/version that has not been configured to be loaded. For instance, this would cause the `aws_cloudtrail` default ruleset to throw an `unknown field json.value` error at loading time if the `json` plugin is not configured to be loaded. This is an implicit and non-expressive way of failing.

This PR proposes a fix for this by letting the Engine perform all the check, and ensures that all rules-plugins dependencies are satisfied at loading time.

Andreagit97:
> I think this fix deserves some additional context. The current plugin/rules loading works as follows:
> 
>     * The Falco Engine loads ruleset files and collects plugin version requirements (e.g. `required_plugin_versions`)
> 
>     * Falco loads all the plugins depending on its configuration
> 
>     * For each loaded plugin, Falco checks if it is compatible with the version requirements in the engine
> 
> 
> This logic has the simple flow that Falco is not able to find out if the loaded rules require a certain plugin/version that has not been configured to be loaded. For instance, this would cause the `aws_cloudtrail` default ruleset to throw an `unknown field json.value` error at loading time if the `json` plugin is not configured to be loaded. This is an implicit and non-expressive way of failing.
> 
> This PR proposes a fix for this by letting the Engine perform all the check, and ensures that all rules-plugins dependencies are satisfied at loading time.

Thank you @jasondellaluce for the additional context, I forgot to add it :smile:

PR #2017, opened 2022-05-24
refactor(userspace): change falco engine design to properly support multiple sources

jasondellaluce:
**What type of PR is this?**

/kind bug

/kind cleanup

/kind design

/kind feature

This PR refactors and cleans up the design of the Falco Engine classes. Since introducing the notion of "multiple sources" due to the plugin system, the state management and the responsibilities of the internal components of the engine became ambiguous. The goal of these changes is to polish the design of the Falco Engine internals without introducing breaking changes in its interface, making it more extendible, and less error-prone. Overall, the changes involve:
- Removing cyclic dependencies and forward declarations inside the engine
- Defining documented structs for falco rules and sources, each in its standalone header
- Separating state management responsibilities in the engine
- Adding a `complete_rule_loading()` method in `falco_engine` to make the engine explicitly know when all rules have been loaded and enabled/disabled by Falco. So far, this knowledge has always been implicit.
- Introducing an interface for filter rulesets named `filter_ruleset`. This allows injecting multiple ruleset implementations inside the engine, each with different indexing/optimization strategies. This will serve us in the future to have plugin-specific indexing strategies for rule matching because the current evttype-based is totally ineffective. The default ruleset implementation is still the evttype-indexing one, which is now a standalone class named `evttype_index_ruleset` with explicit responsibilities.
- Documenting all the newly-introduced classed and some of the legacy ones too

/area engine

Fixes #

The PR is big in LOC due to some parts of code being moved, however ~200 of those are related to unit test being ported to the new method definitions.

```release-note
NONE
```


userspace/engine/ruleset.h
```diff
@@ -16,108 +16,140 @@ limitations under the License.
...
-#include "event.h"
-
-#include "gen_filter.h"
-
-class falco_ruleset
+/*!
+	\brief Manages a set of rulesets. A ruleset is a set of
+	enabled rules that is able to process events and find matches for those rules.
+*/
+class filter_ruleset
```
leogr, line 29: Shall we rename the filename to `filter_ruleset.h`?

leogr (commented):
I can only say thank you for the effort you have spent in improving those parts of the codebase. 
:pray: 

I reserve some time to take a second look, but it SGTM. Just left a suggestion.

leogr (approved):
LGTM

I give it several tries and everything works as expected.

Thank you!

/approve

mstemm (approved):
Thanks for making this valuable set of changes!

leogr (approved):
LGTM

Thank you!

yay! :partying_face:

PR #2019, opened 2022-05-25
fix(userspace/falco): enable k8s and mesos clients only when syscall source is enabled

jasondellaluce:
**What type of PR is this?**

/kind bug

/area engine

These changes ensure that the k8s and mesos clients are inited only if the syscal event source is active. Currently, they would be activated even if using a plugin source, which is unwanted behavior and resource consuming.

Parts of the code of the `init_inspector` action has been moved to a new `init_clients` action. Git history has been preserved on the moved code.

```release-note
fix(userspace/falco): enable k8s and mesos clients only when syscall source is enabled
```


leogr (approved):
LGTM

FedeDP (commented):
I cannot approve but this LGTM! Thanks Jason!

mstemm (approved):

PR #2021, opened 2022-05-26
rule(Java Process JNDI Connection): detect potential log4shell exploitation

pirxthepilot:
**What type of PR is this?**

/kind rule-create

/area rules

This detects potential log4shell exploitation by monitoring `sendto()` syscall and the buffer data from Java processes. See [this](https://pirx.io/posts/2022-05-20-syscall-based-detection-of-log4shell-cve-2021-44228-on-linux/) blog post for more details.

N/A

The rule as it is written in the PR can detect jndi for Java versions lower than 17. On 17, Java switched from using `sendto` to `write` syscalls (this is based on my own testing but happy to be corrected here).

The ideal `java_network_write` macro is actually:

```
- macro: java_network_write
  condition: (evt.type in (write, sendto) and evt.dir=< and fd.type in (ipv4, ipv6) and proc.name=java)
```

but falco needs to run with the `-A` flag.

```release-note
NONE
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -3177,6 +3177,24 @@
   priority: CRITICAL
   tags: [container, mitre_privilege_escalation, mitre_lateral_movement]
 
+# Rule for detecting potential successful Log4Shell (CVE-2021-44228) exploitation
+- macro: java_network_write
+  condition: (evt.type = sendto and evt.dir=< and fd.type in (ipv4, ipv6) and proc.name=java)
+
+- macro: jndi_ldap_indicator
+  condition: (evt.buffer contains "2.16.840")
+
+- macro: jndi_rmi_indicator
+  condition: (evt.buffer startswith "JRMI")
+
+- rule: Java Process JNDI Connection
+  desc: Potential exploitation of the log4shell Log4j vulnerability (CVE-2021-44228)
+  condition: >
+        java_network_write and (jndi_ldap_indicator or jndi_rmi_indicator)
+  output: Java process JNDI connection (user=%user.name user_loginname=%user.loginname user_loginuid=%user.loginuid event=%evt.type connection=%fd.name server_ip=%fd.sip server_port=%fd.sport proto=%fd.l4proto process=%proc.name command=%proc.cmdline parent=%proc.pname buffer=%evt.buffer container_id=%container.id image=%container.image.repository)
+  priority: CRITICAL
+  tags: [mitre_initial_access]
+
 # Application rules have moved to application_rules.yaml. Please look
 # there if you want to enable them by adding to
 # falco_rules.local.yaml.


leogr:
Hey @pirxthepilot !

Thank you for this PR :) 

I have noticed that integration tests are failing. We discovered an issue with `evt.buffer` that was fixed by I guess that https://github.com/falcosecurity/falco/pull/2022.

So, could you rebase your PR on the current `master` branch? That should fix the failing tests.

Let me know if you need any help. I may rebase the PR for you, eventually.

pirxthepilot:
Hi @leogr, looks like it's now working as expected, but failed the test [here](https://app.circleci.com/pipelines/github/falcosecurity/falco/2570/workflows/67da8151-7e90-49da-824c-f01478ef04d5/jobs/23534?invite=true#step-103-2640). Apache Cassandra making RMI connections it seems. Not sure if this is acceptable tolerance and/or something we can add an exception for? Thanks!

loresuso:
Hi @pirxthepilot, thank you for this PR! This seems to be something really interesting!
I just wanted to ask: with this rule, we will alert every time someone is using LDAP or RMI. In the Java enterprise world, I think these two technologies are pretty much used, so my concern is that we are going to alert many false positives. Is this correct? Also, in this case also LDAPS should be considered.

Another way to do this would be using the `bcontains` operator to spot the download of a Java class, which is the step of the attack that really leads to RCE and it is not something happening frequently (if not at all). My idea would be to detect that behavior using the Java .class magic bytes, so something like this might work well:
``` 
evt.buffer bcontains CAFEBABE
```
For what concerns syscall involved, I think the same reasoning applies, so `recvfrom` and `read`. Thank you a lot for noticing this!
What do you think about this? 🙂

pirxthepilot:
@loresuso yes that's a great idea! I was actually eagerly awaiting the `bcontains` feature. It didn't exist yet when I came up with this rule :)

And your assessment is correct, it does detect straight up Java LDAP and RMI (and not LDAPS). I can understand that this is a fairly common feature for a lot of organizations and so is not practical as a general detection.

If it's alright with you, I can update this PR to implement your idea. I'll fire up my lab environment again to test and verify (and also update my blog post), unless you've already done the work?

Thanks!

Kaizhe:
@pirxthepilot thanks for you contribution, I like your blog a lot. Just a quick question, how do you differentiate a normal JNDI connection from a log4j exploit JNDI connection?

pirxthepilot:
Hi @Kaizhe, thanks for reading! It doesn't distinguish between the two and is based on the assumption that JNDI is not commonly used in one's organization. I realize not that this might not be a good generalization to make.

I also followed up @loresuso 's idea and did a quick analysis on a PoC. I found the Java class being downloaded thanks to the presence of `cafebabe`, just as they described:

```
sysdig -X "evt.type in (write, sendto, read, recvfrom) and fd.type in (ipv4, ipv6) and proc.name=java" \
    -p"%fd.name (%evt.type, %evt.buflen bytes) %evt.buffer"
```

cafebabe
 
As I understand it, Java downloading a Java class is pretty rare which should make this a good detection? One caveat though is I don't think this will catch downloads via HTTPS (or would it?)

loresuso:
Hi @pirxthepilot, no worries, feel free to modify the PR with the `bcontains` operator,  I will be happy to test it! 
For the HTTPS problem, I think we should try to modify the PoC and see what happens, but you might be right

leogr:
Btw, this rule gets triggered by two fixtures we have in the integration tests :point_down: 
Screenshot from 2022-05-30 10-52-20

Not sure if it's a rule issue or the fixtures contains the exploits :smile_cat:

pirxthepilot:
@leogr , I got intrigued by your suggestion so I traced (no pun intended) and downloaded the fixtures from [here](https://download.falco.org/fixtures/trace-files/traces-positive_20200831.zip) and examined the scap in question which is `kube-demo.scap`. As far as I can tell, it is operating normally (not simulating an exploit),but I could also be wrong. kube-demo scap

Also, sorry for not submitting the bcontains PR yet, I'm away from my main workstation for the whole week. Hopefully I will have something over the weekend!

pirxthepilot:
Decided to create a separate PR #2041 to preserve the code and discussion here in case we want to review/compare/look back at some point :)

leogr:
Closed in favor of https://github.com/falcosecurity/falco/pull/2041.

PR #2041, opened 2022-06-07
rule(Java Process Class Download): detect potential log4shell exploitation

pirxthepilot:
**What type of PR is this?**

/kind rule-create

/area rules

This detects potential log4shell exploitation by monitoring recvfrom() syscall and the presence of `cafebabe` in the buffer data from Java processes, indicating the download of a Java class. This is an alternate rule to #2021 . (cc @loresuso and @Kaizhe )

N/A

As with #2021, recvfrom() applies to Java versions less than 17. In 17, Java uses read() which is ignored by falco by default.

In addition, this is not able to detect downloads via HTTPS. Also, I have not fully tested whether or not `cafebabe` can be obscured by making the HTTP server return a large chunk of headers (beyond the buffer falco is able to see by default), though it does work in my tests with a python simple HTTP server,

```release-note
NONE
```


loresuso:
Hi @pirxthepilot, thanks for spending your time on this!
This looks good to me, although you are right, this rule has some limitations:
- It cannot detect downloads via HTTPS, but this at least requires the attacker to do something more to avoid being detected
- Java uses the `read` syscalls since version 17 as you said. Maybe we can add this as a comment
- It is also true that `bcontains` can be blind if the bytes we are matching are past the size of the buffer we are passing to userspace from our drivers.

Although these limitations, I think it can work well in the majority of cases and I think it is a very good contribution.

pirxthepilot:
@loresuso great! I updated the diff with the note about the Java version. Let me know if you need anything else and thanks so much for the guidance!

jasondellaluce (approved):
/approve

PR #2043, opened 2022-06-07
Flatcar driver loader support

jepio:
**What type of PR is this?**

/kind bug

/area build

This PR ensure falco-driver-loader supports Flatcar. There are two commits that address two parts of this:
1. fix the kernel module build with dkms, which currently doesn't work due to the driver-loader container having a glibc that is older than the kernel module build directory was compiled against. This is solved by relocating the build tools in the kernel module directory, to run using host linker.
2. when searching for pre-built kernel module/eBPF probe, use the /etc/os-release VERSION_ID field instead of kernel release. This is consistent with kernel-crawler output and the convention chosen in https://github.com/falcosecurity/driverkit/pull/131, due to the unique design of Flatcar: you can find the right kernel config/kernel version from the OS version, but not the other way around (not without guessing)

Fixes #

```release-note
fix(falco-driver-loader): building falco module with DKMS on Flatcar and supporting fetching pre-built module/eBPF probe
```

### DIFF
--- a/docker/falco/Dockerfile
+++ b/docker/falco/Dockerfile
@@ -34,6 +34,7 @@ RUN apt-get update \
 	libssl-dev \
 	llvm-7 \
 	netcat \
+	patchelf \
 	xz-utils \
 	&& rm -rf /var/lib/apt/lists/*
 
--- a/docker/local/Dockerfile
+++ b/docker/local/Dockerfile
@@ -40,6 +40,7 @@ RUN apt-get update \
 	liblsan0 \
 	libtsan0 \
 	libcc1-0 \
+	patchelf \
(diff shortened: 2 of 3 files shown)


leogr (dismissed):
:+1:

jasondellaluce (dismissed):
/approve

jasondellaluce:

/close

leogr:
This PR needs to be rebased. I'll do that for you.

leogr (approved):
Rebased. 

cc @jasondellaluce Could you approve again please? :pray:

jasondellaluce (approved):
/approve

jepio:
Thanks a lot!

PR #2080, opened 2022-06-21
refactor!(userspace/falco): remove output rate limiter and make output engine explicitly thread-safe

### LINKED ISSUE #1333: The outputs throttling mechanism skips notifications without any notice

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind design

/kind feature

/area engine

This PR removes the rate limiter based on token bucket from the output engine of Falco and from the default configuration.

Since removing the rete limiting logic, the output engine became fully thread-safe. The class interface and its documentation have been update to reflect the thread-safety properties (and Doxygen-compliant).

Fixes #1333 

Thread-safety guarantees is something we should aim for in some parts of the Falco codebase in order to achieve the feature proposed in https://github.com/falcosecurity/falco/issues/2074. At any level, thread-safety in the output engine is high in priority, because it would become a synchronization point almost in any event source parallelization scenario.

```release-note
refactor!(userspace/falco): remove output rate limiter and make output engine explicitly thread-safe
```

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -173,22 +173,6 @@ syscall_event_timeouts:
 
 output_timeout: 2000
 
-# A throttling mechanism implemented as a token bucket limits the
-# rate of falco notifications. This throttling is controlled by the following configuration
-# options:
-#  - rate: the number of tokens (i.e. right to send a notification)
-#    gained per second. Defaults to 1.
-#  - max_burst: the maximum number of tokens outstanding. Defaults to 1000.
-#
-# With these defaults, falco could send up to 1000 notifications after
-# an initial quiet period, and then up to 1 notification per second
-# afterward. It would gain the full burst back after 1000 seconds of
-# no activity.
-
-outputs:
-  rate: 1
-  max_burst: 1000
-
 # Where security notifications should go.
 # Multiple outputs can be enabled.
 
--- a/userspace/falco/app_actions/init_outputs.cpp
+++ b/userspace/falco/app_actions/init_outputs.cpp
@@ -40,19 +40,16 @@ application::run_result application::init_outputs()
 		hostname = c_hostname;
 	}
 
-	m_state->outputs->init(m_state->engine,
-			      m_state->config->m_json_output,
-			      m_state->config->m_json_include_output_property,
-			      m_state->config->m_json_include_tags_property,
-			      m_state->config->m_output_timeout,
-			      m_state->config->m_buffered_outputs,
-			      m_state->config->m_time_format_iso_8601,
-			      hostname);
-
-	for(auto output : m_state->config->m_outputs)
-	{
-		m_state->outputs->add_output(output);
-	}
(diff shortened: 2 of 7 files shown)


jasondellaluce:
> Isn't this a breaking change since stuff from the config file was deleted? I'd expect the "!" in the release notes!

You're right. Even if the configuration parsing function is loose and will not complain if useless fields are present, the behavior of Falco would still differ. Luckily, rate limiting is a minor non-functional feature (it is indeed user-facing though).

jasondellaluce:
Opened https://github.com/falcosecurity/falco/pull/2139 as an alternative to this PR that puts the rate limiter as optional instead of removing it.

leogr:
> Opened #2139 as an alternative to this PR that puts the rate limiter as optional instead of removing it.

I like the idea of making it optional, instead of completely removing it. So +1 from me.

leogr:
Closing in favor of https://github.com/falcosecurity/falco/pull/2139
/close

Issue #2074, opened 2022-06-20
[Tracking] Opening multiple event sources in the same Falco instance


jasondellaluce:
# Motivation

The plugin system allows Falco to open new kinds of event sources that go beyond the historical syscall use case. Recently, this has been leveraged to port the k8s audit log event source to a plugin (see: https://github.com/falcosecurity/plugins/tree/master/plugins/k8saudit, and https://github.com/falcosecurity/falco/pull/1952). One of the core limitation that comes from the plugin system implementation of the libraries, is that a given Falco instance is capable of opening **only one event source**. In the example above, this implies that Falco instances are not able to ingest both syscalls and k8s audit logs together. This can instead be accomplished by deploying two distinct Falco instances, one for each event source.

# Feature Requirements

- **(R1)** A single Falco instance should become able to open more than one event source at once and in parallel
- **(R2)** There should be **feature parity** and **performance parity** between having 2+ source active in parallel in a single Falco instances and having 2+ single-source Falco instances with the same event sources

# Proposed Solution

## Release Goals

~_To be defined_. This is out of reach for [Falco 0.32.1](https://github.com/falcosecurity/falco/issues/2070).~ [Falco 0.33.0](https://github.com/falcosecurity/falco/issues/2230)

## Terminology

- **Capture Mode**: A configuration of `sinsp` inspectors that reads event from a trace file
- **Live Mode**: A configuration of `sinsp` inspectors that reads event from one of the supported modes (kmod, ebpf, gvisor, plugin)

## Design

- **(D1)** The feature is implemented in Falco only, and mostly only affects the codebase of `falcosecurity/falco`. Both libsinsp and libscap will keep working in single-source mode
- **(D2)** Falco manages multiple sinsp instances, one in each thread
- **(D3)** Falco manages one or more instances of `sinsp` inspectors
  - If the # of inspectors is 1, everything runs in the main thread just like now
  - If the # of inspectors is 2+, each inspector runs in its own a separate thread (see **(R1)**). The whole event data path happens in parallel within each thread (event production, data enrichment, event-rule matching, and output formatting)
- **(D4)** If in capture mode, Falco runs a only 1 inspector configured to read events from a trace file
- **(D5)** If in live mode, Falco runs 1 inspector for each active event source
  - If an event source terminates due to EOF being reached, Falco waits for the other event sources to terminate too
  - If an event source terminates with an error, Falco forces the termination of all the other event sources
[...]

PR #2082, opened 2022-06-21
refactor(userspace/engine): strengthen and document thread-safety guarantees of falco_engine::process_event

jasondellaluce:
**What type of PR is this?**

/kind feature

/area engine

This PR is focused on guaranteeing (and documenting) that concurrent access to `falco_engine::process_event()` is safe for different event sources.

Since the new plugin system got implemented, and due to the recent refactoring of the Falco engine for Falco v0.32, the engine is not internally split by event sources. This means that the engine contains multiple rulesets, each mapped 1-1 with an event source. Accordingly, it is possible to assume that `falco_engine::process_event()` is now thread-safe with some assumptions:
1. Each concurrent invoker must use a distinct and unique `source_idx`
2. Each concurrent invoker must not switch values of `source_idx` across subsequent calls to `falco_engine::process_event()`
3. Each concurrent invoker must use pass a different event (which is an implicit consequence of point 1)
4. Filterchecks and formatters used to populate the conditions for a given event-source ruleset are not reused across rulesets of other event sources

Considering the above, the only synchronization point is `stats_manager::on_event()`, which has been modified to be thread-safe and lock-free.

Thread safety is not a requirement for Falco, but future developments such as the ones described in https://github.com/falcosecurity/falco/issues/2074 could require it. Among the many things that can be useful to make thread safe, the most crucial one is the rule-event matching routine, of which `falco_engine::process_event()` is entrypoint. With this implementation, it is possible to consider processing events coming from different event sources in different parallel threads.

Assuming a single-threaded scenario, the changes involved in `stats_manager::on_event()` are a no-op in both expected behavior and performance. This is true because `stats_manager::on_event()` is invoked when an event matches a given rule exclusively, which happens with less frequency than the production of events by some orders of magnitude. As such, we can consider these changes as non-noticeable for single-thread scenarios (which is what we have right now). Moreover, when events are matched with a rule we then have to queue them in the Falco output engine, which is still an asynchronous sequential synchronization point.

```release-note
refactor(userspace/engine): strengthen and document thread-safety guarantees of falco_engine::process_event
```

### DIFF
--- a/userspace/engine/falco_engine.cpp
+++ b/userspace/engine/falco_engine.cpp
@@ -332,6 +332,12 @@ std::shared_ptr<gen_event_formatter> falco_engine::create_formatter(const std::s
 unique_ptr<falco_engine::rule_result> falco_engine::process_event(std::size_t source_idx, gen_event *ev, uint16_t ruleset_id)
 {
 	falco_rule rule;
+
+	// note: there are no thread-safety guarantees on the filter_ruleset::run()
+	// method, but the thread-safety assumptions of falco_engine::process_event()
+	// imply that concurrent invokers use different and non-switchable values of
+	// source_idx, which means that at any time each filter_ruleset will only
+	// be accessed by a single thread.
 	if(should_drop_evt() || !find_source(source_idx)->ruleset->run(ev, rule, ruleset_id))
 	{
 		return unique_ptr<struct rule_result>();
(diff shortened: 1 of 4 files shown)


leogr (approved):
Great!

FedeDP (approved):
/approve

PR #2085, opened 2022-06-22
new: add option to enable event sources selectively

jasondellaluce:
**What type of PR is this?**

/kind feature

/area engine

/area tests

This PR introduces a new cmd line option `--enable-source=XXX` that is dual to the already-existing `--disable-source=XXX`. Both of them are opt-out: it is up to Falco to choose the default enable event sources depending on the provided configuration. So far, only one can be enabled per time (either `syscall`, or a plugin-based one).

The semantic of `--disable-source` is of disabling event sources selectively starting from the set of the ones enabled by Falco by default. On the opposite, `--enable-source` erases the Falco default enable set and allows enabling sources selectively.

The feature of selecting a custom set of event sources has been introduced in the regression test suite too (`falco_tests.py`). Tests relying on plugins have been updated accordingly.

For the current setup, this is sort of a huge no-op. However, this introduces preliminary work for https://github.com/falcosecurity/falco/issues/2074.

```release-note
new: add option to enable event sources selectively
```


leogr (approved):

userspace/falco/app_actions/select_event_sources.cpp
```diff
@@ -0,0 +1,81 @@
...
+		{
+			return run_result::fatal("Can not enable more than one event source");
+		}
+		if(*m_state->enabled_sources.begin() == falco_common::syscall_source)
+		{
+			m_state->inspector->m_input_plugin = nullptr;
+		}
+
+		/* Print all enabled sources. */
+		std::ostringstream os;
```
FedeDP, line 73: For a future PR: perhaps move this part to a different app_action?

FedeDP (approved):
/approve

userspace/falco/app_actions/select_event_sources.cpp
jasondellaluce: Yeah, I agree. Plus, one day we should also decide if we want to enable/disable event source for capture files too (A single capture file can potentially contain events from more than one source).

leogr (approved):

PR #2109, opened 2022-07-01
refactor(userspace/falco): re-design stats writer and make it thread-safe

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind design

/area engine

As for the requirements of https://github.com/falcosecurity/falco/issues/2074, this PR refactors the stats writer of Falco to make it thread-safe. The class has been re-designed as a whole with a multi-producer single-consumer model with a worker and a queue (similar to what we have in the output framework). Moreover, this approach should also perform a little better because during the event processing loop we now push on the queue instead of writing on the file buffer (it just happens once in a while when the periodic timeout ticks, so it's not a big deal). This design also make it easier to add new stats field if needed in the future.

```release-note
refactor(userspace/falco): re-design stats writer and make it thread-safe
```


FedeDP (commented):
Very nice and clean PR! Thank you Jason!
I'll wait for approval until 0.32.1 is out :)

userspace/falco/stats_writer.cpp
```diff
@@ -0,0 +1,187 @@
+/*
+Copyright (C) 2019 The Falco Authors.
```
leogr, line 2:
```suggestion
Copyright (C) 2022 The Falco Authors.
```

userspace/falco/stats_writer.h
```diff
@@ -0,0 +1,128 @@
+/*
+Copyright (C) 2019 The Falco Authors.
```
leogr, line 2:
```suggestion
Copyright (C) 2022 The Falco Authors.
```

leogr (commented):
SGMT, just the copyright year to be updated :)

FedeDP (approved):
/approve

leogr (approved):
Thank you

PR #2112, opened 2022-07-05
Use total events to calculate drop_pct

### LINKED ISSUE #2111: Unexpected drop_pct value
**Describe the bug**

The expected value for drop_pct is the percentage of total events that were dropped, but that doesn't appear to be the value reported. While troubleshooting high numbers of dropped events on some systems, I found that the drop_pct sometimes exceeded 100%. It looks like Falco is not including the dropped events in the total when calculating drop_pct.

**How to reproduce it**

```
falco --stats-interval 60000
```
Note: this will reproduce the problem as long as it drops at least one event in one minute.

**Expected behaviour**

I expect the drop_pct to show dropped events as a percent of total events (e.g. if 1000 events are processed and 10 are dropped then drop_pct should be 1).

**Screenshots**

```
log: {"sample": 15, "cur": {"events": 7316740312, "drops": 13564074902, "preemptions": 0}, "delta": {"events": 488585741, "drops": 904962208, "preemptions": 0}, "drop_pct": 185.221},
```

**Environment**

- Falco version: 0.31.0

- System info:

- Cloud provider or hardware configuration: GKE
- OS: Debian GNU/Linux 10 (buster)

- Kernel: 5.10.109+

- Installation method: 

**Additional context**

shane-lawrence:
**What type of PR is this?**

/kind cleanup

/area engine

My expectation for `drop_pct` was that it showed the percent of total events dropped by the Falco engine. While investigating some misbehaving Falco processes, I found that many were reporting more than 100% of events being dropped:
```
log: {"sample": 15, "cur": {"events": 7316740312, "drops": 13564074902, "preemptions": 0}, "delta": {"events": 488585741, "drops": 904962208, "preemptions": 0}, "drop_pct": 185.221},
```

I think that Falco is counting only the successfully processed events in the total when calculating drop_pct. This PR changes that calculation to include both the successful events (n_events) and the dropped events (n_drops) in the total. It also moves the drop_pct value into the delta subfield of the stats output since it is calculated from the delta, not the cumulative total.

Fixes #2111

```release-note
use total events to calculate drop_pct and move to delta subfield
```

### DIFF
--- a/userspace/falco/statsfilewriter.cpp
+++ b/userspace/falco/statsfilewriter.cpp
@@ -129,7 +129,7 @@ void StatsFileWriter::handle()
 			"\"events\": " << delta.n_evts <<
 			", \"drops\": " << delta.n_drops <<
 			", \"preemptions\": " << delta.n_preemptions <<
-			"}, \"drop_pct\": " << (delta.n_evts == 0 ? 0 : (100.0*delta.n_drops/delta.n_evts)) <<
+			"}, \"drop_pct\": " << (delta.n_evts == 0 ? 0 : (100.0*delta.n_drops/(delta.n_drops + delta.n_evts))) <<
 			"}," << endl;
 
 		m_last_stats = cstats;
--- a/userspace/falco/statsfilewriter.cpp
+++ b/userspace/falco/statsfilewriter.cpp
@@ -129,8 +129,8 @@ void StatsFileWriter::handle()
 			"\"events\": " << delta.n_evts <<
 			", \"drops\": " << delta.n_drops <<
 			", \"preemptions\": " << delta.n_preemptions <<
-			"}, \"drop_pct\": " << (delta.n_evts == 0 ? 0 : (100.0*delta.n_drops/(delta.n_drops + delta.n_evts))) <<
-			"}," << endl;
+			", \"drop_pct\": " << (delta.n_evts == 0 ? 0 : (100.0*delta.n_drops/(delta.n_drops + delta.n_evts))) <<
+			"}}," << endl;
 
 		m_last_stats = cstats;
 	}


Andreagit97:
Hi, @shane-lawrence thanks for this PR. The situation here is a little bit complicated, let's proceed step by step:
* first of all I have seen in the issue that you are using `Falco 0.31.0` and `Cloud provider or hardware configuration: GKE` so I guess the eBPF probe.
* `Falco 0.31.0` was affected by an issue, in a nutshell, we counted twice the number of drops, one-time kernel-side, and another userspace-side. You can find more info about it [in this PR](https://github.com/falcosecurity/libs/pull/224).  This explains the strange data you have faced `"events": 7316740312, "drops": 13564074902` as you can see the number of drops is greater than the number of events and this is impossible since the events seen by the drivers should contain also drops. From this point, we can understand also the strange `drpo_pct` value.
* Please note that our `n_evts` fields, both in `kmod` and in `bpf` should be intended as "events seen by the drivers" so they should contain also the number of drops. This is for example what we do in our bpf probe after [this variable increment](https://github.com/falcosecurity/libs/blob/aa8589da33a63927505100340a1a4adfe1418a73/driver/bpf/plumbing_helpers.h#L511) we will increment also the drops number in case the event cannot [be pushed into the buffer](https://github.com/falcosecurity/libs/blob/aa8589da33a63927505100340a1a4adfe1418a73/driver/bpf/fillers.h#L103) for example.  
* In the end we can say that probably the way in which we compute the percentage is good, the problem is that Falco 0.31.0 with the bpf probe was affected by a bug, as I said before.

BTW we probably need to double-check all the cases in which we manage these numbers in order to avoid inconsistencies  in the computation

shane-lawrence:
Thanks for the info @Andreagit97. Is it possible that the drops are counted more than twice? I have also seen drop_pct values greater than 200, as in this sample:
```
log: {"sample": 45, "cur": {"events": 2174324593, "drops": 4030518781, "preemptions": 0}, "delta": {"events": 36712191, "drops": 109353131, "preemptions": 0}, "drop_pct": 297.866},
```

FedeDP:
IMHO `drop_pct` there should be computed using `cstats` instead of `delta`.
We can also add a `delta_drop_pct` that explicitly expresses the drop_pct between 2 consecutive points in time.

Moreover, together with @jasondellaluce , we discovered this: https://github.com/falcosecurity/falco/blob/120027dc2e3a638ef5dda7be9241bad549e2e4e1/userspace/falco/event_drops.cpp#L94

@Andreagit97

Andreagit97:
> IMHO `drop_pct` there should be computed using `cstats` instead of `delta`. We can also add a `delta_drop_pct` that explicitly expresses the drop_pct between 2 consecutive points in time.
> 
> Moreover, together with @jasondellaluce , we discovered this:
> 
> https://github.com/falcosecurity/falco/blob/120027dc2e3a638ef5dda7be9241bad549e2e4e1/userspace/falco/event_drops.cpp#L94

Yeah, these are the inconsistencies I was talking about

> @Andreagit97

@shane-lawrence more than twice seems strange probably there are also some errors in the percentage computation, I will take a deeper look into it. Anyway thank you for all these data, they surely help us in finding bugs :eye:

shane-lawrence:
Thanks! I dug into the code a bit more. I see that [the definition](https://github.com/falcosecurity/libs/blob/656eefb5c94fccbdb7b2f329798eb4a9e2da1719/userspace/libscap/scap.h#L117) of `n_evts` describes it as the `total number of events received by the driver` so I don't think my approach of summing `n_drops` + `n_evts` is correct. The count must not obey this in some places though or we could never get `n_drops` > `n_evts`.

I'm not very familiar with the codebase so it's possible that I just missed parts where each is counted, but it looks like the eBPF probe is dropping events before they are counted in the total in [bpf/plumbing_helpers.h](https://github.com/falcosecurity/libs/blob/5373eb1e8364626b5591d961bfbc7794338ec18d/driver/bpf/plumbing_helpers.h#L508-L511).

If this sounds accurate to you, I can update this PR to increment `n_evts` before dropping the event in that file. If not, I will summarize our conversation in #2111 and close this PR.

Please let me know what you think.

Andreagit97:
> Thanks! I dug into the code a bit more. I see that [the definition](https://github.com/falcosecurity/libs/blob/656eefb5c94fccbdb7b2f329798eb4a9e2da1719/userspace/libscap/scap.h#L117) of `n_evts` describes it as the `total number of events received by the driver` so I don't think my approach of summing `n_drops` + `n_evts` is correct. The count must not obey this in some places though or we could never get `n_drops` > `n_evts`.

Yeah. completely agree with you. Thanks to your issue we have already found an inconsistent logic [here](https://github.com/falcosecurity/falco/blob/120027dc2e3a638ef5dda7be9241bad549e2e4e1/userspace/falco/event_drops.cpp#L94), and we are still searching for other possible problems. More precisely we are focusing our attention on this strange `drop_pct: 297.866`! Looking at the actual code we have not found anything strange since you are using an old Falco version there is the possibility that different issues (2-times BPF drops + something wrong in the percentage computation) mixed together could cause this strange number. It would be amazing if you could test the last Falco release (`0.32.1`) to see if the issue is still there, this would mean that there is something wrong in the actual logic for sure!
 
> I'm not very familiar with the codebase so it's possible that I just missed parts where each is counted, but it looks like the eBPF probe is dropping events before they are counted in the total in [bpf/plumbing_helpers.h](https://github.com/falcosecurity/libs/blob/5373eb1e8364626b5591d961bfbc7794338ec18d/driver/bpf/plumbing_helpers.h#L508-L511).

Oh yes, that's a good point, this logic is a little bit confusing because these events are not counted at all, nor in the `n_evts` nor in the `drops`, for this reason, we are thinking to re-implement it in some other ways so right now I wouldn't change it if it's ok for you.

> If this sounds accurate to you, I can open a PR in [falcosecurity/libs](https://github.com/falcosecurity/libs) to increment `n_evts` before dropping the event in that file. If not, I will summarize our conversation in #2111 and close this PR.
> 
> Please let me know what you think.

It would be amazing if you could summarize the conversation on the issue and maybe test the latest Falco just to be sure that the problem is still here. For what concern this PR I think you can close it, thank you very much for all your effort, you helped us in finding some bugs!

PR #2118, opened 2022-07-08
new(rules): Directory traversal monitored file read

incertum:

**What type of PR is this?**

/kind feature

/kind rule-create

/area rules

Add new rule `Directory traversal monitored file read`.

Web applications can be vulnerable to directory traversal attacks that allow accessing files outside of the web app's root directory (e.g. Arbitrary File Read bugs). System directories like /etc are typically accessed via absolute paths. Access patterns outside of this (here path traversal) can be regarded as suspicious. This for example will allow monitoring the entire `/etc` directory. For instance while `/etc/passwd` is not regarded a sensitive file like `/etc/shadow` - reading `/etc/passwd` via directory traversal is often performed as reconnaissance / probing for Arbitrary File Read vulnerabilities especially in web applications, because `/etc/passwd` is always there and it is world readable. In addition, vulnerability scanners will often try reading `/etc/passwd` via directory traversal `../../../../../../../etc/passwd`.

Thanks @Kaizhe and @darryk10 for feedback to derive final version.

Example simulated log 
```
{"priority":"Warning","rule":"Directory traversal monitored file read","source":"syscall","time":"2022-07-15T00:00:34.758753699Z", "output_fields": {"container.id":"host","container.image.repository":null,"evt.res":"SUCCESS","evt.time":1657843234758753699,"fd.name":"/etc/shadow","fd.nameraw":".././././.././././../././././../etc/shadow","proc.aname[2]":null,"proc.cmdline":"cat ../././//.////.././././.././././///./../etc/shadow","proc.cwd":"/home/vagrant/","proc.exepath":"/bin/cat","proc.name":"cat","proc.pname":"sudo","user.loginuid":1000,"user.name":"root","user.uid":0}}
```

Needs new filter/display field `fd.nameraw` https://github.com/falcosecurity/libs/pull/468

```release-note
new rule(Directory traversal monitored file read)
new rule(macro: open_file_failed)
new rule(macro: directory_traversal)
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -1366,6 +1366,19 @@
 - macro: user_known_read_sensitive_files_activities
   condition: (never_true)
 
+- rule: Directory traversal monitored file read
+  desc: >
+    Web applications can be vulnerable to directory traversal attacks that allow accessing files outside of the web app's root directory (e.g. Arbitrary File Read bugs).
+    System directories like /etc are typically accessed via absolute paths. Access patterns outside of this (here path traversal) can be regarded as suspicious.
+  condition: open_read and (fd.directory startswith "/etc" or fd.name contains ".ssh/" or fd.name contains "id_rsa") and fd.nameraw contains "../" and fd.nameraw glob *../*../* and not proc.pname in (shell_binaries)
+  enabled: true
+  output: >
+    Read monitored file via directory traversal (username=%user.name useruid=%user.uid user_loginuid=%user.loginuid program=%proc.name exe=%proc.exepath
+    command=%proc.cmdline parent=%proc.pname file=%fd.name fileraw=%fd.nameraw parent=%proc.pname
+    gparent=%proc.aname[2] container_id=%container.id image=%container.image.repository returncode=%evt.res cwd=%proc.cwd)
+  priority: WARNING
+  tags: [filesystem, mitre_discovery, mitre_exfiltration, mitre_credential_access]
+
 - rule: Read sensitive file trusted after startup
   desc: >
     an attempt to read any sensitive file (e.g. files containing user/password/authentication
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -29,13 +29,20 @@
 #   condition: (syscall.type=read and evt.dir=> and fd.type in (file, directory))
 
 - macro: open_write
-  condition: evt.type in (open,openat,openat2) and evt.is_open_write=true and fd.typechar='f' and fd.num>=0
+  condition: (evt.type in (open,openat,openat2) and evt.is_open_write=true and fd.typechar='f' and fd.num>=0)
 
 - macro: open_read
-  condition: evt.type in (open,openat,openat2) and evt.is_open_read=true and fd.typechar='f' and fd.num>=0
+  condition: (evt.type in (open,openat,openat2) and evt.is_open_read=true and fd.typechar='f' and fd.num>=0)
 
 - macro: open_directory
-  condition: evt.type in (open,openat,openat2) and evt.is_open_read=true and fd.typechar='d' and fd.num>=0
+  condition: (evt.type in (open,openat,openat2) and evt.is_open_read=true and fd.typechar='d' and fd.num>=0)
+
+# Failed file open attempts, useful to detect threat actors making mistakes
+# https://man7.org/linux/man-pages/man3/errno.3.html
+# evt.res=ENOENT - No such file or directory
+# evt.res=EACCESS - Permission denied
+- macro: open_file_failed
+  condition: (evt.type in (open,openat,openat2) and fd.typechar='f' and fd.num=-1 and evt.res startswith E)
 
 - macro: never_true
   condition: (evt.num=0)
@@ -51,32 +58,32 @@
   condition: (proc.name!="<NA>")
 
 - macro: rename
-  condition: evt.type in (rename, renameat, renameat2)
+  condition: (evt.type in (rename, renameat, renameat2))
 
 - macro: mkdir
-  condition: evt.type in (mkdir, mkdirat)
+  condition: (evt.type in (mkdir, mkdirat))
 
 - macro: remove
-  condition: evt.type in (rmdir, unlink, unlinkat)
+  condition: (evt.type in (rmdir, unlink, unlinkat))
 
 - macro: modify
-  condition: rename or remove
+  condition: (rename or remove)
 
 - macro: spawned_process
-  condition: evt.type in (execve, execveat) and evt.dir=<
+  condition: (evt.type in (execve, execveat) and evt.dir=<)
 
 - macro: create_symlink
-  condition: evt.type in (symlink, symlinkat) and evt.dir=<
+  condition: (evt.type in (symlink, symlinkat) and evt.dir=<)
 
 - macro: create_hardlink
-  condition: evt.type in (link, linkat) and evt.dir=<
+  condition: (evt.type in (link, linkat) and evt.dir=<)
 
 - macro: chmod
   condition: (evt.type in (chmod, fchmod, fchmodat) and evt.dir=<)
 
 # File categories
 - macro: bin_dir
-  condition: fd.directory in (/bin, /sbin, /usr/bin, /usr/sbin)
+  condition: (fd.directory in (/bin, /sbin, /usr/bin, /usr/sbin))
 
 - macro: bin_dir_mkdir
   condition: >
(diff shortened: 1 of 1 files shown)


Kaizhe:
This is a really good one @incertum , do you think adding a condition about non shell parent process would make more sense? I am thinking a scenario a path traverse is an exploit through a web app.

incertum:
Thanks @Kaizhe. Absolutely can add the additional filter statement in and concur this rule is only relevant for web applications that have an Arbitrary File Read bug.

rules/falco_rules.yaml
```diff
@@ -3194,6 +3194,16 @@
...
   tags: [mitre_initial_access]
 
+- condition: open_read and (fd.name="/etc/passwd" and fd.nameraw contains "../.")
+  desc: etc passwd read via directory traversal
+  enabled: true
+  output: (user=%user.name user_loginuid=%user.loginuid program=%proc.name exe=%proc.exepath
+    command=%proc.cmdline parent=%proc.pname file=%fd.name fileraw=%fd.nameraw parent=%proc.pname
+    gparent=%proc.aname[2] container_id=%container.id image=%container.image.repository returncode=%evt.res)
+  priority: WARNING
+  rule: Directory traversal etc passwd read
```
darryk10, line 3204: Hi, 
could you please follow the best practice on the rule syntax with the name on the top?
Thank you for the change and the contribution

darryk10 (changes requested):

incertum:
One liner to simulate if rule triggers `python -c 'import subprocess; subprocess.Popen("cat ../../../////etc/passwd", shell=True)'`.

incertum:
@darryk10 indeed, this seems more economic to just create one more complete rule, good call :)
New suggestion below. Would you have additional or other ideas?

```
- rule: Directory traversal monitored file read
  desc: monitored file read via directory traversal
  condition: open_read and (fd.name="/etc/passwd" or sensitive_files or fd.directory="/etc/security" or fd.name contains ".ssh/") and (fd.nameraw contains "../." and fd.nameraw contains "./..") and not proc.pname in (shell_binaries)
  enabled: true
  output: monitored file read via directory traversal (username=%user.name useruid=%user.uid user_loginuid=%user.loginuid program=%proc.name exe=%proc.exepath
    command=%proc.cmdline parent=%proc.pname file=%fd.name fileraw=%fd.nameraw parent=%proc.pname
    gparent=%proc.aname[2] container_id=%container.id image=%container.image.repository returncode=%evt.res cwd=%proc.cwd)
  priority: WARNING
  tags: [filesystem, mitre_discovery, mitre_exfiltration]
```
(1) Number of occurrences of a substring could be a neat future filter option. Meanwhile while not perfect matching for `../.` and `./..` could perhaps be a better way to approximate at least 2 levels traversals for scenarios where you have `../././//.////.././././.././././///./../` which is being sanitized to `.././././.././././../././././../`

(2) `.pem` files and the like can be of interest as well, but since those belong more to the user's directory, relative paths with traversal are much more likely to be benign -> perhaps something end users may want to consider for custom local rules.

(3) Would need to place this rule before rule `Read sensitive file trusted after startup` as it would be the broader / more rich signal that should take precedence.

(4) Any additional output fields we should add?

(5) Failed file access attempts seem not to be logged -> likely need `libs` src adjustments for that.

darryk10:
Hi @incertum, really good summary and I really like the condition you came up with. 
I agree with the `fd.nameraw glob *../*../*` in (1) and (3). 
Nothing to add in (4) the output looks good to me.
I was just thinking that another used of path traversal is to read the id_rsa (private key) to gett SSH access to the host.
What do you think on adding `fd.name="id_rsa"`, to add the specific SSH private key use case?

darryk10:
> Tweaked it a bit, proposing that it could be even simpler - just look for `/etc` any file. Also suspecting `glob` is a more expensive filter, `fd.nameraw contains "../" and fd.nameraw glob *../*../*` filter may give perf boosts.
> 
> ```
> # Web applications can be vulnerable to directory traversal attacks that allow 
> # accessing files outside of the web app's root directory (e.g. Arbitrary File Read bugs). 
> # System directories like /etc are typically accessed via absolute paths, 
> # access patterns outside of this (here path traversal) can be regarded as suspicious.
> 
> - rule: Directory traversal monitored file read
>   desc: monitored file read via directory traversal
>   condition: open_read and (fd.directory startswith "/etc" or fd.name contains ".ssh/" or fd.name contains "id_rsa") and fd.nameraw contains "../" and fd.nameraw glob *../*../* and not proc.pname in (shell_binaries)
>   enabled: true
>   output: (username=%user.name useruid=%user.uid user_loginuid=%user.loginuid program=%proc.name exe=%proc.exepath
>     command=%proc.cmdline parent=%proc.pname file=%fd.name fileraw=%fd.nameraw parent=%proc.pname
>     gparent=%proc.aname[2] container_id=%container.id image=%container.image.repository returncode=%evt.res cwd=%proc.cwd)
>   priority: WARNING
>   tags: [filesystem, mitre_discovery, mitre_exfiltration]
> ```

LGTM @incertum. Amazing job and thenks for all the changed :)

darryk10 (approved):
LGTM

jasondellaluce:
Hey @incertum we just bumped the libs version on master, so now the changes of https://github.com/falcosecurity/libs/pull/468 should be merged into Falco. After rebasing this branch on master, I think tests will pass!

incertum:
@darryk10 placed the rule before `Read ssh information` and used `etc_dir` macro instead, please re-review, thanks :)

rules/falco_rules.yaml
```diff
@@ -956,6 +956,19 @@
   priority: ERROR
   tags: [filesystem, mitre_persistence]
 
+- rule: Directory traversal monitored file read
+  desc: >
+    Web applications can be vulnerable to directory traversal attacks that allow accessing files outside of the web app's root directory (e.g. Arbitrary File Read bugs).
+    System directories like /etc are typically accessed via absolute paths. Access patterns outside of this (here path traversal) can be regarded as suspicious.
+  condition: open_read and (etc_dir or fd.name contains ".ssh/" or fd.name contains "id_rsa") and fd.nameraw contains "../" and fd.nameraw glob *../*../* and not proc.pname in (shell_binaries)
```
darryk10, line 963: great the `etc_dir` macro. What you think of using `user_ssh_directory` macro instead of `fd.name contains ".ssh/"` to check the user ssh directory?
also I agree with you re the position 👍🏻

darryk10 (changes requested):

rules/falco_rules.yaml
incertum: Done, now also `directory_traversal` macro, we are slowly closing in 🙃 .

Suggesting to optimize `user_ssh_directory`? Running glob against `fd.name` across few rules may have perf impacts. For instance, Big Data technologies are best used via pushing down cheap filters like `startswith`, `contains`, `endswith` ... before adding more expensive expressions.

incertum:
> (5) Failed file access attempts seem not to be logged -> likely need `libs` src adjustments for that.

Mystery solved, not a `libs` issue. Just required some experimentation and the resulting new macro `open_file_failed`.

In addition, included a minor cleanup (adding parentheses to important macros just to be consistent and safe).

@jasondellaluce @leodido @Kaizhe @darryk10 - Let's brainstorm if the new macro `open_file_failed` can be of interest for other rules as well? In general, also looking at "failures" or something really out of the norm can help detect suspicious behavior more accurately by making the signal more rich.

---

Quick Simulation

[removed `and not proc.pname in (shell_binaries)` from condition for simulation]

```
{"priority":"Warning","rule":"Directory traversal monitored file read","source":"syscall", "output_fields": {"container.id":"host","container.image.repository":null,"evt.res":"EACCES","fd.name":"/etc/shadow","fd.nameraw":"../../../../../../../etc/shadow","proc.aname[2]":"gnome-terminal-","proc.cmdline":"cat ../../../../../../../etc/shadow","proc.cwd":"/tmp/","proc.exepath":"/usr/bin/cat","proc.name":"cat","proc.pname":"bash","user.loginuid":1000, "user.uid":1000}}
{"priority":"Warning","rule":"Directory traversal monitored file read","source":"syscall", "output_fields": {"container.id":"host","container.image.repository":null,"evt.res":"SUCCESS", "fd.name":"/etc/shadow","fd.nameraw":"../../../../../../../etc/shadow","proc.aname[2]":"sudo","proc.cmdline":"cat ../../../../../../../etc/shadow","proc.cwd":"/tmp/","proc.exepath":"/usr/bin/cat","proc.name":"cat","proc.pname":"sudo","user.loginuid":1000,"user.name":"root","user.uid":0}}
```

jasondellaluce (approved):
LGTM! I also agree on trying to see if `open_file_failed` can be applied elsewhere too, but let's go with these changes for now!

leogr (approved):
This PR is great :star_struck: 

Thank you!

jasondellaluce (approved):
/approve

PR #2128, opened 2022-07-12
fix(userspace): compute the `drop ratio` in the right way

Andreagit97:
Co-authored-by: Shane Lawrence <shane@lawrence.dev>

**What type of PR is this?**

/kind bug

/kind cleanup

/area engine

First of all, I would to thank  @shane-lawrence for spotting this issue in #2111! Great catch!

As I said [here](https://github.com/falcosecurity/falco/pull/2112#issuecomment-1175101375) the `n_evts` fields, both in `kmod` and in `bpf` should be intended as "events seen by the drivers" so they should contain also the number of drops. This is for example what we do in our bpf probe after [this variable increment](https://github.com/falcosecurity/libs/blob/aa8589da33a63927505100340a1a4adfe1418a73/driver/bpf/plumbing_helpers.h#L511) we will increment also the drops number in case the event cannot [be pushed into the buffer](https://github.com/falcosecurity/libs/blob/aa8589da33a63927505100340a1a4adfe1418a73/driver/bpf/fillers.h#L103) for example. 

According to it, we need to change the way in which we compute the `drop ratio`

```release-note
fix: compute the `drop ratio` in the right way
```

### DIFF
--- a/userspace/falco/event_drops.cpp
+++ b/userspace/falco/event_drops.cpp
@@ -91,8 +91,8 @@ bool syscall_evt_drop_mgr::process_event(std::shared_ptr<sinsp> inspector, sinsp
 		if(delta.n_drops > 0)
 		{
 			double ratio = delta.n_drops;
-			// Assuming the number of event does not contains the dropped ones
-			ratio /= delta.n_drops + delta.n_evts;
+			// The `n_evts` always contains the `n_drops`.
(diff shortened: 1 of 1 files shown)


FedeDP (approved):
That's a great catch!
/approve

jasondellaluce (approved):
/approve

PR #2130, opened 2022-07-13
new: add a `drop_pct` referred to the global number of events

Andreagit97:
**What type of PR is this?**

/kind cleanup

/kind feature

/area engine

Looking at this [PR]() from @shane-lawrence I thought that we need a little bit of cleanup here, and moreover, it could be useful to have the `drop_percentage` computed on the global statistics and not only on the delta. I used the `nlohmann json` library just to simplify the output :) Let me know what you think about that 

```release-note
new: add a `drop_pct` referred to the global number of events
```

### DIFF
--- a/userspace/falco/statsfilewriter.cpp
+++ b/userspace/falco/statsfilewriter.cpp
@@ -126,10 +126,11 @@ void StatsFileWriter::handle()
 		jmsg["cur"]["events"] = cstats.n_evts;
 		jmsg["cur"]["drops"] = cstats.n_drops;
 		jmsg["cur"]["preemptions"] = cstats.n_preemptions;
+		jmsg["cur"]["drop_pct"] = (cstats.n_evts == 0 ? 0 : (100.0*cstats.n_drops/cstats.n_evts));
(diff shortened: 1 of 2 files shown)


userspace/falco/statsfilewriter.cpp
```diff
@@ -67,30 +67,6 @@ bool StatsFileWriter::init(std::shared_ptr<sinsp> inspector, string &filename, u
 		return false;
 	}
 
-	// (Undocumented) feature. Take any environment keys prefixed
```
FedeDP, line 70: Can you spend a couple of words about the reasoning behind this removal?

FedeDP (commented):
This LGTM thanks!
Left a minor question :)

FedeDP (approved):
/approve

userspace/falco/statsfilewriter.cpp
Andreagit97: yeah the `run_performance_tests.sh` was the only script that used it, so I think we could remove it :thinking:

shane-lawrence:
👍👍👍

PR #2139, opened 2022-07-21
refactor(userspace/falco)!: make output rate limiter optional and output engine explicitly thread-safe

### LINKED ISSUE #1333: The outputs throttling mechanism skips notifications without any notice

jasondellaluce:
**What type of PR is this?**

/kind feature

/area engine

This is an alternative to https://github.com/falcosecurity/falco/pull/2080, that makes the token-bucket alert rate limiter optional by default. When enabled, the output framework now uses one token bucket for each inspector. Once we have multi-source support, alerts coming from one source won't influence the rate limiting of the others. At the same time, this will also be thread-safe because each thread will be own its own token bucket.

Fixes #1333

```release-note
refactor(userspace/falco): make output rate limiter optional and output engine explicitly thread-safe
update(falco.yaml)!: notification rate limiter disabled by default.
```

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -190,16 +190,14 @@ output_timeout: 2000
 # rate of falco notifications. This throttling is controlled by the following configuration
 # options:
 #  - rate: the number of tokens (i.e. right to send a notification)
-#    gained per second. Defaults to 1.
+#    gained per second. When 0, the throttling mechanism is disabled.
+#    Defaults to 0.
 #  - max_burst: the maximum number of tokens outstanding. Defaults to 1000.
 #
-# With these defaults, falco could send up to 1000 notifications after
-# an initial quiet period, and then up to 1 notification per second
-# afterward. It would gain the full burst back after 1000 seconds of
-# no activity.
+# With these defaults, the throttling mechanism is disabled.
 
 outputs:
-  rate: 1
+  rate: 0
   max_burst: 1000
 
 # Where security notifications should go.
--- a/falco.yaml
+++ b/falco.yaml
@@ -187,14 +187,19 @@ syscall_event_timeouts:
 output_timeout: 2000
 
 # A throttling mechanism implemented as a token bucket limits the
-# rate of falco notifications. This throttling is controlled by the following configuration
-# options:
+# rate of Falco notifications. One rate limiter is assigned to each event
+# source, so that alerts coming from one can't influence the throttling
+# mechanism of the others. This is controlled by the following options:
 #  - rate: the number of tokens (i.e. right to send a notification)
 #    gained per second. When 0, the throttling mechanism is disabled.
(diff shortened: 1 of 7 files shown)


falco.yaml
```diff
@@ -187,19 +187,22 @@ syscall_event_timeouts:
 output_timeout: 2000
 
 # A throttling mechanism implemented as a token bucket limits the
-# rate of falco notifications. This throttling is controlled by the following configuration
-# options:
+# rate of Falco notifications. One rate limiter is assigned to each event
+# source, so that alerts coming from one can't influence the throttling
```
FedeDP, line 191: We are already talking about multiple event sources; isn't it premature here? :D

leogr (approved):
LGTM!

I prefer this over #2080 

/approve

FedeDP (approved):
Left a minor comment :) it LGTM!
/approve

PR #2151, opened 2022-08-03
improve falco files loading performance

VadimZy:
**What type of PR is this?**

> /kind design

> /kind feature

/area engine

Improve rules loading/parsing performance

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/engine/filter_evttype_resolver.h
+++ b/userspace/engine/filter_evttype_resolver.h
@@ -16,10 +16,130 @@ limitations under the License.
 
 #pragma once
 
+#include <sinsp.h>
+
 #include <filter/parser.h>
 #include <string>
 #include <set>
 #include <memory>
+#include <functional>
+
+class falco_event_types
+{
+private:
+	using vec_t = std::vector<uint8_t>;
+	vec_t m_types{};
+
+	static inline void check_range(uint16_t e)
+	{
+		if(e > PPM_EVENT_MAX)
+		{
+			throw std::range_error("invalid event type");
+		}
+	}
+
+public:
+	falco_event_types(falco_event_types&&) = default;
+	falco_event_types(const falco_event_types&) = default;
+	falco_event_types& operator=(falco_event_types&&) = default;
+	falco_event_types& operator=(const falco_event_types&) = default;
+
+	inline falco_event_types():
+		m_types(PPM_EVENT_MAX + 1, 0)
+	{
+	}
+
+	inline void insert(uint16_t e)
+	{
+		check_range(e);
+		m_types[e] = 1;
+	}
+
+	void merge(const falco_event_types& other)
+	{
+		for(int i = 0; i <= PPM_EVENT_MAX; ++i)
+		{
+			m_types[i] |= other.m_types[i];
+		}
+	}
+
+	void merge(const std::set<uint16_t>& other)
+	{
+		for(const auto& e : other)
+		{
+			insert(e);
+		}
+	}
+
+	inline bool contains(uint16_t e) const
+	{
+		check_range(e);
+		return m_types[e] != 0;
+	}
+
+	void clear()
+	{
+		for(auto& v : m_types)
+		{
+			v = 0;
+		}
+	}
+
+	bool equals(const falco_event_types& other) const
+	{
+		return m_types == other.m_types;
+	}
+
+	falco_event_types diff(const falco_event_types& other)
+	{
+		falco_event_types ret;
+		for(size_t i = 0; i <= PPM_EVENT_MAX; ++i)
+		{
+			if(m_types[i] == 1 && other.m_types[i] == 0)
+			{
+				ret.m_types[i] = 1;
+			}
+		}
+		return ret;
+	}
+
+	falco_event_types intersect(const falco_event_types& other)
+	{
+		falco_event_types ret;
+		for(size_t i = 0; i <= PPM_EVENT_MAX; ++i)
+		{
+			if(m_types[i] == 1 && other.m_types[i] == 1)
+			{
+				ret.m_types[i] = 1;
+			}
+		}
+		return ret;
+	}
+
+	void for_each(std::function<bool(uint16_t)> consumer) const
+	{
+		for(uint16_t i = 0; i < m_types.size(); ++i)
+		{
+			if(m_types[i] != 0)
+			{
+				if(!consumer(i))
+				{
+					return;
+				}
+			}
+		}
+	}
+};
+
+inline bool operator==(const falco_event_types& lhs, const falco_event_types& rhs)
+{
+	return lhs.equals(rhs);
+}
+
+inline bool operator!=(const falco_event_types& lhs, const falco_event_types& rhs)
+{
+	return !(lhs == rhs);
+}
 
 /*!
 	\brief Helper class for finding event types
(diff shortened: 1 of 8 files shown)


userspace/engine/filter_evttype_resolver.h
```diff
@@ -21,6 +21,124 @@ limitations under the License.
 #include <set>
 #include <memory>
 
+class falco_event_types
+{
+private:
+    static const int PPM_EVENT_MAX = 368;
```
jasondellaluce, line 27: We should use the libscap constant here!

userspace/engine/filter_evttype_resolver.h
```diff
@@ -21,6 +21,124 @@ limitations under the License.
 #include <set>
 #include <memory>
 
+class falco_event_types
```
jasondellaluce, line 24: This is awesome!

userspace/engine/rule_loader.cpp
```diff
@@ -326,19 +273,89 @@ static void resolve_macros(
...
-			l.used = true;
+			if((re_match.rm_so == 0 || delims.contains(cond[start + re_match.rm_so -1]))
+			   && ( start + re_match.rm_eo >= cond.size() || delims.contains(cond[start + re_match.rm_eo]) ))
+			{
+				auto *li  = lists.at(cond.substr(start + re_match.rm_so+1, re_match.rm_eo - re_match.rm_so-2));
+				if (li != nullptr)
+				{
+					li->used = true;
+					ret += cond.substr(start,  re_match.rm_so);
+					ret += "(";
```
jasondellaluce, line 306: I didn't run this, but I think you may be skipping cases when lists are referenced in other lists like:
```
evt.type in (open, some_types_list, connect)
```
Unfortunately, this is legit and supported.

userspace/engine/rule_loader.cpp
```diff
@@ -326,19 +273,89 @@ static void resolve_macros(
 	}
 }
 
-// note: there is no visibility order between filter conditions and lists
-static shared_ptr<ast::expr> parse_condition(
-	string condition,
-	indexed_vector<rule_loader::list_info>& lists)
+struct macro_list_inserter
```
jasondellaluce, line 276: If this is only for lists, why not naming this `list_inserter`? Macros and lists are different things in Falco

userspace/engine/rule_loader.cpp
```diff
@@ -347,8 +364,9 @@ static shared_ptr<ast::expr> parse_condition(
 	}
 	catch (const sinsp_exception& e)
 	{
-		throw falco_exception("Compilation error when compiling \""
-			+ condition + "\": " + to_string(p.get_pos().col) + ": " + e.what());
+		throw falco_exception("Compilation error when compiling \n"
```
jasondellaluce, line 367: Watch out for conflicts with https://github.com/falcosecurity/falco/pull/2098.

mstemm:
Should we bother changing any of the json_evt code at all, given that it's effectively dead with the transition of k8s audit support to a plugin?

userspace/engine/rule_loader.cpp
VadimZy: fixed

userspace/engine/rule_loader.cpp
VadimZy: renamed

userspace/engine/rule_loader.cpp
VadimZy: resolved

VadimZy:
> Should we bother changing any of the json_evt code at all, given that it's effectively dead with the transition of k8s audit support to a plugin?

it will exist in the code for a while, sysdig might use it for some time as well.

userspace/engine/rule_loader.cpp
jasondellaluce: This is not solved, the changes of #2098 introduced a new form of error reposting. We should throw a `rule_loader::rule_load_exception` with proper err code here.

userspace/engine/rule_loader.cpp
```diff
@@ -21,7 +21,12 @@ limitations under the License.
 #include "filter_evttype_resolver.h"
 #include "filter_warning_resolver.h"
 #include <version.h>
-#include <sstream>
+
+#ifndef _WIN32
```
jasondellaluce, line 25: This is not enough and that's also the reason why the CI builds are failing. We have a very weird setup in the libraries for regex support that depends on:
- old compiler versions, that don't support `<regex>`
- inclusion or not of `oniguruma` in the build, which is a dependency of `jq`

To solve all the combinations, this trick worked for me:
- https://github.com/falcosecurity/libs/blob/d08daff1bf69c2388b7474c1fed96bd6e98b900f/userspace/libsinsp/filter/parser.cpp#L35
- https://github.com/falcosecurity/libs/blob/d08daff1bf69c2388b7474c1fed96bd6e98b900f/userspace/libsinsp/CMakeLists.txt#L272 (I think you should be able to use the `USE_BUNDLED_ONIGURUMA` definition from libsinsp's cmake file)

Mind that this would also cause the regex standard used in the code to differ, depending on what regex library we are building with. So you need to write two copies of each regex, and also handle them in two different ways:
- https://github.com/falcosecurity/libs/blob/d08daff1bf69c2388b7474c1fed96bd6e98b900f/userspace/libsinsp/filter/parser.cpp#L46
- https://github.com/falcosecurity/libs/blob/d08daff1bf69c2388b7474c1fed96bd6e98b900f/userspace/libsinsp/filter/parser.cpp#L532

Good news is that the CI and the tests won't pass until this is addressed: we have different build types that help us test all the cases.

Bad news is that this is truly terrible, and I really hope we'll be able to do better than this as soon as possible. The bottleneck is the presence of `oniguruma` brought in by `jq`, we we manage to remove that we should be able to just stick to one regex standard (posix I guess).

userspace/engine/rule_loader.cpp
```diff
@@ -728,20 +662,185 @@ static void resolve_macros(
...
+
+	~list_inserter()
+	{
+		regfree(&re_list);
+		regfree(&re_sub);
+	}
+
+	// split string found in insert_lists by delimiters
+	// concatenate lists expansion
+	void cat_lists(std::string& ret, const std::string& cond, lists_map_t &lists) const
```
jasondellaluce, line 739: I think this will work now, thanks. The regression tests must be able to spot any remaining issue, if any.

userspace/engine/rule_loader.cpp
VadimZy: dropping fix for the list inserter altogether - I don't have resources to fight regex portability here.
@jasondellaluce keep in mind that the current list expansion routine is quite inefficient - it keeps on looking for all available lists (100s) in the continuously growing string and does it from the string beginning each time.

jasondellaluce: Unfortunately, that's what we had in the past with Lua as well. This part will need to be substituted with proper AST traversal+substitutions in the future, however that would introduce ruleset breaking changes so it's something that we'll tackle in the future.

jasondellaluce (approved):

jasondellaluce:
@VadimZy now we introduced RE2. Let me know if you intend to re-introduce your `list_interter` now that portable regexs are available. If not, this still needs to be rebased on top of master.

jasondellaluce (approved):
I rebased this on master for you. LGTM, thanks!

/approved

leogr (approved):

PR #2154, opened 2022-08-04
cleanup(rules): cleanup rules disabled by default

incertum:

**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

/kind rule-update

> /kind rule-create

> /area build

> /area engine

> /area rules

> /area tests

> /area proposals

Be more consistent with explicitly tagging rules that are disabled by default with `enabled: false`, should give some perf boosts as well. Spotted 3 rules eligible for cleanup. Are there more rules eligible for cleanup?

Fixes #

Depends on how end users customizes these rules. It should be communicated that tag `enabled: false` was added to some rules.

```release-note
rule(Disallowed SSH Connection)!: disabled by default
rule(macro: consider_ssh_reads)!: macro removed
rule(macro: do_unexpected_udp_check)!: macro remove
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -360,6 +360,7 @@
 - rule: Disallowed SSH Connection
   desc: Detect any new ssh connection to a host other than those in an allowed group of hosts
   condition: (inbound_outbound) and ssh_port and not allowed_ssh_hosts
+  enabled: false
   output: Disallowed SSH Connection (command=%proc.cmdline connection=%fd.name user=%user.name user_loginuid=%user.loginuid container_id=%container.id image=%container.image.repository)
   priority: NOTICE
   tags: [network, mitre_remote_service]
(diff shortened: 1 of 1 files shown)


jasondellaluce (approved):
/approve

This LGTM! Disabling rules definitely play a key role for better performance. I think we should perform some sort of static analysis in the ruleset and spot which rules end up including `never_true`.

> Does this PR introduce a user-facing change?:
Depends on how end users customizes these rules. It should be communicated that tag enabled: false was added to some rules. 

We usually address this issue through the release note! We have the internal convention of marking rule changes with the prefixes:
- `rule(RULE_NAME): ` for rules
- `rule(macro MACRO_NAME): ...` for macros
- `rule(list LIST_NAME): ...` for lists
This usually allows us to create more expressive changelogs at every Falco release. In your case, I think the release note should be something like:
```
rule(Disallowed SSH Connection): disable rule by default
rule(macro consider_ssh_reads): remove unused macro
rule(Read ssh information): disable rule by default
rule(macro do_unexpected_udp_check): remove unused macro
rule(Unexpected UDP Traffic): disable rule by default
```

leogr (approved):
Great!

/approve

/milestone 0.33.0

PS
I manually fixed the release note block and added `!` since those are kind of breaking changes for our users.

PR #2155, opened 2022-08-04
Support condition parse errors in rule loading results

mstemm:
In #2098, we reworked how rules loading errors/warnings were returned
to provide a richer set of information, including locations/context
for the errors/warnings.

That did *not* include locations within condition expressions,
though. When parsing a condition expression resulted in a
warning/error, the location simply pointed to the condition property
of the rule.

This commit improves this to handle parse errors:

- When libsinsp::filter::parser::parse() throws an exception, use
  get_pos() to get the position within the condition string.
- Add a new context() constructor that takes a filter pos_info instead
  of a YAML::Mark.

Doing this required some small changes to the context struct,
though. Previously, the name (e.g. yaml filename) was held separate
from the list of locations. However, condition strings don't directly
reflect any content in the yaml file. Yaml block scalars remove
whitespace/unwrap strings. Macro references/list references are
replaced with their contents.

To handle this, move the "name" (e.g. filename) from the context into
each location. This allows a chain of locations to start with file
positions but transition to offsets within a condition expression.

Also allow a context to contain an alternate content string which is
used to build the snippet. For contexts related to condition strings,
the content is the condition.

Finally, when printing snippets that contain very long lines (> a
static const 160 chars), instead of printing the entire line, print
the 160 chars surrounding the position.

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind rule-update

> /kind rule-create

> /area build

/area engine

> /area rules

> /area tests

> /area proposals

Add better support for parse errors in rule conditions when displaying rule load errors.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/engine/rule_loader.cpp
+++ b/userspace/engine/rule_loader.cpp
@@ -21,6 +21,7 @@ limitations under the License.
 #include "filter_evttype_resolver.h"
 #include "filter_warning_resolver.h"
 #include <version.h>
+#include <string>
 #include <sstream>
 
 #define MAX_VISIBILITY		((uint32_t) -1)
@@ -35,16 +36,45 @@ static string s_default_extra_fmt  = "%container.name (id=%container.id)";
 using namespace std;
 using namespace libsinsp::filter;
 
+static const std::string item_type_strings[] = {
+	"value for",
+	"exceptions",
+	"exception",
+	"exception values",
+	"exception value",
+	"rules content",
+	"rules content item",
+	"required_engine_version",
+	"required plugin versions",
+	"required plugin versions entry",
+	"required plugin versions alternative",
+	"list",
+	"list item",
+	"macro",
+	"macro condition",
+	"rule",
+	"rule condition",
+	"condition expression",
+	"rule output",
+	"rule output expression",
+	"rule priority"
+};
+
+const std::string& rule_loader::context::item_type_as_string(enum item_type it)
+{
+	return item_type_strings[it];
+}
+
 rule_loader::context::context(const std::string& name)
 {
 	// This ensures that every context has one location, even if
 	// that location is effectively the whole document.
-	location loc = {name, position(), "rules content", ""};
+	location loc = {name, position(), rule_loader::context::RULES_CONTENT, ""};
 	m_locs.push_back(loc);
 }
 
 rule_loader::context::context(const YAML::Node &item,
-			      const std::string item_type,
+			      const item_type item_type,
 			      const std::string item_name,
 			      const context& parent)
 {
@@ -64,7 +94,6 @@ rule_loader::context::context(const libsinsp::filter::parser::pos_info& pos,
 	std::replace(name.begin(), name.end(), '\n', ' ');
 	std::replace(name.begin(), name.end(), '\r', ' ');
 
-	std::string item_type = "condition expression";
 	std::string item_name = "";
 
 	// Convert the parser position to a context location. Both
@@ -76,7 +105,7 @@ rule_loader::context::context(const libsinsp::filter::parser::pos_info& pos,
 	condpos.line = pos.line-1;
 	condpos.column = pos.col-1;
 
-	init(name, condpos, item_type, item_name, parent);
+	init(name, condpos, rule_loader::context::CONDITION_EXPRESSION, item_name, parent);
 }
 
 const std::string& rule_loader::context::name() const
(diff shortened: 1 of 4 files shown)


userspace/engine/rule_loader.cpp
```diff
@@ -36,34 +36,80 @@ using namespace std;
 using namespace libsinsp::filter;
 
 rule_loader::context::context(const std::string& name)
-	: name(name)
 {
+	location loc = {name, YAML::Mark(), "rules content", ""};
```
jasondellaluce, line 40: Out of curiosity, why we do this here?

userspace/engine/rule_loader.cpp
```diff
@@ -36,34 +36,80 @@ using namespace std;
...
+{
+
+	// Contexts based on conditions don't use the
+	// filename. Instead the "name" is just the condition, and
+	// uses a short prefix of the condition.
+	std::string name = "\"" + condition.substr(0, 20) + "...\"";
+	std::replace(name.begin(), name.end(), '\n', ' ');
+	std::replace(name.begin(), name.end(), '\r', ' ');
+
+	std::string item_type = "condition expression";
```
jasondellaluce, line 65: We're starting to have many of these types. Can't we have an enum or something to centralize them?

userspace/engine/rule_loader.cpp
```diff
@@ -750,10 +811,15 @@ static shared_ptr<ast::expr> parse_condition(
 	}
 	catch (const sinsp_exception& e)
 	{
+		libsinsp::filter::parser::pos_info pos;
+		p.get_pos(pos);
```
jasondellaluce, line 815: I think we have a 0-params overload that allows:
```suggestion
		auto pos = p.get_pos();
```

jasondellaluce (commented):
This is a very good improvement and LGTM for the most part! I left you few comments on the code, and I'd add:
- Instead of having a static `snippet_width`, can't we make it a parameter of the `snippet()` method?
- Seems like we use `YAML::Mark` just to access `pos`, `line`, and `column`. Can't we create a small struct for this, so that we don't depend on `YAML::Mark` and on other position-like structs like the one from libsinsp's parser. For sake of simplicity, even a type alias would work for now. This would work better for when a third case of position struct will come, and would also make it easier to write unit tests for `rule_loader` (still a TODO) without dealing with the YAML library.

mstemm:
The test failure was not related to these changes, and was actually a design flaw in the earlier PR that wasn't detected until I changed how very long snippet lines were returned. 

If reading rules files A and B, it's possible that reading file A has no warnings, but after reading file B there are warnings in file A, because B overrides macros/rules/lists from A. Two files that show this are https://github.com/falcosecurity/falco/blob/master/test/rules/single_rule.yaml and https://github.com/falcosecurity/falco/blob/master/test/rules/override_macro.yaml. override_macro overrides the macro is_cat, but doing that causes the list used in the original definition of is_cat to become unused.

I'm going to fix that in a separate PR. It's going to have a lot of the changes that are in this PR (moving filenames into the context positions, handling long snippet lines), but it won't have the specific changes for handling condition parse errors.

So putting this one on hold for now.

userspace/engine/rule_loader.cpp
mstemm: Just so every context has some location, So for the top level context, it's a simple "somewhere in this file" location.

userspace/engine/rule_loader.cpp
mstemm: Yes, I made an enum.

userspace/engine/rule_loader.cpp
mstemm: I created a standalone location struct and YAML::Marks and parser::pos_infos are converted now.

userspace/engine/rule_loader.cpp
```diff
@@ -96,67 +172,61 @@ nlohmann::json rule_loader::context::as_json()
...
+		throw falco_exception("rule_loader::context without location?");
 	}
 
 	rule_loader::context::location loc = m_locs.back();
+	auto it = rules_contents.find(loc.name);
 
-	auto it = rules_contents.find(name);
+	if(alt_content.empty() && it == rules_contents.end())
+	{
+		if(it == rules_contents.end())
```
jasondellaluce, line 215: The nested if statement seems useless here, am I wrong?

jasondellaluce (commented):
Changes LGTM! I have a minor suggestion but other than that I think this is ready for approval.

userspace/engine/rule_loader.cpp
mstemm: Yeah, this is redundant. I removed it.

jasondellaluce (approved):
/approve

FedeDP (approved):
/approve

PR #1499, opened 2020-12-02
Add CodeQL Security Scanning

caniszczyk:
**What type of PR is this?**

/kind feature

**Any specific area of the project related to this PR?**

/area build

**What this PR does / why we need it**:

This PR introduces CodeQL security scanning to Falco.

### How to try it out

Visit https://github.com/falcosecurity/falco/security/code-scanning

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

/hold

**Does this PR introduce a user-facing change?**:

```release-note
Add CodeQL security scanning
```

Signed-off-by: Chris Aniszczyk <caniszczyk@gmail.com>

PR #2171, opened 2022-08-19
new(CI): add CodeQL security scanning to Falco.

Andreagit97:
**What type of PR is this?**

/kind feature

/area CI

This PR recalls the one opened by @caniszczyk  here :point_right:  https://github.com/falcosecurity/falco/pull/1499 

Don't know why this has not been merged at the time, since I find it very useful for us :thinking: BTW I will try to reopen it now, thank you again @caniszczyk for your great job! I've just updated the version of some GH actions :) 

```release-note
new(CI): add CodeQL security scanning to Falco.
```

### DIFF
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -21,7 +21,7 @@ on:
 jobs:
   analyze:
     name: Analyze
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     permissions:
       actions: read
       contents: read
@@ -37,6 +37,8 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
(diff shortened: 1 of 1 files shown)


leogr:
Good catch! :star_struck: 
Thank you, @Andreagit97 

/milestone 0.33.0

Andreagit97:
Unforunately CodeQL run the analysis also on our libraries since they are part of the build process... I've also tried to exclude some paths through the configuration file with `paths` and `paths-ignore` keywords but I faced the following warning :( 

```
The "paths"/"paths-ignore" fields of the config only have effect for JavaScript, Python, and Ruby
```

.github/workflows/codeql.yaml
```diff
@@ -0,0 +1,75 @@
...
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "master", release/0.26.2, release/0.32.2 ]
```
jasondellaluce, line 16: I don't think we need to also point release branches here. After all, release branches will be expected to contain only cherry-picked commits that have been merged in mailine first. What do you think?

Alternatively, is something like `release/*` a viable option in GHA?

jasondellaluce (commented):
> Unfortunately CodeQL runs the analysis also on our libraries since they are part of the build process...

I see. If we can't do anything different, I think this may be acceptable. After all the job should not be required, so even if we have some noise we can live with it. Plus, this may help us spot issues in libs too.

.github/workflows/codeql.yaml
Andreagit97: Agree I left them from the previous PR but I think we don't need the release branches since they are just cherry-picked commits, I will remove them, thank you :rocket:

jasondellaluce (approved):
/approve

leogr (approved):
LGTM :star_struck:

PR #2182, opened 2022-08-30
update: support running multiple event sources in parallel

jasondellaluce:
**What type of PR is this?**

/kind design

/kind feature

/area engine

See https://github.com/falcosecurity/falco/issues/2074. This is the last piece of "glue code" that uses all the preliminary work to support multiple parallel event processing loops coming from different sources.

From the UX known so far for Falco, this is a big no-op. However, this now supports loading multiple plugin with event sourcing capability and running all of them in parallel along with the traditional `syscall` event source. Each event source runs in its dedicated thread and does not influence the others. For example, this re-enables having both the `k8s_audit` and the `syscall` event sources on the same Falco instance.

By default, Falco enables the `syscall` event source and all the event sources implemented by the loaded plugin. The set of event sources enabled for event processing can be influenced with the `--enable-source` and `--disable-source` CLI options.

There is a user-facing change that needed to be introduced in the stats writer (`-s` and `--stats-interval` CLI options). Considering that multiple event source can now run in parallel, the old stats formatting didn't make sense anymore. As such, the periodic sample is now partitioned by event source. Here's an example of the expected output:
```json
{
  "sample": 71,
  "k8s_audit": {
    "cur": {
      "events": 1
    },
    "delta": {
      "events": 1
    }
  },
  "syscall": {
    "cur": {
      "drop_pct": 0,
      "drops": 0,
      "events": 9525,
      "preemptions": 0
    },
    "delta": {
      "drop_pct": 0,
      "drops": 0,
      "events": 137,
      "preemptions": 0
    }
  }
}
```

```release-note
update: support running multiple event sources in parallel
update(userspace/falco)!: adapt stats writer for multiple parallel event sources
```


userspace/falco/app_actions/init_falco_engine.cpp
```diff
@@ -57,25 +56,80 @@ void application::configure_output_format()
...
+			// in the engine. This makes the source index assignment easier.
+			add_source_to_engine(src);
+		}
+	}
 
-	m_state->syscall_source_idx = m_state->engine->add_source(falco_common::syscall_source, syscall_filter_factory, syscall_formatter_factory);
-	
-	if(m_state->config->m_json_output)
+	// add syscall as last source
+	add_source_to_engine(falco_common::syscall_source);
```
FedeDP, line 104: Do we want to always add `syscall_source` unconditionally, without checking if it is loaded?

userspace/falco/app_actions/load_plugins.cpp
```diff
@@ -27,121 +27,34 @@ application::run_result application::load_plugins()
...
+	auto empty_src = state::source_info{};
 
-	// By default only the syscall event source is loaded and enabled
-	m_state->loaded_sources = {falco_common::syscall_source};
-	m_state->enabled_sources = {falco_common::syscall_source};
+	// Initialize the set of loaded event sources. 
+	// By default, the set includes the 'syscall' event source
+	m_state->sources.clear();
+	m_state->sources.insert(empty_src, falco_common::syscall_source);
+	m_state->loaded_sources = { falco_common::syscall_source };
```
FedeDP, line 36: Oh, we always load `syscall_source`, i see.
Is there any reason behind that?

userspace/falco/app_actions/load_plugins.cpp
```diff
@@ -27,121 +27,34 @@ application::run_result application::load_plugins()
...
+		falco_logger::log(LOG_INFO, "Loading plugin '" + p.m_name + "' from file " + p.m_library_path + "\n");
+		auto plugin = m_state->offline_inspector->register_plugin(p.m_library_path);
+		m_state->plugin_configs.insert(p, plugin->name());
+		if(plugin->caps() & CAP_SOURCING)
 		{
-			return run_result::fatal("Plugin '" + p->name()
-				+ "' has field extraction capability but is not compatible with any enabled event source");
+			auto sname = plugin->event_source();
+			m_state->sources.insert(empty_src, sname);
+			m_state->loaded_sources.insert(sname);
```
FedeDP, line 57: So, what's the difference between `sources` and `loaded_sources` ?

userspace/falco/app_actions/process_events.cpp
```diff
@@ -38,20 +40,32 @@ using namespace falco::app;
...
 	sinsp_evt* ev;
 	stats_writer::collector stats_collector(statsw);
 	uint64_t duration_start = 0;
 	uint32_t timeouts_since_last_success_or_msg = 0;
-	std::size_t source_idx;
-	bool source_idx_found = false;
 	token_bucket rate_limiter;
 	bool rate_limiter_enabled = m_state->config->m_notifications_rate > 0;
+	bool source_engine_idx_found = false;
+	bool is_capture_mode = source.empty();
```
FedeDP, line 60: This is startup time fixed value right? Can't we store it somewhere in application class?

userspace/falco/app_actions/process_events.cpp
```diff
@@ -38,20 +40,32 @@ using namespace falco::app;
...
+	bool syscall_source_engine_idx = m_state->sources.at(falco_common::syscall_source)->engine_idx;
```
FedeDP, line 61: Same as above; i am worried about the new `do_inspect` performances.

userspace/falco/app_actions/select_event_sources.cpp
```diff
@@ -17,65 +17,54 @@ using namespace falco::app;
...
-		std::copy(m_state->enabled_sources.begin(), m_state->enabled_sources.end(), std::ostream_iterator<std::string>(os, ","));
-		std::string result = os.str();
-		result.pop_back();
-		falco_logger::log(LOG_INFO, "Enabled event sources: " + result + "\n");
+	if(m_state->enabled_sources.empty())
+	{
+		return run_result::fatal("Must enable at least one event source");
 	}
 
+	/* Print all enabled sources. */
```
FedeDP, line 62: I'd move this to another app action :)

FedeDP (commented):
First-pass light review; i will surely need a second deeper review.

userspace/falco/app_actions/select_event_sources.cpp
jasondellaluce: I'm not sure! I think as app actions as pieces of Falco behavior that depend on its options and configurations. Here we print the list of sources unconditionally and I'm not sure it's worth isolating this code in an action just for this little purpose. WDYT?

userspace/falco/app_actions/init_falco_engine.cpp
jasondellaluce: Good point I'm gonna try to respond to this point and some of the other questions below. So, for event sources we maintain in the app state three things:
- `loaded_sources`: a set of names all loaded event sources. By default this contains `syscall` that is provided out-of-the-box by Falco, and the event sources coming from the loaded plugins.
- `enabled_sources`: a set of names all event sources and that are enabled for event processing (either in capture/offline or in live mode). By default this is the same as `loaded_sources`, and can then be changed with the `--enable-source` and `--disable-source` options.
- `sources`: a list indexed by source name that contains all information mapped to a given source. For example, we map 1 inspector for each source name, 1 filtercheck list, and so on

The action of adding an event source to the rule engine relies on the `loaded_sources` list, because the rule engine must be aware of all sources no matter if they are actually enabled for event processing or not, because otherwise it will not be able to load the configured rulesets. Doing otherwise will deteriorate the UX as Falco would stop with error at startup if configured with even a single rule attached to a non-enabled event source. This is also why the `syscall` event source is added unconditionally.

userspace/falco/app_actions/load_plugins.cpp
jasondellaluce: See comment above

userspace/falco/app_actions/load_plugins.cpp
jasondellaluce: See comment above. I agree that `sources` is a poor naming decision. Do you have any suggestions? Maybe we should go for something simple like `source_infos` (which is what this really is about)?

userspace/falco/app_actions/process_events.cpp
jasondellaluce: We can't store this in the app state because `do_inspect` is now potentially run in different threads, each with a different value of `source` (this is also why the string is not passed as reference anymore).

However, this is just at startup time when the event processing loop is started, but it's out of the loop itself. As such, this does not impact performance at runtime.

userspace/falco/app_actions/process_events.cpp
jasondellaluce: Same as above. This is out of the event processing loop and is done before entering in it, so this should not affect runtime performance

userspace/falco/app_actions/init_falco_engine.cpp
jasondellaluce: Note that the "all enabled by default" and the `--disable-source` options behavior are the same as when we had k8s_audit support bundled into Falco. I thought it made sense to reproduce the same UX.

FedeDP: Makes sense! Thanks Jason!

userspace/falco/app_actions/load_plugins.cpp
FedeDP: Yeah i agree, `sources_info` is a much better name imho :)

userspace/falco/app_actions/process_events.cpp
FedeDP: I see, thank you! That makes totally sense

userspace/falco/app_actions/select_event_sources.cpp
FedeDP: No problem, not a big deal anyway!

FedeDP:
Local build is failing with 
```
In file included from /home/federico/Work/falco/userspace/engine/filter_evttype_resolver.cpp:17:
/home/federico/Work/falco/userspace/engine/filter_evttype_resolver.h: In static member function ‘static void falco_event_types::check_range(uint16_t)’:
/home/federico/Work/falco/userspace/engine/filter_evttype_resolver.h:36:36: error: ‘range_error’ is not a member of ‘std’
36 |                         throw std::range_error("invalid event type");
```
Do i miss anything? 
This is a clean build!

jasondellaluce:
> Local build is failing with
> 
> ```
> In file included from /home/federico/Work/falco/userspace/engine/filter_evttype_resolver.cpp:17:
> /home/federico/Work/falco/userspace/engine/filter_evttype_resolver.h: In static member function ‘static void falco_event_types::check_range(uint16_t)’:
> /home/federico/Work/falco/userspace/engine/filter_evttype_resolver.h:36:36: error: ‘range_error’ is not a member of ‘std’
> 36 |                         throw std::range_error("invalid event type");
> ```
> 
> Do i miss anything? This is a clean build!

Huh, I think this happens after rebasing on top of https://github.com/falcosecurity/falco/pull/2151. Does this happen on `master` too?

FedeDP:
> Does this happen on master too?

Confirm! Thanks! Weird that the issue is not spotted by the CI though.

FedeDP:
Fixed by https://github.com/falcosecurity/falco/pull/2197

jasondellaluce:
> Fixed by https://github.com/falcosecurity/falco/pull/2197

Tests are back being all-green!

leogr (approved):
This is awesome! :star_struck:

PR #2190, opened 2022-09-05
update(userspace/engine): support alternative plugin version requirements in checks

### LINKED ISSUE #2150: improve rule engine plugin dependencies
**Motivation**

Since rule engine version 10, Falco allows specifying in each ruleset a list of required plugin and their version. Example (https://github.com/falcosecurity/plugins/blob/fd296e843f7516939fe6201bf2f4646eb0485055/plugins/k8saudit/rules/k8s_audit_rules.yaml#L20):
```yaml
- required_plugin_versions:
  - name: k8saudit
    version: 0.1.0
  - name: json
    version: 0.3.0
```

This is good for automatically tracking down and verifying that all dependencies are satisfied. However, this is not sufficient due to https://falco.org/docs/plugins/#plugin-event-sources-and-interoperability:
> Events returned by plugins with event sourcing capability have an event source that describes the event's information. This is distinct from the plugin name to allow for multiple plugin implementations to generate the same kind of events.

We have few problems:
- Each ruleset can only specify a **conjunction** of required plugin names and versions, and does not support alternatives (e.g. `k8saudit` or `eks-k8saudit`, `gke-k8saudit`, etc)
- In the plugin system we don't have an explicit naming for the "set of supported fields of the plugin", so we can only rely either on the plugin name or on its event source (if available).

**Feature**

Make the rule engine more expressive and support either:
- Specifying more than one plugin name/version in disjunction, so that users can configure Falco with plugin alternatives
- Specifying event sources to depend on (those are not versioned though)
- Both

**Alternatives**

Not yet, I'm wide open for discussion.

jasondellaluce:
**What type of PR is this?**

/kind feature

/area engine

/area tests

This PR updates the rule engine to support multiple alternatives plugin version requirements definitions in rulesets. The problem was first mentioned in https://github.com/falcosecurity/plugins/pull/134#discussion_r935806326, and documented in the #2150 issue.

Given that a new feature is added, the rule engine version has been bumped (to v15). Moreover, unit tests have been introduced to make sure that the plugin version checks work as expected.

Fixes #2150

In this proposal, rulesets will be able to support the following syntax:
```yaml
- required_plugin_versions:
  - name: k8saudit
    version: 0.1.0
    alternatives:
      - name: k8saudit-other
        version: 0.4.0
```

Here, `alternatives` represents a list of alternative plugin name-version pairs to be checked as alternatives to the primary one. Maintaining a primary plugin requirement definition, and making `alternatives` optional, allow backward compatibile definitions such as:
```yaml
- required_plugin_versions:
  - name: k8saudit
    version: 0.1.0
```

Given so, `required_plugin_versions` now implicitly defines a CNF ([Conjunctive Normal Form](https://en.wikipedia.org/wiki/Conjunctive_normal_form). Rulesets are correctly loaded if **all** the `required_plugin_versions` definitions are satisfied. Each entry is satisfied if **at least one** of its alternatives is satisfied.

```release-note
update(userspace/engine): support alternative plugin version requirements in checks
```

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -22,6 +22,7 @@ set(
   engine/test_filter_macro_resolver.cpp
   engine/test_filter_evttype_resolver.cpp
   engine/test_filter_warning_resolver.cpp
+  engine/test_plugin_requirements.cpp
   falco/test_configuration.cpp
 )
 
--- a/tests/engine/test_plugin_requirements.cpp
+++ b/tests/engine/test_plugin_requirements.cpp
@@ -1,0 +1,269 @@
+/*
+Copyright (C) 2022 The Falco Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include <memory>
+#include <catch.hpp>
+#include "falco_engine.h"
+
+static void check_requirements(
+        bool expect_success,
+        const std::vector<falco_engine::plugin_version_requirement>& plugins,
+        const std::string& ruleset_content)
+{
+    std::string err;
+    std::unique_ptr<falco_engine> e(new falco_engine());
+    falco::load_result::rules_contents_t c = {{"test", ruleset_content}};
+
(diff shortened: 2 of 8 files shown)


leogr (approved):
Thank you!

/milestone 0.33.0

FedeDP:
I understood it correctly that we are expected to ship proper `alternatives` in default config when we support multiple plugins offering similar functionalities?

jasondellaluce:
> I understood it correctly that we are expected to ship proper alternatives in default config when we support multiple plugins offering similar functionalities?

By `default config` do you mean `default rulesets`? If so, yes, in our default rulesets we'll provide all the officially-supported known plugin alternatives for the rulesets we maintain. A first-class example are `k8saudit` and `k8saudit-eks` that's in the working (https://github.com/falcosecurity/plugins/pull/134).

FedeDP:
Uh yes! thank you! Makes sense!

FedeDP (approved):
/approve

PR #2193, opened 2022-09-06
Add new rule that detect retrieving environment variables from /proc files

hi120ki:
Signed-off-by: Hi120ki <12624257+hi120ki@users.noreply.github.com>

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind release

> /kind rule-update

/kind rule-create

> /area build

> /area engine

/area rules

> /area tests

> /area proposals

> /area CI

These days, it's common to pass a secret to an application running on a container via an environment variable. Secrets store very sensitive information such as cloud credentials and JWT secretkeys, so it is important to detect when environment variables have been retrieved by an attack.

And if the application has SSRF or path traversal vulnerabilities, the attacker can steal the environment variables of the application process by reading `/proc/self/environ` and `/proc/1/environ` under the `/proc` directory.

<https://linux.die.net/man/5/proc>

For example, SSRF:

```
curl -X POST -d "url=file:///proc/self/environ" https://example.com/ssrf
curl -X POST -d "url=file:///proc/1/environ" https://example.com/ssrf
```

Path traversal:

```
curl --path-as-is https://example.com/path/../../proc/self/environ
curl --path-as-is https://example.com/path/../../proc/1/environ
```

In MITRE ATT&CK, this technique is categorised as [T1552 : Unsecured Credentials](https://attack.mitre.org/techniques/T1552/), and Sub-technique [T1552.001 : Unsecured Credentials: Credentials In Files](https://attack.mitre.org/techniques/T1552/001/), and Detection is [DS0022 - File - File Access](https://attack.mitre.org/datasources/DS0022/), and component is [T1003.007 OS Credential Dumping: Proc Filesystem](https://attack.mitre.org/techniques/T1003/007/).

I propose the following as a rule to detect environment variable reading from files under the `/proc` directory.

```yaml
- list: proc_environ_file_names
  items: [/proc/self/environ, /proc/1/environ]

- macro: proc_environ_files
  condition: >
    fd.name in (proc_environ_file_names)

- rule: Read environment variable from /proc files
  desc: An attempt to read process environment variables from /proc files
  condition: >
    container and open_read and proc_environ_files
  enabled: true
  output: >
    Environment variables were retrieved from /proc files (user=%user.name user_loginuid=%user.loginuid program=%proc.name
    command=%proc.cmdline file=%fd.name parent=%proc.pname gparent=%proc.aname[2] ggparent=%proc.aname[3] gggparent=%proc.aname[4] container_id=%container.id image=%container.image.repository)
  priority: WARNING
  tags: [filesystem, mitre_credential_access, mitre_discovery]
```

> Adding a host to the detection target causes false positives by `systemd`, so limited to container.

I'm waiting for your comments on detecting reading these files and my proposed rules.

Thank you.

Fixes #

```release-note
Add rule to detect an attempt to read process environment variables from /proc files
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -3202,3 +3202,21 @@
 # Application rules have moved to application_rules.yaml. Please look
 # there if you want to enable them by adding to
 # falco_rules.local.yaml.
+
+- list: proc_environ_file_names
+  items: [/proc/self/environ, /proc/1/environ]
+
+- macro: proc_environ_files
+  condition: >
+    fd.name in (proc_environ_file_names)
+
+- rule: Read environment variable from /proc files
+  desc: An attempt to read process environment variables from /proc files
+  condition: >
+    container and open_read and proc_environ_files
+  enabled: true
+  output: >
+    Environment variables were retrieved from /proc files (user=%user.name user_loginuid=%user.loginuid program=%proc.name
+    command=%proc.cmdline file=%fd.name parent=%proc.pname gparent=%proc.aname[2] ggparent=%proc.aname[3] gggparent=%proc.aname[4] container_id=%container.id image=%container.image.repository)
+  priority: WARNING
+  tags: [filesystem, mitre_credential_access, mitre_discovery]
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -3204,7 +3204,7 @@
 # falco_rules.local.yaml.
 
 - list: known_binaries_to_read_environment_variables_from_proc_files
-  items: [scsi_id]
+  items: [scsi_id, argoexec]
 
 - rule: Read environment variable from /proc files
   desc: An attempt to read process environment variables from /proc files
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -3203,17 +3203,11 @@
 # there if you want to enable them by adding to
 # falco_rules.local.yaml.
 
-- list: proc_environ_file_names
-  items: [/proc/self/environ, /proc/1/environ]
-
-- macro: proc_environ_files
-  condition: >
-    fd.name in (proc_environ_file_names)
-
 - rule: Read environment variable from /proc files
   desc: An attempt to read process environment variables from /proc files
   condition: >
-    container and open_read and proc_environ_files
+    open_read and (fd.name glob /proc/*/environ)
+    and not proc.name in (systemctl, systemd-detect-, cloud-id)
   enabled: true
   output: >
     Environment variables were retrieved from /proc files (user=%user.name user_loginuid=%user.loginuid program=%proc.name
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -3211,7 +3211,6 @@
   condition: >
     container and open_read and (fd.name glob /proc/*/environ)
     and not proc.name in (known_binaries_to_read_environment_variables_from_proc_files)
-  enabled: true
   output: >
     Environment variables were retrieved from /proc files (user=%user.name user_loginuid=%user.loginuid program=%proc.name
     command=%proc.cmdline file=%fd.name parent=%proc.pname gparent=%proc.aname[2] ggparent=%proc.aname[3] gggparent=%proc.aname[4] container_id=%container.id image=%container.image.repository)
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -3209,7 +3209,7 @@
 - rule: Read environment variable from /proc files
   desc: An attempt to read process environment variables from /proc files
   condition: >
-    container and open_read and (fd.name glob /proc/*/environ)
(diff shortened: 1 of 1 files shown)


darryk10:
Hi @hi120ki, thanks for contributing first of all :) 
Amazing rule and I agree with you on  the security use case.
I also do like the container scoping to avoid the noise since we need to rely on `open_read` events. 
Just wondering if we can open the rule to hosts, just whitelisting `systemd` as you mentioned, or you feel it might introduce too much default noise.
I'm going to test the rule as well to further reduce the noise if possible :) 
Thanks

hi120ki:
@darryk10 @jasondellaluce I update my proposed rule to expand host. 
```yaml
- rule: Read environment variable from /proc files
  desc: An attempt to read process environment variables from /proc files
  condition: >
    open_read and (fd.name glob /proc/*/environ)
    and not proc.name in (systemctl, systemd-detect-, cloud-id)
  enabled: true
  output: >
    Environment variables were retrieved from /proc files (user=%user.name user_loginuid=%user.loginuid program=%proc.name
    command=%proc.cmdline file=%fd.name parent=%proc.pname gparent=%proc.aname[2] ggparent=%proc.aname[3] gggparent=%proc.aname[4] container_id=%container.id image=%container.image.repository)
  priority: WARNING
  tags: [filesystem, mitre_credential_access, mitre_discovery]
```
Host machine has many processes, and there is many `/proc/*/environ` files.(`*` is process id).
So, updated rule uses use `glob` instead of `in` and enables to detect any process's environ files.

Then, in host machine some system process such as `systemd` uses `/proc/*/environ` files, and I added allowlist into rule.
If you find other  system processes, cloud you add to allowlist?

hi120ki:
@darryk10 @jasondellaluce I tryed to apply my rule to my testbed, there are so many false positive detection from host machine. (For example, `systemd-tmpfile` `systemd-user-se` `systemd-update-`)
Of cource, it is nice to add all systemd and other system-level process binary to allowlist if we can, but systemd updates fast, and it is impossible to maintain this allowlist.

So, we should restrict this rule only container. I'm waiting for your comments.

darryk10:
Hi @hi120ki, thanks for the further test. I've been testing the host condition and I agree the rule generates too noise. Even though would be good to have a role covering host as well, I would suggest to keep this use case for container only to reduce the noise as you suggested in the PR. 
Regarding the condition for container, instead, it's very accurate and very low false positive. In addition I observed good true positives on malwares and bad activities so definitely a rule we want in our ruleset. 
Happy to approve the PR once reverted on container use case only.
Really appreciate your work and looking forward to receiving new rules from you :)

hi120ki:
@darryk10 Hi, I run this rule on my testbed, and found one false positive detection in `scsi_id` proc.
This proc runs by [gcp-compute-persistent-disk-csi-driver](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver), it is common in GKE cluster.
I added this into allowlist, could you review it?

rules/falco_rules.yaml
```diff
@@ -3202,3 +3202,18 @@
...
+
+- list: known_binaries_to_read_environment_variables_from_proc_files
+  items: [scsi_id]
+
+- rule: Read environment variable from /proc files
+  desc: An attempt to read process environment variables from /proc files
+  condition: >
+    container and open_read and (fd.name glob /proc/*/environ)
+    and not proc.name in (known_binaries_to_read_environment_variables_from_proc_files)
+  enabled: true
```
darryk10, line 3214:
```suggestion
```

darryk10: This isn't necessary since rules are enabled by default.

darryk10:
@hi120ki I think you also need to specify the change log in the PR otherwise the PR isn't mergeable.

darryk10 (approved):
Thanks for your contribution :)

jasondellaluce (dismissed):
/approve

rules/falco_rules.yaml
```diff
@@ -3202,3 +3202,17 @@
...
 # there if you want to enable them by adding to
 # falco_rules.local.yaml.
+
+- list: known_binaries_to_read_environment_variables_from_proc_files
+  items: [scsi_id, argoexec]
+
+- rule: Read environment variable from /proc files
+  desc: An attempt to read process environment variables from /proc files
+  condition: >
+    container and open_read and (fd.name glob /proc/*/environ)
```
Kaizhe, line 3212: to optimize rule checking performance, it is good to put `open_read` in the beginning of the rule.

hi120ki: Thank you! I put open_read in the beginning of the rule.

darryk10 (approved):

jasondellaluce (approved):
/approve

leogr (approved):

leogr:
/close

leogr:
/hold

for the CI to complete

PR #2200, opened 2022-09-12
fix(userspace/engine): avoid reading duplicate exception values

jasondellaluce:
**What type of PR is this?**

/kind bug

/area engine

The recursive function that reads rule exception values does not clear out the value list, so consecutive list-type exception values may end up containing in duplicate items.

The regression seems to be introduced in https://github.com/falcosecurity/falco/pull/2098

A minimum example to reproduce the issue is reported below:
```yaml
- rule: Sample Rule
  desc: A sample rule
  output: Sample output (type=%evt.type)
  priority: ERROR
  condition: >
    evt.dir = <
  exceptions:
    - name: sample_exception
      fields: [evt.type, proc.name]
      comps: [in, in]
      values:
        - [[openat], [ls]]
```
This rule and its exceptions will be compiled into the following condition (note the rolled-over duplication of `openat`):
```
(evt.dir = < and not (evt.type in (openat) and proc.name in (openat, ls)))
```
With this fix, the rule is correctly compiled as:
```
(evt.dir = < and not (evt.type in (openat) and proc.name in (ls)))
```

```release-note
fix(userspace/engine): avoid reading duplicate exception values
```

### DIFF
--- a/userspace/engine/rule_reader.cpp
+++ b/userspace/engine/rule_reader.cpp
@@ -139,9 +139,9 @@ static void decode_exception_info_entry(
 	if (val.IsSequence())
 	{
 		out.is_list = true;
-		rule_loader::rule_exception_info::entry tmp;
 		for(const YAML::Node& v : val)
 		{
+			rule_loader::rule_exception_info::entry tmp;
 			rule_loader::context lctx(v, rule_loader::context::EXCEPTION, "", valctx);
 
 			// Optional is always false once you get past the outer values


FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

VadimZy:
Technically speaking, the implementation of `decode_exception_info_entry` is error-prone.
- The return value is ignored at the root level, 
- a case where `tmp.items.empty()` is ignored (left to the clients)
- node checks only at the root level, no validation after in the tree branches and leafs 
- multiple copying of tmp objects, even with possible moving it will require a lot of allocations for the inner entry structures; 

```
std::unique_ptr<rule_loader::rule_exception_info::entry> decode_exception_info_entry(
	const YAML::Node& node)
{
	using entry_t = rule_loader::rule_exception_info::entry;
	if (node.IsDefined())
	{
		if (node.IsScalar())
		{
			std::unique_ptr<entry_t> ret(new entry_t());
			
			ret->is_list = false;
			if (YAML::convert<string>::decode(node, ret->item))
			{
				return ret;
			}
			return {};
		}
		if (node.IsSequence())
		{
			std::unique_ptr<entry_t> ret(new entry_t());
			ret->is_list = true;
			for(const YAML::Node& v : node)
			{
				auto item = decode_exception_info_entry(v);
				if (item != nullptr)
				{
					ret->items.push_back(std::move(item));
				}
			}
			return ret->items.empty() ? std::unique_ptr<entry_t>() : std::move(ret);
		}
	}
	return {};
}
```
same for `read_rule_exceptions` function
 
`rule_exception_info, rule_exception_info::entry` should have their constructors with throwing validations.

PR #2201, opened 2022-09-12
update(userspace/falco, cmake): updated libs to latest master.

FedeDP:
**What type of PR is this?**

/kind feature

/area build

/area engine

Adapted API to sinsp::open API break, and simple consumer API break.

Fixes #

```release-note
update: bumped libs to fd46dd139a8e35692a7d40ab2f0ed2016df827cf.
```


userspace/falco/app_actions/init_inspectors.cpp
```diff
@@ -48,12 +48,11 @@ static void init_syscall_inspector(
 
 	if(!opts.all_events)
 	{
-		// Drop EF_DROP_SIMPLE_CONS kernel side
-		inspector->set_simple_consumer();
-		// Eventually, drop any EF_DROP_SIMPLE_CONS event
-		// that reached userspace (there are some events that are not syscall-based
-		// like signaldeliver, that have the EF_DROP_SIMPLE_CONS flag)
-		inspector->set_drop_event_flags(EF_DROP_SIMPLE_CONS);
+		m_state->ppm_sc_of_interest = inspector->enforce_simple_ppm_sc_set();
```
FedeDP, line 51: New syscalls/tp of interest APIs.

userspace/falco/app_actions/load_rules_files.cpp
```diff
@@ -39,7 +39,7 @@ void application::check_for_ignored_events()
 			continue;
 		}
 
-		if(!sinsp::simple_consumer_consider_evtnum(evtnum))
+		if(!simple_consumer_consider(etable[evtnum].flags, false))
```
FedeDP, line 42: `simple_consumer_consider` is now a Falco only concept.

userspace/falco/app_actions/open_inspector.cpp
```diff
@@ -22,13 +22,15 @@ limitations under the License.
...
+#define FALCO_BPF_ENV_VARIABLE "FALCO_BPF_PROBE"
+
 using namespace falco::app;
 
 application::run_result application::open_offline_inspector()
 {
 	try
 	{
-		m_state->offline_inspector->open(m_options.trace_filename);
+		m_state->offline_inspector->open_savefile(m_options.trace_filename, 0);
```
FedeDP, line 33: TODO: this should really be an optional param. A libs PR will fix it.

userspace/falco/app_actions/open_inspector.cpp
```diff
@@ -67,14 +68,47 @@ application::run_result application::open_live_inspector(
...
+					const char *home = std::getenv("HOME");
+					if(!home)
+					{
+						return run_result::fatal("Cannot get the env variable 'HOME'");
+					}
+					char full_path[PATH_MAX];
+					snprintf(full_path, PATH_MAX, "%s/%s", home, FALCO_PROBE_BPF_FILEPATH);
+					bpf_probe_path = full_path;
+				}
+				inspector->open_bpf(2048, bpf_probe_path, m_state->ppm_sc_of_interest, m_state->tp_of_interest);
```
FedeDP, line 93: TODO: `2048` will be a value parsed from Falco config; see #2164 ; it will be ported in this PR.

Moreover, IMHO i'd update `open_bpf` signature, putting the most important param, ie: `bpf_probe_path`, as first param; all the others should be optional ones.

A 0-valued `buffer_dimension` should enforce a default one in libscap.

userspace/falco/app_actions/open_inspector.cpp
```diff
@@ -67,14 +68,47 @@ application::run_result application::open_live_inspector(
...
+			{
+				return run_result::fatal("Cannot use the BPF probe: " + std::string(e.what()));
+			}
+		}
+		else /* Kernel module (default). */
 		{
-			inspector->open();
+			try
+			{
+				inspector->open_kmod(2048, m_state->ppm_sc_of_interest, m_state->tp_of_interest);
```
FedeDP, line 105: A 0-valued `buffer_dimension` should enforce a default one in libscap.

userspace/falco/app_actions/open_inspector.cpp
```diff
@@ -22,13 +22,15 @@ limitations under the License.
 
 #include "application.h"
 
+#define FALCO_BPF_ENV_VARIABLE "FALCO_BPF_PROBE"
```
FedeDP, line 25: TODO: should we drop anything in the cmake files?

FedeDP: Yep, we needed to drop also `SCAP_BPF_PROBE_ENV_VAR_NAME` cmake variable, as it was unsued now.
Dropped in latest commit.

userspace/falco/app_actions/init_inspectors.cpp
```diff
@@ -48,12 +48,11 @@ static void init_syscall_inspector(
...
+		m_state->tp_of_interest = inspector->enforce_sinsp_state_tracepoints();
+		// We are not interested in sched_switch tracepoint,
+		// highly noisy and not useful for state/events enrichment.
+		m_state->tp_of_interest.erase(SCHED_SWITCH);
```
Andreagit97, line 55: Maybe we can remove it directly from the `sinsp_state_set` since it is not so necessary and quite noisy for every client :thinking:

userspace/falco/app_actions/load_rules_files.cpp
Andreagit97: Uhm here probably is not so simple, we need to check if this event is considered or not by Falco, so we need to check our set `ppm_sc_of_interest`. The problem is that Falco has no concepts of `event` but it only has concept of `ppm_sc` code, maybe we can provide a method in sinps to translate it :thinking:

userspace/falco/app_actions/print_ignored_events.cpp
```diff
@@ -28,7 +28,7 @@ void application::print_all_ignored_events()
 	std::set<string> ignored_event_names;
 	for(uint32_t j = 0; j < PPM_EVENT_MAX; j++)
 	{
-		if(!sinsp::simple_consumer_consider_evtnum(j))
+		if(!simple_consumer_consider(etable[j].flags))
```
Andreagit97, line 31: Probably also here we need to find a way to check which events are in the Falco interesting set, the only problem here is that we don't have the Falco set populated at this point in time

Andreagit97, line 31: In this function we print together event names and an syscall names :thinking:

userspace/falco/app_actions/init_inspectors.cpp
FedeDP: Yep i think that could be a neat idea. It seems like sched_switch events count between 5-10% of the total number of events, so we could save both "wake up time" and ring buffer space for everyone.

userspace/falco/app_actions/print_ignored_events.cpp
FedeDP: Yep...i am not sure how useful is this function, but it was there and we are not going to break user cmdline API :)

userspace/falco/app_actions/load_rules_files.cpp
FedeDP: I agree with you; but we are already storing set used by Falco; i think enriching this check by also checking that the `ppm_sc` linked to each evtnum are present in Falco ppm_sc set should work.

userspace/falco/app_actions/init_inspectors.cpp
jasondellaluce: Totally agree on this

userspace/falco/app_actions/load_rules_files.cpp
FedeDP: Waiting on https://github.com/falcosecurity/libs/pull/609 for this

userspace/falco/app_actions/open_inspector.cpp
FedeDP: Done.

userspace/falco/app_actions/open_inspector.cpp
FedeDP: This is not present in sinsp/scap (and i am not sure if it would even be merged) so let's skip it for now.

userspace/falco/app_actions/init_inspectors.cpp
FedeDP: I think we should let clients remove it, just like Falco is doing, because `sched_switch` somewhat enriches sinsp state, but they enrich something that is unused by Falco.
And of course dropping it is a good perf tradeoff.

userspace/falco/app_actions/process_events.cpp
```diff
@@ -189,7 +189,9 @@ application::run_result application::do_inspect(
 			return run_result::fatal("Drop manager internal error");
 		}
 
-		if(!ev->simple_consumer_consider() && !m_options.all_events)
+		/* If we have not set the `-A` flag and the event is unused or old, interrupt the flow */
+		uint16_t evt_type = ev->get_type();
+		if(!m_options.all_events && (sinsp::is_unused_event(evt_type) || sinsp::is_old_version_event(evt_type)))
```
Andreagit97, line 194: Not sure about the utility of this check... here we check for an old version event or for an unused event and we interrupt the flow, but this should be always true also if the `-A` flag is enabled :thinking:

leogr: I agree that having this check now is a bit weird because it is always true, or we are saying sinsp is not consistent.
Could we introduce a simple regression test to be relatively sure of our code without checking the same logic at runtime twice?

(If it were too late for this release, I would propose to leave it as-is and  `todo:`).

Andreagit97: my question here is why we are dropping old version events?

Andreagit97: I think that this check is no more useful I will remove it, unused event should be filtered by sinsp and old events should be considered if received and not dropped

leogr (approved):

userspace/falco/app_cmdline_options.cpp
```diff
@@ -155,7 +155,7 @@ void cmdline_options::define()
 #else
 		("c",                             "Configuration file. If not specified tries " FALCO_SOURCE_CONF_FILE ", " FALCO_INSTALL_CONF_FILE ".", cxxopts::value(conf_filename), "<path>")
 #endif
-		("A",                             "Monitor all events, including those with EF_DROP_SIMPLE_CONS flag.", cxxopts::value(all_events)->default_value("false"))
+		("A",                             "Monitor all events, including not interesting ones. Please use the `--i` command line option to see the ignored events. This option has no effect if the capture is not live", cxxopts::value(all_events)->default_value("false"))
```
jasondellaluce, line 158:
```suggestion
		("A",                             "Monitor all events, including not interesting ones. Please use the `-i` command line option to see the ignored events. This option has no effect if the capture is not live", cxxopts::value(all_events)->default_value("false"))
```

jasondellaluce (commented):
Just a minor suggestion

userspace/falco/app_cmdline_options.cpp
Andreagit97: :facepalm: great catch!

leogr: I'd change the last sentence with `This option is implicit when the capture is not live.` or something similar.

leogr (approved):

jasondellaluce (approved):
/approve

Thanks!!

PR #2212, opened 2022-09-16
Set Default Driver Download Location

distortedsignal:
TL;DR: Set a default driver download location to
https://download.falco.org/driver to facilitate a more download-and-go user experience for the driver-loader script.

Every few months or so I come back to the falco project and need to get reacquainted with the structure of where things are. How do I install the kernel modules? Where should I download the kernel modules from? Where are the eBPF probes? It's pretty consistently 20 minutes to an hour of work that needs to be repeated two or more times per year.

One of the things that promised to help with this was the falco-driver-loader script. It's really good! As soon as you know all the inputs to the script. So instead of relying on future me to reverse engineer the inputs from a lack of knowledge, present me is going to update the script to include the public distribution point of eBPF probes and kernel modules.

Signed-off-by: Tom Kelley <distortedsignal@gmail.com>

**What type of PR is this?**

/kind bug

/kind feature

/area scripts

Provide a good default for the download URL for eBPF probes and kernel modules. This helps users to get up and running with Falco faster.

N/A - it's easier to open a PR than to open an issue.

```release-note

The `falco-driver-loader` script now uses the `download.falco.org/driver` distribution point as the default driver distribution point.

```

### DIFF
--- a/scripts/falco-driver-loader
+++ b/scripts/falco-driver-loader
@@ -602,6 +602,8 @@ print_usage() {
 	echo ""
 	echo "Environment variables:"
 	echo "  DRIVERS_REPO             specify different URL(s) where to look for prebuilt Falco drivers (comma separated)"
+	echo "                           The default value is https://download.falco.org/driver - the public distribution"
+	echo "                           point of downloadable falco kernel drivers"
 	echo "  DRIVER_NAME              specify a different name for the driver"
 	echo "  DRIVER_INSECURE_DOWNLOAD whether you want to allow insecure downloads or not"
 	echo ""
@@ -621,7 +623,7 @@ if ! hash sed > /dev/null 2>&1; then
 fi
 KERNEL_VERSION=$(uname -v | sed 's/#\([[:digit:]]\+\).*/\1/')
 
-DRIVERS_REPO=${DRIVERS_REPO:-"@DRIVERS_REPO@"}
+DRIVERS_REPO=${DRIVERS_REPO:-"https://download.falco.org/driver"}
 
 if [ -n "$DRIVER_INSECURE_DOWNLOAD" ]
 then


jasondellaluce:
cc @leogr, @FedeDP

/milestone 0.33.0

I'm gonna insert this in the current release milestone, and eventually slip it to 0.34.0 should the discussion require more time.

FedeDP:
Hi @distortedsignal ! 
The `falco-driver-loader` script gets build time configured through cmake `configure_file`; the `DRIVERS_REPO` variable is set here: https://github.com/falcosecurity/falco/blob/master/CMakeLists.txt#L115
The script itself is configured here: https://github.com/falcosecurity/falco/blob/master/scripts/CMakeLists.txt#L32

As you can see, you can also use the `DRIVERS_REPO` env variable to bypass the default value.
IMHO there is no need for this patch.
/cc @leogr

distortedsignal:
@FedeDP - I agree with all of your points except the last one. Because the `falco-driver-loader` script is [mentioned in the official docs as the way to load the kernel module](https://falco.org/docs/getting-started/installation/#install-driver), I think there should be some kind of pointer to the official download links for the presently available kernel modules. Even if this is just a doc change (adding something to the help text would be sufficient), I think there needs to be a statement of "here is how to do this with the present up-to-date kernel modules."

leogr:
> @FedeDP - I agree with all of your points except the last one. Because the `falco-driver-loader` script is [mentioned in the official docs as the way to load the kernel module](https://falco.org/docs/getting-started/installation/#install-driver), I think there should be some kind of pointer to the official download links for the presently available kernel modules. Even if this is just a doc change (adding something to the help text would be sufficient), I think there needs to be a statement of "here is how to do this with the present up-to-date kernel modules."

Hey @distortedsignal 

The file you are trying to change is the script's source code (the source code is not intended to be used as-is). As explained by @FedeDP, the source code is then processed by cmake, and all placeholders are replaced with default values. Finally, the built script is included in [all Falco packages](https://falco.org/docs/getting-started/download/#packages), so it automatically gets installed with Falco.

If you installed Falco as intended by the documentation, you use `falco-driver-loader` directly (you don't need to download it separately), and it will already point to https://download.falco.org/driver

Thus, unless I'm missing something, there's no need for this patch.

PR #2214, opened 2022-09-20
new: configure syscall buffer dimension from Falco

### LINKED ISSUE #2208: add an option to pick the buffer dimension
**Motivation**

As a natural consequence that sinsp is adding the ability to set the ring buffer dimension (see https://github.com/falcosecurity/libs/pull/584), we should enable the end user to use this option.

From initial discussions ( refs https://github.com/falcosecurity/falco/pull/2164#issuecomment-1220479171, https://github.com/falcosecurity/falco/pull/2164#issuecomment-1220792012), it was clear that using bytes would not have been the best choice for the users. However, as recently discussed (https://github.com/falcosecurity/libs/pull/584#discussion_r970580068, https://github.com/falcosecurity/libs/pull/584#issuecomment-1246407597), sinsp will still use bytes to leave a certain degree of freedom and avoid future API break in case we have to change the criteria in the future.

From a user perspective, specifying the buffer size in absolute value does not make much sense. It would be hard to calculate the correct value, and it may be different in some systems, and the optimal value might change from version to version. So adding a `--buffer-dim` flag using bytes is not an option on the table.

One possible solution (initially proposed in https://github.com/falcosecurity/falco/pull/2164#issuecomment-1220792012) is to provide some hardcoded (or automatically calculated) presets and let the user pick one of them. I think it's the best solution from a UX perspective.

Moreover, if the default preset matches the current buffer size value (8MB), this new feature will not change the default behavior, so it won't disrupt those blindly upgrading Falco.

**Feature**

Add a fixed array of presets of buffer dimensions that the user can easily pick using a command line flag.

Ideally, the array should contain about 10 values.

One of those values must match the current default (8MB). It must be used if the user does not specify any preset.

**Additional context**

We would like to delivery this feature within
/milestone 0.33.0
Assuming the code freeze starts on Sept 26th we have a week and half to implement this feature :crossed_fingers: 

TODO: We still have to think about how to calculate values for this array. The calculation should consider some constraints (values must be a power of 2, multiple of PAGE_SIZE, etc,)

@Andreagit97 @incertum @FedeDP @gnosek @jasondellaluce 

Andreagit97:
**What type of PR is this?**

/kind feature

As discussed in various places (https://github.com/falcosecurity/falco/pull/2164#issuecomment-1220479171, https://github.com/falcosecurity/falco/pull/2164#issuecomment-1220792012, https://github.com/falcosecurity/falco/issues/2208)  this PR adds the possibility to change the syscall buffer dimension through the Falco configuration file.

The approach used is the one described here https://github.com/falcosecurity/falco/issues/2208#issuecomment-1248354349. Users will have an array of indexes that map to precise buffer dimensions in bytes. You can find more information in the `falco.yaml` file, but just for simplicity I report them here:

```yaml
# --- [Description]
#
# This is an index that controls the dimension of the syscall buffers.
# The syscall buffer is the shared space between Falco and its drivers where all the syscall events
# are stored.
# Falco uses a syscall buffer for every online CPU, and all these buffers share the same dimension.
# So this parameter allows you to control the size of all the buffers!
#
# --- [Usage]
#
# You can choose between different indexes: from `1` to `10` (`0` is reserved for future uses).
# Every index corresponds to a dimension in bytes:
#
# [(*), 1 MB, 2 MB, 4 MB, 8 MB, 16 MB, 32 MB, 64 MB, 128 MB, 256 MB, 512 MB]
#   ^    ^     ^     ^     ^     ^      ^      ^       ^       ^       ^
#   |    |     |     |     |     |      |      |       |       |       |
#   0    1     2     3     4     5      6      7       8       9       10
#
# As you can see the `0` index is reserved, while the index `1` corresponds to
# `1 MB` and so on.
#
# These dimensions in bytes derive from the fact that the buffer size must be:
# (1) a power of 2.
# (2) a multiple of your system_page_dimension.
# (3) greater than `2 * (system_page_dimension)`.
#
# According to these constraints is possible that sometimes you cannot use all the indexes, let's consider an
# example to better understand it:
# If you have a `page_size` of 1 MB the first available buffer size is 4 MB because 2 MB is exactly
# `2 * (system_page_size)` -> `2 * 1 MB`, but this is not enough we need more than `2 * (system_page_size)`!
# So from this example is clear that if you have a page size of 1 MB the first index that you can use is `3`.
#
# Please note: this is a very extreme case just to let you understand the mechanism, usually the page size is something
# like 4 KB so you have no problem at all and you can use all the indexes (from `1` to `10`).
#
# To check your system page size use the Falco `--page-size` command line option. The output on a system with a page
# size of 4096 Bytes (4 KB) should be the following:
#
# "Your system page size is: 4096 bytes."
#
# --- [Suggestions]
#
# Before the introduction of this param the buffer size was fixed to 8 MB (so index `4`, as you can see
# in the default value below).
# You can increase the buffer size when you face syscall drops. A size of 16 MB (so index `5`) can reduce
# syscall drops in production-heavy systems without noticeable impact. Very large buffers however could
# slow down the entire machine.
# On the other side you can try to reduce the buffer size to speed up the system, but this could
# increase the number of syscall drops!
# As a final remark consider that the buffer size is mapped twice in the process' virtual memory so a buffer of 8 MB
# will result in a 16 MB area in the process virtual memory.
# Please pay attention when you use this parameter and change it only if the default size doesn't fit your use case.

syscall_buf_size_preset: 4
```

I added also a new command line option `--page-size` to let the user easily obtain the page size of their system

If you have any suggestions I'm all ears :) 

Fixes #2208 

```release-note
new: configure syscall buffer dimension from Falco
```

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -169,6 +169,61 @@ syscall_event_drops:
 syscall_event_timeouts:
   max_consecutives: 1000
 
+# --- [Description]
+#
+# This is an index that controls the dimension of the syscall buffers.
+# The syscall buffer is the shared space between Falco and its drivers where all the syscall events
+# are stored.
+# Falco uses a syscall buffer for every online CPU, and all these buffers share the same dimension.
+# So this parameter allows you to control the size of all the buffers!
+#
+# --- [Usage]
+#
+# You can choose between different indexes: from `1` to `10` (`0` is reserved for future uses).
+# Every index corresponds to a dimension in bytes:
+#
+# [(*), 1 MB, 2 MB, 4 MB, 8 MB, 16 MB, 32 MB, 64 MB, 128 MB, 256 MB, 512 MB]
+#   ^    ^     ^     ^     ^     ^      ^      ^       ^       ^       ^
+#   |    |     |     |     |     |      |      |       |       |       |
+#   0    1     2     3     4     5      6      7       8       9       10
+#
+# As you can see the `0` index is reserved, while the index `1` corresponds to
+# `1 MB` and so on.
+#
+# These dimensions in bytes derive from the fact that the buffer size must be:
+# (1) a power of 2.
+# (2) a multiple of your system_page_dimension.
+# (3) greater than `2 * (system_page_dimension)`.
+#
+# According to these constraints is possible that sometimes you cannot use all the indexes, let's consider an
+# example to better understand it:
+# If you have a `page_size` of 1 MB the first available buffer size is 4 MB because 2 MB is exactly
+# `2 * (system_page_size)` -> `2 * 1 MB`, but this is not enough we need more than `2 * (system_page_size)`!
+# So from this example is clear that if you have a page size of 1 MB the first index that you can use is `3`.
+#
+# Please note: this is a very extreme case just to let you understand the mechanism, usually the page size is something
+# like 4 KB so you have no problem at all and you can use all the indexes (from `1` to `10`).
+#
+# To check your system page size use the Falco `--page-size` command line option. The output on a system with a page
+# size of 4096 Bytes (4 KB) should be the following:
+#
+# "Your system page size is: 4096 bytes."
+#
+# --- [Suggestions]
+#
+# Before the introduction of this param the buffer size was fixed to 8 MB (so index `4`, as you can see 
+# in the default value below).
+# Unless you are sure about what you are doing please keep this value as it is, Falco should work as
+# well as it always has with this value!
+# You can try to increase the buffer size when you face a lot of syscalls drops, but remember that this has
+# a price, larger buffers could slow down the entire machine. Moreover, consider that the buffer size is mapped
+# twice in the process' virtual memory so a buffer of 8 MB will result in a 16 MB area in the process virtual memory.
+# On the other side you can try to reduce the buffer size to speed up the system, but this could
+# increase the number of syscall drops!
+# So just to conclude, change this index only if you have the necessity otherwise leave it as it is!
+
+syscall_buffer_index: 4
+
 # Falco continuously monitors outputs performance. When an output channel does not allow
 # to deliver an alert within a given deadline, an error is reported indicating
 # which output is blocking notifications.
--- a/falco.yaml
+++ b/falco.yaml
@@ -211,18 +211,18 @@ syscall_event_timeouts:
 #
 # --- [Suggestions]
 #
-# Before the introduction of this param the buffer size was fixed to 8 MB (so index `4`, as you can see 
+# Before the introduction of this param the buffer size was fixed to 8 MB (so index `4`, as you can see
 # in the default value below).
-# Unless you are sure about what you are doing please keep this value as it is, Falco should work as
-# well as it always has with this value!
-# You can try to increase the buffer size when you face a lot of syscalls drops, but remember that this has
-# a price, larger buffers could slow down the entire machine. Moreover, consider that the buffer size is mapped
-# twice in the process' virtual memory so a buffer of 8 MB will result in a 16 MB area in the process virtual memory.
+# You can increase the buffer size when you face syscall drops. A size of 16 MB (so index `5`) can reduce
+# syscall drops in production-heavy systems without noticeable impact. Very large buffers however could
+# slow down the entire machine.
 # On the other side you can try to reduce the buffer size to speed up the system, but this could
 # increase the number of syscall drops!
-# So just to conclude, change this index only if you have the necessity otherwise leave it as it is!
+# As a final remark consider that the buffer size is mapped twice in the process' virtual memory so a buffer of 8 MB
+# will result in a 16 MB area in the process virtual memory.
+# Please pay attention when you use this parameter and change it only if the default size doesn't fit your use case.
 
-syscall_buffer_index: 4
+syscall_buf_size_preset: 4
 
 # Falco continuously monitors outputs performance. When an output channel does not allow
 # to deliver an alert within a given deadline, an error is reported indicating
--- a/userspace/falco/app_actions/compute_syscall_buffer_size.cpp
+++ b/userspace/falco/app_actions/compute_syscall_buffer_size.cpp
@@ -1,0 +1,68 @@
+/*
+Copyright (C) 2022 The Falco Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "application.h"
+
+using namespace falco::app;
+
+/* These indexes could change over the Falco releases. */
+#define MIN_INDEX 1
+#define MAX_INDEX 10
+
+application::run_result application::configure_syscall_buffer_size()
+{
+	/* We don't need to compute the syscall buffer dimension if we are in capture mode or if the
+	 * the syscall source is not enabled.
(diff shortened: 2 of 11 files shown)


Andreagit97:
@incertum @leogr @FedeDP @jasondellaluce

userspace/falco/app_cmdline_options.cpp
```diff
@@ -203,7 +203,9 @@ void cmdline_options::define()
 		("u,userspace",                   "Parse events from userspace. To be used in conjunction with the ptrace(2) based driver (pdig)", cxxopts::value(userspace)->default_value("false"))
 		("V,validate",                    "Read the contents of the specified rules(s) file and exit. Can be specified multiple times to validate multiple files.", cxxopts::value(validate_rules_filenames), "<rules_file>")
 		("v",                             "Verbose output.", cxxopts::value(verbose)->default_value("false"))
-		("version",                       "Print version number.", cxxopts::value(print_version_info)->default_value("false"));
+		("version",                       "Print version number.", cxxopts::value(print_version_info)->default_value("false"))
+		("page-size",                     "Print the system page size used to choose the syscall buffer size.", cxxopts::value(print_page_size)->default_value("false"));
```
leogr, line 207:
```suggestion
		("page-size",                     "Print the system page size (may assist choose the syscall buffer size).", cxxopts::value(print_page_size)->default_value("false"));
```

userspace/falco/app_actions/compute_syscall_buffer_size.cpp
```diff
@@ -0,0 +1,68 @@
...
+	/* Sizes from `1 MB` to `512 MB`. The index `0` is reserved, users cannot use it! */
+	std::vector<uint32_t> vect{0, 1 << 20, 1 << 21, 1 << 22, 1 << 23, 1 << 24, 1 << 25, 1 << 26, 1 << 27, 1 << 28, 1 << 29};
+
+	uint64_t chosen_size = vect[index];
+
+	/* If the page size is not valid we return here. */
+	long page_size = getpagesize();
+	if(page_size <= 0)
+	{
+		return run_result::fatal("\nUnable to get the system page size through 'getpagesize()'\n");
```
FedeDP, line 50: So, we would break Falco on system where we are not able to fetch page size (while it worked before, right?). Ok this is an extreme case, still i'd prefer to just leave the default value and stop try to change the buffer size, instead of halting Falco.

FedeDP (commented):
Left a comment; otherwise this LGTM!

falco.yaml
```diff
@@ -169,6 +169,61 @@ syscall_event_drops:
...
+# Unless you are sure about what you are doing please keep this value as it is, Falco should work as
+# well as it always has with this value!
+# You can try to increase the buffer size when you face a lot of syscalls drops, but remember that this has
+# a price, larger buffers could slow down the entire machine. Moreover, consider that the buffer size is mapped
+# twice in the process' virtual memory so a buffer of 8 MB will result in a 16 MB area in the process virtual memory.
+# On the other side you can try to reduce the buffer size to speed up the system, but this could
+# increase the number of syscall drops!
+# So just to conclude, change this index only if you have the necessity otherwise leave it as it is!
+
+syscall_buffer_index: 4
```
leogr, line 225:
```suggestion
syscall_buffer_index: 4
```

The term `index` may be misleading, IMO. What about something like `syscall_buffer_preset` or the more verbose `syscall_buf_size_preset`? 

leogr (commented):
This looks great.

I've just left some suggestions (nitpicking :smile_cat:)

falco.yaml
```diff
@@ -169,6 +169,61 @@ syscall_event_drops:
...
+# To check your system page size use the Falco `--page-size` command line option. The output on a system with a page
+# size of 4096 Bytes (4 KB) should be the following:
+#
+# "Your system page size is: 4096 bytes."
+#
+# --- [Suggestions]
+#
+# Before the introduction of this param the buffer size was fixed to 8 MB (so index `4`, as you can see 
+# in the default value below).
+# Unless you are sure about what you are doing please keep this value as it is, Falco should work as
```
incertum, line 216: Perhaps adjust this section? 

```
You can increase the buffer size when you face syscall drops. The preset one size larger than the default (16MB) can reduce syscall drops in production heavy systems without noticeable impact. Very large buffers however could slow down the entire machine.
```

Btw hard-coded 16MB and didn't see any impact in terms of CPU usage or slowdown, so that would probably only happen if you go extreme ... would mention that 16MB can solve the issues some end users face and make that option sound reasonable. What do you think?

In addition, @leogr worth mentioning that `k8s` plugin can impact the machine and as a consequence also cause higher syscall drops?

falco.yaml
incertum: Very good call @leogr!  `syscall_buf_size_preset` sounds great.

userspace/falco/app_actions/compute_syscall_buffer_size.cpp
Andreagit97: uhm yeah let's say this seems a really strange case, but yeah we can start with 8 MB in this case :)

falco.yaml
Andreagit97: > Btw hard-coded 16MB and didn't see any impact in terms of CPU usage or slowdown, so that would probably only happen if you go extreme ... would mention that 16MB can solve the issues some end users face and make that option sound reasonable. What do you think?

Yeah make sense

> In addition, @leogr worth mentioning that k8s plugin can impact the machine and as a consequence also cause higher syscall drops?

make sense but probably not here :thinking:

falco.yaml
Andreagit97: changed :)

falco.yaml
Andreagit97: changed :)

FedeDP (dismissed):
/approve

userspace/falco/app_actions/compute_syscall_buffer_size.cpp
```diff
@@ -0,0 +1,70 @@
...
+	if(page_size <= 0)
+	{
+		falco_logger::log(LOG_WARNING, "Unable to get the system page size through 'getpagesize()'. Try to use the default syscall buffer dimension: " + std::to_string(DEFAULT_BYTE_SIZE) + " bytes.\n");
+		return run_result::ok();
+	}
+
+	/* Check if the chosen size is a multiple of the page size. */
+	if(chosen_size % page_size != 0)
+	{
+		return run_result::fatal("The chosen size '" + std::to_string(chosen_size) + "' is not a multiple of your system page '" + std::to_string(page_size) + "'. Please choose a greater index.\n");
```
jasondellaluce, line 58:
```suggestion
		return run_result::fatal("The chosen syscall buffer size '" + std::to_string(chosen_size) + "' is not a multiple of your system page size '" + std::to_string(page_size) + "'. Please configure a greater syscall_buf_size_preset value.\n");
```
This error may be too generic for the average user. We need to be specific about _which_ size is wrong, and what specific action item can fix the issue.

This applies to the check below too.

userspace/falco/configuration.h
```diff
@@ -269,6 +269,9 @@ class falco_configuration
 	uint32_t m_metadata_download_chunk_wait_us;
 	uint32_t m_metadata_download_watch_freq_sec;
 
+	// Index corresponding to the syscall buffer dimension.
+	uint64_t m_syscall_buf_size_preset;
```
jasondellaluce, line 273: nit: uint64_t seems a bit excessive here

userspace/falco/app_actions/compute_syscall_buffer_size.cpp
Andreagit97: Solved, ty :)

userspace/falco/configuration.h
Andreagit97: changed into `uint16_t` ty :)

Andreagit97:
Rebased on actual master just to have this fix https://github.com/falcosecurity/falco/pull/2215 
I've also fixed the case in which we are not able to get the page system size in the last commit

FedeDP (approved):
/approve

PR #2219, opened 2022-09-22
fix(scripts): force falco-driver-loader script to try to compile the driver anyway even on unsupported platforms

FedeDP:
**What type of PR is this?**

/kind cleanup

Currently, when falco-driver-loader is not able to fetch OS_NAME from multiple files under /etc, like `/etc/os-release`, it exit with an error.
I think that the best approach is trying to compile the requested driver in any case; consider that `TARGET_ID` is only really used to concatenate the correct name for the driver on the s3 bucket.
So, in case we are not able to fetch any TARGET_ID, just fallback at disabling `ENABLE_DOWNLOAD` and go on to compile the driver.
This is mostly useful when using `HOST_ROOT` but the `/etc/` folder is not a shared volume under it.

Fixes #

```release-note
NONE
```

### DIFF
--- a/scripts/falco-driver-loader
+++ b/scripts/falco-driver-loader
@@ -710,8 +710,13 @@ if [ -z "$source_only" ]; then
 	get_target_id
 	res=$?
 	if [ $res != 0 ]; then
-		>&2 echo "Detected an unsupported target system, please get in touch with the Falco community. Trying to compile anyway."
-		ENABLE_DOWNLOAD=
+		if [ -n "$ENABLE_COMPILE" ]; then
+			ENABLE_DOWNLOAD=
+			>&2 echo "Detected an unsupported target system, please get in touch with the Falco community. Trying to compile anyway."
+		else
+			>&2 echo "Detected an unsupported target system, please get in touch with the Falco community."
+			exit 1
+		fi
 	fi
 
 	if [ -n "$clean" ]; then
--- a/scripts/falco-driver-loader
+++ b/scripts/falco-driver-loader
@@ -114,8 +114,7 @@ get_target_id() {
 		# Older CentOS distros
 		OS_ID=centos
 	else
-		>&2 echo "Detected an unsupported target system, please get in touch with the Falco community"
-		exit 1
+		return 1
 	fi
 
 	# Overwrite the OS_ID if /etc/VERSION file is present.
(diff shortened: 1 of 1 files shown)


jasondellaluce (approved):
LGTM!

scripts/falco-driver-loader
```diff
@@ -711,6 +707,13 @@ if [ -z "$source_only" ]; then
 		exit 1
 	fi
 
+	get_target_id
+	res=$?
+	if [ $res != 0 ]; then
+		>&2 echo "Detected an unsupported target system, please get in touch with the Falco community. Trying to compile anyway."
+		ENABLE_DOWNLOAD=
```
leogr, line 714:
```suggestion
		ENABLE_DOWNLOAD="no"
```

scripts/falco-driver-loader
```diff
@@ -711,6 +707,13 @@ if [ -z "$source_only" ]; then
 		exit 1
 	fi
 
+	get_target_id
+	res=$?
+	if [ $res != 0 ]; then
+		>&2 echo "Detected an unsupported target system, please get in touch with the Falco community. Trying to compile anyway."
```
leogr, line 713: `Trying to compile anyway` is true only if `ENABLE_COMPILE` is yes.

Perhaps it is worth printing a different message if both compile&download are disabled, and give up immediately.

leogr (approved):
Just found two nits, otherwise LGTM.

Putting
/hold
in case you want to apply my suggestions

scripts/falco-driver-loader
FedeDP: Ehy Leo! It would be great if this worked :D Unfortunately, in the script we check if it is not emtpy:
```
if [ -n "$ENABLE_DOWNLOAD" ]; then
```
I wanted to make the patch with the smallest diff possible, therefore i avoided to touch it.

scripts/falco-driver-loader
FedeDP: That could be a great idea; but i didn't want to embed more logic inside the script :) 
If this works for you, i'd leave it as is; otherwise i can make the small patch.

All in all, i am hoping for `falco-driver-loader` to be kicked out by `falcoctl` sooner or later!

leogr (approved):
LGTM :+1:

jasondellaluce (approved):
/approve

PR #2220, opened 2022-09-22
cleanup(docker): avoid linking /lib/modules to /host/lib/modules at docker image creation time

FedeDP:
**What type of PR is this?**

/kind cleanup

Avoid linking /lib/modules to /host/lib/modules at docker image creation time.
Instead, do it in docker-entrypoint scripts, so that even users using different HOST_ROOT than "/host", will still have a working image.

Moreover, if HOST_ROOT is set, but "$HOST_ROOT/proc" is not present, soft link "/proc" to "HOST_ROOT/proc", to allow Falco to run. Otherwise, scap_procs would exit with error and kill the istance.

Fixes #

```release-note
cleanup: allow users that use a different HOST_ROOT to still have a working falco-driver-loader driver compilation.
```

### DIFF
--- a/docker/driver-loader/docker-entrypoint.sh
+++ b/docker/driver-loader/docker-entrypoint.sh
@@ -25,4 +25,19 @@ do
     ln -s "$i" "/usr/src/$base"
 done
 
+if [ -n "$HOST_ROOT" ]; then
+    echo "* Setting up /lib/modules links from host"
+    ln -s /lib/modules $HOST_ROOT/lib/modules 
+    
+    # If HOST_ROOT is set, but HOST_ROOT/proc does not exist
+    # link real /proc to HOST_ROOT/proc, so that Falco can run gracefully.
+    # This is mostly useful when dealing with an hypervisor, like aws Fargate,
+    # where the container running Falco does not need to bind-mount the host proc volume,
+    # and its /proc already sees all task processes because it shares the same namespace.
+    if [ ! -d "$HOST_ROOT/proc" ]; then
+        echo "* Setting up /proc links from host"
+        ln -s "/proc" "$HOST_ROOT/proc"
+    fi
+fi
+
 /usr/bin/falco-driver-loader "$@"
--- a/docker/driver-loader/docker-entrypoint.sh
+++ b/docker/driver-loader/docker-entrypoint.sh
@@ -25,7 +25,7 @@ do
     ln -s "$i" "/usr/src/$base"
 done
 
-if [ -n "$HOST_ROOT" ]; then
+if [ -n "$HOST_ROOT" ] && [ "$HOST_ROOT" != "/" ]; then
     echo "* Setting up /lib/modules links from host"
     ln -s /lib/modules $HOST_ROOT/lib/modules 
     
--- a/docker/falco/Dockerfile
+++ b/docker/falco/Dockerfile
@@ -103,8 +103,7 @@ RUN sed -e 's/time_format_iso_8601: false/time_format_iso_8601: true/' < /etc/fa
 # Some base images have an empty /lib/modules by default
 # If it's not empty, docker build will fail instead of
 # silently overwriting the existing directory
-RUN rm -df /lib/modules \
-	&& ln -s $HOST_ROOT/lib/modules /lib/modules
+RUN rm -df /lib/modules
 
 # debian:stable head contains binutils 2.31, which generates
 # binaries that are incompatible with kernels < 4.16. So manually
--- a/docker/falco/docker-entrypoint.sh
+++ b/docker/falco/docker-entrypoint.sh
@@ -30,4 +30,19 @@ if [[ -z "${SKIP_DRIVER_LOADER}" ]]; then
     /usr/bin/falco-driver-loader
 fi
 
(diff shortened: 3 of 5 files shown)


docker/driver-loader/docker-entrypoint.sh
```diff
@@ -25,4 +25,19 @@ do
     ln -s "$i" "/usr/src/$base"
 done
 
-/usr/bin/falco-driver-loader "$@"
\ No newline at end of file
+if [ -n "$HOST_ROOT" ]; then
+    echo "* Setting up /lib/modules links from host"
+    ln -s /lib/modules $HOST_ROOT/lib/modules
```
FedeDP, line 30: drivers makefile uses `/lib/modules`; in case `HOST_ROOT` is set, we link /lib/modules to its relative path under HOST_ROOT.
This was done before at docker image build time.

leogr: I don't understand why we need this. 

Moreover, `$HOST_ROOT` can be equal to `/` which would produce:
```
ln -s /lib/modules //lib/modules
```


Here's a legit use case for `/`:
```
docker pull falcosecurity/falco-no-driver:latest
docker run --rm -i -t \
    -e HOST_ROOT=/ \
    --cap-add SYS_PTRACE --pid=host $(ls /dev/falco* | xargs -I {} echo --device {}) \
    -v /var/run/docker.sock:/var/run/docker.sock \
    falcosecurity/falco-no-driver:latest
```


Ref https://falco.org/docs/getting-started/running/#docker-least-privileged

FedeDP: Yep, the issue here is when you have an HOST_ROOT that is not "/host". Before, you'd have "/host/lib/modules" linked to "/lib/modules" from Dockerfile, even when, at runtime, you'd got HOST_ROOT="/foo".
So, you'd have a Falco running with HOST_ROOT="/foo", but /lib/modules/ symlinked to "/host/lib/modules".
Of course, this ain't gonna work :)

docker/driver-loader/docker-entrypoint.sh
```diff
@@ -25,4 +25,19 @@ do
...
+    
+    # If HOST_ROOT is set, but HOST_ROOT/proc does not exist
+    # link real /proc to HOST_ROOT/proc, so that Falco can run gracefully.
+    # This is mostly useful when dealing with an hypervisor, like aws Fargate,
+    # where the container running Falco does not need to bind-mount the host proc volume,
+    # and its /proc already sees all task processes because it shares the same namespace.
+    if [ ! -d "$HOST_ROOT/proc" ]; then
+        echo "* Setting up /proc links from host"
+        ln -s "/proc" "$HOST_ROOT/proc"
+    fi
```
leogr, line 40: It seems legit. Anyway, just setting `HOST_ROOT=/` (or `HOST_ROOT=`) should work without handling it here. Is there any compelling use case in which `HOST_ROOT=/` is insufficient? If yes, can we document it here in the comment or in the official docs?

docker/driver-loader/docker-entrypoint.sh
FedeDP: You are right, i think we should update the 
`if [ -n "$HOST_ROOT" ]; then`
with
`if [[ -n "$HOST_ROOT" -a "$HOST_ROOT" != "/" ]]; then`

FedeDP: Fixed!

leogr (commented):
I retook a look at this, and I'm still not totally convinced (not saying the PR is not ok).

Questions :thinking: 
- Since `/lib/modules` is only required to build the driver, why not put this logic inside `falco-driver-loader`? (or am I missing something?)
- Although the `/proc` symlink makes sense in the case you mentioned, how can we generally ensure it matches the same namespace? (ie. mounting the wrong `/proc` might be worse than not mounting it at all).

FedeDP:
> Since /lib/modules is only required to build the driver, why not put this logic inside falco-driver-loader? (or am I missing something?)

We can probably do this too; but since we already do some other stuff in the entrypoint, i assumed that was the best place  actually

> Although the /proc symlink makes sense in the case you mentioned, how can we generally ensure it matches the same namespace? (ie. mounting the wrong /proc might be worse than not mounting it at all).

That's a good point actually; i am not sure what we can do (aside from failing badly :D )

leogr:
> > Since /lib/modules is only required to build the driver, why not put this logic inside falco-driver-loader? (or am I missing something?)
> 
> We can probably do this too; but since we already do some other stuff in the entrypoint, i assumed that was the best place actually

I think there's no best practice here :angel: and I'm not yet sure putting it into `falco-driver-loader` is the best thing to do. 
I would like to manage this more consistently. @falcosecurity/falco-maintainers WDYT?

> > Although the /proc symlink makes sense in the case you mentioned, how can we generally ensure it matches the same namespace? (ie. mounting the wrong /proc might be worse than not mounting it at all).
> 
> That's a good point actually; i am not sure what we can do (aside from failing badly :D )

FedeDP:
I'll move this to 0.36 since we are still looking for a general consensus :D 
/milestone 0.36.0

PR #2226, opened 2022-09-28
Rule: PTRACE attached to process

Brucedh:
Signed-off-by: Alessandro Brucato <alessandro.brucato@sysdig.com>

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind release

> /kind rule-update

/kind rule-create

> /area build

> /area engine

/area rules

> /area tests

> /area proposals

> /area CI

The ptrace system call can be used to alter a target's memory and execution which make it popular for injecting code. It is often used by malware for this purpose in order to be more stealthy, or steal information from the process. Legitimate tools, such as GDB or security software, may also leverage this method.

Fixes #

```release-note
New rule to detect attempts to inject code into a process using PTRACE
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -3238,3 +3238,10 @@
     command=%proc.cmdline pid=%proc.pid file=%fd.name parent=%proc.pname gparent=%proc.aname[2] ggparent=%proc.aname[3] gggparent=%proc.aname[4] container_id=%container.id image=%container.image.repository)
   priority: WARNING
   tags: [filesystem, mitre_credential_access, mitre_discovery]
+
+- rule: PTRACE attached to process
+  desc: "This rule detects an attempt to inject code into a process using PTRACE."
+  condition: evt.type=ptrace and evt.dir=> and evt.arg.request=11 and proc_name_exists
+  output: Detected ptrace PTRACE_ATTACH attempt (proc.cmdline=%proc.cmdline container=%container.info evt.type=%evt.type evt.arg.request=%evt.arg.request proc.pid=%proc.pid proc.cwd=%proc.cwd proc.ppid=%proc.ppid proc.pcmdline=%proc.pcmdline proc.sid=%proc.sid proc.exepath=%proc.exepath user.uid=%user.uid user.loginuid=%user.loginuid user.loginname=%user.loginname user.name=%user.name group.gid=%group.gid group.name=%group.name container.id=%container.id container.name=%container.name image=%container.image.repository)
+  priority: WARNING
+  tags: [process]
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -3239,9 +3239,15 @@
   priority: WARNING
   tags: [filesystem, mitre_credential_access, mitre_discovery]
 
+- list: known_ptrace_binaries
+  items: []
+
+- marco: known_ptrace_procs
+  condition: (proc.name in (known_ptrace_binaries))
+
 - rule: PTRACE attached to process
   desc: "This rule detects an attempt to inject code into a process using PTRACE."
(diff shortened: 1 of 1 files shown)


rules/falco_rules.yaml
```diff
@@ -3238,3 +3238,10 @@
     command=%proc.cmdline pid=%proc.pid file=%fd.name parent=%proc.pname gparent=%proc.aname[2] ggparent=%proc.aname[3] gggparent=%proc.aname[4] container_id=%container.id image=%container.image.repository)
   priority: WARNING
   tags: [filesystem, mitre_credential_access, mitre_discovery]
+
+- rule: PTRACE attached to process
+  desc: "This rule detects an attempt to inject code into a process using PTRACE."
+  condition: evt.type=ptrace and evt.dir=> and evt.arg.request=11 and proc_name_exists
```
Kaizhe, line 3244: This is a really initiative!! Thanks!
 
Here are my two cents/questions:
1. Should we include `PTRACE_SEIZE` as well?
2. Why do we care `proc_name_exists`?
3. Security tooling might use ptrace capability (e.g. Falco in certain scenario), maybe we should add macro to exclude the known processes?

loresuso, line 3244: Nice idea @Brucedh! 

Just wanted to add something to this. Code injection is usually performed by issuing the `ptrace` syscall and using the following three requests in my opinion: `PTRACE_ATTACH`, `PTRACE_POKETEXT` and `PTRACE_SETREGS`. The poke text requests usually copy a shell code to the memory of the victim process and the set regs request usually modifies the instruction pointer to execute from that shell code. As you were saying, the attach requests could possibly lead to false positives, so I think we can use the two other requests to improve this rule. 

WDYT?

nicklangsysdig: Hey, thanks for the feedback. Do you know why Falco thinks that "POKETEXT" is evt.arg.request=5 and not 4, same with SETREGS being 27 from Falco's point of view but 13 from a C programmer's perspective? https://elixir.bootlin.com/linux/v4.8/source/arch/arm/include/uapi/asm/ptrace.h#L16

loresuso: Hello @nicklangsysdig, the reason is that Falco maps those flags to its internal representations. This is done for portability so that regardless of the system Falco is running, the flags will all have a unique representation. This is useful if flags can change from one OS to another. If you want to dig deeper and know about how Falco redefines these flags, please check it here: https://github.com/falcosecurity/libs/blob/master/driver/ppm_events_public.h#L416-L454
keep in mind that the same reasoning applies to all flags, so this is an advice you can re-use in the future 🙂

Brucedh: Hey @Kaizhe , thank you for your feedback!
1. Sure, PTRACE_SEIZE makes sense to me, I added it in the condition
2. We saw some false positives without that macro
3. Sure, I added the macro (as an empty list)

Brucedh: Thank you for your feedback @loresuso !
I agree it’s a good idea to cover also PTRACE_POKETEXT and PTRACE_SETREGS. I added them along with PTRACE_POKEDATA, which is equivalent to PTRACE_POKETEXT

nicklangsysdig:
@Kaizhe @loresuso Looking good so far?

rules/falco_rules.yaml
loresuso: It makes sense, thank you!

darryk10:
LGTM

darryk10 (approved):

jasondellaluce (approved):
/approve

leogr (approved):

PR #2227, opened 2022-09-28
Rule: PTRACE anti-debug attempt

Brucedh:
Signed-off-by: Alessandro Brucato <alessandro.brucato@sysdig.com>

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind release

> /kind rule-update

/kind rule-create

> /area build

> /area engine

/area rules

> /area tests

> /area proposals

> /area CI

The ptrace system call is used by debuggers, such as GDB, to examine a target process. Malicious programs will often try to prevent this by using the ptrace system call first, preventing other programs from using it on them. This rule detects that behavior. Legitimate processes may also do this for DRM or diagnostic purposes. 

Fixes #

```release-note
New rule to detect usages of the PTRACE system call with the PTRACE_TRACEME argument, indicating a program actively attempting to avoid debuggers attaching to the process.
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -3253,3 +3253,10 @@
   output: Detected ptrace PTRACE_ATTACH attempt (proc.cmdline=%proc.cmdline container=%container.info evt.type=%evt.type evt.arg.request=%evt.arg.request proc.pid=%proc.pid proc.cwd=%proc.cwd proc.ppid=%proc.ppid proc.pcmdline=%proc.pcmdline proc.sid=%proc.sid proc.exepath=%proc.exepath user.uid=%user.uid user.loginuid=%user.loginuid user.loginname=%user.loginname user.name=%user.name group.gid=%group.gid group.name=%group.name container.id=%container.id container.name=%container.name image=%container.image.repository)
   priority: WARNING
   tags: [process]
+
+- rule: PTRACE anti-debug attempt
+  desc: "Detect usage of the PTRACE system call with the PTRACE_TRACEME argument, indicating a program actively attempting to avoid debuggers attaching to the process. This behavior is typically indicative of malware activity."
+  condition: evt.type=ptrace and evt.dir=> and evt.arg.request=1 and proc_name_exists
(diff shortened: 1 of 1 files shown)


darryk10:
@loresuso @Kaizhe can you provide your feedback for this rule as well?
We have tested the rule for a while and it looks pretty accurate withouf FPs. If you have any ways we can improve the rule to cover other use cases would be great.
Thanks

leogr:
> @loresuso @Kaizhe can you provide your feedback for this rule as well? We have tested the rule for a while and it looks pretty accurate withouf FPs. If you have any ways we can improve the rule to cover other use cases would be great. Thanks

:+1: 

cc @loresuso @darryk10 @Kaizhe

loresuso:
My opinion on this is that this anti-debugging technique is the most simple one and most people know about it, and I think it is rarely used because it can be easily bypassed by simply patching calls to this syscall. 

I believe malware writers have many other options for anti-debugging, which are more meaningful to me, like timing analysis (if the process is being debugged, execution times will differ from normal execution flow) or just checking if parent process is a known debugger binary, which is also easy to implement like this one. Furthermore, as you pointed out, legitimate processes may use this. 

So what I think is that this rule triggering wouldn't be so much relevant for spotting malware, although it can be useful in some cases, I don't know if we want it to be in the default Falco ruleset.

jasondellaluce:
This needs to be ported to https://github.com/falcosecurity/rules

leogr:
> This needs to be ported to https://github.com/falcosecurity/rules

cc @darryk10 @loresuso

leogr (changes requested):
Please, this PR needs to be ported to https://github.com/falcosecurity/rules

darryk10:
Due to what happened here https://github.com/falcosecurity/rules/pull/37/files I would revaluate this rule once the PR is merged and tested. @Brucedh feel free to start closing this here and open a new one in the new repo falcosecurity/rules.
Thanks

Brucedh:
Alright, I close this PR after having ported it to https://github.com/falcosecurity/rules/pull/40

PR #2235, opened 2022-10-06
fix(userspace/falco): make signal termination more robust with multi-threading

jasondellaluce:
**What type of PR is this?**

/kind bug

/area engine

The termination signal (SIGINT) used to put Falco in deadlock if triggered while loading rules, plugins, or the config. Reproducing the deadlock was not deterministic. The solution is to attach the signal handlers as later as possible (after initialization time), but before spawning any additional thread (webserver, grpc, k8s client, per-source thread, etc...).

```release-note
NONE
```


FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #2238, opened 2022-10-10
fix(falco_service): falco service needs to write under `/sys/module/falco`

Andreagit97:
**What type of PR is this?**

/kind bug

/area build

Trying to run the latest Falco deb/rpm package there is an issue:

Since this PR https://github.com/falcosecurity/falco/pull/2214, Falco service needs to write under `/sys/module/falco/parameters/g_buffer_bytes_dim`  the variable buffer dimension. The problem is that right now the path `/sys/module` is mounted read-only so Falco will fail with `Errno 30` in the attempt to write `/sys/module/falco/parameters/g_buffer_bytes_dim`:

```
Oct 07 23:41:26 ubuntu2204.localdomain falco[34605]: Falco version: 0.32.1-241+251b4d5 (x86_64)
Oct 07 23:41:26 ubuntu2204.localdomain falco[34605]: Fri Oct  7 23:41:26 2022: Falco version: 0.32.1-241+251b4d5 (x86_64)
Oct 07 23:41:26 ubuntu2204.localdomain falco[34605]: Fri Oct  7 23:41:26 2022: Falco initialized with configuration file: /etc/falco/falco.yaml
Oct 07 23:41:26 ubuntu2204.localdomain falco[34605]: Falco initialized with configuration file: /etc/falco/falco.yaml
Oct 07 23:41:26 ubuntu2204.localdomain falco[34605]: Loading rules from file /etc/falco/falco_rules.yaml
Oct 07 23:41:26 ubuntu2204.localdomain falco[34605]: Fri Oct  7 23:41:26 2022: Loading rules from file /etc/falco/falco_rules.yaml
Oct 07 23:41:26 ubuntu2204.localdomain falco[34605]: Loading rules from file /etc/falco/falco_rules.local.yaml
Oct 07 23:41:26 ubuntu2204.localdomain falco[34605]: Fri Oct  7 23:41:26 2022: Loading rules from file /etc/falco/falco_rules.local.yaml
Oct 07 23:41:26 ubuntu2204.localdomain falco[34605]: The chosen syscall buffer dimension is: 8388608 bytes (8 MBs)
Oct 07 23:41:26 ubuntu2204.localdomain falco[34605]: Fri Oct  7 23:41:26 2022: The chosen syscall buffer dimension is: 8388608 bytes (8 MBs)
Oct 07 23:41:26 ubuntu2204.localdomain falco[34605]: Fri Oct  7 23:41:26 2022: Starting health webserver with threadiness 6, listening on port 8765
Oct 07 23:41:26 ubuntu2204.localdomain falco[34605]: Starting health webserver with threadiness 6, listening on port 8765
Oct 07 23:41:26 ubuntu2204.localdomain falco[34605]: Fri Oct  7 23:41:26 2022: Enabled event sources: syscall
Oct 07 23:41:26 ubuntu2204.localdomain falco[34605]: Fri Oct  7 23:41:26 2022: Opening capture with Kernel module
Oct 07 23:41:26 ubuntu2204.localdomain falco[34605]: Enabled event sources: syscall
Oct 07 23:41:26 ubuntu2204.localdomain falco[34605]: Opening capture with Kernel module
Oct 07 23:41:26 ubuntu2204.localdomain falco[34605]: Trying to inject the Kernel module and opening the capture again...
Oct 07 23:41:26 ubuntu2204.localdomain falco[34605]: Fri Oct  7 23:41:26 2022: Trying to inject the Kernel module and opening the capture again...
Oct 07 23:41:26 ubuntu2204.localdomain falco[34605]: Error: unable to open '/sys/module/falco/parameters/g_buffer_bytes_dim': Errno 30. Please ensure the kernel module is already loaded.
```

Considering the falco service config:

```
[Unit]
Description=Falco: Container Native Runtime Security
Documentation=https://falco.org/docs/

[Service]
Type=simple
User=root
ExecStartPre=/sbin/modprobe falco
ExecStart=/usr/bin/falco --pidfile=/var/run/falco.pid
ExecStopPost=/sbin/rmmod falco
UMask=0077
TimeoutSec=30
RestartSec=15s
Restart=on-failure
PrivateTmp=true
NoNewPrivileges=yes
ProtectHome=read-only
ProtectSystem=full
ProtectKernelTunables=true
RestrictRealtime=true
RestrictAddressFamilies=~AF_PACKET

[Install]
WantedBy=multi-user.target

```

there are some solutions to this problem:
1. The most aggressive one is to put `ProtectKernelTunables` to `false`
2. [PROPOSED ONE] The cleanest one should be to inject the kmod in a separate unit and call the falco unit only when the path `/sys/module/falco` is already there, in this way the falco unit could set `ReadWritePaths` to only `/sys/module/falco`
3. The third solution sets `ReadWritePaths` to  `/sys/module` because the falco subfolder is not yet created at startup-time and we cannot `mkdir` this directory into another unit because we cannot create folders under `sys/module` only the kernel can do that. 

```release-note
NONE
```

### DIFF
--- a/cmake/cpack/CMakeCPackOptions.cmake
+++ b/cmake/cpack/CMakeCPackOptions.cmake
@@ -1,11 +1,13 @@
 if(CPACK_GENERATOR MATCHES "DEB")
 	list(APPEND CPACK_INSTALL_COMMANDS "mkdir -p _CPack_Packages/${CPACK_TOPLEVEL_TAG}/${CPACK_GENERATOR}/${CPACK_PACKAGE_FILE_NAME}/usr/lib/systemd/system")
 	list(APPEND CPACK_INSTALL_COMMANDS "cp scripts/debian/falco.service _CPack_Packages/${CPACK_TOPLEVEL_TAG}/${CPACK_GENERATOR}/${CPACK_PACKAGE_FILE_NAME}/usr/lib/systemd/system")
+	list(APPEND CPACK_INSTALL_COMMANDS "cp scripts/debian/falco_inject_kmod.service _CPack_Packages/${CPACK_TOPLEVEL_TAG}/${CPACK_GENERATOR}/${CPACK_PACKAGE_FILE_NAME}/usr/lib/systemd/system")
 endif()
 
 if(CPACK_GENERATOR MATCHES "RPM")
 	list(APPEND CPACK_INSTALL_COMMANDS "mkdir -p _CPack_Packages/${CPACK_TOPLEVEL_TAG}/${CPACK_GENERATOR}/${CPACK_PACKAGE_FILE_NAME}/usr/lib/systemd/system")
 	list(APPEND CPACK_INSTALL_COMMANDS "cp scripts/rpm/falco.service _CPack_Packages/${CPACK_TOPLEVEL_TAG}/${CPACK_GENERATOR}/${CPACK_PACKAGE_FILE_NAME}/usr/lib/systemd/system")
+	list(APPEND CPACK_INSTALL_COMMANDS "cp scripts/rpm/falco_inject_kmod.service _CPack_Packages/${CPACK_TOPLEVEL_TAG}/${CPACK_GENERATOR}/${CPACK_PACKAGE_FILE_NAME}/usr/lib/systemd/system")
 endif()
 
 if(CPACK_GENERATOR MATCHES "TGZ")
--- a/scripts/CMakeLists.txt
+++ b/scripts/CMakeLists.txt
@@ -22,11 +22,17 @@ configure_file(debian/prerm.in debian/prerm)
 file(COPY "${PROJECT_SOURCE_DIR}/scripts/debian/falco.service"
 	DESTINATION "${PROJECT_BINARY_DIR}/scripts/debian")
 
+file(COPY "${PROJECT_SOURCE_DIR}/scripts/debian/falco_inject_kmod.service"
+	DESTINATION "${PROJECT_BINARY_DIR}/scripts/debian")
+
 configure_file(rpm/postinstall.in rpm/postinstall)
 configure_file(rpm/postuninstall.in rpm/postuninstall)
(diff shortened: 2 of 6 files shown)


Andreagit97:
PS: I forgot to mention that `systemd` fail at all when something in a `ReadWritePath` directive doesn't exist :point_down: 
https://github.com/systemd/systemd/issues/10972#issuecomment-576016402

Andreagit97:
The first commit propose the `3` solution
> The third solution sets ReadWritePaths to /sys/module because the falco subfolder is not yet created at startup-time and we cannot mkdir this directory into another unit because we cannot create folders under sys/module only the kernel can do that.

The last commit propose the `2` solution that should be better :thinking: 
> [PROPOSED ONE] The cleanest one should be to inject the kmod in a separate unit and call the falco unit only when the path /sys/module/falco is already there, in this way the falco unit could set ReadWritePaths to only /sys/module/falco

Happy-Dude:
Does `PermissionsStartOnly=true` allow writing into `/sys/module` or is still mounted as read only?

While I don't love having multiple service units, I do appreciate the clarity that comes from splitting the `modprobe` tasks.
(I personally modify the service unit to remove the `modprobe` commands because I want to start the service in BPF mode.)

FedeDP:
Hi @Happy-Dude !
I think `PermissionsStartOnly=true` is only useful for Pre/Post Exec, right?
In this case, it won't help because the write to `/sys/module` is being done by Falco itself (https://github.com/falcosecurity/libs/blob/master/userspace/libscap/engine/kmod/scap_kmod.c#L79).

Btw i am working on an improvements over the current falco systemd unit, to support bpf too. I will hopefully remember to tag you when the PR it's open :D

FedeDP:
The changes LGTM. I'll wait some more feedback!

FedeDP (approved):
/approve

jasondellaluce (approved):
/approve

PR #2239, opened 2022-10-10
fix(userspace/falco): make termination and signal handlers more stable

jasondellaluce:
**What type of PR is this?**

/kind bug

/kind cleanup

/area engine

Signal handlers are used in Falco to force termination, hot reload, or output reloading. However, the current implementation is not stable for neither thread-safety and async-signal safety (more here: https://docs.oracle.com/cd/E19455-01/806-5257/gen-26/index.html). This PR refactors the way signals are handled to make sure it's safe both on single-threaded (e.g. capturing only syscalls) and multi-threaded use cases (e.g. multiple event sources in parallel).

Accordingly, the event source termination stage has been refactored to be more stable by using atomic values to check 1) that the event source has actually finished processing events, and 2) that the corresponding thread has already been joined. This is safer than simply relying on the `thread.joinable()` and `thread.join()` primitives.

```release-note
NONE
```

### DIFF
--- a/userspace/falco/app_actions/create_signal_handlers.cpp
+++ b/userspace/falco/app_actions/create_signal_handlers.cpp
@@ -30,26 +30,24 @@ using namespace falco::app;
 // provided application, and in unregister_signal_handlers it will be
 // rebound back to the dummy application.
 
-static application dummy;
-static std::reference_wrapper<application> s_app = dummy;
 static int inot_fd;
 
-static void signal_callback(int signal)
+static void terminate_signal_handler(int signal)
 {
-	falco_logger::log(LOG_INFO, "SIGINT received, exiting...\n");
-	s_app.get().terminate();
+	ASSERT(falco::app::g_terminate.is_lock_free());
+	falco::app::g_terminate.store(APP_SIGNAL_SET, std::memory_order_seq_cst);
 }
 
-static void reopen_outputs(int signal)
+static void reopen_outputs_signal_handler(int signal)
 {
-	falco_logger::log(LOG_INFO, "SIGUSR1 received, reopening outputs...\n");
-	s_app.get().reopen_outputs();
+	ASSERT(falco::app::g_reopen_outputs.is_lock_free());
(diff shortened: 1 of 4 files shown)


userspace/falco/app_actions/create_signal_handlers.cpp
```diff
@@ -74,21 +72,32 @@ bool application::create_handler(int sig, void (*func)(int), run_result &ret)
...
 	run_result ret;
-	s_app = *this;
-	if(! create_handler(SIGINT, ::signal_callback, ret) ||
-	   ! create_handler(SIGTERM, ::signal_callback, ret) ||
-	   ! create_handler(SIGUSR1, ::reopen_outputs, ret) ||
-	   ! create_handler(SIGHUP, ::restart_falco, ret))
+	if(! create_handler(SIGINT, ::terminate_signal_handler, ret) ||
+	   ! create_handler(SIGTERM, ::terminate_signal_handler, ret) ||
+	   ! create_handler(SIGUSR1, ::reopen_outputs_signal_handler, ret) ||
+	   ! create_handler(SIGHUP, ::restart_signal_handler, ret))
```
Andreagit97, line 90: what about something like:
```suggestion
	return !create_handler(SIGINT, ::terminate_signal_handler, ret) ||
	   !create_handler(SIGTERM, ::terminate_signal_handler, ret) ||
	   ! create_handler(SIGUSR1, ::reopen_outputs_signal_handler, ret) ||
	   ! create_handler(SIGHUP, ::restart_signal_handler, ret)) ? run_result::fatal("") : run_result::ok();
```

Or if we don't want to return fatal just `run_result::ok()` in both cases :thinking: just a nitpick feel free to ignore

jasondellaluce: `create_handler` already sets its own error results in the passed-in `ret`, so errors are already catched.

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #2240, opened 2022-10-10
chore(userspace/falco): do not allow multiple Falco instances.

FedeDP:
**What type of PR is this?**

/kind cleanup

/area engine

Create a lock file to deny multiple Falco instances.

Fixes #

```release-note
new: allow a single Falco instance
```

### DIFF
--- a/userspace/falco/app_actions/gain_lock.cpp
+++ b/userspace/falco/app_actions/gain_lock.cpp
@@ -1,0 +1,45 @@
+/*
+Copyright (C) 2022 The Falco Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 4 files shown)


userspace/falco/app_actions/gain_lock.cpp
```diff
@@ -0,0 +1,45 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "application.h"
+#include <sys/file.h>
+
+using namespace falco::app;
+
+#define FALCO_LOCK_FILE "falco.lock"
```
FedeDP, line 22: Note: this is in the current working directory, otherwise build would fail after linking (or, we should always build as root).
It is not great though, because one can easily escape the check by changing cwd.
This is something that needs to be addressed.

FedeDP:
/close

Imho this does not make much sense ahha

PR #2241, opened 2022-10-10
feat: Support for detecting outbound connection to c2 servers with FQ…

Nicolas-Peiffer:
…DN domains and IP addresses.

**What type of PR is this?**

/kind rule-update

/kind rule-create

/area rules

This is a modification and enhancement of the [Falco rule](https://github.com/falcosecurity/falco/blob/79d875c28fdfb144ac8f417ee668fe2c4df175a0/rules/falco_rules.yaml#L3052-L3060) _outbound connection to c2 servers_. The rule was only considering IP addresses as a list of C2 servers. The rule was not considering Domain names (FQDN).

Now the rule both considers FQDN and IP address. This is espacially convenient for public list of C2 servers with both FQDN and IP addresses and that can be found on the internet, like this one:
https://feodotracker.abuse.ch/downloads/ipblocklist_recommended.json

The new  _outbound connection to c2 servers_ rules is inspired by other outbound rules like:
https://github.com/falcosecurity/falco/blob/master/rules/falco_rules.yaml#L392

No issue.

I am not sure that there is a user-facing change. Else I would say:
_Rule: Adding support for detecting outbound connection to c2 servers with both FQDN domains and IP addresses._

```release-note
Update the "Outbound connection to C2 server" rule to match both FQDN and IP adresses. Prior to this change, the rule only matched IP adresses and not FQDN.
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -3052,10 +3052,39 @@
 - list: c2_server_ip_list
   items: []
 
+- list: c2_server_fqdn_list
+  items: []
+
+# For example, you can fetch a list of IP addresses and FQDN on this website:
+# https://feodotracker.abuse.ch/downloads/ipblocklist_recommended.json.
+# Use Falco HELM chart to update (append) the c2 server lists with your values.
+# See an example below.
+#
+#  ```yaml
+#  # values.yaml Falco HELM chart file
+#  [...]
+#  customRules:
+#    c2-servers-list.yaml: |-
+#      - list: c2_server_ip_list
+#        append: true
+#        items: 
+#        - "'51.178.161.32'"
+#        - "'46.101.90.205'"
+#        
+#      - list: c2_server_fqdn_list
+#        append: true
+#        items: 
+#        - "srv-web.ffconsulting.com"
+#        - "57.ip-142-44-247.net"
+#  ```
+
 - rule: Outbound Connection to C2 Servers
-  desc: Detect outbound connection to command & control servers
-  condition: outbound and fd.sip in (c2_server_ip_list)
-  output: Outbound connection to C2 server (command=%proc.cmdline pid=%proc.pid connection=%fd.name user=%user.name user_loginuid=%user.loginuid container_id=%container.id image=%container.image.repository)
+  desc: Detect outbound connection to command & control servers thanks to a list of IP addresses & a list of FQDN.
+  condition: >
+    outbound and 
+    ((fd.sip in (c2_server_ip_list)) or
+     (fd.sip.name in (c2_server_fqdn_list)))
+  output: Outbound connection to C2 server (c2_domain=%fd.sip.name c2_addr=%fd.sip command=%proc.cmdline connection=%fd.name user=%user.name user_loginuid=%user.loginuid container_id=%container.id image=%container.image.repository)
   priority: WARNING
   tags: [network]
 


rules/falco_rules.yaml
```diff
@@ -3052,10 +3052,38 @@
...
+#        - "'46.101.90.205'"
+#        
+#      - list: c2_server_fqdn_list
+#        append: true
+#        items: 
+#        - "srv-web.ffconsulting.com"
+#        - "57.ip-142-44-247.net"
+#  ```
+
+- rule: Outbound Connection to C2 Servers IPs and FQDNs
```
darryk10, line 3077: I honestly don't think we need to change the rule name since it's accurate as it is. 
we can just add in the existent description that the rules match IPs and FQDN.
I would just update the rule condition as done to support fqdn as well

darryk10 (changes requested):
Hi,
thanks for contributing :) I agree with your idea and it can be helpful to use public lists.
I added some comments to simplify the change without rewriting the overall rule.

darryk10:
@Nicolas-Peiffer can you also pls add something in the PR section for the release note?
`Does this PR introduce a user-facing change?
`

rules/falco_rules.yaml
Nicolas-Peiffer: Hi, I agree  with your suggestions. The name of the rule can stay the same (`Outbound Connection to C2 Servers`), as well as the `output:` field.

If you prefer, I can revert to the previous rule name.

***

Question: I understand you prefer this

```yaml
  desc: Detect outbound connection to command & control servers thanks to a list of IP addresses & a list of FQDN.
```

over this?

```yaml
  desc: >
    Detect outbound connection to command & control servers. For example, fetch
    a list of IP addresses and FQDN on this website:
    https://feodotracker.abuse.ch/downloads/ipblocklist_recommended.json.
```

I have no opinion on this one. I understand it is maybe better not to include a link to `feodotracker.abuse.ch`.

darryk10: @Nicolas-Peiffer I'll go with this one
`  desc: Detect outbound connection to command & control servers thanks to a list of IP addresses & a list of FQDN.
`
I feel the desc you added is more a comment for the rule so since it is actually a useful comment I think you can add it on top fo the rule as a proper comment as done in other rules. What do you think?

Nicolas-Peiffer:
@darryk10 I updated the release note.

And I cleaned my commits and modifed the rule taking into account your suggestions.

rules/falco_rules.yaml
```diff
@@ -3052,10 +3052,39 @@
...
 - rule: Outbound Connection to C2 Servers
-  desc: Detect outbound connection to command & control servers
-  condition: outbound and fd.sip in (c2_server_ip_list)
-  output: Outbound connection to C2 server (command=%proc.cmdline pid=%proc.pid connection=%fd.name user=%user.name user_loginuid=%user.loginuid container_id=%container.id image=%container.image.repository)
+  desc: Detect outbound connection to command & control servers thanks to a list of IP addresses & a list of FQDN.
+  condition: >
+    outbound and 
+    ((fd.sip in (c2_server_ip_list)) or
+     (fd.sip.name in (c2_server_fqdn_list)))
+  output: Outbound connection to C2 server (command=%proc.cmdline connection=%fd.name user=%user.name user_loginuid=%user.loginuid container_id=%container.id image=%container.image.repository)
```
darryk10, line 3087: I think it would be also useful to print out `fd.sip` and `fd.sip.name` for troubleshooting and investigation.
WDYT?

Nicolas-Peiffer: I did not think of it but you are right about that enhancement of the `output:`.
However, I am not sure what is the behaviour if there is only an IP addresse or only an FQDN: would either `%fd.sip.name` or `%fd.sip` be blank in that case?

But indeed, we can consider something like this:
```
output: Outbound connection to C2 server domain=%fd.sip.name addr=%fd.sip (command=%proc.cmdline connection=%fd.name user=%user.name user_loginuid=%user.loginuid container_id=%container.id image=%container.image.repository)
```
or this (moving the parenthesis):

```
output: Outbound connection to C2 server (domain=%fd.sip.name addr=%fd.sip command=%proc.cmdline connection=%fd.name user=%user.name user_loginuid=%user.loginuid container_id=%container.id image=%container.image.repository)
```

We could also be more specific by adding a prefix `c2_` like this: `c2_domain=%fd.sip.name c2_addr=%fd.sip` .
I have no particular preference.

darryk10: I agree the option with c2_domain and c2_addr would be specific and clear so I would go in this direction. `
I would also go for this output you mentioned 
`output: Outbound connection to C2 server (domain=%fd.sip.name addr=%fd.sip command=%proc.cmdline connection=%fd.name user=%user.name user_loginuid=%user.loginuid container_id=%container.id image=%container.image.repository)`
As far as the empty fields goes I don't think it would be a problem if left empty but if you could do a quick test to see if all makes sense from your side would be awesome.
Thanks again

Nicolas-Peiffer: Okay no big deal with blank IP addresses or blank FQDN : falco will output `<NA> ` if there are no values.

In the following, we can see `c2_domain=<NA> c2_addr=46.101.90.205` because the rule was triggered only with the IP adress.

I am pushing the modification.

```
falco 18:25:57.134661204: Warning Outbound connection to C2 server (c2_domain=<NA> c2_addr=46.101.90.205 command
=curl 46.101.90.205 connection=10.42.0.7:59170->46.101.90.205:80 user=<NA> user_loginuid=-1 container_id=62513c2
df2f2 image=<NA>) k8s.ns=default k8s.pod=tmp-shell container=62513c2df2f2
```

darryk10 (approved):
LGTM
Thanks @Nicolas-Peiffer for the changes :)

leogr (approved):

jasondellaluce (approved):
/approve

leogr:
/close

PR #2255, opened 2022-10-14
fix(userspace/falco): avoid using CPU when main thread waits for parallel event sources

jasondellaluce:
**What type of PR is this?**

/kind bug

/area engine

When synchronizing with parallel event sources running in their own thread, the main thread was not put to sleep and consumed CPU with no purpose. Using a semaphore, we're able to let it sync with the additional threads only when at least one signals termination.

```release-note
NONE
```


Andreagit97 (approved):
/approve

FedeDP (approved):
/approve

PR #2258, opened 2022-10-14
update(userspace/falco): turn error into warning log when plugins are loaded but unused

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind design

/area engine

/area tests

The error `plugin XXX has field extraction capability but is not compatible with any loaded event source` has been troublesome for many users. Although it's semantically correct to prevent Falco from loading plugins that are not actually used, this causes the UX to be worse for users who are getting started with the plugin system. So, this PR proposes turning the error into a warning-level log message, so that Falco does not get blocked.

This is a proposal open for discussion, with the goal to insert it in Falco 0.33.0 eventually given the small amount of changes involved.

/hold

/milestone 0.33.0

```release-note
update(userspace/falco): turn error into warning log when plugins are loaded but unused
```

### DIFF
--- a/test/falco_tests_plugins.yaml
+++ b/test/falco_tests_plugins.yaml
@@ -54,13 +54,6 @@ trace_files: !mux
       - 'Cloudtrail Create Instance': 1
     conf_file: BUILD_DIR/test/confs/plugins/cloudtrail_json_create_instances_bigevent.yaml
 
-  incompatible_extract_sources:
-    exit_status: 1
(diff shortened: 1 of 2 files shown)


leogr (approved):

jasondellaluce:
/close

I feel like these changes are not really necessary. It is good to make users' life easier, but I'm concerned that by making this fail silently will make issues related to plugins config harder/impossible to debug and troubleshoot.

PR #2210, opened 2022-09-14
Rules matching perf improvements

mstemm:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind release

> If contributing rules or changes to rules, please make sure to also uncomment one of the following line:

> /kind rule-update

> /kind rule-create

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area build

/area engine

> /area rules

> /area tests

> /area proposals

> /area CI

**What this PR does / why we need it**:

Some small changes to rules matching that reduce cpu usage with high event volumes (> 1M syscalls/sec)

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Some small changes to rules matching that reduce cpu usage with high event volumes (> 1M syscalls/sec)
```

PR #2272, opened 2022-10-25
fix(engine): save syscall source only when processing events

mstemm:
The optimization in https://github.com/falcosecurity/falco/pull/2210 had a bug when the engine uses multiple sources at the same time--m_syscall_source is a pointer to an entry in the indexed vector m_sources, but if add_source is called multiple times, the vector is resized, which copies the structs but invalidates any pointer to the vector entries.

So instead of caching m_syscall_source in add_source(), cache it in process_events(). m_sources won't change once processing events starts.

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind release

> /kind rule-update

> /kind rule-create

> /area build

/area engine

> /area rules

> /area tests

> /area proposals

> /area CI

Fixes #

```release-note
Avoid crash related to caching syscall source when the falco engine uses multiple sources at the same time.
```

### DIFF
--- a/userspace/engine/falco_engine.cpp
+++ b/userspace/engine/falco_engine.cpp
@@ -346,6 +346,11 @@ unique_ptr<falco_engine::rule_result> falco_engine::process_event(std::size_t so
 
 	if(source_idx == m_syscall_source_idx)
 	{
+		if(m_syscall_source == NULL)
+		{
+			m_syscall_source = find_source(m_syscall_source_idx);
+		}
+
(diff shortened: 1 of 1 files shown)


jasondellaluce (approved):
/approve 

FYI, this didn't cause trouble in Falco so far because the syscall source is always loaded as the last one.

FedeDP (approved):
/approve

PR #2282, opened 2022-11-07
[WIP] update(CI): Integration of modern BPF probe into Falco

Andreagit97:
**What type of PR is this?**

/kind feature

/area build

/area CI

This PR tries to integrate the modern BPF probe into Falco, 3 main aspects will be evaluated:
1. CI integration
2. build integration
3. code integration (If Falco needs some fixes to work correctly with the modern probe)

```release-note
update(CI): Integration of modern BPF probe into Falco
```


Andreagit97:
cc @leogr

.github/workflows/ci.yml
```diff
@@ -167,4 +167,48 @@ jobs:
...
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      # Build the modern BPF skeleton
+      - name: Build modern BPF skeleton 🐝
+        run: |
+          mkdir -p skeleton-build
+          cd skeleton-build && cmake -DUSE_BUNDLED_DEPS=ON -DBUILD_FALCO_MODERN_BPF=ON -DCREATE_TEST_TARGETS=Off ../
+          make ProbeSkeleton
```
incertum, line 192: @Andreagit97: Would building on ubuntu-22.04 work for example for running Falco w/ modern_bpf on let's say a centos7 6.0 kernel? When we debugged the possible "heisenbug" I recall it would need to be built in a centos7 container because of old GLIBC version `2.17` constraints (given the modern probe is baked into scap). In that case need to build newer clang versions from source in the centos7 container or maybe curl pre-built clang artifacts from Falco's Artifact store?

incertum: Or given below when you use the centos7 container this is all not a problem and it works. In that case please disregard.

Andreagit97: yeah, the trick here is to build the header file (so the BPF skeleton) on a recent machine with the latest clang/llvm versions like ubuntu 22.04 since building it directly on centos7 would be a real pain. After that, we use this header file to build Falco on centos7 in this way the modern probe should be compliant with lower versions of GLIBC like `2.17` as you mentioned :)

Andreagit97:
Just a quick update: 
This PR will be used to ship the packages (`x86`, `aarch64`) thanks to the Circle CI jobs. The CI doesn't generate the docker images until the PR is merged, for this reason, I will manually download the `tag.gz` packages for both architectures, create the docker images locally and push them to my personal dockerhub :) We will notify you on the Falco slack channel when all is in place.

After a short period of testing, we will merge this PR and the modern probe will be regularly shipped into Falco 0.34 as another possible syscall source :partying_face:

Andreagit97:
We will leave this PR open until Falco 0.34 is released after that we will close it. The PR with the modern probe integration within Falco is this one https://github.com/falcosecurity/falco/pull/2320

Andreagit97:
You can find the new packages and docker images directly on the master branch
/close

PR #2290, opened 2022-11-14
rules: use list of Falco containers instead of repeating them

oscr:
**What type of PR is this?**

/kind cleanup
/kind rule-update

/area rules

In the default rules I noticed the Falco containers where repeated in multiple rules. This creates a list and uses it instead which is similar to how the ibm_cloud_containers list is implemented and used.   

Fixes #

```release-note
rules: add Falco container lists
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -1844,18 +1844,31 @@
           registry.access.redhat.com/sematext/agent,
           registry.access.redhat.com/sematext/logagent]
 
+# Falco containers
+- list: falco_containers
+  items:
+    - falcosecurity/falco
+    - docker.io/falcosecurity/falco
+    - public.ecr.aws/falcosecurity/falco
+
+# Falco no driver containers
+- list: falco_no_driver_containers
+  items:
+    - falcosecurity/falco-no-driver
+    - docker.io/falcosecurity/falco-no-driver
+    - public.ecr.aws/falcosecurity/falco-no-driver
+
 # These container images are allowed to run with --privileged and full set of capabilities
 - list: falco_privileged_images
   items: [
+    falco_containers,
     docker.io/calico/node,
     calico/node,
     docker.io/cloudnativelabs/kube-router,
     docker.io/docker/ucp-agent,
-    docker.io/falcosecurity/falco,
     docker.io/mesosphere/mesos-slave,
     docker.io/rook/toolbox,
     docker.io/sysdig/sysdig,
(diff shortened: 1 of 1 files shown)


rules/falco_rules.yaml
```diff
@@ -2395,6 +2393,20 @@
     - registry.ng.bluemix.net/armada-master/metrics-server-amd64
     - registry.ng.bluemix.net/armada-master/olm
 
+# Falco containers
+- list: falco_containers
```
darryk10, line 2397: Can we move the list declaration on top before to actually use it. this would raise some errors and it would be better to find the macro if declared before to use it.

darryk10 (changes requested):
Thanks for the contribution. Cleaning is always appreciated and it might help with future whitelists.
I just dropped some comments.
Thanks

rules/falco_rules.yaml
oscr: No problem! I moved it before the first usage.

darryk10: @oscr I still see the list where it was. Do you mind do a double check pls?

oscr: @darryk10 I'm sorry! I forgot to actually add the file before amending and pushing :sweat_smile: Fixed now

rules/falco_rules.yaml
```diff
@@ -1838,18 +1838,31 @@
           registry.access.redhat.com/sematext/agent,
           registry.access.redhat.com/sematext/logagent]
 
+# Falco containers
+- list: falco_containers
```
darryk10, line 1842: @jasondellaluce are the lists ok with this syntax or we would prefer to use the same syntax we have for the other lists?

oscr: @darryk10 Are you thinking `containers` vs `images`?

I followed the example of `- list: ibm_cloud_containers` but there seems to be others `- list: redhat_io_images_privileged`.

leogr (approved):

FedeDP (approved):
/approve

leogr:
Closing and reopening to trigger circleCI
/close

leogr (approved):
I had to rebase to fix some issues with the CI.

Reapproved.

@FedeDP could you again, pls?

jasondellaluce (approved):
/approve

oscr:
Thank you everyone for feedback and help with this pr! :pray:

PR #2291, opened 2022-11-14
rules: add salt-call as a trusted process

vin01:
**What type of PR is this?**

/kind bug
/kind rule-update

> /area rules

`salt-minion` was added as a trusted process in https://github.com/falcosecurity/falco/commit/85f51cf38c504772e00cc8918442ba6323ec9b89 and https://github.com/falcosecurity/falco/commit/1feae90c74943a7152972d94a39ed80965c26614, this however does not cover [salt-call](https://docs.saltproject.io/en/latest/ref/cli/salt-call.html) which is also a saltstack component and used on minions to apply changes from client-side. It should also be a trusted process along-with salt-minion as part of Saltstack.

```release-note
rule(Read sensitive file untrusted): let salt-call read sensitive files
rule(macro: rpm_procs): let salt-call write to rpm database
```

Thanks!

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -220,7 +220,7 @@
   items: [probe_rpminfo, probe_rpmverify, probe_rpmverifyfile, probe_rpmverifypackage]
 
 - macro: rpm_procs
-  condition: (proc.name in (rpm_binaries, openscap_rpm_binaries) or proc.name in (salt-minion))
+  condition: (proc.name in (rpm_binaries, openscap_rpm_binaries) or proc.name in (salt-call, salt-minion))
 
 - list: deb_binaries
   items: [dpkg, dpkg-preconfigu, dpkg-reconfigur, dpkg-divert, apt, apt-get, aptitude,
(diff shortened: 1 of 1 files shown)


darryk10:
Hi @vin01 , thanks for contributing :) 
I'm wondering if you are suggesting to add `salt-call` since you have seen some noise coming form it. If this is the case, can you share which kind of noise have you experienced?
thanks

vin01:
> Hi @vin01 , thanks for contributing :) I'm wondering if you are suggesting to add `salt-call` since you have seen some noise coming form it. If this is the case, can you share which kind of noise have you experienced? thanks

Hi @darryk10, right. It was primarily `Warning` alerts as a result of attempts by `salt-call` to read files like `/etc/shadow`, `/etc/sudo.*` which is expected as salt is used to manage users, sudo rights using in-built [modules](https://docs.saltproject.io/en/latest/ref/modules/all/salt.modules.user.html). Typically I would just use a custom rule override but since `salt-minion` is trusted in this rule-set, I think it makes sense to have `salt-call` as well present in the same. `salt-call` is used on the client-side to apply states, while `salt-minion` is used when applying the same changes from Salt master itself.

```
Warning Sensitive file opened for reading by non-trusted program (user=root user_loginuid=-1 program=salt-call command=salt-call /usr/bin/salt-call state.apply ..
```

darryk10:
Hi @vin01 thanks for the extra info and thanks to contributing to lower the noise :) 
LGTM

darryk10 (approved):
LGTM

jasondellaluce (approved):
/approve

jasondellaluce:
closing and reopening to trigger the CI

/close

leogr (approved):

leogr (approved):

PR #2321, opened 2022-12-13
update(ci): enable gh actions jobs on `maintainers/` branches

Andreagit97:
**What type of PR is this?**

/kind feature

/area CI

This PR allows maintainers to create upstream branches to test new features. The testing branch must start with `maintainers/` prefix

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,9 @@ on:
   pull_request:
     branches: [master]
   push:
-    branches: [master]
+    branches:
+      - master
+      - 'maintainers/**'
   workflow_dispatch:
 
 jobs:


.github/workflows/ci.yml
```diff
@@ -3,7 +3,9 @@ on:
   pull_request:
     branches: [master]
   push:
-    branches: [master]
+    branches:
+      - master
+      - 'maintainers/**'
```
leogr, line 8: Since only maintainers can create branches, why not just enable that on all branches? :thinking:

Andreagit97: Good question @leogr! this is because if you open a pull request from an upstream branch GH jobs will run twice since you have both a pull-request and an upstream branch. The idea here is to use  `maintainers/` branches as test and then open the pull request from another branch called in a different way, in this way jobs will run just one time

leogr: No way to tell the GHA system to dedup jobs with the same commit id? :thinking: 

I'm not contrary to this solution, anyway there're two things I don't like:
- This introduces a convention that maintainers should know, and it's easy to forget
- Occasionally, a testing branch may become PR; in that case, if one by mistake reuses the same branch, the jobs will still run twice

In other repositories, we sometimes open dummy PRs that will never be merged (i.e, `DO NOT MERGE THIS PR`) for the same purpose (just running the CI jobs). 
IMO, it's an alternative to consider since that wouldn't require an explicit CI config, makes the practice of manually testing some code more transparent, and can be used by non-maintainers too.

Just my 2 cents.

what do other maintainers think? cc @falcosecurity/falco-maintainers

Andreagit97:
Probably not so necessary, i can close it :)

PR #2325, opened 2022-12-15
new(falco): add more version fields to --support and --version

LucaGuerra:
**What type of PR is this?**

/kind feature

/area engine

Since a few Falco versions we have fields indicating libs, drivers, plugin APIs and so forth. These should be reflected into the `--support` option of Falco. In addition, the `--version` command should contain the engine version as well in my opinion since users may want to easily check whether a rule file is compatible with their version of Falco without necessarily running against it.

Fixes #

```release-note
new(falco): add more version fields to --support and --version
```

### DIFF
--- a/userspace/falco/app_actions/print_support.cpp
+++ b/userspace/falco/app_actions/print_support.cpp
@@ -37,6 +37,8 @@ application::run_result application::print_support()
 		nlohmann::json support;
 		struct utsname sysinfo;
 		std::string cmdline;
+		std::unique_ptr<sinsp> s(new sinsp());
+		char driver_api_version_string[32], driver_schema_version_string[32];
 
 		if(uname(&sysinfo) != 0)
 		{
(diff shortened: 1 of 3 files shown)


leogr:
/milestone 0.34.0

:+1:

userspace/falco/app_actions/print_support.cpp
```diff
@@ -37,13 +37,36 @@ application::run_result application::print_support()
...
+		unsigned long driver_api_patch = PPM_API_VERSION_PATCH(driver_api_version);
+		snprintf(driver_api_version_string, sizeof(driver_api_version_string), "%lu.%lu.%lu", driver_api_major, driver_api_minor, driver_api_patch);
+
+		support["driver_api_version"] = driver_api_version_string;
+
+		auto driver_schema_version = s->get_scap_schema_version();
+		unsigned long driver_schema_major = PPM_API_VERSION_MAJOR(driver_schema_version);
+		unsigned long driver_schema_minor = PPM_API_VERSION_MINOR(driver_schema_version);
+		unsigned long driver_schema_patch = PPM_API_VERSION_PATCH(driver_schema_version);
+		snprintf(driver_schema_version_string, sizeof(driver_schema_version_string), "%lu.%lu.%lu", driver_schema_major, driver_schema_minor, driver_schema_patch);
```
Andreagit97, line 65: it would be amazing if we had a sinsp API to do that instead of using low levels macro, but we can do that in a next step :)

Andreagit97 (approved):
/approve

leogr (approved):

userspace/falco/app_actions/print_support.cpp
LucaGuerra: Completely agree, since I wrote this when libs 0.10.0 was already about to be released I didn't want to add an API dependency right now, as harmless as it may be.

Issue #2326, opened 2022-12-15
[Bug] Incorrect log output on arm64 in application::check_for_ignored_events


Happy-Dude:
**Describe the bug**

When starting Falco on a arm64 host, I see the following log message:

```
Dec 15 12:44:28 debian falco_execstart.sh[735]: Loaded rules match the following events: symlink, mkdir, unlink, dup2, rmdir, link, open, rename, creat
Dec 15 12:44:28 debian falco_execstart.sh[735]: But these events are not returned unless running falco with -A
```

This log message is generated from https://github.com/falcosecurity/falco/blob/5e531870a94131abc092a7b08c49e8b127630ef8/userspace/falco/app_actions/load_rules_files.cpp#L39

However, this seems incorrect as `symlink, mkdir, unlink, dup2, rmdir, link, open, rename, creat` include syscalls that should be part of the simple set https://github.com/falcosecurity/libs/blob/50340f031fd9c16b27941bb23fc2d799194c9820/userspace/libsinsp/sinsp_ppm_sc.cpp#L58-L149

When discussed in Falco Slack channel, https://kubernetes.slack.com/archives/CMWH3EH32/p1671130569476529?thread_ts=1668629889.534749&cid=CMWH3EH32, these syscalls may not be defined in arm64

**How to reproduce it**

* Install Falco on arm64 node

**Expected behaviour**

* Either Falco supports these syscalls and/or emits the correct log message

**Environment**

- Falco version:
```
schan@debian:~$ falco --version
Falco version: 0.33.1-1-48a2c44b
Libs version:  0.9.0
Plugin API:    2.0.0
Driver:
  API version:    2.0.0
  Schema version: 2.0.0
  Default driver: 3.0.1+driver
```
- System info:
```
Thu Dec 15 13:03:43 2022: Falco version: 0.33.1-1-48a2c44b (aarch64)
Thu Dec 15 13:03:43 2022: Falco initialized with configuration file: /etc/falco/falco.yaml
Thu Dec 15 13:03:43 2022: Loading rules from file /etc/falco/falco_rules.yaml
Thu Dec 15 13:03:43 2022: Loading rules from file /etc/falco/falco_rules.local.yaml
Rules match ignored syscall: warning (ignored-evttype):
Loaded rules match the following events: symlink, mkdir, unlink, dup2, rmdir, link, open, rename, creat
But these events are not returned unless running falco with -A
{
  "machine": "aarch64",
  "nodename": "debian",
[...]

PR #2329, opened 2022-12-16
doc(userspace): provide users with a correct message when some syscalls are not defined

Andreagit97:
**What type of PR is this?**

/kind bug

/kind documentation

/area engine

As reported by @Happy-Dude here https://github.com/falcosecurity/falco/issues/2326, when running Falco on `aarch64`, Falco logs a misleading error message. Take as an example the issue log:

```
Loaded rules match the following events: symlink, mkdir, unlink, dup2, rmdir, link, open, rename, creat
```

All these syscalls are not defined on `aarch64` architectures but Falco is not yet able to detect it! As a first step, I would fix the log just to let the user know what is happening under the hood. Obviously, this is not the final solution, which will require some changes in the libraries to detect this sort of behavior, for this reason, the final fix will be shipped only in `Falco 0.35`

```release-note
doc(userspace): provide users with a correct message when some syscalls are not defined
```

### DIFF
--- a/userspace/falco/app_actions/load_rules_files.cpp
+++ b/userspace/falco/app_actions/load_rules_files.cpp
@@ -67,7 +67,7 @@ void application::check_for_ignored_events()
 		std::cerr << (first ? "" : ", ") << it.c_str();
 		first = false;
 	}
-	std::cerr << std::endl << "These events could be associated with syscalls not defined on your architecture (please take a look here: https://marcin.juszkiewicz.com.pl/download/tables/syscalls.html). If syscalls are defined you have to run Falco with `-A` to catch these events" << std::endl;
+	std::cerr << std::endl << "These events might be associated with syscalls undefined on your architecture (please take a look here: https://marcin.juszkiewicz.com.pl/download/tables/syscalls.html). If syscalls are instead defined, you have to run Falco with `-A` to catch these events" << std::endl;
 }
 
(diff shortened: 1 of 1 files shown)


userspace/falco/app_actions/load_rules_files.cpp
```diff
@@ -67,7 +67,7 @@ void application::check_for_ignored_events()
 		std::cerr << (first ? "" : ", ") << it.c_str();
 		first = false;
 	}
-	std::cerr << std::endl << "But these events are not returned unless running falco with -A" << std::endl;
+	std::cerr << std::endl << "These events could be associated with syscalls not defined on your architecture (please take a look here: https://marcin.juszkiewicz.com.pl/download/tables/syscalls.html). If syscalls are defined you have to run Falco with `-A` to catch these events" << std::endl;
```
FedeDP, line 70:
```suggestion
	std::cerr << std::endl << "These events might be associated with syscalls undefined on your architecture (please take a look here: https://marcin.juszkiewicz.com.pl/download/tables/syscalls.html). If syscalls are instead defined, you have to run Falco with `-A` to catch these events" << std::endl;
```

Andreagit97: ty :)

FedeDP (approved):
/approve

leogr (approved):

Issue #2222, opened 2022-09-27
[New Feature]: Falco native resource utilization metrics logs support


incertum:
**Motivation**

Support for "Falco native resource utilization metrics" is high up on the wish list of SREs I have the pleasure to work with.
While many end users sustaining a large deployment already pull such metrics from their systems using other mechanisms, there is always some loss of information. In addition, it can be cumbersome to join information from different sources and typically specialized metrics are not supported. 

Falco could very easily emit basic aggregated resource utilization metrics scheduled on a cron tab (or simpler alternative similar to existing methods). Additional overhead should be low since many metrics are already available.

Finally, it would make it easier to perform ad-hoc performance studies, especially as it seems LSM hooks are favored as additional event sources for the next Falco iterations. That way both tool developers and end users can optimize the tool better for threat detection use cases and derive SLO (Service Level Objectives) that can be the basis for resource overhead budgeting. Such logs with more specialized metrics can also help disentangling factors that can cause higher resource utilization that are however outside of the tool developer's influence, such as hardware, kernel version or actual workload footprint (event rate).

**Feature**

[Edited May 23, 2023]:

Consider these key points about the new `metrics` feature in Falco:

- It introduces a redesigned stats/metrics system.
- Native support for resource utilization metrics and specialized performance metrics.
- Metrics are emitted as monotonic counters at predefined intervals (snapshots).
- All metrics are consolidated into a single log message, adhering to the established  rules schema and naming conventions.
- Additional info fields complement the metrics and facilitate customized statistical analyses and correlations.
-  The metrics framework is designed for easy future extension.

Additional highlights:
- Includes old / established stats / counters in one place
- For bpf drivers we now support `libbpf` stats
- More extensions (syscall counters and prometheus exporter option) are planned

Navigate to the `metrics` key in https://github.com/falcosecurity/falco/blob/master/falco.yaml.

Example metrics snapshot schema for Falco 0.35 release using `bpf` driver:
```json
{
  "hostname": "x",
  "output": "Falco metrics snapshot",
  "output_fields": {
    "evt.source": "syscall",
    "evt.time": 1684842387420399900,
    "falco.container_memory_used": 0.0,
    "falco.cpu_usage_perc": 1.500000,
    "falco.duration_sec": 26,
    "falco.evts_rate_sec": 213.677443,
    "falco.host_boot_ts": 1682760026000000000,
    "falco.host_num_cpus": 20,
    "falco.hostname": "x",
[...]

PR #2333, opened 2022-12-19
new(falco): introduce new metrics w/ `Falco internal: metrics snapshot` option and new `metrics` config

incertum:

**What type of PR is this?**

> /kind bug

> /kind cleanup

/kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind release

> /kind rule-update

> /kind rule-create

> /area build

/area engine

> /area rules

> /area tests

> /area proposals

> /area CI

Initial solution to [Falco native resource utilization metrics logs support](https://github.com/falcosecurity/falco/issues/2222).

Consider these key points about the new `metrics` feature in Falco:

- It introduces a redesigned stats/metrics system.
- Native support for resource utilization metrics and specialized performance metrics.
- Metrics are emitted as monotonic counters at predefined intervals (snapshots).
- All metrics are consolidated into a single log message, adhering to the established  rules schema and naming conventions.
- Additional info fields complement the metrics and facilitate customized statistical analyses and correlations.
-  The metrics framework is designed for easy future extension.

Additional highlights:
- Includes old / established stats / counters in one place
- For bpf drivers we now support `libbpf` stats
- More extensions (syscall counters and prometheus exporter option) are planned

https://github.com/falcosecurity/falco/issues/2222

Fixes #

```release-note
new(falco): introduce new metrics w/ Falco internal: metrics snapshot option and new metrics config
```


incertum:
Continuing to stage here, but the expectation would be to move CPU and memory resource utilization to `libsinsp` and expose via proper APIs, besides it's still unclear how the concept of a unified stats module will look like in Falco.

Meanwhile, wondering about `memory`, to say the least it is a bit confusing how different tools compute it and knowing not just what it means, but also what implications for tool optimizations are etc. Perhaps worth a discussion to decide what we think is meaningful for Falco optimizations, especially the Kubernetes use case?

Added reading `/sys/fs/cgroup/memory/memory.usage_in_bytes`, but there is more potential in `/sys/fs/cgroup/memory/memory.stat` and have been reading through a couple of [resources](https://blog.freshtracks.io/a-deep-dive-into-kubernetes-metrics-part-3-container-resource-metrics-361c5ee46e66) and I am still unclear how `container_memory_working_set_bytes` is calculated, also cAdvisor source fetches info differently, not from fs ... @gnosek just curious if you had more thoughts as well. I think it would be important to optimize Falco for setting the right limits, so end users don't run into unexpected OOM kills, also something I have been chatting with @Andreagit97 . It would be wonderful having all these metrics in one Falco log and not needing to stitch together numbers from 2-3 external sources.

Happy-Dude:
Would this be the right place to request metrics/counters on syscalls inspected?

A scenario that played out in my organization is that a endpoint-detection tool was experiencing abnormally high CPU usage, and emitted stats of syscalls it was inspecting. 

One particular syscall was encountered much more than the rest, and we did some further digging (via other tooling and auditd) and discovered that the latest version of that service daemon was experiencing errors and needed to be patched.

While Falco's high CPU usage was not the source of the issue, being able to get an aggregate summary of what falco is doing during it's uptime is helpful. This is slightly different from getting a direct read of events Falco is triggering on (which I believe is already exported as a metric), because an event does not have to be a true positive to affect Falco's overall performance.

Happy-Dude:
Would this be the right place to request metrics/counters on syscalls inspected?

A scenario that played out in my organization is that a endpoint-detection tool was experiencing abnormally high CPU usage, and emitted stats of syscalls it was inspecting.

One particular syscall was encountered much more than the rest, and we did some further digging (via other tooling and auditd) and discovered that the latest version of that service daemon was experiencing errors and needed to be patched.

While Falco's high CPU usage was not the source of the issue, being able to get an aggregate summary of what falco is doing during it's uptime is helpful. This is slightly different from getting a direct read of events Falco is triggering on (which I believe is already exported as a metric), because an event does not have to be a true positive to affect Falco's overall performance.

gnosek:
@incertum, ages ago, the most relevant memory metric for me turned out to be PSS (not RSS), which accounts for page sharing between processes (e.g. a 2MB mapping shared between two processes is accounted as 1MB to each one), but I don't think there's one metric that fits all cases (e.g. PSS fails when you want to determine how much memory you'll get back by killing that process, which I suppose the OOM killer is interested in).

TL;DR, every choice we make will be wrong for somebody.

@Happy-Dude, yeah, some more detailed stats would be nice. Still, to track down the issue you were having, you'd need per (tid-or-pid, syscall) stats and I'm afraid that's overkill in the general case (there's a pretty big chance you'd need to (re)allocate memory for every event while the stats table grows).

I'm ->this<- close to suggesting that we use chisels for this (so that we can run arbitrary lua code on each event) but that will only make the perf issues worse, if any. Or we take the memory hit and we shove the stats in sinsp_threadinfo, but I'd rather make this optional.

BTW, sysdig (as opposed to falco) would be a better tool IMO to track down performance issues. At the very least, it supports chisels out of the box ;) (there's e.g. `topscalls` which does roughly what you want)

Happy-Dude:
> Existing stats for kernel side [event_drops](https://github.com/falcosecurity/falco/blob/master/userspace/falco/event_drops.cpp#L164) had been extended to be more verbose, they now expose some coarse-grained syscall categories counters. However, we should have more options in userspace. Would similar bucketing (but more buckets) of syscalls work?

This is great! Agree that dropped events gives an overview of the cost for monitoring the syscall/ currently loaded rules.

At the cost of overhead, I'd like a verbose/debug flag that exposes per-syscall being inspected by Falco.
I agree with @gnosek that this is likely excessive and is better investigated with other tooling. 

> Meanwhile, the [scap-open](https://github.com/falcosecurity/libs/blob/master/userspace/libscap/examples/01-open/scap_open.c) util binary has become really good for testing stuff and measuring perf under different configurations as the kernel tracing accounts for most CPU.

I'm catching up on the discussion with scap-open, and it seems pretty useful. Could be something my org can incorporate scap-open as part of our linting workflow as a proactive measure when crafting new rules 🤔

incertum:
@Happy-Dude re `scap-open` not quite yet given atm a static set of syscalls is activated regardless of Falco rules. As @FedeDP confirmed such a capability to only activate the syscalls needed based on Falco rules MUST land for 0.35.

Started a [HackMD](https://hackmd.io/jmjIG5zLRvS-TPuDzxH7Qw?view) document to further discuss such counters per event type categories. Would you be able to help creating the right buckets? As I will open a libs PR with proper APIs could try adding them as well. Ty!

incertum:
Will work on this next week, idea is to move to sinsp API, will split into 3 PRs:
1. all cpu and memory metrics exposed over proper APIs or otherwise integrated into existing stuff
2. new userspace syscall categories counters
3. bpf libbpf stats / metrics

Will squash commits in this PR then and refactor a bit.

Also looked more into `container_memory_working_set_bytes` that OOM killer uses and values from `/sys/fs/cgroup/memory/memory.usage_in_bytes` are close enough (carefully examined numbers from multiple sources in a real-life Kubernetes cluster), would say that suffices for the cgroup memory metrics. Will also make it super clear in definitions and docs that `kubectl top pod` numbers would reflect the sum of all containers and typically folks have some sort of sidekick container that can use memory too ... and these metrics are for the falco container and falco binary only.

incertum:
Hi :wave: lot's of work happened in `libs` leading up to supporting these new Falco stats. 

Revised plan is to only support new internal rules for Falco 0.35. while postponing the Prometheus integration to the next Falco release 0.36.

Highlights:
- Relevant wrapper fields to put metrics in context are included
- For the critical counters have current counts and previous counts and event rates as determined by what happened between snapshots
- Regardless given we emit literally all counters, end users can calculate custom event rates or other calculations in a data lake / compute environment
- Hijacked existing stats_writer without breaking it @jasondellaluce , we both probably aren't yet too happy with it, but would urge to postpone a more major refactor after merging this PR if ok.

Items requiring some discussion:
- CMD arg `stats-interval` should be moved as proper config to falco.yaml, we could keep supporting `msec` and / or offer some presets similar to the buffer size ... old stats writer to a file probably shouldn't be promoted, we likely want to rather promote Prometheus later
- Still thinking about how to expose unit conversions for the memory metrics which right now were kept raw, but it's a mix of bytes or kb ... the typical unit SREs work with would be MB.
- Also still open to discuss how much config control to provide the end user with, e.g. expose options to subscribe to resource utilization metrics only, or in combination with the kernel side counters or plus the libbpf stats, right now it's all of them if applicable. 
- What is missing compared to the old stats writer is including plugin stats counters ... big TBD

New schema:

```
{
  "hostname": "x",
  "output": "",
  "output_fields": {
    "container_memory_used": "0",
    "cpu_usage_perc": "1.700000",
    "driver": "bpf",
    "evt.time": "1682517563904238818",
    "falco_evts_drop_rate_kernel_sec": "0",
    "falco_evts_rate_kernel_sec": "134",
    "falco_evts_rate_sec": "132",
    "falco_num_evts": "2890",
    "falco_num_evts_prev": "1561",
    "falco_start_ts": "1682517543901322593",
    "falco_version": "0.34.1-140+5c1062b",
    "host_boot_ts": "1679600728000000000",
    "host_num_cpus": "x",
    "hostname": "x",
    "kernel_release": "x",
    "memory_pss": "34613",
    "memory_rss": "37984",
    "memory_vsz": "1064472",
    "n_drops": "0",
    "n_drops_buffer_clone_fork_enter": "0",
    "n_drops_buffer_clone_fork_exit": "0",
    "n_drops_buffer_connect_enter": "0",
    "n_drops_buffer_connect_exit": "0",
    "n_drops_buffer_dir_file_enter": "0",
    "n_drops_buffer_dir_file_exit": "0",
    "n_drops_buffer_execve_enter": "0",
    "n_drops_buffer_execve_exit": "0",
    "n_drops_buffer_open_enter": "0",
    "n_drops_buffer_open_exit": "0",
    "n_drops_buffer_other_interest_enter": "0",
    "n_drops_buffer_other_interest_exit": "0",
    "n_drops_buffer_total": "0",
    "n_drops_bug": "0",
    "n_drops_page_faults": "0",
    "n_drops_prev": "0",
    "n_drops_scratch_map": "0",
    "n_evts": "2942",
    "n_evts_prev": "1599",
    "page_fault_kern": "0",
    "page_fault_kern.run_cnt": "0",
    "page_fault_kern.run_time_ns": "0",
    "page_fault_user": "0",
    "page_fault_user.run_cnt": "0",
    "page_fault_user.run_time_ns": "0",
    "sched_process_e.avg_time_ns": "16262",
    "sched_process_e.run_cnt": "5",
    "sched_process_e.run_time_ns": "81313",
    "sched_switch": "0",
    "sched_switch.run_cnt": "0",
    "sched_switch.run_time_ns": "0",
    "signal_deliver": "0",
    "signal_deliver.run_cnt": "0",
    "signal_deliver.run_time_ns": "0",
    "sys_enter.avg_time_ns": "439",
    "sys_enter.run_cnt": "72461",
    "sys_enter.run_time_ns": "31873687",
    "sys_exit.avg_time_ns": "833",
    "sys_exit.run_cnt": "72447",
    "sys_exit.run_time_ns": "60405277"
  },
  "priority": "Debug",
  "rule": "Falco internal: resource utilization stats metrics",
  "time": "2023-04-26T13:59:23.904238818Z"
}
```

incertum:
TODO:
- Discuss few details also around naming
- Finalize description after more feedback, e.g. be more specific with what settings you can use these stats aka in the internal rule case need debug priority etc ...

jasondellaluce (commented):
Left some comments on the config and UX. Will try to work on the stats writer internal improvements by proposing commits.

cmake/modules/falcosecurity-libs.cmake
```diff
@@ -50,6 +50,8 @@ if(MUSL_OPTIMIZED_BUILD)
 endif()
 
 set(SCAP_HOST_ROOT_ENV_VAR_NAME "HOST_ROOT")
+set(SCAP_HOSTNAME_ENV_VAR "FALCO_HOSTNAME")
```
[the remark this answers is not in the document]
incertum: There is an existing `FALCO_HOSTNAME` env variable which could be replaced by the new scap agent info we added to scap (not in this PR obviously).

The other one is completely new and open to any naming conventions as as you said it will become a contract.

If you also referred to the `_NAME` suffix yes we can add that.

falco.yaml
```diff
@@ -544,3 +544,84 @@ metadata_download:
...
+# and might change until stats_v2 reaches a stable release.
+#
+# --- [Usage]
+#
+# Disabled by default.
+#
+# Env variable `-E FALCO_HOSTNAME=myhostname` customizes hostname in case
+# you deploy in Kubernetes where the hostname can be equivalent to the pod name.
+#
+# `stats_interval_preset`:
```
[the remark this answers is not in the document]
incertum: Lovely suggestions, let's go with them. It makes sense especially because we want to also offer a Prometheus integration once the new metrics become stable for Falco 0.36.

falco.yaml
```diff
@@ -544,3 +544,84 @@ metadata_download:
...
+stats_v2:
+  enabled: false
+  stats_interval_preset: 3
+  stats_interval_ms: 0
+  stats_internal_rule: true
+  # stats_filename: /tmp/falco_stats.jsonl
+  include_resource_utilization: true
+  include_kernel_evts_counters: true
+  include_libbpf_stats: true
+  convert_memory_to_mb: true
```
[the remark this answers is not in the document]
incertum: Agreed.

incertum (commented):
@jasondellaluce thanks for the great comments. Also having limited availability atm, but we still have some time for the Falco metrics vs possible further `libs` addition (e.g. syscall counters) need to be added sooner.

Please feel free to push commits as you have time and when I have time I'll check in what items are still outstanding as well + will try to address them. Thanks for your help!

incertum:
@jasondellaluce got back to working on it.

> Counter proposal: why not supporting the "duration" syntax of prometheus? https://prometheus.io/docs/prometheus/latest/configuration/configuration/#configuration-file
... It can be parsed with a simple regex with catching groups (defined in that link), for which we can use RE2 imported from libsinsp ...

Didn't read this carefully and implemented the regex to be agnostic of duplicate time units plus atm not requiring ordering time units from longest to shortest duration. Since I already implemented it, pushed it that way, but I can change it.  It's not a perf critical regex as we parse only once at start up. Hence we can afford being more resilient here.

Addition follow up question: Is legacy interval settings vs new config support now more clear and as desired?

incertum:
@Happy-Dude could we kindly request your help to make the falco.yaml documentation better and more clear for the new `metrics` section? Much appreciated and thanks in advance!
[FYI we will add the syscall counters in June ...]

falco.yaml
```diff
@@ -544,3 +544,91 @@ metadata_download:
...
+# emitted at a pre-defined interval. Each metric is captured in the same consolidated log message.
+# In addition, relevant wrapper fields are added, allowing the end user to perform customized
+# statistical analyses and correlations. Lastly, the metrics framework can be easily extended in the future.
+#
+# Notice: Schema and output field names are not guaranteed to be stable
+# and might change until `metrics` reaches a stable release.
+#
+# --- [Usage]
+#
+# Disabled by default.
```
Happy-Dude, line 564: Add a header above this line to be consistent with the other options:

```
# `enabled`:
# Disabled by default.
```

falco.yaml
```diff
@@ -544,3 +544,91 @@ metadata_download:
...
+#
+# Env variable `-E FALCO_HOSTNAME=myhostname` customizes hostname for Kubernetes
```
Happy-Dude, line 566: Not sure if I understand where this is applicable / useful (in other words, why would I set something like this?). Since this is an env var set outside of `falco.yaml`, should this be moved above or below the settings configured in this file?

incertum: will add more info around it and maybe for here just saying you can set an environment variable is enough as there are multiple ways, for example you could do `sudo -FALCO_HOSTNAME=myhostname -E FALCO_BPF_PROBE=/root/.falco/falco-bpf.o ./userspace/falco/falco` ... and as said Linux and bash offer many ways.

falco.yaml
```diff
@@ -544,3 +544,91 @@ metadata_download:
...
+# metrics: periodic stats and resource utilization metrics, initial release
+#
+# --- [Description]
+#
+# `metrics` reflects a stats/metrics re-design from the ground up. Falco natively supports
+# resource utilization metrics and enhanced specialized metrics to monitor Falco's
+# performance in production. Metrics are exposed as monotonic counters or snapshots
+# emitted at a pre-defined interval. Each metric is captured in the same consolidated log message.
+# In addition, relevant wrapper fields are added, allowing the end user to perform customized
+# statistical analyses and correlations. Lastly, the metrics framework can be easily extended in the future.
```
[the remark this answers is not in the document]
incertum: This is good, will change it!

jasondellaluce (commented):
@incertum I rebased all your commits on top of the current `master` branch, and I added two more commits of my own to 1) solve a couple of issues, and 2) organize the code in a safer and more optimized way.

Overall, I think this looks good now on my end. As for what's missing, here's what I think we still need to make this release-ready:
- `falco_metrics.h`/`falco_metrics.cpp` have the sole purpose of wrapping an utility function, which is actually a prometheus time period parser (not related to our notion of metrics). I suggest generalizing the name of that function so that it can maybe be reused, and move it under `falco_utils.h/cpp` and the `falco::utils` namespace. I'm not a huge fan of how the regex parsing happens, IMO we could do it with one single expression, but I will not fight over it. Be mindful that this implementation will potentially accept out of order periods such as `5ms10d3h8w` 😃 
- Tests are totally missing. To the bare minimum we should test the time period parsing function in all the corner cases we can
- Some old stats output have not been preserved, such as the `delta` and `drop_pct`. Can we at least support the drop percentage? I think that's valuable for many community members (I recall issues about it with @shane-lawrence). Example of old JSON samples: ```{"sample":1,"syscall":{"cur":{"drop_pct":0.0,"drops":0,"events":5738,"preemptions":0},"delta":{"drop_pct":0.0,"drops":0,"events":5738,"preemptions":0}}}```
- Stats names are very inconsistent and arbitrary, and it's hard to understand what's the expectations depending on the output (e.g. rule's output fields or file json). My big ask is to have coherent naming rules for each metric name, for each output (rules alert, file, and prometheus in the future), and for syscall/non-syscall event sources. I attached a JSON sample below as an example. For instance, rules' output fields have a grammar constraint of being dot-separated and lowercase, and I think we should respect that. Also, what's the basis on which something has a `falco_` prefix or does not? Please let's work on making this very clear. I'll leave the discussion open from this point on. @Happy-Dude @incertum 


```json
{
  "example": {
    "container_memory_used": "13769.152344",
    "cpu_usage_perc": "126.300000",
    "driver": "source_plugin",
    "evt.time": "1684510672862762546",
    "falco_duration_sec": "5",
    "falco_num_evts": "3170770",
    "falco_num_evts_prev": "0",
    "falco_start_ts": "1684510667604175473",
    "falco_version": "0.34.1-238+359c86c",
    "host_boot_ts": "0",
    "host_num_cpus": "8",
    "hostname": "ubuntu-focal",
    "kernel_release": "5.4.0-148-generic",
    "memory_pss": "45.992188",
    "memory_rss": "48.710938",
    "memory_vsz": "2255.636719",
    "src": "example"
  },
  "sample": 1,
  "syscall": {
    "container_memory_used": "13769.281250",
    "cpu_usage_perc": "126.200000",
    "driver": "kmod",
    "evt.time": "1684510672878849714",
    "falco_duration_sec": "5",
    "falco_evts_drop_rate_kernel_sec": "0",
    "falco_evts_rate_kernel_sec": "0",
    "falco_num_evts": "9608",
    "falco_num_evts_prev": "0",
    "falco_start_ts": "1684510667604175473",
    "falco_version": "0.34.1-238+359c86c",
    "host_boot_ts": "1684491254000000000",
    "host_num_cpus": "8",
    "hostname": "ubuntu-focal",
    "kernel_release": "5.4.0-148-generic",
    "memory_pss": "46.066406",
    "memory_rss": "48.710938",
    "memory_vsz": "2319.636719",
    "n_drops": "0",
    "n_drops_buffer_clone_fork_enter": "0",
    "n_drops_buffer_clone_fork_exit": "0",
    "n_drops_buffer_connect_enter": "0",
    "n_drops_buffer_connect_exit": "0",
    "n_drops_buffer_dir_file_enter": "0",
    "n_drops_buffer_dir_file_exit": "0",
    "n_drops_buffer_execve_enter": "0",
    "n_drops_buffer_execve_exit": "0",
    "n_drops_buffer_open_enter": "0",
    "n_drops_buffer_open_exit": "0",
    "n_drops_buffer_other_interest_enter": "0",
    "n_drops_buffer_other_interest_exit": "0",
    "n_drops_buffer_total": "0",
    "n_drops_bug": "0",
    "n_drops_page_faults": "0",
    "n_drops_prev": "0",
    "n_evts": "9648",
    "n_evts_prev": "0",
    "n_preemptions": "0",
    "src": "syscall"
  }
}
```

incertum:
> @incertum I rebased all your commits on top of the current `master` branch, and I added two more commits of my own to 1) solve a couple of issues, and 2) organize the code in a safer and more optimized way.
> 

Thank you! Added few minor things back in such as an info message around the metrics time interval that was set (similar to the buffer size and others).

>    `falco_metrics.h`/`falco_metrics.cpp` have the sole purpose of wrapping an utility function, which is actually a prometheus time period parser (not related to our notion of metrics). I suggest generalizing the name of that function so that it can maybe be reused, and move it under `falco_utils.h/cpp` and the `falco::utils` namespace. 

Done.

> I'm not a huge fan of how the regex parsing happens, IMO we could do it with one single expression, but I will not fight over it. Be mindful that this implementation will potentially accept out of order periods such as `5ms10d3h8w` smiley
> 

Anticipated it here https://github.com/falcosecurity/falco/pull/2333#issuecomment-1549554688, was a mistake on my end to not read the prometheus time ordering constraint and I was out of time the other day, but changed it now.

>  Tests are totally missing. To the bare minimum we should test the time period parsing function in all the corner cases we can
> 

Added some initial tests, let me also check how we can test this better as we just recently refactored things to be able to better test app actions etc. Re the `libs` stats portions, we had added some tests in libs.

>  Some old stats output have not been preserved, such as the `delta` and `drop_pct`. Can we at least support the drop percentage? I think that's valuable for many community members (I recall issues about it with @shane-lawrence). Example of old JSON samples: `{"sample":1,"syscall":{"cur":{"drop_pct":0.0,"drops":0,"events":5738,"preemptions":0},"delta":{"drop_pct":0.0,"drops":0,"events":5738,"preemptions":0}}}`
> 

Added `n_drops_perc`, please take a look.


>   Stats names are very inconsistent and arbitrary, and it's hard to understand what's the expectations depending on the output (e.g. rule's output fields or file json). My big ask is to have coherent naming rules for each metric name, for each output (rules alert, file, and prometheus in the future), and for syscall/non-syscall event sources. I attached a JSON sample below as an example. For instance, rules' output fields have a grammar constraint of being dot-separated and lowercase, and I think we should respect that. Also, what's the basis on which something has a `falco_` prefix or does not? Please let's work on making this very clear. I'll leave the discussion open from this point on. @Happy-Dude @incertum
> 

Few thoughts:
- Wouldn't be a big fan of introducing a new nested schema at all and here is why: Let's be consistent with the regular rules `output_fields`, because then we do not break adopter's ETL's. In addition in my experience for efficient data storage and processing you flatten nested incoming JSON schemas out again anyways ...
- But I do agree wrt the inconsistencies and a lack of clarity around naming. Removed all `falco_` prefixes and my suggestion is to add a `k.` prefix for metrics around kernel side counters and `u.` for userspace, all other fields have no extra prefix and should be more self explanatory. Open to other prefix or naming conventions just again please no new nested JSON to ensure a seamless adoption just like a regular Falco rule ...
- On that note for the file output I would just keep the `output_fields` key and schema the same and we can add more unique wrapper fields such as we currently do with `sample`, but that way there should be no need for discussing differences (note that `src` is a field in the  `output_fields` already)?


@Happy-Dude made a few changes to the config description based on your feedback as well.

Current example output:

```json
{
  "hostname": "x",
  "output": "Falco metrics snapshot",
  "output_fields": {
    "container_memory_used": "0.0",
    "cpu_usage_perc": "3.500000",
    "driver": "bpf",
    "duration_sec": "10",
    "evt.time": "1684676985437250046",
    "host_boot_ts": "1682760026000000000",
    "host_num_cpus": "x",
    "hostname": "x",
    "k.evts_drop_rate_sec": "0",
    "k.evts_rate_sec": "242.977928",
    "k.n_drops": "0",
    "k.n_drops_buffer_clone_fork_enter": "0",
    "k.n_drops_buffer_clone_fork_exit": "0",
    "k.n_drops_buffer_connect_enter": "0",
    "k.n_drops_buffer_connect_exit": "0",
    "k.n_drops_buffer_dir_file_enter": "0",
    "k.n_drops_buffer_dir_file_exit": "0",
    "k.n_drops_buffer_execve_enter": "0",
    "k.n_drops_buffer_execve_exit": "0",
    "k.n_drops_buffer_open_enter": "0",
    "k.n_drops_buffer_open_exit": "0",
    "k.n_drops_buffer_other_interest_enter": "0",
    "k.n_drops_buffer_other_interest_exit": "0",
    "k.n_drops_buffer_total": "0",
    "k.n_drops_bug": "0",
    "k.n_drops_page_faults": "0",
    "k.n_drops_perc": "0.000000",
    "k.n_drops_prev": "0",
    "k.n_drops_scratch_map": "0",
    "k.n_evts": "2502",
    "k.n_evts_prev": "1994",
    "k.page_fault_kern.avg_time_ns": "0",
    "k.page_fault_kern.run_cnt": "0",
    "k.page_fault_kern.run_time_ns": "0",
    "k.page_fault_user.avg_time_ns": "0",
    "k.page_fault_user.run_cnt": "0",
    "k.page_fault_user.run_time_ns": "0",
    "k.sched_process_e.avg_time_ns": "4872",
    "k.sched_process_e.run_cnt": "9",
    "k.sched_process_e.run_time_ns": "43851",
    "k.sched_switch.avg_time_ns": "0",
    "k.sched_switch.run_cnt": "0",
    "k.sched_switch.run_time_ns": "0",
    "k.signal_deliver.avg_time_ns": "0",
    "k.signal_deliver.run_cnt": "0",
    "k.signal_deliver.run_time_ns": "0",
    "k.sys_enter.avg_time_ns": "393",
    "k.sys_enter.run_cnt": "35366",
    "k.sys_enter.run_time_ns": "13918695",
    "k.sys_exit.avg_time_ns": "579",
    "k.sys_exit.run_cnt": "35367",
    "k.sys_exit.run_time_ns": "20498056",
    "kernel_release": "6.0.16-200.fc36.x86_64",
    "memory_pss": "44.209961",
    "memory_rss": "47.386719",
    "memory_vsz": "1121.835938",
    "src": "syscall",
    "start_ts": "1684676975343843549",
    "u.evts_rate_sec": "238.194898",
    "u.num_evts": "2416",
    "u.num_evts_prev": "1918",
    "version": "0.34.1-245+ae94b75"
  },
  "priority": "Informational",
  "rule": "Falco internal: metrics snapshot",
  "source": "internal",
  "time": "2023-05-21T13:49:45.437250046Z"
}

```

userspace/falco/stats_writer.h
```diff
@@ -18,11 +18,15 @@ limitations under the License.
...
-#include <map>
+#include <unordered_map>
 
 #include <sinsp.h>
 
 #include "tbb/concurrent_queue.h"
+#include "falco_outputs.h"
+#include "configuration.h"
+
+#define MAX_STATS_PRESET_INDEX 6
```
jasondellaluce, line 29:
```suggestion
```
Leftover?

userspace/falco/stats_writer.cpp
```diff
@@ -25,7 +25,8 @@ limitations under the License.
 #include "stats_writer.h"
 #include "logger.h"
 #include "banned.h" // This raises a compilation error when certain functions are used
-#include "logger.h"
+#include "config_falco.h"
+#include <re2/re2.h>
```
jasondellaluce, line 29:
```suggestion
```
Leftover?

userspace/falco/stats_writer.cpp
```diff
@@ -161,39 +182,219 @@ void stats_writer::worker() noexcept
...
+	output_fields["evt.time"] = std::to_string(now); /* Some ETLs may prefer a consistent timestamp within output_fields. */
+	output_fields["version"] = FALCO_VERSION;
+	output_fields["start_ts"] = std::to_string(agent_info->start_ts_epoch);
+	output_fields["duration_sec"] = std::to_string((now - agent_info->start_ts_epoch) / ONE_SECOND_IN_NS);
+	output_fields["kernel_release"] = agent_info->uname_r;
+	output_fields["host_boot_ts"] = std::to_string(machine_info->boot_ts_epoch);
+	output_fields["hostname"] = machine_info->hostname; /* Explicitly add hostname to log msg in case hostname rule output field is disabled. */
+	output_fields["host_num_cpus"] = std::to_string(machine_info->num_cpus);
 
+	output_fields["src"] = src;
```
jasondellaluce, line 212: Starting from this release we'll have a new field named `evt.source` coming from libsinsp, so just like you did with `evt.time`, I'd try to be consistent.
```suggestion
	output_fields["evt.source"] = src;
```

jasondellaluce (commented):
LGTM, left you two minor suggestions.

My biggest concerns now is still just output fields name. If we want to be compliant with the "fields" as we supported in our condition syntax, then we need to stick to the same regex used by the condition grammar ([[a-zA-Z]+[a-zA-Z0-9_]*(\.[a-zA-Z]+[a-zA-Z0-9_]*)+](https://github.com/falcosecurity/libs/blob/2e9e6346eefeddd0afce7b6a06cb42d4265615dd/userspace/libsinsp/filter/parser.h#L62)). So:

- All fields must have a "field class" prefix
- Whatever field is `evt.***`, we need to be compliant with the same value format used in libsinsp
- Whatever new field class we had, it will become sort of a "reserved name" for field class prefixes (e.g. what you did with `u.` and `k.`.
- We should not format every value as string: json supports numbers, so I think we should format numbers as numbers, and strings as strings. This could probably be attained easily by passing down nlohman json array instead of a map<string,string> to `handle_msg()` and to the output queue control message.

On that JSON output example you posted above, my re-naming proposal is as follows (here I'm reserving the `falco.` and `driver.` (can also be `kernel.` or `capture.` maybe), and I'm using the right number formatting:

```json
{
"evt.time": 1684676985437250046,
"evt.source": "syscall",
"falco.container_memory_used": 0.0,
"falco.cpu_usage_perc": 3.500000,
"falco.version": "0.34.1-245+ae94b75",
"falco.duration_sec": 10,
"falco.host_boot_ts": 1682760026000000000,
"falco.host_num_cpus": 8,
"falco.hostname": "x",
"falco.memory_pss": 44.209961,
"falco.memory_rss": 47.386719,
"falco.memory_vsz": 1121.835938,
"falco.start_ts": 1684676975343843549,
"falco.evts_rate_sec": 238.194898,
"falco.num_evts": 2416,
"falco.num_evts_prev": 1918,
"falco.kernel_release": "6.0.16-200.fc36.x86_64",
"driver.name": "bpf",
"driver.evts_drop_rate_sec": 0,
"driver.evts_rate_sec": 242.977928,
"driver.n_drops": 0,
"driver.n_drops_buffer_clone_fork_enter": 0,
"driver.n_drops_buffer_clone_fork_exit": 0,
"driver.n_drops_buffer_connect_enter": 0,
"driver.n_drops_buffer_connect_exit": 0,
"driver.n_drops_buffer_dir_file_enter": 0,
"driver.n_drops_buffer_dir_file_exit": 0,
"driver.n_drops_buffer_execve_enter": 0,
"driver.n_drops_buffer_execve_exit": 0,
"driver.n_drops_buffer_open_enter": 0,
"driver.n_drops_buffer_open_exit": 0,
"driver.n_drops_buffer_other_interest_enter": 0,
"driver.n_drops_buffer_other_interest_exit": 0,
"driver.n_drops_buffer_total": 0,
"driver.n_drops_bug": 0,
"driver.n_drops_page_faults": 0,
"driver.n_drops_perc": 0.000000,
"driver.n_drops_prev": 0,
"driver.n_drops_scratch_map": 0,
"driver.n_evts": 2502,
"driver.n_evts_prev": 199,
"driver.page_fault_kern.avg_time_ns": 0,
"driver.page_fault_kern.run_cnt":0,
"driver.page_fault_kern.run_time_ns": 0,
"driver.page_fault_user.avg_time_ns": 0,
"driver.page_fault_user.run_cnt": 0,
"driver.page_fault_user.run_time_ns": 0,
"driver.sched_process_e.avg_time_ns": 4872,
"driver.sched_process_e.run_cnt": 9,
"driver.sched_process_e.run_time_ns": 43851,
"driver.sched_switch.avg_time_ns": 0,
"driver.sched_switch.run_cnt": 0,
"driver.sched_switch.run_time_ns": 0,
"driver.signal_deliver.avg_time_ns": 0,
"driver.signal_deliver.run_cnt": 0,
"driver.signal_deliver.run_time_ns": 0,
"driver.sys_enter.avg_time_ns": 393,
"driver.sys_enter.run_cnt": 35366,
"driver.sys_enter.run_time_ns": 13918695,
"driver.sys_exit.avg_time_ns": 579,
"driver.sys_exit.run_cnt": 35367,
"driver.sys_exit.run_time_ns": 20498056,
}
```

leogr:
> On that JSON output example you posted above, my re-naming proposal is as follows (here I'm reserving the `falco.` and `scap.` and I'm using the right number formatting:

:+1:

Happy-Dude:
> polished the config description a lot, please take another look.

Read through the `falco.yaml` doc changes and it is clear what the config entails; thanks for the improvements 🥳 !

jasondellaluce (approved):
/approve

jasondellaluce:
I'll take care of right number formatting in a  follow-up PR.

userspace/falco/stats_writer.cpp
```diff
@@ -161,39 +181,220 @@ void stats_writer::worker() noexcept
...
-			msg.stop = false;
-			msg.source = src;
-			inspector->get_capture_stats(&msg.stats);
-			m_samples++;
-			if(m_samples == 1)
+			for(uint32_t stat = 0; stat < nstats; stat++)
 			{
-				msg.delta = msg.stats;
+				char metric_name[STATS_NAME_MAX] = "falco.";
+				strncat(metric_name, utilization[stat].name, sizeof(metric_name) - strlen(metric_name));
```
FedeDP, line 254:
```
If src contains n or more bytes, strncat() writes n+1 bytes to
       dest (n from src plus the terminating null byte).  Therefore, the
       size of dest must be at least strlen(dest)+n+1.
```

We should do `sizeof(metric_name) - strlen(metric_name) - 1` here to avoid missing the terminator char.
We can do it in a subsequent PR but please do that :) @jasondellaluce

FedeDP (approved):
/approve

incertum:
> LGTM, `strncat` note aside. We can fix it in a subsequent PR!

Same comment here as in another PR: 

@FedeDP please note in libs we do it the way it currently is e.g. `kmod_path = strncat(cwd, KMOD_DEFAULT_PATH, FILENAME_MAX - strlen(cwd))` hence we would need a commit in libs to uniformly change it as well.

PR #2351, opened 2023-01-12
update: print JSON version output by using json_output config field

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind feature

/area engine

/area tests

This PR proposes removing the recently-introduced `version-json` CLI option and makes Falco print a JSON version output depending on the `json_output` configuration field such as follows:
```
# userspace/falco/falco --version 
Falco version: 0.33.1-111+faa878d
Libs version:  0.10.0
Plugin API:    2.0.0
Engine:        15
Driver:
  API version:    3.0.0
  Schema version: 3.0.0
  Default driver: 4.0.0+driver
```

```
# userspace/falco/falco --version -o json_output=true
{"default_driver_version":"4.0.0+driver","driver_api_version":"3.0.0","driver_schema_version":"2.0.0","engine_version":"15","falco_version":"0.33.1-111+faa878d","libs_version":"0.10.0","plugin_api_version":"2.0.0"}
```

The goal here is to avoid duplication of UX elements, since `json_output` is already used other Falco features to control whether the output is in plain text of JSON format.

This PR can be the first of a series where we make every Falco output support a JSON format depending on the `json_output` configuration. In the past, that field was only used to control the JSON formatting of the alert outputs, and it recently was also used to output the result of rule files validation in JSON format. Since we now have a third use case, I think in the future we might use `json_output` to make every Falco output formatted as JSON and thus machine-readable.

Also, this PR refactors a little bit the `falco_configuration` class (responsible of ~250 LOC to preserve the git blame), which can not be initialized and supports a more explicit default state. Moreover, loading the configuration has been moved as the first action performed by Falco, so that all actions can rely on configuration fields (such as `json_output`). Accordingly, blocking validation checks (such as requiring at least one enabled output or at least one rules file) are moved in their own action, so that early actions such as `print_version` are not blocked and can proceed even with an empty configuration.

```release-note
update: print JSON version output when json_output is enabled
```


userspace/falco/app_actions/init_outputs.cpp
```diff
@@ -23,6 +23,11 @@ using namespace falco::app;
 
 application::run_result application::init_outputs()
 {
+	if (m_state->config->m_outputs.empty())
+	{
+		return run_result::fatal("No outputs configured. Please configure at least one output file output enabled but no filename in configuration block");
```
FedeDP, line 28: Mmmh is this sentence ok? :laughing:

userspace/falco/app_actions/load_config.cpp
```diff
@@ -20,29 +20,48 @@ using namespace falco::app;
...
-			falco_logger::log(LOG_DEBUG, "CLI args: " + m_state->cmdline);
+			m_state->config->init(m_options.cmdline_config_options);
 		}
+	}
+	catch (std::exception& e)
+	{
+		return run_result::fatal(e.what());
+	}
+
+	// log after config init because config determines where logs go
```
FedeDP, line 39: :top:

userspace/falco/application.cpp
```diff
@@ -172,13 +172,14 @@ bool application::run(std::string &errstr, bool &restart)
...
 	std::list<std::function<run_result()>> run_steps = {
+		std::bind(&application::load_config, this),
 		std::bind(&application::print_help, this),
 		std::bind(&application::print_version, this),
 		std::bind(&application::print_page_size, this),
 		std::bind(&application::print_generated_gvisor_config, this),
 		std::bind(&application::print_ignored_events, this),
 		std::bind(&application::print_syscall_events, this),
-		std::bind(&application::load_config, this),
+		std::bind(&application::require_config_file, this),
```
FedeDP, line 182: Do i read this right that we now allow to run Falco without a config file to just print various infos?

FedeDP (commented):
Minor comments/questions aside, LGTM!

FedeDP:
Perhaps @LucaGuerra is interested, because he made the `version-json` change ;)

userspace/falco/app_actions/init_outputs.cpp
jasondellaluce: This was moved as-is from the configuration parsing code. You're right, let me rephrase this better 😂

userspace/falco/application.cpp
FedeDP: (it was already like this, right. Ignore me!)

jasondellaluce: Correct! Basically:

- Falco loads its default config and applies all the CLI option overrides (`-o` option) right at start time. If a config file is specified, then Falco loads it instead of using the default config
- Alle the print actions are executed by not strictly requiring a config file (this was the original behavior), but will not have access to the configuration loaded up to that point (either the default one or the file-loaded one)
- Finally, at a certain point the Falco actions will require a non-default configuration to be loaded, thus will first check that the currently-loaded one came from a File

This was my best proposal for both allowing early actions to access the Falco config, while also preserving all the original behavior that required a config file to proceed.

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

Happy-Dude:
Can there be an additional documentation update that makes it clear that `json_output` influences the `--version` flag?

This was surprising to me initially when pulling in the latest upstream changes because I expected `json_output` to only influence the output of the falco_events / alerts / logs; I did not expect this option to influence the output of `--version`.

(My confusion was also compounded a user runs `falco --version` and doesn't have permissions to read `/etc/falco/falco.yaml`, so their output appeared as what I was familiar with whereas a script executing w/ the right context/permissions had the output in JSON.)

PR #2386, opened 2023-01-27
fix(scripts): fixed falco-driver-loader kmod cleanup

FedeDP:
**What type of PR is this?**

/kind bug

Properly cleanup leftover kmod files under /lib/modules.

Fixes #

```release-note
fix(scripts): properly cleanup all kmod leftover instances
```

### DIFF
--- a/scripts/falco-driver-loader
+++ b/scripts/falco-driver-loader
@@ -394,7 +394,29 @@ clean_kernel_module() {
 			echo "[WARNING] Removing '${KMOD_NAME}' version '${CURRENT_VER}' failed."
 		fi
 	done
-
+	
+	# While modinfo sees falco modules installed,
+	# clean them up.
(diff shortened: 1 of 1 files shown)


leogr (commented):
As discussed privately, the root cause should be the usage of `modprobe` introduced by https://github.com/falcosecurity/falco/pull/2242/files#diff-7adb35e11672e18c2abdb4b2a7a9e24758301c4080499fa6b47379710b10bd49.

IIRC, in the past, `modprobe` was removed in favor of `insmod` because the latter ensures that the correct build of the module will be loaded.

My proposal is to revert that change, and use only locally built drivers for deb and rpm packages. In this way, the kmod is only tracked by dkms, and no other moving parts will be involved (as it was before).

FedeDP:
Fully agree with you @leogr ! I am going to revert some of the changes you highlighted! 
Thank you!
/close

PR #2398, opened 2023-02-02
fix(scripts): make /usr writable

therealbobo:
Signed-off-by: Roberto Scolaro <roberto.scolaro21@gmail.com>

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind release

> /area build

> /area engine

> /area tests

> /area proposals

> /area CI

`ProtectSystem` directive makes `/usr` directory read-only. By the way `falcoctl` needs write permission.

Fixes #

```release-note
NONE
```


scripts/systemd/falcoctl-artifact-follow.service
```diff
@@ -13,7 +13,6 @@ RestartSec=15s
 Restart=on-failure
 PrivateTmp=true
 NoNewPrivileges=yes
-ProtectSystem=true
```
FedeDP, line 16: Can we try `ReadWritePaths` instead?

FedeDP: It looks like `ReadWritePaths` was introduced in systemd 231: https://github.com/systemd/systemd/blob/main/NEWS#L9059

I think we should use `ReadWriteDirectories` that is kept for backward compatibility:
> The old names of these settings remain
          available for compatibility.

FedeDP: In falco-kmod.service too!

scripts/systemd/falco-kmod.service
```diff
@@ -19,7 +19,7 @@ NoNewPrivileges=yes
 ProtectHome=read-only
 ProtectSystem=full
 ProtectKernelTunables=true
-ReadWritePaths=/sys/module/falco
+ReadWriteDirectories=/sys/module/falco
```
FedeDP, line 22: Thanks!
@Andreagit97 we will need a quick test of the falco-kmod service after this change (from packages built by this pr ci)
This should work with super old versions of systemd as well (it was not working before 231).

FedeDP (approved):
Tested using CI packages. It works fine!
/approve

leogr (approved):

PR #2409, opened 2023-02-07
fix(cmake): fixed tag fetching fallback (that is indeed needed)

FedeDP:
**What type of PR is this?**

/kind bug

/area build

Fixes issue introduced by #2292 ; the fallback for the tag fetch was indeed not a real fallback :/

Fixes #

```release-note
NONE
```

### DIFF
--- a/cmake/modules/GetFalcoVersion.cmake
+++ b/cmake/modules/GetFalcoVersion.cmake
@@ -19,29 +19,35 @@ if(NOT FALCO_VERSION)
   # Try to obtain the exact git tag
   git_get_exact_tag(FALCO_TAG)
   if(NOT FALCO_TAG)
-      # Fetch current hash
-      get_git_head_revision(refspec FALCO_HASH)
-      if(NOT FALCO_HASH OR FALCO_HASH MATCHES "NOTFOUND$")
-        set(FALCO_VERSION "0.0.0")
-      else()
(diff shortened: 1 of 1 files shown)


cmake/modules/GetFalcoVersion.cmake
```diff
@@ -19,29 +19,36 @@ if(NOT FALCO_VERSION)
...
-      get_git_head_revision(refspec FALCO_HASH)
-      if(NOT FALCO_HASH OR FALCO_HASH MATCHES "NOTFOUND$")
-        set(FALCO_VERSION "0.0.0")
-      else()
-        # Obtain the closest tag
-        git_get_latest_tag(FALCO_LATEST_TAG)
-        if(NOT FALCO_LATEST_TAG OR FALCO_LATEST_TAG MATCHES "NOTFOUND$")
+      # Obtain the closest tag
+      git_describe(FALCO_VERSION "--always" "--tags" "--abbrev=7")
+      string(REGEX MATCH "^[0-9]+.[0-9]+.[0-9]+$" FALCO_TAG ${FALCO_VERSION})
```
FedeDP, line 24: Only account for `git describe --always --tags --abbrev=7` when we are on a tag, otherwise compute the current dev version using the new method.

Andreagit97 (approved):
/approve

PR #2413, opened 2023-02-09
new: driver selection in falco.yaml

therealbobo:
Signed-off-by: Roberto Scolaro <roberto.scolaro21@gmail.com>

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind release

> /area build

> /area engine

> /area tests

> /area proposals

> /area CI

It might be useful to configure the driver mode via the configuration file `falco.yaml`

Fixes #

```release-note
new: driver selection in falco.yaml
```


Andreagit97:
Thank you for this! @therealbobo 
We discussed many times to uniform how we open engines and this could be a possible way 
cc @leogr

therealbobo:
I think that driver selected in `falco.yaml` should be considered if and only if the driver to be used is not specified in any other way (e.g. env variable or args). Moreover, using the config file, it will possibile to simplify the service management. 🤔

FedeDP:
We actually need this for the shiny new falco-driver-loader integrated inside falcoctl: https://github.com/falcosecurity/falcoctl/issues/327

FedeDP:
I'd suggest a rebase, and then:
* DEPRECATION NOTICE for `FALCO_BPF_PROBE`
* DEPRECATION NOTICE for cmd line options (note that `-u` was already dropped in the meantime, so i think that only the `modern-bpf` related one survives today)

Imho config option should always take precedence over deprecated cmdline/env var, since otherwise falcoctl would not be able to correctly instrument the right driver for Falco.

Andreagit97:
> I'd suggest a rebase, and then:
> 
>     * DEPRECATION NOTICE for `FALCO_BPF_PROBE`
> 
>     * DEPRECATION NOTICE for cmd line options (note that `-u` was already dropped in the meantime, so i think that only the `modern-bpf` related one survives today)
> 
> 
> Imho config option should always take precedence over deprecated cmdline/env var, since otherwise falcoctl would not be able to correctly instrument the right driver for Falco.
> 
> /cc @leogr

I agree with the proposed changes. Since we are here, we could evaluate if we should put also `gvisor` and `no-driver` in this config, in this way we have a knob for all possible engines. Moreover we need to find a way to specify additional config for each drivers, for example the bpf probe path for the `bpf` option

FedeDP:
> Moreover we need to find a way to specify additional config for each drivers, for example the bpf probe path for the bpf option

Yep; we already have the `modern_bpf` options block; we might do the same for the other, ie: support `kmod:` and `bpf:`.
It is simple yet effective, since one can specify options for all drivers and then switch between them by just updating the `driver.type` option, without further changes to eg (non-existent): `driver.options`.

userspace/falco/configuration.cpp
```diff
@@ -105,8 +106,28 @@ void falco_configuration::init(const std::string& conf_filename, const std::vect
 	load_yaml(conf_filename, config);
 }
 
+static driver_mode_type get_driver_mode(const std::string& input){
+	// Set driver mode if not already set.
+	const std::unordered_map<std::string, driver_mode_type> driver_mode_lut = {
+		{"kmod",driver_mode_type::KMOD},
+		{"bpf",driver_mode_type::BPF},
+		{"modern_bpf",driver_mode_type::MODERN_BPF},
```
FedeDP, line 114: I'd call this one `modern-bpf`. 
That is how it is used today in rpm/deb packages scripts, and this is what i will use for https://github.com/falcosecurity/falcoctl/pull/343.
Also, the systemd unit is called `modern-bpf` and the CLI flag too.

falco.yaml
```diff
@@ -773,6 +775,39 @@ metrics:
...
+# - `gvisor`: gVisor (gVisor sandbox)
+# - `custom`: Custom Driver (Specify a custom driver module)
+
+# Example usage:
+# driver_mode: kmod
+
+# Select the appropriate driver mode by uncommenting the corresponding line.
+# Make sure to specify only one driver mode at a time.
+
+driver_mode: kmod
```
leogr, line 805:
```suggestion
driver: 
   kind: kmod
```

I'd prefer to have the ability to include suboptions to `driver` to avoid polluting the config with first-level keys. For example:

```yaml
driver:
  kind: ebpf
  ebpf:
    # Path where the eBPF probe is located. 
    path: /root/.falco/
  modern_ebpf:
    cpus_for_each_syscall_buffer: 2
```

This will mimic [the Falco Helm chart config](https://github.com/falcosecurity/charts/blob/master/charts/falco/values.yaml#L177), which is fine IMO
 
**N.B. re. naming convention**
I believe we als have to fix the naming convention of this stuff for the last time. My proposal is as follows and include some alias to avoid headaches:

- `nodriver` (alias: `no-driver`)
- `module` (alias: `kmod`)
- `ebpf` (alias: `bpf`)
- `modern-ebpf`: (alias: `modern-bpf`)
- `gvisor` (which is technically not a driver, but I don't think we have a better option atm)

Andreagit97: i tend to agree with this design and naming, we should follow the same approach for helm charts and systemd units

therealbobo: I really like it! I'll update it!

Andreagit97: when we decide something definitive, we should update this issue https://github.com/falcosecurity/falco/issues/2840: "deprecation" in Falco 0.37 means "breaking change" in Falco 0.38, so I would directly update this issue

FedeDP: I agree with the design; about the naming, i'd go with:
* `nodriver` alias `no-driver`
* `module` alias `kmod`
* `bpf`
* `modern-bpf`
* `gvisor`

We never used `ebpf` nor `modern-ebpf` anywhere afaik. We have `FALCO_BPF_PROBE` and `--modern-bpf` today in Falco, and i'd keep them.

Falcoctl will use: `kmod`, `bpf`, `modern-bpf`.
Once again remember the naming of our systemd services too (https://github.com/falcosecurity/falco/tree/master/scripts/systemd)!

Andreagit97: we `ebpf` in the helm chart and we tend to use `ebpf` instead of `bpf` in all the documentation https://falco.org/tags/ebpf/ because technically it is more appropriate... but I have no strong opinion here we just need to pick one

FedeDP: Oh i see; thanks for pointing that out!

falco.yaml
```diff
@@ -773,6 +775,39 @@ metrics:
...
+# Falco with your system. Choose the appropriate driver mode based on your
+# system's configuration and requirements.
+#
+# Available driver modes:
+# - `kmod`: Kernel Module (Kernel Module)
+# - `bpf`: eBPF (Extended Berkeley Packet Filter)
+# - `modern_bpf`: Modern eBPF (Modern Extended Berkeley Packet Filter)
+# - `nodriver`: No Driver (No driver, just for testing)
+# - `gvisor`: gVisor (gVisor sandbox)
+# - `custom`: Custom Driver (Specify a custom driver module)
```
Andreagit97, line 797: not sure about the usage of this `custom` what would be the final use case?

falco.yaml
```diff
@@ -773,6 +775,39 @@ metrics:
...
+# Falco supports different driver modes for capturing syscall events. The choice
+# of driver mode can significantly impact the performance and compatibility of
+# Falco with your system. Choose the appropriate driver mode based on your
+# system's configuration and requirements.
+#
+# Available driver modes:
+# - `kmod`: Kernel Module (Kernel Module)
+# - `bpf`: eBPF (Extended Berkeley Packet Filter)
+# - `modern_bpf`: Modern eBPF (Modern Extended Berkeley Packet Filter)
+# - `nodriver`: No Driver (No driver, just for testing)
```
Andreagit97, line 795: I would probably rename the `nodriver` in something else. Today the `nodriver` is a workaround to load plugins with `syscall` source, but to the final user this is not clear using a word like `nodriver`, probably something like `syscall_plugin` would be more meaningful.

falco.yaml
```diff
@@ -773,6 +775,39 @@ metrics:
...
+# [Stable] `driver_mode`
+#
+# --- [Description]
+#
+# Falco supports different driver modes for capturing syscall events. The choice
+# of driver mode can significantly impact the performance and compatibility of
+# Falco with your system. Choose the appropriate driver mode based on your
+# system's configuration and requirements.
+#
+# Available driver modes:
```
Andreagit97, line 791: I would avoid the word `driver`. Moreover, in this list we still miss:
* the capability to read `scap-file` so something like ` mode: replay`
* the capability to open a source plugin, but this is not so easy since we have also the `load_plugin` config

falco.yaml
leogr: Agree. I'd avoid introducing the `custom` option.

falco.yaml
leogr: The _no driver_ option is not related to plugins. It can be helpful in case of testing or running Falco in CI.
If we believe it is misleading, we can choose a more appropriate label, still expressing that Falco will not rely on a driver. 

For example, `none` might be a good alternative:

```
driver:
  kind: none
```

Alternatively, we can use a different config key:

```
driver:
  enabled: false
```

falco.yaml
leogr: I'm not totally convinced about this. The `driver` word expresses the need for an external running component to the instrument to collect the data. On the other hand, a plugin is a component that runs together with `falco` in the same process. A scap file is an entirely different use case. Moreover, I believe the `replay` mode should only be available from the command line as a flag (why one should configure it in `falco.yaml` :thinking: ?).

 Likely, your proposal tried to mimic the _scap open_ behavior. However, we should refrain from exposing how scap open works to the end users and keep the _driver_ distinct from other modes (as it is now), IMO.

falco.yaml
leogr: I'm ok with picking `ebpf` as the standard one. Also, I don't see any issue with maintaining aliases.

falco.yaml
Andreagit97: > The no driver option is not related to plugins. It can be helpful in case of testing or running Falco in CI.

Looking at the code it seems also related to source plugins with "syscall" source

```cpp
		else if (s.options.nodriver) /* nodriver engine. */
		{
			// when opening a capture with no driver, Falco will first check
			// if a plugin is capable of generating raw events from the libscap
			// event table (including system events), and if none is found it
			// will use the nodriver engine.
			for (const auto& p: inspector->get_plugin_manager()->plugins())
			{
				if (p->caps() & CAP_SOURCING && p->id() == 0)
				{
					auto cfg = s.plugin_configs.at(p->name());
					falco_logger::log(LOG_INFO, "Opening '" + source + "' source with plugin '" + cfg->m_name + "'");
					inspector->open_plugin(cfg->m_name, cfg->m_open_params);
					return run_result::ok();
				}
			}
			falco_logger::log(LOG_INFO, "Opening '" + source + "' source with no driver\n");
			inspector->open_nodriver();
		}
```

jasondellaluce: The no-driver option related to plugins implementing the `syscall` raw event source. Since Falco never had a per-plugin selector for opening a capture, but rather a per-source one, the logic has always been as follows:
- When enabling the `syscall` source, it is opened by using the selected driver
- When enabling any non-`syscall` source, it sis opened by using the loaded plugin implementing it
- When having a event-sourcing plugin implementing the `syscall` source itself, the priority is still given to the selected driver for non-ambiguity and backward compatibility with the previously existing behavior. In such case, if the selected driver is `no-driver`, then the plugin is used for opening the `syscall` event source. This was introduced in Falco 0.35.0

leogr: @jasondellaluce @Andreagit97 

My point was that `--no-driver` itself is not strictly related to plugins. On the other hand, it's related to `syscall` since the `syscall` data source needed a driver historically. The fact that we need `nodriver` for loading the event-sourcing plugin implementing the syscall source is the consequence of maintaining the backward compatibility, but we can change this behavior in the future, can't we? :thinking:

Moreover, there is no actual implementation of a syscall plugin, while `nodriver` for the CI purpose is a compelling use case currently [in use](https://github.com/falcosecurity/charts/blob/master/tests/falco-test-ci.yaml#L14C8-L14C8).


So I still think the `nodriver` option is a compelling option that deserves to exist regardless of the plugin use case. Let me know if you disagree!

falco.yaml
FedeDP: I'd say: `source` instead of `driver`, and `source.mode: plugin` and deprecate `load_plugin` in next release; for now it will be supported along side (ie: `load_plugin` would override the `source.mode`.

FedeDP:
TODO: once this is merged, we will need to update https://github.com/falcosecurity/falco/issues/2840 with new DEPRECATED options.

userspace/falco/app/actions/helpers_inspector.cpp
FedeDP, line 1: In this file we will always use config values, since CLI options have already been merged and stored in relevant config structures by `apply_deprecated_options` below.

userspace/falco/app/actions/load_config.cpp
```diff
@@ -18,15 +18,42 @@ limitations under the License.
...
 
-// applies legacy/in-deprecation options to the current config
-static void apply_deprecated_options(
-		const falco::app::options& opts,
-		const std::shared_ptr<falco_configuration>& cfg)
+// applies legacy/in-deprecation options to the current state
+static falco::app::run_result apply_deprecated_options(falco::app::state& s)
 {
-	// Keep for future use cases.
+	// If overridden from CLI options (soon to be removed),
```
FedeDP, line 30: Here we are merging CLI options into relevant config structures.
This way, we can trust `config` values.

userspace/falco/configuration.cpp
```diff
@@ -380,11 +443,14 @@ void falco_configuration::load_yaml(const std::string& config_name, const yaml_h
 	/* We put this value in the configuration file because in this way we can change the dimension at every reload.
 	 * The default value is `4` -> 8 MB.
 	 */
-	m_syscall_buf_size_preset = config.get_scalar<uint16_t>("syscall_buf_size_preset", 4);
-
-	m_cpus_for_each_syscall_buffer = config.get_scalar<uint16_t>("modern_bpf.cpus_for_each_syscall_buffer", 2);
-
-	m_syscall_drop_failed_exit = config.get_scalar<bool>("syscall_drop_failed_exit", false);
+	// TODO: remove in Falco 0.38 since they are deprecated.
+	m_kmod.m_buf_size_preset = config.get_scalar<uint16_t>("syscall_buf_size_preset", 4);
```
FedeDP, line 447: We load `m_buf_size_preset` and `m_drop_failed_exit` 3 times, one for each supporting driver; this is just to avoid a switch on `m_driver_mode`.
It will be dropped in Falco 0.38 anyway.

FedeDP:
This PR is quite ready; we are just choosing a better name than `driver` for the new config section :)

FedeDP:
Rebased on top of master.

FedeDP:
Renamed `driver.` to `engine.` in config; morever, renamed `engine.replay.scap_file` to `engine.replay.trace_file`.
This is ready.
/cc @leogr

userspace/falco/app/actions/configure_syscall_buffer_num.cpp
```diff
@@ -34,10 +34,10 @@ falco::app::run_result falco::app::actions::configure_syscall_buffer_num(falco::
 		return run_result::fatal("cannot get the number of online CPUs from the system\n");
 	}
 
-	if(s.config->m_cpus_for_each_syscall_buffer > online_cpus)
+	if(s.config->m_modern_bpf.m_cpus_for_each_syscall_buffer > online_cpus)
```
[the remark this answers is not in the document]
Andreagit97: I should have solved all these issues in the commits I added :)

FedeDP:
Great catch Andre, thanks!

Andreagit97 (dismissed):
This LGTM!
/approve

Andreagit97 (commented):
LGMT!

userspace/falco/app/actions/helpers_inspector.cpp
```diff
@@ -20,30 +20,28 @@ limitations under the License.
...
 using namespace falco::app::actions;
 
 falco::app::run_result falco::app::actions::open_offline_inspector(falco::app::state& s)
 {
 	try
 	{
-		s.offline_inspector->open_savefile(s.options.trace_filename);
-		falco_logger::log(falco_logger::level::INFO, "Reading system call events from file: " + s.options.trace_filename + "\n");
+		s.offline_inspector->open_savefile(s.config->m_replay.m_capture_file);
+		falco_logger::log(falco_logger::level::INFO, "Reading system call events from file: " + s.config->m_replay.m_capture_file + "\n");
```
leogr, line 39:
```suggestion
		falco_logger::log(falco_logger::level::INFO, "Replaying events from the capture file: " + s.config->m_replay.m_capture_file + "\n");
```

N.B. _capture files_ are not tied to system call

userspace/falco/app/actions/load_config.cpp
```diff
@@ -18,15 +18,67 @@ limitations under the License.
...
+static falco::app::run_result apply_deprecated_options(falco::app::state& s)
 {
-	// Keep for future use cases.
+	// Please note: is not possible to mix command line options and configs to obtain a configuration
+	// we need to use only one method. For example, is not possible to set the gvisor-config through
+	// the command line and the gvisor-root through the config file. For this reason, if we detect
+	// at least one change in the default config we don't allow to use the command line options.
+	if(s.config->m_changes_in_engine_config)
+	{
+		return run_result::ok();
```
leogr, line 36: I'm fine with this behavior. However, I don't like silently ignoring deprecated options here (users may not get what's happening).

So, when the user mistakenly uses _BOTH_ the command line flag _AND_ change the config, we should print something like:

_The '-g,--gvisor-config' command line option is deprecated and will be removed in Falco version 0.38. Note that this option is ineffective when 'engine' settings are specified via the configuration file._

userspace/falco/app/actions/load_config.cpp
```diff
@@ -18,15 +18,67 @@ limitations under the License.
...
+	{
+		falco_logger::log(falco_logger::level::WARNING, "DEPRECATION NOTICE: the '--modern-bpf' cmdline option is deprecated and will be removed in Falco 0.38!\n");
+		s.config->m_engine_mode = engine_kind_t::MODERN_EBPF;
+		s.config->m_modern_ebpf.m_drop_failed_exit = s.config->m_syscall_drop_failed_exit;
+		s.config->m_modern_ebpf.m_buf_size_preset = s.config->m_syscall_buf_size_preset;
+		s.config->m_modern_ebpf.m_cpus_for_each_buffer = s.config->m_cpus_for_each_syscall_buffer;
+	}
+	if (!s.options.gvisor_config.empty())
+	{
+		falco_logger::log(falco_logger::level::WARNING, "DEPRECATION NOTICE: the '-g,--gvisor-config' cmdline option is deprecated and will be removed in Falco 0.38!\n");
```
leogr, line 65:
```suggestion
		falco_logger::log(falco_logger::level::WARNING, "DEPRECATION NOTICE: the '-g,--gvisor-config' command line option is deprecated and will be removed in Falco 0.38!\n");
```

just nit :)

userspace/falco/app/actions/load_config.cpp
```diff
@@ -18,15 +18,67 @@ limitations under the License.
...
+
+	// Replace the kmod default values in case the engine was open with the kmod.
+	// We don't have a command line option to open the kmod so we have to always enforce the
+	// default values.
+	s.config->m_kmod.m_drop_failed_exit = s.config->m_syscall_drop_failed_exit;
+	s.config->m_kmod.m_buf_size_preset = s.config->m_syscall_buf_size_preset;
+
+	// If overridden from CLI options (soon to be removed),
+	// use the requested driver.
+	if (getenv(FALCO_BPF_ENV_VARIABLE))
```
leogr, line 47: What will happen when [this](https://github.com/falcosecurity/falco/pull/2918) gets merged, and perhaps some of the users will use the same env var in the config :point_down: ? 

```yaml
engine:
  kind: kmod
  kmod:
    buf_size_preset: 4
    drop_failed_exit: false
  ebpf:
    # path to the elf file to load.
    probe: ${FALCO_BPF_PROBE}
```

:exploding_head:

userspace/falco/configuration.cpp
```diff
@@ -105,8 +102,101 @@ void falco_configuration::init(const std::string& conf_filename, const std::vect
...
+		{"ebpf",engine_kind_t::EBPF},
+		{"modern-ebpf",engine_kind_t::MODERN_EBPF},
+		{"replay",engine_kind_t::REPLAY},
+		{"gvisor",engine_kind_t::GVISOR},
+		{"none",engine_kind_t::NONE},
+	};
+
+	constexpr int16_t default_buf_size_preset = 4;
+	constexpr int16_t default_cpus_for_each_syscall_buffer = 2;
+	constexpr bool default_drop_failed_exit = false;
```
leogr, line 119: These defaults are a bit hidden. However, not a big issue.

userspace/falco/configuration.h
```diff
@@ -37,6 +37,16 @@ limitations under the License.
...
 #include "falco_outputs.h"
 
+enum class engine_kind_t : uint8_t
+{
+	KMOD,
+	EBPF,
+	MODERN_EBPF,
+	REPLAY,
+	GVISOR,
+	NONE
```
leogr, line 47: Do you know if this is a particular reason for this ordering?

Not saying it is an issue, just I don't get it.

leogr (commented):
Love this. It's a huge improvement, and we will finally get a bit of consistency across our configuration. So, first of all, a big thanks to @therealbobo, @Andreagit97, and @FedeDP :clap: 

The implementation is already good. I've just left some comments about minor issues I've found when reviewing.

Other than that, I want to ensure that a warning message is printed to the user any time a deprecated option is used (either command line or config). I've already identified a corner case (see my comment below) and I think another similar corner case can happen if the user modified both the old and the new field in the configuration file.

Ping me privately, I can provide more details if needed.

userspace/falco/configuration.h
FedeDP: No special reason!

userspace/falco/configuration.cpp
FedeDP: Yep, they will go away in next release cycle.

userspace/falco/app/actions/load_config.cpp
FedeDP: I'd add a single log, something like:
```
if(s.config->m_changes_in_engine_config)
	{
                   falco_logger::log(falco_logger::level::WARNING, "Since you are using the new engine configuration key, deprecated command line options [-e,-g,...] and FALCO_BPF_PROBE env variable will be ignored.\n");
		return run_result::ok(); 
```

falco.yaml
```diff
@@ -145,6 +147,196 @@ rules_file:
...
+    probe: /path/to/probe.o
+    buf_size_preset: 4
+    drop_failed_exit: false
+  modern-ebpf:
+    cpus_for_each_buffer: 2
+    buf_size_preset: 4
+    drop_failed_exit: false
+  replay:
+    # path to the capture file to replay.
+    capture_file: /path/to/file.scap
```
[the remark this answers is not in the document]
FedeDP: > I would make Falco output a clear error message instructing the user to pass a value to replay.capture_file

There is already one. Agree about the default!

falco.yaml
```diff
@@ -145,6 +147,196 @@ rules_file:
...
+  gvisor:
+    # A Falco-compatible configuration file can be generated with
+    # '--gvisor-generate-config' and utilized for both runsc and Falco.
+    config: /path/to/gvisor_config.yaml
```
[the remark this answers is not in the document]
FedeDP: I'd leave it without any example, we can iterate it further in a follow up PR if needed.

userspace/falco/app/actions/load_config.cpp
FedeDP: Added in latest commit; example output:
```
Thu Nov 23 15:15:50 2023: Falco version: 0.37.0-146+9bf0565 (x86_64)
Thu Nov 23 15:15:50 2023: Falco initialized with configuration file: ../falco.yaml
Thu Nov 23 15:15:50 2023: Since the new 'engine' config key is being used, deprecated CLI options [-e,-g,--gvisor-config,--nodriver,--modern-bpf] and FALCO_BPF_PROBE environment variable will be ignored.
Thu Nov 23 15:15:50 2023: Loading rules from file ../rules/falco_rules.yaml
```

userspace/falco/app/actions/load_config.cpp
leogr: > Nothing, i mean the deprecated env variable takes precedence.

Will it print a warning even if it shouldn't? :thinking:

FedeDP: Yes it will print the newly introduced warning about the fact that FALCO_BPF_PROBE env variable will be ignored, and that is correct(since we won't directly use the env variable but the new config key that resolves to the env variable value)

userspace/falco/app/actions/load_config.cpp
leogr: Ok. I was thinking of printing the warning only if one of those options was used. But your solution is fine anyway :+1:

userspace/falco/app/actions/load_config.cpp
FedeDP: Oh i mis-read (i thought `kind: bpf` was in use). In this case, as i stated earlier, the deprecated env variable takes precedence (since the `bpf` kind configuration is not even parsed, because the `kind` is kmod).
No warning in this case (except for the `FALCO_BPF_PROBE` env var deprecation )

LucaGuerra (approved):
Thank you for this! 😮 LGTM, so we can test it better

leogr (approved):
For sure we need to deeply test it, but LGTM at the moment!

Thank you'll!

PR #2414, opened 2023-02-09
refactor(userspace/falco): turn app actions in simple functions

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind design

/area engine

This PRs **proposes** turning all the Falco application actions into simple stateless functions. There is a standalone struct defining the application's state (`falco::app::state`), and each app action is a simple function in the form `(state) => run_result`.

The motives for this changes are:
- The class-based application definition started as a great solution but has quickly grown. For example, this caused the `application.h` header to grow in a much less-readable state, and made the code sticky due to the need of adding a new private method everytime a new helper or funcionality needed to be introduced
- There is no place for helper functions. Everytime we need to share some piece of logic, we just add a new private method to the application class instead of creating a new standalone and reusable piece of source code
- Being the code a bit sticky due to everything being defined in the same class, the amount of LOC changes and the odds of conflicting PRs requiring to be rebased recently became quite higher
- The class-based approach turned out not being very testable due to the action flow being dictated by the class itself. The mix of private and public methods and fields made the visibility of many parts of code harder. With the new function-based model, writing a test for an action is just a matter of: 1) setting state pre-conditions, 2) run the action, and 3) checking state post-conditions.
- Due to the new-ly introduced continuous-delivery features of falcoctl, we may want to introduce a sort of "dry-run" feature in Falco that allows pre-checking new configuration changes before an hot-reload. Due to the points above, the class-based approach would make such a feature harder to implement

This is a huge no-op from a functional standpoint. I'm aware the changes are a lot of LOC, but I can ensure that 90%+ of this is just moving pieces of code around. The git blame has been preserved wherever possible.

```release-note
refactor(userspace/falco): turn app actions in simple functions
```


userspace/falco/app/signals.h
```diff
@@ -0,0 +1,56 @@
...
+#include <functional>
+
+#define APP_SIGNAL_NOT_SET          0   // The signal flag is not set
+#define APP_SIGNAL_SET              1   // The signal flag has been set
+#define APP_SIGNAL_ACTION_TAKEN     2   // The signal flag has been set and the application took action
+
+namespace falco {
+namespace app {
+
+// todo(jasondellaluce): hide this into a class
```
jasondellaluce, line 29: I know this piece is a bit ugly now that it's out of a class. I already have a follow-up PR ready to make signal handlers more polished  and stable, to open after/if we go through with this one.

userspace/falco/falco.cpp
```diff
@@ -42,26 +42,11 @@ static void display_fatal_err(const std::string &&msg)
 //
 int falco_init(int argc, char **argv, bool &restart)
```
FedeDP, line 43: WDYT about renaming this to `falco_loop` or `falco_run`?

FedeDP (commented):
I am pretty much on board with the changes; i really like the new tree, namespace and all in all, i think this PR improves code readability and followability a lot!

userspace/falco/falco.cpp
jasondellaluce: Agreed! 👍

FedeDP (dismissed):
/approve

Andreagit97 (approved):
/approve

PR #2415, opened 2023-02-09
cleanup(userspace/falco): do not enter dropping mode.

FedeDP:
**What type of PR is this?**

/kind cleanup

When not run in `-A` mode, Falco will start dropping mode.
Dropping mode enables following logic in drivers:
* parameter passed to `start_dropping_mode` is the "sampling_ratio", ie: how much a second is "sampled"; if you pass `1` (what we are doing in Falco), you are saying that you want each second to be fully sampled; if you pass `2`, it means you want to sample each 0.5s, and so on.
* So, Falco is passing `1` basically disabling the sampling logic (that can be found here: https://github.com/falcosecurity/libs/blob/master/driver/main.c#L1695)
* instead, it is just enabling the dropping logic: https://github.com/falcosecurity/libs/blob/master/driver/main.c#L1677
* what it does is: drop all `UF_ALWAYS_DROP` events + drop some more events, by using following logic: https://github.com/falcosecurity/libs/blob/master/driver/main.c#L1607

Considering that nowadays all `UF_ALWAYS_DROP` events are already dropped by syscalls of interest when not in `-A` mode (and we are looking to expand current behavior with adaptive syscalls), and that the `drop_nostate_event` dropping logic should not have a big impact on perf, i think we can entirely skip all this (weird) logic altogether.
Note, moreover, that the `drop_nostate_event` logic could possibly be fully superseeded by https://github.com/falcosecurity/libs/pull/834 (that would of course be a opt-in feature).

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/falco/app_actions/open_inspector.cpp
+++ b/userspace/falco/app_actions/open_inspector.cpp
@@ -123,11 +123,5 @@ application::run_result application::open_live_inspector(
 		return run_result::fatal(e.what());
 	}
 
-	// This must be done after the open
-	if (!m_options.all_events)
-	{
-		inspector->start_dropping_mode(1);
-	}
-
 	return run_result::ok();
 }


Andreagit97 (approved):
/approve

jasondellaluce (approved):
/approve

PR #2416, opened 2023-02-09
Refactor Falco unit tests

Andreagit97:
**What type of PR is this?**

/kind cleanup

/area tests

This PR introduces `gtest` in our unit test framework. This simplifies a little bit the building phase and allows us to use the more intuitive `gtest` framework to write our Falco unit tests

```release-note
NONE
```


FedeDP:
Really nice! Thank you @Andreagit97 !! Great work, as always :)

Andreagit97:
for the unit tests the `MINIMAL` build is enough. We run them in GitHub actions so there is no reason to run them also in CircleCI while building Falco...

FedeDP (approved):
I love the effort and i love the cmake modules cleanups :) 
/approve

unit_tests/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
...
+    PRIVATE
+    ${CMAKE_SOURCE_DIR}/userspace
+    ${CMAKE_BINARY_DIR}/userspace/falco # we need it to include indirectly `config_falco.h` file
+    ${CMAKE_SOURCE_DIR}/userspace/engine # we need it to include indirectly `falco_common.h` file
+)
+
+set(FALCO_UNIT_TESTS_DEPENDENCIES
+  gtest
+  gtest_main
+  falco_engine
```
jasondellaluce, line 44: Reminder for ourselves: we don't need to address this in this PR, however we will not be able to test sources under the `userspace/falco` directory if we don't create a cmake library target for it. The reason why we're able to have `test_configuration.cpp` is that the falco config helpers are header-only.

My proposal for a future PR would be to have a new library target alongside `falco_engine`, and then link `falco.cpp` with both to have the final `falco` executable target. Would it make sense?

jasondellaluce (approved):
/approve

This is lovely Andrea, really a major and much-needed cleanup. Thanks!

PS: I left a minor comment as a reminder for one of the future directions of this.

Issue #2373, opened 2023-01-25
[New Feature]: User input option for additional syscalls overriding default sinsp state enforcement


incertum:
**Motivation**

Falco's sinsp state enforcement is currently not conditioned by Falco rules and may activate too many syscalls for a particular use case / Falco rules configurations. The dilemma is that there is never an unlimited compute overhead budget and the end user currently has no options to tailor Falco over configs in this regard. In addition, the adaptive syscalls activations behaviors may not suit some use cases (even when improved). The reason why the syscalls activations mechanisms are so important for CPU and memory utilizations is that the more syscalls are activated, the more work has to be done in the kernel as well as in userspace (processing and applying filters). This is especially true for more high volume syscalls.

**Feature**

Add an option to put the end user in the driver seat aka empower the end user to 100% select all syscalls. Initial discussions started in https://github.com/falcosecurity/falco/pull/2361 as well as here https://github.com/falcosecurity/libs/pull/826 and initial co-authored code changes (@FedeDP @Happy-Dude) are preserved here https://github.com/incertum/falco/tree/archive1-rules-evttypes-push-down, containing a bunch of text and suggestions in the modified falco.yaml

FedeDP:
I was thinking about the name; perhaps `base_syscalls` is much more relatable?
Btw thanks for opening this one, to track it and have a discussion!

@leogr @jasondellaluce

incertum:
Really liking `base_syscalls`, was trying to think of alternatives, but nothing better came to mind. It's a great choice and would make the explanations more clear from a technical perspective.

Few more questions before opening the new PR after the other first PR is merged:
@jasondellaluce mentioned Falco shall not error out on checks, so what are our fall-back conditions? 
   - all strings in  `base_syscalls` are invalid
   - `base_syscalls` only contains exclude syscalls
Fall-back to adaptive syscalls in those cases or let it just pass as this option is "use at won risk" with clear suggestion on how to use it while knowing the risks.

PR #2428, opened 2023-02-21
new(app_actions): introduce base_syscalls user option

incertum:
**What type of PR is this?**

> /kind bug

/kind cleanup

/kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind release

> /area build

/area engine

> /area tests

> /area proposals

> /area CI

See https://github.com/falcosecurity/falco/issues/2373 for context.

Fixes #

```release-note
new(app_actions): introduce base_syscalls user option
```

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -442,3 +442,68 @@ metadata_download:
   max_mb: 100
   chunk_wait_us: 1000
   watch_freq_sec: 1
+
+
+# base_syscalls ! Use with caution !
+#
+# --- [Description]
+#
+# With this option you are in full control of the total set of syscalls that
+# Falco will enable in the kernel for active tracing.
+
+# All syscalls and events from each enabled Falco rule will automatically be activated
+# even when choosing this option. This option provides full end user control to specifically
+# define a static set of base syscalls that will be activated in addition to the
+# syscalls defined in the rules.
+#
+# When using this option, Falco does not add any other syscalls that may be needed for
+# Falco's state engine. The union of all syscalls from the rules (including resolved macros)
+# and the ones specified here compose the final set of syscalls that are traced in the kernel.
+# This puts the end user in the driver seat, but if not used correctly Falco logs may be
+# incomplete or wrong. This option however can be very useful to lower CPU utilization and
+# allowing you to tailor Falco to specific environments according to your
+# organization's threat model and security posture as well as cost budget.
+
+# !!! When NOT using this option, Falco defaults to adding a static set of syscalls in addition
+# to the rules system calls you need for Falco's state engine build-up and life-cycle management.
+#
+#
+# --- [Usage]
+#
+# List of system calls names (<syscall-name>) plus negative ("!<syscall-name>") notation supported.
+#
+# base_syscalls: [<syscall-name>, <syscall-name>, "!<syscall-name>"]
+#
+#
+# --- [Suggestions]
+#
+# Here are a few recommendations that may help you to use the full power of this option:
+#
+# Consider to at minimum add the following syscalls regardless of the syscalls used in the rules.
+#
+# [clone, clone3, fork, vfork, execve, execveat, close]
+#
+# This is because some Falco fields you may output for an execve* system call are retrieved
+# from the associated "clone", "clone3", "fork", "vfork" syscall when spawning a new process.
+# The "close" system call is used to purge file descriptors from Falco's internal
+# thread / process cache table and therefore should always be added when you have rules around fds
+# (e.g. open, openat, openat2, socket, connect, accept, accept4 ... and many more)
+#
+# When network syscalls are used in rules we recommend to at minimum set
+#
+# [clone, clone3, fork, vfork, execve, execveat, close, socket, bind, getsockopt]
+#
+# It turns out that while you absolutely can log connect or accept* syscalls without the socket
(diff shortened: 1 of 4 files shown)


falco.yaml
```diff
@@ -442,3 +442,68 @@ metadata_download:
...
+# Lastly, if you care about the correct uid, gid or sid, pgid of a process when that process then
+# opens a file or makes a network connection or any other action, consider also
+# adding the following syscalls:
+#
+# setresuid, setsid, setuid, setgid, setpgid, setresgid, setsid, capset, chdir, chroot, fchdir
+#
+# Only exclude syscalls, e.g. "!mprotect" if you need a fast deployment update (overriding rules),
+# else rather remove unwanted or not needed syscalls from the Falco rules.
+
+base_syscalls: []
```
incertum, line 509: The research behind the guidance was happening here https://github.com/falcosecurity/libs/pull/826.

userspace/falco/app/actions/helpers_interesting_sets.cpp
```diff
@@ -75,15 +88,51 @@ static void select_event_set(falco::app::state& s, const libsinsp::events::set<p
...
+	if (!user_base_syscalls_names.empty())
+	{
+		auto valid_events = libsinsp::events::names_to_event_set(user_base_syscalls_names);
+		auto negative_names = extract_negative_base_syscalls_names(user_base_syscalls_names);
+		auto valid_negative_events = libsinsp::events::names_to_event_set(negative_names);
+
+		auto n_invalid_positive_names = (user_base_syscalls_names.size() - negative_names.size()) - libsinsp::events::names_to_sc_set(user_base_syscalls_names).size();
+		if (n_invalid_positive_names > 0)
+		{
+			std::cerr << "User config base_syscalls includes ("
```
incertum, line 101: We discussed to only emit light warnings if there is invalid user input, but there will be no fall-back. This option is 100% override.

userspace/falco/app/actions/helpers_interesting_sets.cpp
```diff
@@ -116,7 +160,15 @@ static void select_event_set(falco::app::state& s, const libsinsp::events::set<p
 
 static void select_syscall_set(falco::app::state& s, const libsinsp::events::set<ppm_event_code>& rules_event_set)
 {
-	s.selected_sc_set = libsinsp::events::event_set_to_sc_set(s.selected_event_set);	
+	s.selected_sc_set = libsinsp::events::event_set_to_sc_set(s.selected_event_set);
+
+	if (!s.selected_sc_set.empty())
```
incertum, line 165: This needs the outstanding fixes from https://github.com/falcosecurity/libs/pull/889.

jasondellaluce: I think the ones required here have been isolated in https://github.com/falcosecurity/libs/pull/921 and already merged.

userspace/falco/app/actions/helpers_interesting_sets.cpp
```diff
@@ -75,14 +88,45 @@ static void select_event_set(falco::app::state& s, const libsinsp::events::set<p
 	auto base_event_set = libsinsp::events::sinsp_state_event_set();
 	s.selected_event_set = rules_event_set.merge(base_event_set);
 
+	auto user_base_syscalls_names = s.config->m_base_syscalls;
+	if (!user_base_syscalls_names.empty())
+	{
+		auto valid_events = libsinsp::events::names_to_event_set(user_base_syscalls_names);
```
jasondellaluce, line 94: The way we get `valid_events` is the only thing that I'm not sure about.

- We should pass down to `libsinsp::events::names_to_event_set` a list of names without the `!xxx` negative ones. Even if the API ignores unknown names, I still think it's not a good practice
- If `base_syscall` is not empty, but contains only negative syscall names, then `valid_events` will be empty, which would cause the selected set to contain only the ruleset-defines sycalls. I don't think this is what we want.

To fix both, I think `extract_negative_base_syscalls_names` should be generalized to split `base_syscall` in two disjunct sets of positive and negative syscalls names. Then, the positive names are used to determine `valid_events`, and if empty will fallback to `libsinsp::events::sinsp_state_event_set()`. Then, all negative syscalls will be removed at the end by `s.selected_event_set` (like you already do).

What do you think?

jasondellaluce:
/milestone 0.35.0

Fixes #2373

incertum:
@jasondellaluce attempted to address each suggestion. Plus minor adjustments to ensure correct state_event_set for all configurations (my bad missed this during the first pass). Now also valid check_for_rules_unsupported_events for all configurations. Lastly, in order to make it easier to review removed base_syscalls validation warnings as this may need a separate discussion. Happy to re-introduce them in a better form in follow up PR.

Andreagit97 (approved):
/approve
thank you

jasondellaluce (approved):
/approve

I think we can proceed with this. Note that this is not blocked by the WIP changes of https://github.com/falcosecurity/libs/pull/889 for three main reasons:
- The PPM_SC selection is not affected by the information loss of old-version events, because live captures will always return the most up-to-date event versions
- The information loss of non-syscall events (meta events, internal events, plugin events, etc...) does not affect the event selection sent to the drivers for live captures
- Tracepoint selection is still handled differently and is not altered

PR #2430, opened 2023-02-22
update(userspace/falco): support dry-run option and implement safety checks before hot reloads

jasondellaluce:
**What type of PR is this?**

/kind feature

/area engine

This adds a new CLI option `--dry-run`, which basically runs all the Falco application steps but skips the event processing part. This can be used to do preliminary checks on Falco for a given configuration, rulesets, plugins, and so on.

Accordingly, this new capability has been leveraged to add safety checks before performing hot restarts. This happens either when Falco receives a SIGHUP or when one of the watched files changes (config file, rules files). When triggering an hot restart, Falco will now simulate a dry-run in parallel and will proceed on reloading only if no issue is encountered. For instance, this can help to avoid reloading Falco with a broken configuration. In case of failure in the dry-run safety checks, Falco will both log a message and send an `internal` alert in the configured outputs for higher visibility.

```release-note
update(userspace/falco): support dry-run option and implement safety checks before hot reloads
```


jasondellaluce:
Of course, as anything dealing with multi-threaded code, this will require a good bunch of testing. We need to check that all the start/stop conditions and resource usage (mem, cpu) are the same as now. Tried to stress it locally, and seems to work for now.

jasondellaluce:
This is ready for review! Only concerns of mine are:
- Do we need a debouncing logic before launching the dry run? I assume not, because we read inotify events all in a single buffer and not one-by-one, but I'm open to any suggestion.
- We need to make sure there is no additional CPU consumption in idle. There wasn't in my tests, but I would ask reviewers to test this and fuzz it for cross-checking.
- There is no deadlock scenario. There shouldn't be any, but human error is always a thing.

jasondellaluce:
> Do we need a debouncing logic before launching the dry run? I assume not, because we read inotify events all in a single buffer and not one-by-one, but I'm open to any suggestion.

This is off the list now, now the restarter handler supports debouncing subsequent events within a certain timeout.

FedeDP (approved):
/approve
I tested it locally, it works really well!

leogr (approved):
:+1:

PR #2432, opened 2023-02-24
update: improve support and tests for live-capture event selection

jasondellaluce:
### **What type of PR is this?**

/kind bug

/kind design

### **Any specific area of the project related to this PR?**

/area engine

/area tests

### **What this PR does / why we need it**:

This follows up the contributions of https://github.com/falcosecurity/falco/pull/2428 by fixing a design issue, adapting the codebase to the latest updates in libsinsp, and by growing the unit tests for the part of code responsible for event selection during live captures.

~400 lines of code of this PR are for tests.

### **Special notes for your reviewer**:

A lot changed this week, so here's quick recap for how Falco deals with events:
- When reading capture files, Falco will receive and match rules for all events in the files
- When opening a live capture for an event source that is not `syscall`, Falco will behave as usual
- When opening a live event capture for the `syscall` event source:
  1. Falco will determine a base set of events to collect. This can be the default minimum set that ensures state consistency and stability in libsinsp, or a custom set of events defined by the user through the `base_syscalls` configuration field.
  2. The final selected set of events is the union of the base set computed in the previous step, and the events matched by the loaded rules. This can now include generic events too (never supported up until now)
  3. If the `-A` flag is not set, performance-heavy syscalls are forcefully removed from the selected set (such as IO-intensive syscalls)
  4. Falco will configure the chosen kernel driver with the selected set of events, and will only receive events matching the set

**What was missing?** The current implementation incorrectly uses the ppm_event_code enumerative as single source of truth for defining the set of selected events. This has some faults, the first being that there can be information loss about generic events. Instead, Falco needs to operate with both **ppm_sc_code** and **ppm_event_code** for their different purposes:
- **ppm_event_code** is required for correctly matching events in the rulesets at runtime. In fact, Falco will receive events of all kinds (internal, container, plugin, old versions from capture files...), and needs to recognize and index them properly.
- **ppm_sc_code** is required to define the set of selected events to configure the kernel drivers with. `ppm_sc_code` is a richer enumerative that has no information loss about generic events and  includes syscalls-related events only. The unit test suite now checks that everything works as expected.

**What's still missing?** The work and design discussion around https://github.com/falcosecurity/libs/pull/889 will lead to fixing two remaining problems:
1. Information loss when converting a ppm_sc_code to a ppm_event_code and viceversa.
2. Have a unique representation for events related to tracepoints (e.g. `switch`, `procexit`) and all the rest of the syscall events.

None of this is blocking for Falco in its current feature development status. Point 1) has no impact because Falco does not rely on the libsinsp conversion utilities, but searches for ppm_sc_codes in the ruleset right on instead. Point 2) will require a minor refactor in Falco because we currently have tracepoint in a different selection sets, however this will only be non-functional changes.

### **Does this PR introduce a user-facing change?**:

```release-note
update: improve support and tests for live-capture event selection
```


userspace/falco/app/actions/configure_interesting_sets.cpp
```diff
@@ -0,0 +1,195 @@
...
+static void select_event_set(falco::app::state& s, const libsinsp::events::set<ppm_sc_code>& rules_sc_set)
+{
+	/* PPM syscall codes (sc) can be viewed as condensed libsinsp lookup table
+	 * to map a system call name to it's actual system syscall id (as defined
+	 * by the Linux kernel). Hence here we don't need syscall enter and exit distinction. */
+	auto rules_names = libsinsp::events::sc_set_to_names(rules_sc_set);
+	if (!rules_sc_set.empty())
+	{
+		falco_logger::log(LOG_DEBUG, "(" + std::to_string(rules_names.size())
+			+ ") events in rules: " + concat_set_in_order(rules_names) + "\n");
```
incertum, line 70: Need to change message back to "syscalls in rules" now, throughout.

userspace/falco/app/actions/configure_interesting_sets.cpp
```diff
@@ -0,0 +1,195 @@
...
+		// sure that bad user inputs are ignored
+		auto user_positive_sc_set_names = libsinsp::events::sc_set_to_names(user_positive_sc_set);
+		falco_logger::log(LOG_DEBUG, "+(" + std::to_string(user_positive_sc_set_names.size())
+			+ ") syscalls added (base_syscalls override): "
+			+ concat_set_in_order(user_positive_sc_set_names) + "\n");
+	}
+
+	// selected events are the union of the rules events set and the
+	// base events set (either the default or the user-defined one)
+	s.selected_sc_set = rules_sc_set.merge(base_sc_set);
```
incertum, line 107: I like how now that we reverted to starting from sc_set we could simplify and clean up "base_syscalls" right away and be more consistent 🎉

userspace/falco/app/actions/configure_interesting_sets.cpp
```diff
@@ -0,0 +1,195 @@
...
+	extract_base_syscalls_names(s.config->m_base_syscalls, user_positive_names, user_negative_names);
+	auto user_positive_sc_set = libsinsp::events::names_to_sc_set(user_positive_names);
+	auto user_negative_sc_set = libsinsp::events::names_to_sc_set(user_negative_names);
+
+	if (!user_positive_sc_set.empty())
+	{
+		// user overrides base event set
+		base_sc_set = user_positive_sc_set;
+
+		// we re-transform from events set to names to make
```
incertum, line 97: "re-transform from sc_set" ?

incertum (commented):
@jasondellaluce nice work! Given master is currently a tad flawed, would vote for merging rather faster and then polishing in a follow up PR if needed. WDYT?

userspace/falco/app/actions/configure_interesting_sets.cpp
```diff
@@ -0,0 +1,195 @@
...
+	if(!s.options.all_events)
+	{
+		auto ignored_sc_set = libsinsp::events::io_sc_set();
+		auto erased_sc_set = s.selected_sc_set.intersect(ignored_sc_set);
+		s.selected_sc_set = s.selected_sc_set.diff(ignored_sc_set);
+		if (!erased_sc_set.empty())
+		{
+			auto erased_sc_set_names = libsinsp::events::sc_set_to_names(erased_sc_set);
+			falco_logger::log(LOG_DEBUG, "-(" + std::to_string(erased_sc_set_names.size())
+				+ ") ignored events (-> activate via `-A` flag): "
```
incertum, line 148: ignored syscalls

userspace/falco/app/actions/configure_interesting_sets.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+	/* Derive the diff between the additional syscalls added via libsinsp state
+	enforcement and the syscalls from each Falco rule. We avoid printing
+	this in case the user specified a custom set of base syscalls */
+	auto non_rules_sc_set = s.selected_sc_set.diff(rules_sc_set);
+	if (!non_rules_sc_set.empty() && user_positive_sc_set.empty())
+	{
+		auto non_rules_sc_set_names = libsinsp::events::sc_set_to_names(non_rules_sc_set);
+		falco_logger::log(LOG_DEBUG, "+(" + std::to_string(non_rules_sc_set_names.size())
+			+ ") events (Falco's state engine set of events): "
```
incertum, line 130: ) syscalls

userspace/falco/app/actions/configure_interesting_sets.cpp
```diff
@@ -0,0 +1,195 @@
...
+	 * performance reasons. */
+	auto unsupported_sc_set = rules_sc_set.diff(s.selected_sc_set);
+	if (unsupported_sc_set.empty())
+	{
+		return;
+	}
+
+	/* Get the names of the events (syscall and non syscall events) that were not activated and print them. */
+	auto names = libsinsp::events::sc_set_to_names(unsupported_sc_set);
+	std::cerr << "Loaded rules match event types that are not activated or unsupported with current configuration: warning (unsupported-evttype): " + concat_set_in_order(names) << std::endl;
```
incertum, line 57: "Loaded rules match syscalls that are not activated (e.g. were removed via config settings such as no -A flag or negative base_syscalls elements) or unsupported with current configuration:"

... or similar ...

Also promised will add the new still outstanding warning messages around base_syscalls soon.

userspace/falco/app/actions/configure_interesting_sets.cpp
```diff
@@ -0,0 +1,195 @@
...
+	std::cerr << "If syscalls in rules include high volume I/O syscalls (-> activate via `-A` flag), else syscalls might be associated with syscalls undefined on your architecture (https://marcin.juszkiewicz.com.pl/download/tables/syscalls.html)" << std::endl;
```
incertum, line 58: ... ", else syscalls may have been removed via base_syscalls option or might be associated ..."

userspace/engine/evttype_index_ruleset.cpp
```diff
@@ -163,11 +170,15 @@ void evttype_index_ruleset::add(
 		wrap->filter = filter;
 		if(rule.source == falco_common::syscall_source)
 		{
-			wrap->evttypes = libsinsp::filter::ast::ppm_event_codes(condition.get());
+			wrap->sc_codes = libsinsp::filter::ast::ppm_sc_codes(condition.get());
```
incertum, line 173: Please see some more libs fixes in https://github.com/falcosecurity/libs/pull/915/commits/338bff09a5bbdd81cbe1a15ede48d77313af94ca that aim to make sure we handle the tough corner cases and special snow flakes. This is probably a perfect reflection of why we were going back and forth between first resolving the evt.type string names to event codes vs syscall codes ... meaning we need a combination of the two approaches after all.

incertum:
Just noticed that `selected_event_set` disappeared completely? Wouldn't we still need to activate all crucial non syscalls events for userspace? In addition, we could consider removing the static sinsp state enforcement from `libsinsp::events::sinsp_state_event_set()` again since we pivoted in the approach? https://github.com/falcosecurity/libs/blob/50aa2a0c9bb5f7e3c61dc0f2470f513521e34ec2/userspace/libsinsp/events/sinsp_events.cpp#L174

jasondellaluce:
> Just noticed that selected_event_set disappeared completely? Wouldn't we still need to activate all crucial non syscalls events for userspace?

This event selection process only concerns events that could be ignored at kernel-level, and for now we have no way to express the selection of non-syscall events (e.g. `container`) . This means that we'll keep receiving all of them even after these changes. Most of non-syscall events are meta-events or tracepoint-based events. The former are generated internally by libsinsp as part of its internal machinery, and the latter can be selected in the separate `selected_tp_set`.

incertum:
> > Just noticed that selected_event_set disappeared completely? Wouldn't we still need to activate all crucial non syscalls events for userspace?
> 
> This event selection process only concerns events that could be ignored at kernel-level, and for now we have no way to express the selection of non-syscall events (e.g. `container`) . This means that we'll keep receiving all of them even after these changes. Most of non-syscall events are meta-events or tracepoint-based events. The former are generated internally by libsinsp as part of its internal machinery, and the latter can be selected in the separate `selected_tp_set`.

Thank you Jason. Fair to say it was never needed? https://github.com/falcosecurity/falco/blob/0.34.1/userspace/falco/app_actions/configure_interesting_sets.cpp#L38

jasondellaluce:
> Thank you Jason. Fair to say it was never needed? https://github.com/falcosecurity/falco/blob/0.34.1/userspace/falco/app_actions/configure_interesting_sets.cpp#L38

At the point in time of 0.34.0/.1, it was needed for printing the output of `-i` and the warning messages related to `-A` with its old behavior:

- https://github.com/falcosecurity/falco/blob/9a4205cc70e449b170d32d3033dc7e74cf4ca613/userspace/falco/app_actions/print_syscall_events.cpp#L76
- https://github.com/falcosecurity/falco/blob/0.34.1/userspace/falco/app_actions/print_ignored_events.cpp#L55

At the current time, storing this in the app state does not have value anymore, as the event set is not used only by the Falco engine.

jasondellaluce:
cc @falcosecurity/falco-maintainers for visibility.

incertum:
@jasondellaluce as we bump libs one more suggesting to update falco engine version as we will have new fields available as well, ty!

jasondellaluce:
Note, the `Code scanning results / CodeQL` failing job is currently failing just reporting some issues with the latest version of TBB recently updated in falcosecurity/libs.

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #2456, opened 2023-03-24
new: driver drop failed

FedeDP:
**What type of PR is this?**

/kind feature

/area engine

This PR bumps libs to latest master (https://github.com/falcosecurity/libs/commit/c8b0d6a8fdc1bb3ea9067bc2fdc3ae5858cff48f).
Moreover, it introduces a new `syscall_drop_failed` config option to instrument drivers to drop failed syscalls exit events, exposing the new feature impl by https://github.com/falcosecurity/libs/pull/834.

Fixes #

```release-note
new(cmake): bumped libs to c8b0d6a8fdc1bb3ea9067bc2fdc3ae5858cff48f
new(userspace): add a new `syscall_drop_failed` config option to drop failed syscalls exit events
```

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -170,6 +170,10 @@ syscall_event_drops:
 syscall_event_timeouts:
   max_consecutives: 1000
 
+# Enabling this option allows Falco to drop failed syscalls exit events.
+# This can enable some small optimization both in CPU usage and ring buffer usage,
+# possibly leading to lower number of event losses.
+# Be careful: enabling it also means losing a bit of visibility on the system.
 syscall_drop_failed: false
 
 # --- [Description]
--- a/falco.yaml
+++ b/falco.yaml
@@ -170,7 +170,8 @@ syscall_event_drops:
 syscall_event_timeouts:
   max_consecutives: 1000
 
-# Enabling this option allows Falco to drop failed syscalls exit events.
+# Enabling this option allows Falco to drop failed syscalls exit events
+# in the kernel driver before the event is pushed onto the ring buffer.
 # This can enable some small optimization both in CPU usage and ring buffer usage,
 # possibly leading to lower number of event losses.
 # Be careful: enabling it also means losing a bit of visibility on the system.
--- a/falco.yaml
+++ b/falco.yaml
@@ -175,7 +175,7 @@ syscall_event_timeouts:
 # This can enable some small optimization both in CPU usage and ring buffer usage,
 # possibly leading to lower number of event losses.
 # Be careful: enabling it also means losing a bit of visibility on the system.
-syscall_drop_failed: false
+syscall_drop_failed_exit: false
 
 # --- [Description]
 #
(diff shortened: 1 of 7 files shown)


falco.yaml
```diff
@@ -170,6 +170,12 @@ syscall_event_drops:
 syscall_event_timeouts:
   max_consecutives: 1000
 
+# Enabling this option allows Falco to drop failed syscalls exit events.
+# This can enable some small optimization both in CPU usage and ring buffer usage,
+# possibly leading to lower number of event losses.
+# Be careful: enabling it also means losing a bit of visibility on the system.
+syscall_drop_failed: false
```
incertum, line 177: Suggestion: `drop_failed_syscall_exit`

userspace/falco/app/actions/helpers_inspector.cpp
```diff
@@ -125,5 +125,10 @@ falco::app::run_result falco::app::actions::open_live_inspector(
 		return run_result::fatal(e.what());
 	}
 
+	if (s.config->m_syscall_drop_failed)
+	{
+		falco_logger::log(LOG_DEBUG, "Failed syscalls exit event will be dropped.\n");
```
incertum, line 130: Suggestion: `Failed syscall exit events are dropped in the kernel driver`

falco.yaml
incertum: Similar to below event, perhaps add more details around where we drop them, e.g.
`Enabling this option allows Falco to drop failed syscall exit events in the kernel driver before the event is pushed onto the ring buffer.`

incertum (commented):
LGTM, just few suggestions to improve wording.

falco.yaml
FedeDP: > drop_failed_syscall_exit

I'd leave the `syscall` name as prefix, as it is the `macro` argument (and we already have eg: `syscall_event_timeouts`).

incertum: `syscall_drop_failed_exit`?

FedeDP: Done!

incertum (approved):
Can't officially approve, but this LGTM!

leogr (dismissed):

userspace/falco/app/actions/helpers_inspector.cpp
```diff
@@ -125,5 +125,10 @@ falco::app::run_result falco::app::actions::open_live_inspector(
 		return run_result::fatal(e.what());
 	}
 
+	if (s.config->m_syscall_drop_failed_exit)
```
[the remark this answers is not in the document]
FedeDP: > Shouldn't we configure this before opening the event capture?

Yes an no, i mean we could set it both ways :) 

> Wouldn't init_syscall_inspector be a better function in which setting this field?

Yep if we want to call it **before**, yes! I can move it there!

> Would it make sense to log this as LOG_INFO level, just like the ring buffer size config messages?

+1, ok!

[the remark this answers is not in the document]
FedeDP: Applied, thanks Jason!

jasondellaluce (approved):
/approve

leogr (approved):

Issue #2433, opened 2023-02-27
[New Feature]: User flag for `base_syscalls.repair` as extension to `base_syscalls` config overriding default sinsp state enforcement


incertum:
**Context / Bigger Picture**

Thanks to @Happy-Dude feedback realized to perhaps first set the stage, please consider reading https://github.com/falcosecurity/falco/issues/2433#issuecomment-1447620920 before continuing reading below as we are diving right into the technical details of what we currently call "Option 3".

**Motivation**

At the time of writing Falco's default libsinsp state enforcement activates a [static set of system calls](https://github.com/falcosecurity/libs/blob/f0468f32d8d730d0e70d8e103d57f97d74dd374f/userspace/libsinsp/test/public_sinsp_API/interesting_syscalls.cpp#L34) [see additional note 1] that are associated with the `EF_MODIFIES_STATE` flag in the libs [event_table](https://github.com/falcosecurity/libs/blob/f0468f32d8d730d0e70d8e103d57f97d74dd374f/driver/event_table.c). The entirety of these system calls are activated regardless of the system calls used in Falco rules.

Compute overhead budgets typically are not unlimited plus Falco's rules event evaluation runs single threaded and eventually drops events kernel side once one entire CPU is consumed. Therefore, the default libsinsp state enforcement can have severe production degradation impacts, especially for configurations that do not activate most system calls and/or when running Falco on very active servers w/ 96 CPUs (see https://github.com/falcosecurity/libs/issues/855).

We introduced `base_syscalls` (see https://github.com/falcosecurity/falco/issues/2373) (scheduled for 0.35 release, see https://github.com/falcosecurity/falco/commit/d6421d4e67fd4a058ccb3b448291767f6b5a6992) which provides the end user with a config option to pass their own set of syscalls that will be activated in addition to the system calls from each Falco rule. The recommendations given for this new option are based on the concept of `adaptive_base_syscalls` introduced here https://github.com/falcosecurity/libs/pull/826.

The question of what the system calls are that are truly needed for Falco's state engine given the rules sets remains open. Plus ideally there would also be such an automated `adaptive_base_syscalls` in addition to `base_syscalls` so that end users do not have to keep a manual set of base syscalls correct and can rely on automatic resourceful syscalls activation that ensures that both (1) data is correct and (2) state engine remains healthy.

The intended outcome of this issue is to derive a clear understanding of what the system call dependencies of libsinsp are given x input filter system calls. 

Notes:
[...]

PR #2457, opened 2023-03-27
new(app_actions)!: adjust base_syscalls option, add base_syscalls.repair

incertum:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind release

> /area build

/area engine

/area tests

> /area proposals

> /area CI

Finalize https://github.com/falcosecurity/falco/issues/2433.
Fixes missing `PPM_SC_SCHED_PROCESS_EXIT` safety enforcement when user_positive_sc_set non empty.

Fixes #

```release-note
new(app_actions)!: adjust base_syscalls option, add base_syscalls.repair
```


falco.yaml
```diff
@@ -444,66 +444,86 @@ metadata_download:
   watch_freq_sec: 1
 
 
-# base_syscalls ! Use with caution !
+# base_syscalls ! Use with caution, read carefully !
```
incertum, line 447: Tried to make the description easier to follow, we can plan for some final touch up closer to Falco 0.35 release.

userspace/falco/app/actions/configure_interesting_sets.cpp
```diff
@@ -100,12 +102,32 @@ static void select_event_set(falco::app::state& s, const libsinsp::events::set<p
...
 			+ concat_set_in_order(user_positive_sc_set_names) + "\n");
+		auto invalid_positive_sc_set_names = unordered_set_difference(user_positive_names, user_positive_sc_set_names);
+		if (!invalid_positive_sc_set_names.empty())
+		{
+			std::cerr << "Invalid (positive) syscall names: warning (base_syscalls override): " + concat_set_in_order(invalid_positive_sc_set_names) << std::endl;
+		}
+
+		/* Hidden safety enforcement for `base_syscalls.custom_set` user input override option -> sched_process_exit trace point activation
+		 * (procexit event) is necessary for continuous state engine cleanup, else memory would grow rapidly and linearly over time. */
+		base_sc_set.insert(PPM_SC_SCHED_PROCESS_EXIT);
```
incertum, line 113: This was missing for the user override scenario ...

unit_tests/falco/app/actions/test_configure_interesting_sets.cpp
```diff
@@ -47,7 +47,7 @@ static std::string s_sample_ruleset = "sample-ruleset";
 static std::string s_sample_source = falco_common::syscall_source;
 
 static strset_t s_sample_filters = {
-	"evt.type=connect or evt.type=accept",
+	"evt.type=connect or evt.type=accept or evt.type=accept4 or evt.type=umount2",
```
incertum, line 50: Small touch up on tests to include some of the overlapping syscalls already. Noticed during testing that `init_module` was troublesome ... but it may be a test setup issue, as when I ran the Falco binary with debug log level `init_module` was correctly included everywhere.

As we refactor the ppm sc API one more time with a subsequent libs bump we should include a generic syscalls in the test mock rules as well as this critical test coverage is currently missing. CC @jasondellaluce

incertum:
Seeing that there are issues with the test cases, because of extending them a bit. Let's investigate, locally they passed, but I posted in another comment that when I tried to add init_module into the mock up rules it was troublesome. Happy to revert test extensions a bit and we can address them in a separate PR after @jasondellaluce refactors final aspects of ppm sc API.

jasondellaluce:
@incertum I had a list of suggestions to propose with where impossible-ish to report through the GH UI, so I proceeded by pushing a commit on your branch with my review suggestion. Please drop it if you feel like it's wrong.

The topic is to fix a couple of logic/preference flaws, and strengthen the test cases (also re-including your interim test enhancements). Hopefully this will fix the CI on GH as well, since it works for me locally too.

userspace/falco/app/actions/configure_interesting_sets.cpp
```diff
@@ -89,11 +89,9 @@ static void select_event_set(falco::app::state& s, const libsinsp::events::set<p
 	auto user_positive_sc_set = libsinsp::events::names_to_sc_set(user_positive_names);
 	auto user_negative_sc_set = libsinsp::events::names_to_sc_set(user_negative_names);
 
-	if (!user_positive_sc_set.empty() || s.config->m_base_syscalls_repair)
+	if (!user_positive_sc_set.empty())
```
jasondellaluce: note: we were wrongly accepting totally-empty positive set, which we didn't allow before and I'm a bit contrary on.

userspace/falco/app/actions/configure_interesting_sets.cpp
```diff
@@ -105,28 +103,20 @@ static void select_event_set(falco::app::state& s, const libsinsp::events::set<p
 		auto invalid_positive_sc_set_names = unordered_set_difference(user_positive_names, user_positive_sc_set_names);
 		if (!invalid_positive_sc_set_names.empty())
 		{
-			std::cerr << "Invalid (positive) syscall names: warning (base_syscalls override): " + concat_set_in_order(invalid_positive_sc_set_names) << std::endl;
+			falco_logger::log(LOG_WARNING, "Invalid (positive) syscall names: warning (base_syscalls override): "
```
jasondellaluce, line 106: This is just to make logging consistent.

userspace/falco/app/actions/configure_interesting_sets.cpp
```diff
@@ -105,28 +103,20 @@ static void select_event_set(falco::app::state& s, const libsinsp::events::set<p
...
-		 * Also applies when base_syscalls.custom_set empty but base_syscalls.repair set
-		 * effectively bypassing the default `libsinsp` state enforcement and using
-		 * `sinsp_repair_state_sc_set` instead. */
-		s.selected_sc_set = libsinsp::events::sinsp_repair_state_sc_set(rules_sc_set).merge(base_sc_set);
-	}
+	/* Hidden safety enforcement for `base_syscalls.custom_set` user input
+	 * override option -> sched_process_exit trace point activation
+	 * (procexit event) is necessary for continuous state engine cleanup,
+	 * else memory would grow rapidly and linearly over time. */
+	s.selected_sc_set.insert(ppm_sc_code::PPM_SC_SCHED_PROCESS_EXIT);
```
jasondellaluce, line 119: We need to repair procexit regardless on whether we use a custom set or not, just as a safety measure at this point.

userspace/falco/app/actions/configure_interesting_sets.cpp
```diff
@@ -177,6 +168,27 @@ static void select_event_set(falco::app::state& s, const libsinsp::events::set<p
 		}
 	}
 
+	/* if a custom set is specified (positive, negative, or both), we attempt
```
jasondellaluce, line 171: Set repairing should happen at the last stage, so that we make sure users don't unwillingly delete an essential syscall and Falco can recover by repairing it (if configured to do so). The test cases now assert this as well by negating `bind` and then repairing it.

unit_tests/falco/app/actions/test_configure_interesting_sets.cpp
```diff
@@ -301,9 +301,10 @@ TEST(ConfigureInterestingSets, selection_custom_base_set)
 		// note: `accept` is not included even though it is matched by the rules,
 		// which means that the custom negation base set has precedence over the
 		// final selection set as a whole
-		"connect", "open", "ptrace", "mmap", "execve", "read", "syncfs", "sched_process_exit"
+		// todo(jasondellaluce): add "accept4" once names_to_sc_set is polished on the libs side
```
jasondellaluce, line 304: When disabling `accept`, Falco also disables `accept4`. That's still expected because I still have to clean up `names_to_sc_set` on libsinsp.

userspace/falco/app/actions/configure_interesting_sets.cpp
incertum: What would be an alternative to allow bypassing default sinsp state enforcement, but use .repair option, but maybe still want one negative syscall?

Either way you can always just add execve in there and that way bypass, so I would always find a way to get the behavior I want :)

userspace/falco/app/actions/configure_interesting_sets.cpp
incertum: Agree we should really do this safety just in case always, it's one insert only ...

incertum:
Thanks a lot for fixing up the tests :pray: will address the other changes shortly when I have some time.

userspace/falco/app/actions/configure_interesting_sets.cpp
jasondellaluce: My points are:
- We need a way for users to satisfy the case "all default sinsp set syscalls, except the negated ones", which as of today we interpret the empty positive set for. If you want to really have an empty custom positive set, what is your proposal to support the above? We do the same kind of reasoning for enabled event sources, for example
- I'm not sure that explicitly allowing user to pass-in totally-empty base sets is the best UX, and as you said you still have means to circumvent it by simply defining a syscall as positive and negative at the same time. I'm not strong on this, but what is the concrete use case that you foresee on having a totally-empty positive base set?

incertum:
@jasondellaluce I think we still have a some problems with the tests. They can interfere with each other and then some of them can fail. It appears to be a known issue, was reading up on https://chromium.googlesource.com/external/github.com/google/googletest/+/refs/tags/release-1.8.0/googletest/docs/FAQ.md -> we may need better SetUp and TearDown mechanisms for these ConfigureInterestingSets tests?

How to replicate issues: This is the interesting part, first all worked again locally, changing order of tests made it then fail for 2 of them. And then without changing anything after rebasing some tests failed, but for example if you comment all other tests out the one that failed will surely pass. Hope one of these conditions helps you to simulate these type of apparently a bit random interferences.

userspace/falco/app/actions/configure_interesting_sets.cpp
```diff
@@ -117,6 +123,25 @@ static void select_event_set(falco::app::state& s, const libsinsp::events::set<p
...
+	if (s.config->m_base_syscalls_repair && s.config->m_base_syscalls_custom_set.empty())
+	{
+		/* If `base_syscalls.repair` is specified, but `base_syscalls.custom_set` is empty we are replacing
+		 * the default `sinsp_state_sc_set()` enforcement with the alternative `sinsp_repair_state_sc_set`.
+		 * This approach only activates additional syscalls Falco needs beyond the
+		 * syscalls defined in each Falco rule that are absolutely necessary based
+		 * on the current rules configuration. */
+
+		// returned set already has rules_sc_set merged
+		s.selected_sc_set = libsinsp::events::sinsp_repair_state_sc_set(rules_sc_set);
```
jasondellaluce, line 144:
```suggestion
	/* REPLACE DEFAULT STATE, nothing else. */
	if (s.config->m_base_syscalls_repair && user_positive_sc_set.empty())
	{
		/* If `base_syscalls.repair` is specified, but `base_syscalls.custom_set` defines no positive syscall we are replacing
		 * the default `sinsp_state_sc_set()` enforcement with the alternative `sinsp_repair_state_sc_set`.
		 * This approach only activates additional syscalls Falco needs beyond the
		 * syscalls defined in each Falco rule that are absolutely necessary based
		 * on the current rules configuration. */

		base_sc_set.clear()
```
Agreed on your proposal. I think this should happen only when `user_positive_sc_set` is empty, so that users can still negate some syscall if they need to (potentially used by the rules for maximum control). Also, this needs to be moved at line 110 as an `else` branch of the current `if (!user_positive_sc_set.empty())`, so that negated syscall checks can be applied below.

What do you think?

jasondellaluce:
> we may need better SetUp and TearDown mechanisms for these ConfigureInterestingSets tests?

I think all the tests should be "stateless", and they just access the static globals for reading. It can be that the static global variables are not initialized in the proper execution order maybe, in which case SetUp and TearDown would definitely help. Since they are passing so far, I would leve this test flakyness improvements for another PR, WDYT?

userspace/falco/app/actions/configure_interesting_sets.cpp
incertum: ACK, since `sinsp_repair_state_sc_set` already merges in the rules_sc_set moving it to line 115 may be better. I think it's ok to just re-assign / override s.selected_sc_set again, but we could add more if else, elseif etc behavior as well.

jasondellaluce (approved):
/approve

leogr (approved):
LGTM.

Thank you!

PR #2458, opened 2023-03-27
new(falco/config): add new configurations for http_output

### LINKED ISSUE #2274: http_output not working, libcurl error: Problem with the SSL CA cert (path? access rights?)
**Describe the bug**
After installing Falco in a Kubernetes cluster using the official Helm chart, whenever I try to output the results in a JSON format to an HTTP endpoint, Falco gives me a SSL error. Isn't this ca-certificate package supposed to come with the default installation?

**How to reproduce it**
1. `helm repo add falcosecurity https://falcosecurity.github.io/charts`
2. `helm repo update`
3. `helm install falco falcosecurity/falco`
4. `exec` into one of the pods running Falco.
5. `falco -o json_ouput=true http_output.enabled=true http_output.url=<REQUESTBIN_ADDRESS>`
6. `exec` into another pod (or even the same pod) to generate a Falco event.

**Expected behaviour**
Falco sends an HTTP call to the endpoint containing the JSON of the syscall it just registered.

**Screenshots**
1. Output when running `curl <MY_REQUEST_BIN>` from inside a pod running Falco:
```json
{"about":"Pipedream is the fastest way to connects APIs. Build and run workflows with code-level control when you need it — and no code when you don't.","event_id":"2GfreizFr8I0hJjji77gNQEoP8Y","workflow_id":"p_LQCDnkR","owner_id":"u_KDh7vnG","deployment_id":"d_zqs7eNxx","timestamp":"2022-10-26T13:22:56.700Z","inspect":"https://pipedream.com/@/p_LQCDnkR","quickstart":"https://pipedream.com/quickstart/"}
```
2. Output when Falco receives an event with `json_output=true` and `http_output=true` enabled to the same requestbin address:
```bash
Fri Oct 21 18:02:03 2022: libcurl error: Problem with the SSL CA cert (path? access rights?)
```
3. Command we are running from within the pod: `falco -o json_ouput=true -o http_output.enabled=true -o http_output.url=<MY_REQUEST_BIN_ADDRESS>`

**Environment**
- Falco version: 0.33.0 (x86_64)
- System info: `"machine":"x86_64","nodename":"falco-2tcqs","release":"5.13.0-1031-aws","sysname":"Linux","version":"#35~20.04.1-Ubuntu SMP Mon Jun 13 22:30:30 UTC 2022"`
- OS: Debian GNU/Linux 11 (bullseye)
- Kernel: `Linux falco-2tcqs 5.13.0-1031-aws #35~20.04.1-Ubuntu SMP Mon Jun 13 22:30:30 UTC 2022 x86_64 GNU/Linux`
- Installation method: Official Helm Chart

**Additional context**
TL;DR: We're trying to use the built-in functionality of HTTP Output that comes with Falco to send JSON output about syscalls that Falco catches inside our containers. The main idea is to understand if this works or Sidekick is the only alternative.

### LINKED ISSUE #2448: Make CA/certs path configurable for http_output
**Motivation**

When connecting to a `https` endpoint Falco(libcurl) performs peer SSL certificate verification by default. In some case the default CA certificate store is not enough for various reasons:
* default path could change based on the environment;
* different format for the CA store, ca-bundle/ca-path;
* self-signed certificates.

**Feature**
We need make the certificate path configurable at start-up time by adding the required flags and config fields in the configuration file. We need to take care of the following configs:
 * --http-cacert=`file,` path to the certificate that can verify the remote server;
 * --http-ca-bundle=`file` path to the file as the CA certificate store;
 * --http-ca-path=`path`, use the specified path as CA certificate store.

**Update - 27/03/2023**

The best way to proceed is to allow the CA certificates to be configured using the falco configuration file. The new config fields are added under the `http_output` sections. The new section will be:

```yaml
http_output:
  enabled: false
  url: http://some.url
  user_agent: "falcosecurity/falco"
  # Tell Falco to not verify the remote server.
  insecure: false
  # Path to the CA certificate that can verify the remote server. 
  ca_cert: ""
  # Path to a specific file that will be used as the CA certificate store.
  ca_bundle: ""
  # Path to a folder that will be used as the CA certificate store. CA certificate need to be
  # stored as indivitual PEM files in this directory.
  ca_path: "/etc/ssl/certs
```
Keep in mind that Falco allows overwriting configurations from the command line using the `-o` flag so no need to add new flags for the above configs.

**Alternatives**
NONE

**Additional context**
Please see the comment: https://github.com/falcosecurity/falco/issues/2274#issuecomment-1477109987

alacuku:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind release

> /area build

> /area engine

> /area tests

> /area proposals

> /area CI

Falco now accepts user-provided Certificates and CA certificate stores that are used to verify the remote server to which the `http_output` is sent.

Four new config fields are available under the `http_output` section in **falco.yaml** configuration file:
```yaml
 # Tell Falco to not verify the remote server.
insecure: false
# Path to the CA certificate that can verify the remote server. 
ca_cert: ""
# Path to a specific file that will be used as the CA certificate store.
ca_bundle: ""
# Path to a folder that will be used as the CA certificate store. CA certificate need to be
# stored as indivitual PEM files in this directory.
ca_path: "/etc/ssl/certs"
```

Furthermore, this PR decouples Falco as a consumer of the `libcurl` from the libs. Making the certificate paths configurable at startup time allows Falco to support `https` endpoints in every scenario as far as a valid certificate is provided for the remote server. 

Fixes #2274
Fixes #2448 

```release-note
new(falco/config): add new configurations for http_output that allow custom CA certificates and stores.
```

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -408,6 +408,15 @@ http_output:
   enabled: false
   url: http://some.url
   user_agent: "falcosecurity/falco"
+  # Tell Falco to not verify the remote server.
+  insecure: false
+  # Path to the CA certificate that can verify the remote server. 
+  ca_cert: ""
+  # Path to a specific file that will be used as the CA certificate store.
+  ca_bundle: ""
+  # Path to a folder that will be used as the CA certificate store. CA certificate need to be
+  # stored as indivitual PEM files in this directory.
+  ca_path: "/etc/ssl/certs"
 
 # Falco supports running a gRPC server with two main binding types
 # 1. Over the network with mandatory mutual TLS authentication (mTLS)
--- a/userspace/falco/configuration.cpp
+++ b/userspace/falco/configuration.cpp
@@ -178,6 +178,22 @@ void falco_configuration::load_yaml(const std::string& config_name, const yaml_h
 		user_agent = config.get_scalar<std::string>("http_output.user_agent","falcosecurity/falco");
 		http_output.options["user_agent"] = user_agent;
 
+		bool insecure;
+		insecure = config.get_scalar<bool>("http_output.insecure", false);
+		http_output.options["insecure"] = insecure? std::string("true") : std::string("false");
+		
+		std::string ca_cert;
+		ca_cert = config.get_scalar<std::string>("http_output.ca_cert", "");
+		http_output.options["ca_cert"] = ca_cert;
+
+		std::string ca_bundle;
+		ca_bundle = config.get_scalar<std::string>("http_output.ca_bundle", "");
+		http_output.options["ca_bundle"] = ca_bundle;
+
+		std::string ca_path;
+		ca_path = config.get_scalar<std::string>("http_output.ca_path", "/etc/ssl/certs");
+		http_output.options["ca_path"] = ca_path;
+
 		m_outputs.push_back(http_output);
 	}
 
--- a/userspace/falco/outputs_http.cpp
+++ b/userspace/falco/outputs_http.cpp
@@ -34,15 +34,58 @@ void falco::outputs::output_http::output(const message *msg)
 		} else {
 			slist1 = curl_slist_append(slist1, "Content-Type: text/plain");
 		}
+		res = curl_easy_setopt(curl, CURLOPT_HTTPHEADER, slist1);
 
-		curl_easy_setopt(curl, CURLOPT_HTTPHEADER, slist1);
-		curl_easy_setopt(curl, CURLOPT_URL, m_oc.options["url"].c_str());
-		curl_easy_setopt(curl, CURLOPT_POSTFIELDS, msg->msg.c_str());
-		curl_easy_setopt(curl, CURLOPT_USERAGENT, m_oc.options["user_agent"].c_str());
-		curl_easy_setopt(curl, CURLOPT_POSTFIELDSIZE, -1L);
+		if(res == CURLE_OK)
+		{
+			res = curl_easy_setopt(curl, CURLOPT_URL, m_oc.options["url"].c_str());
+		}
+		
+		if(res == CURLE_OK)
+		{
+			res = curl_easy_setopt(curl, CURLOPT_POSTFIELDS, msg->msg.c_str());
+		}
 
+		if(res == CURLE_OK)
+		{
+			res = curl_easy_setopt(curl, CURLOPT_USERAGENT, m_oc.options["user_agent"].c_str());
+		}
 
-		res = curl_easy_perform(curl);
+		if(res == CURLE_OK)
+		{
+		   res = curl_easy_setopt(curl, CURLOPT_POSTFIELDSIZE, -1L);
+		}
+		
+		if(res == CURLE_OK)
+		{
+			if(m_oc.options["insecure"] == std::string("true"))
+			{
+				res = curl_easy_setopt(curl,CURLOPT_SSL_VERIFYPEER, 0L);
+
+				if(res == CURLE_OK)
+				{
+					res = curl_easy_setopt(curl, CURLOPT_SSL_VERIFYHOST, 0L);
+				}
+			}
+		}
+
+		if(res == CURLE_OK)
+		{
+			if (!m_oc.options["ca_cert"].empty())
+			{
+				res = curl_easy_setopt(curl, CURLOPT_CAINFO, m_oc.options["ca_cert"].c_str());
+			}else if(!m_oc.options["ca_bundle"].empty())
+			{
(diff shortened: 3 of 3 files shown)


userspace/falco/configuration.cpp
```diff
@@ -177,6 +177,22 @@ void falco_configuration::load_yaml(const std::string& config_name, const yaml_h
 		user_agent = config.get_scalar<std::string>("http_output.user_agent","falcosecurity/falco");
 		http_output.options["user_agent"] = user_agent;
 
+		bool insecure;
+		insecure = config.get_scalar<bool>("http_output.insecure", false);
+		http_output.options["insecure"] = insecure? std::string("True") : std::string("False");
```
jasondellaluce, line 182: Isn't there a pair of string <-> bool primitives that we can reuse to avoid custom formats? If not, can we at least make this case-insensitive (or tied to lower case)?

userspace/falco/outputs_http.cpp
```diff
@@ -34,15 +34,58 @@ void falco::outputs::output_http::output(const message *msg)
...
 			slist1 = curl_slist_append(slist1, "Content-Type: text/plain");
 		}
+		res = curl_easy_setopt(curl, CURLOPT_HTTPHEADER, slist1);
 
-		curl_easy_setopt(curl, CURLOPT_HTTPHEADER, slist1);
-		curl_easy_setopt(curl, CURLOPT_URL, m_oc.options["url"].c_str());
-		curl_easy_setopt(curl, CURLOPT_POSTFIELDS, msg->msg.c_str());
-		curl_easy_setopt(curl, CURLOPT_USERAGENT, m_oc.options["user_agent"].c_str());
-		curl_easy_setopt(curl, CURLOPT_POSTFIELDSIZE, -1L);
+		if(res == CURLE_OK)
```
jasondellaluce, line 39: Thanks for checking all the results. I was wondering if there was a more elegant way to attain this without all this if-chain, but if there is not I'm ok in going this route.

falco.yaml
```diff
@@ -401,6 +401,15 @@ http_output:
   enabled: false
   url: http://some.url
   user_agent: "falcosecurity/falco"
+  # Tell Falco to not verify the remote server.
+  insecure: false
```
jasondellaluce, line 405: My only concrete concern is: do we want to nest these config fields under the `http_output` config field, or do we want to keep them separate? For the future, my fear is that we would need the certificate configuration for other use cases as well not related to the http output, which would make this a bit ambiguous.

What about having a new field nesting level called `http`? After all, we do the same with `grcp_output` and `grpc` for this exact reason. Does anyone have better ideas? cc @falcosecurity/falco-maintainers

userspace/falco/configuration.cpp
alacuku: The `config` type for `outputs` uses a` map<string, string>` flexible enough to hold arbitrary values as long as we represent them as strings. https://github.com/falcosecurity/falco/blob/4836978ba9c054cfb752616a252c8d2bda8f1870/userspace/falco/outputs.h#L34-L38
Also the string representation of the `bool` type is internal, we are still requiring the users to use a `bool` value in the configuration. 

I didn't find a more elegant/appropriate way to pass the `bool` value to the `http_output` object. If we stick with this solution we should make the string values lowercase.

userspace/falco/outputs_http.cpp
alacuku: This is the best I came out with. 

Most of these options could be set only once at the creation time of the `http_output` object and reuse the same `curl` object when sending output to the `http endpoint`. We instantiate and initialize a new `curl` object each time we emit a message. Maybe @leogr has more info on why we doing things this way. 
Anyway, this refactoring/optimization will be for another PR.

falco.yaml
alacuku: The idea here is that we could have different needs for different endpoints to which Falco connects. Moving the `http` configuration to a global one it will force all the `http clients` in Falco to use the same SSL configuration. Instead having them per `http client` allows configurations for different scenarios. Just my two cents. Open to other ideas.

jasondellaluce: In that case, then I'm onboard with keeping it under the http output config field.

userspace/falco/configuration.cpp
jasondellaluce: Let's take the simple route and just make it lowercase then.

jasondellaluce (approved):
/approve

PR #2499, opened 2023-04-16
fix(scripts/systemd): restored stdout in unit files

therealbobo:
**What type of PR is this?**

/kind bug

/area engine

Redirecting stdout to null in the unit files results in partial logs. E.g.: in the case that a rule is broken, no hint is printed to the logs.

Fixes #

```release-note
NONE
```

### DIFF
--- a/scripts/systemd/falco-bpf.service
+++ b/scripts/systemd/falco-bpf.service
@@ -20,7 +20,6 @@ ProtectSystem=full
 ProtectKernelTunables=true
 RestrictRealtime=true
 RestrictAddressFamilies=~AF_PACKET
-StandardOutput=null
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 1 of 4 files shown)


Andreagit97:
This seems reasonable, to be honest, I don't remember why we put `StandardOutput=null`, @FedeDP do you remember it?

FedeDP:
This was something already present that we didn't feel like touching since the systemd changes were merged like 2d from the releases :) 
@leogr should have more context on this.

leogr:
@therealbobo @FedeDP @Andreagit97 

I recall pretty clear that it was required to avoid flooding system logs. I can't find the old issue/PR right now, because I'm at the airport heading to KubeconEU Amsterdam  😅

Please double check, but I bet we shouldn't remove it

/hold

PR #2501, opened 2023-04-18
new(ci): gha master and release workflows

### LINKED ISSUE #1876: CI issue: data race in publishing jobs for `-dev` packages
**Describe the bug**

When multiple PRs got merged at the same time, then the CircleCI may run multiple workflows in parallel.
In such a situation, a data race occurs while publishing `-dev` packages to https://download.falco.org/?prefix=packages/
Since the publishing operation also updates RPM and DEB repository metadata, the process may result in corrupted repositories.

This problem happened with #1871 and #1872

However, once another PR gets merged, repository metadata is computed again and the problem disappears.

Only `-dev` packages are affected. Stable packages are not affected because we just tag a release at a time.

**How to reproduce it**

No way to reproduce it locally.

**Expected behaviour**

**Additional context**

I have opened this PR mostly for the record. Likely, we won't fix it.
Final users are not affected by this problem in any way.

FedeDP:
**What type of PR is this?**

/kind feature

/area CI

This PR tries to enable master and release build + packages + docker images CI on gha instead of circleci.
It surely needs some tests hence the `wip`.

To do so, it uses `reusable` workflows that are shared between master and release; therefore testing CI on master should be enough to test also the release one, since they are exactly the same.
Moreover, people could also reuse our provided workflows in their own CI.

Fixes #1876 

Steps:
- [x] fix arm64 build
- [x] fix arm64 job build Falco version (it sees 0.0.0 somehow...)
- [x] add GPG_KEY as gha variable (needs repo admin)
- [x] fix master and release yaml issues
- [x] understand how to docker in docker (reusable_build_docker workflow) -> just go on host. Tested that `docker` and `awscli` can be used in those jobs
- [x] Fix packages publishing (see https://github.com/falcosecurity/test-infra/pull/1102)
- [x] Make `reusable_publish_docker` the only  docker publishing job (ie: `build_docker` today is publishing images too, while `publish_docker` is using the docker manifest API to merge architectures specific image into a multi arch tag)
- [x] What about ecr docker publishing?
- [ ] test master :) 

Other notes:
* `reusable_build_packages` job was tested with a fake job in the PR ci on the arm64 node: https://github.com/falcosecurity/falco/actions/runs/4734095566
The operation was canceled, but it worked just fine.
See the commit with the code: https://github.com/falcosecurity/falco/pull/2501/commits/75759c81d03fd43ba32932833f944de67c0d987a

Also, as you can see, `reusable_build_packages` is not using the falco modern builder image; instead it is directly building in a `centos:7` container (that is the base image for the falco modern builder). This is needed because we cannot run docker in docker on our prow nodes (ie: the arm64 ones).
Instead, the skeleton building phase is made on a fedora:latest image, because it already ships `bpftool` in its repo with a recent version.
 
* `reusable_build_docker` workflow has been tested through a fake job in the PR ci on both arm and x86: https://github.com/falcosecurity/falco/actions/runs/4743816658
It was failing because `{{ github.ref_name }}` variable was not `master` nor a tag name, instead it was `2501/merge`:
> ERROR: invalid tag "falcosecurity/falco-no-driver:aarch64-2501/merge": invalid reference format

But you see it was failing with the same error at the same stage on both x86 and arm; therefore i expect it to work just fine on master.
See the commit with the code: https://github.com/falcosecurity/falco/pull/2501/commits/dc372540e321b91b332a2bf0038d31d5c9c84603

```release-note
new(ci): ported master and release artifacts publishing CI to gha
```


.circleci/config.yml
```diff
@@ -755,78 +755,78 @@ workflows:
...
-            - test-infra
-          filters:
-            tags:
-              ignore: /.*/
-            branches:
-              only: master
-          requires:
-            - "build-docker-dev"
-            - "build-docker-dev-arm64"
+      # - "rpm-sign":
```
FedeDP, line 758: Avoid conflicts; disable circleCI jobs.

.github/workflows/ci.yml
```diff
@@ -2,17 +2,71 @@ name: CI Build
 on:
   pull_request:
     branches: [master]
-  push:
```
FedeDP, line 5: We now have a dedicated `master.yaml` for that.

.github/workflows/ci.yml
```diff
@@ -2,17 +2,71 @@ name: CI Build
...
 
 # Checks if any concurrent jobs under the same pull request or branch are being executed
-# NOTE: this will cancel every workflow that is being ran as group is just the github ref (without the workflow name)
+# NOTE: this will cancel every workflow that is being ran against a PR as group is just the github ref (without the workflow name)
 concurrency:
   group: ${{ github.head_ref || github.run_id }}
   cancel-in-progress: true  
 
 jobs:
+  test-arm:
```
FedeDP, line 14: This is just to test.

FedeDP: Will be removed before removing `wip`.

.github/workflows/master.yaml
```diff
@@ -0,0 +1,67 @@
+name: Dev Packages and Docker images
+on:
+  push:
+    branches: [master]
+
+# Checks if any concurrent jobs under the same pull request or branch are being executed
+concurrency:
```
FedeDP, line 7: This fixes #1876

.github/workflows/reusable_publish_packages.yaml
```diff
@@ -0,0 +1,105 @@
...
+          pip install awscli==1.19.47
+      
+      - name: Download all artifacts
+        uses: actions/download-artifact@v3
+        with:
+          path: $RUNNER_TEMP
+        
+      - name: Import gpg key 
+        run: |
+          echo $GPG_KEY | base64 -d | gpg --import
```
FedeDP, line 42: We miss the `GPG_KEY` gha variable.

.github/workflows/reusable_publish_packages.yaml
```diff
@@ -0,0 +1,105 @@
...
+          spawn rpmsign --addsign {*}\$argv
+          expect -exact "Enter pass phrase: "
+          send -- "\n"
+          expect eof
+          EOF
+          chmod +x ~/sign
+          ~/sign $RUNNER_TEMP/falco-*.rpm
+          rpm --qf %{SIGPGP:pgpsig} -qp $RUNNER_TEMP/falco-*.rpm | grep SHA256
+          
+      - name: Publish rpm
```
FedeDP, line 60: To publish, we also need S3 credentials and access.

.github/workflows/master.yaml
```diff
@@ -0,0 +1,67 @@
...
+    with:
+      arch: aarch64
+    secrets: inherit
+    
+  publish-dev-packages:
+    needs: [build-dev-packages, build-dev-packages-arm64]
+    outputs:
+      version: ${{ steps.expose_version.outputs.version }}
+    
+    steps:
```
FedeDP, line 29: It seems like this won't work: 
```
reusable workflows should be referenced at the top-level `jobs.*.uses' key, not within steps
```
Same applies to release.yaml.

FedeDP:
Steps:
- [ ] fix arm64 build
- [ ] add GPG_KEY as gha variable
- [ ] fix master and release yaml issues
- [ ] Fix publishing (s3 access!)

.github/workflows/master.yaml
```diff
@@ -0,0 +1,57 @@
...
+    
+  publish-dev-packages:
+    needs: [build-dev-packages, build-dev-packages-arm64]
+    uses: falcosecurity/falco/.github/workflows/reusable_publish_packages.yaml@main
+    with:
+      bucket: '-dev'
+      version: ${{ needs.build-dev-packages.outputs.version }}
+    secrets: inherit
+    
+  build-dev-docker:
```
FedeDP, line 32: Both `build-dev-docker` and its arm64 counterpart require `build-dev-packages` because they use its output.

FedeDP:
Also, we definitely need more powerful gha self-hosted runners for arm (and more replicas).

FedeDP:
Uh this time build-arm64 is much faster. Don't know what happened :/

FedeDP:
Still struggling to fix test-arm64 build: 
```
-- [MODERN BPF] USE_BUNDLED_MODERN_BPF: OFF, using skeleton dir: /tmp
-- [MODERN BPF] full skeleton path: /tmp/bpf_probe.skel.h
```
but then:
```
/__w/falco/falco/build/falcosecurity-libs-repo/falcosecurity-libs-prefix/src/falcosecurity-libs/userspace/libpman/src/state.h:23:10: fatal error: bpf_probe.skel.h: No such file or directory
   23 | #include <bpf_probe.skel.h>
      |          ^~~~~~~~~~~~~~~~~~
compilation terminated.
make[3]: *** [libpman/CMakeFiles/pman.dir/src/capture.c.o] Error 1
make[2]: *** [libpman/CMakeFiles/pman.dir/all] Error 2
```
This is weird.

FedeDP:
Ok `test-build-arm64` job is now fixed.

.github/workflows/reusable_build_docker.yaml
```diff
@@ -0,0 +1,131 @@
...
+        type: string
+      version:
+        description: 'Falco version extracted from userspace/falco/config_falco.h'
+        required: true
+        type: string
+
+jobs:
+  build-docker:
+    # See https://github.com/actions/runner/issues/409#issuecomment-1158849936
+    runs-on: ${{ (inputs.arch == "aarch64") && fromJSON('[ "self-hosted", "linux", "ARM64" ]') || 'ubuntu-latest' }}
```
FedeDP, line 26: Here we are trying to run a container on our runner that then needs to use docker to build and push images. I don't think this is feasible :/ 
We need to find another way to do this (perhaps splitting the jobs for x86 and arm and going natively on our node?)

.github/workflows/reusable_build_packages.yaml
```diff
@@ -0,0 +1,164 @@
...
+        value: ${{ jobs.build-packages.outputs.version }}    
+
+jobs:
+  build-modern-bpf-skeleton:
+    # See https://github.com/actions/runner/issues/409#issuecomment-1158849936
+    runs-on: ${{ (inputs.arch == "aarch64") && fromJSON('[ "self-hosted", "linux", "ARM64" ]') || 'ubuntu-latest' }}
+    container: fedora:latest
+    steps:
+      - name: Install build dependencies
+        run: |
```
FedeDP, line 21: Always install deps before invoking `checkout` action, to properly perform a full clone.

FedeDP: (same goes for other jobs that invoke `cmake` ie: `build-packages` and `build-musl-package`.

.github/workflows/reusable_publish_packages.yaml
```diff
@@ -0,0 +1,111 @@
...
+        run: |
+          yum install epel-release -y
+          yum update -y
+          yum install rpm-sign expect which createrepo gpg python python-pip -y
+          pip install awscli==1.19.47
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          role-to-assume: "arn:aws:iam::292999226676:role/github_actions-falco${{ inputs.bucket }}-s3"
```
FedeDP, line 36: The role is either `github_actions-falco-s3` or `github_actions-falco-dev-s3`; see https://github.com/falcosecurity/test-infra/pull/1102.

The dev one can only push dev packages.

.github/workflows/reusable_build_docker.yaml
```diff
@@ -0,0 +1,122 @@
...
+
+jobs:
+  build-docker:
+    # See https://github.com/actions/runner/issues/409#issuecomment-1158849936
+    runs-on: ${{ (inputs.arch == 'aarch64' && fromJSON('[ "self-hosted", "linux", "ARM64" ]')) || 'ubuntu-latest' }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        
+      - name: Install deps
```
FedeDP, line 27: While `ubuntu-latest` already ships awscli, make sure to install it from `pip` even when we run on our own amazonlinux arm node.

.github/workflows/reusable_build_packages.yaml
FedeDP: This is needed to actually get correct Falco version in cmake configure step, instead of `0.0.0`.

.github/workflows/reusable_build_docker.yaml
```diff
@@ -0,0 +1,108 @@
...
+        required: false
+        default: ''
+        type: string
+      version:
+        description: 'Falco version extracted from userspace/falco/config_falco.h'
+        required: true
+        type: string
+
+jobs:
+  build-docker:
```
FedeDP, line 20: Here we just build all docker images as tarballs, then we upload all the tarballs to be later downloaded by `reusable_publish_docker` workflow.

LucaGuerra:
This is PHENOMENAL @FedeDP ❤️ ❤️ 

I'm looking at the PR and also helping you address the open points.
✅  I have added the `GPG_KEY` variable for the package signing key.
Also, I already set up ECR with GHA in the falco org so we can do the same here

.github/workflows/reusable_build_docker.yaml
```diff
@@ -0,0 +1,111 @@
+# This is a reusable workflow used by master and release CI
+on:
+  workflow_call:
+    inputs:
+      arch:
+        description: x86_64 or aarch64
+        required: true
+        type: string
+      bucket:
```
LucaGuerra, line 9: I would also change the variable name since calling it "bucket" is a bit error prone, someone editing this workflow in the future might assume it's the bucket name!

FedeDP: Yep, perhaps bucket_suffix? :)

LucaGuerra:
Another little comment: some of the inline comments could be helpful to be made as regular comments in the code, because they're helpful for the next person that will modify this

.github/workflows/reusable_build_docker.yaml
FedeDP: or just `suffix` would suffice (eheh)

FedeDP:
@LucaGuerra I am going to push some commits to address your review <3 Thanks!

FedeDP:
@LucaGuerra everything was addressed :) PTAL! :pray:

.github/workflows/reusable_publish_docker.yaml
```diff
@@ -0,0 +1,169 @@
...
+        uses: aws-actions/amazon-ecr-login@2f9f10ea3fa2eed41ac443fee8bfbd059af2d0a4 # v1.6.0
+        with:
+          registry-type: public    
+          
+      - name: Login to Amazon ECR Public
+        run: |
+          aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/falcosecurity
+          
+      - name: Create and push no-driver manifest
+        uses: Noelware/docker-manifest-action@master
```
LucaGuerra, line 52: `@master` without the commit hash is a bit risky. I'm also sure that it's there for a reason, since I see that fixes have been pushed to that repo after the latest release. Can you add the hash like you did in `login-ecr-public`?

FedeDP: Since this is the first occurrence in falcosecurity org that this action is used (https://github.com/search?q=org%3Afalcosecurity+%22docker-manifest-action%22&type=code) i think i just copy pasted from the README :D 
I am going to use latest tag :)

FedeDP: Done!

.github/workflows/reusable_publish_packages.yaml
```diff
@@ -0,0 +1,119 @@
...
+      
+      - name: Download deb artifacts
+        uses: actions/download-artifact@v3
+        with:
+          name: falco-${{ inputs.version }}-*.deb
+          path: /tmp
+        
+      - name: Import gpg key 
+        run: |
+          echo $GPG_KEY | base64 -d | gpg --import
```
LucaGuerra, line 113: Same as above ^

.github/workflows/reusable_publish_packages.yaml
```diff
@@ -0,0 +1,119 @@
...
+          
+      - name: Download all artifacts
+        uses: actions/download-artifact@v3
+        with:
+          name: falco-*
+          path: /tmp
+        
+      - name: Import gpg key 
+        run: |
+          echo $GPG_KEY | base64 -d | gpg --import
```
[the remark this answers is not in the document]
FedeDP: Done!

.github/workflows/reusable_publish_packages.yaml
FedeDP: Done!

LucaGuerra:
LGTM, I think we can test it on master :)

leogr (approved):
Couldn't try it in detail, but this is :star_struck: 

I think we should give it a try, then fix any issues that will arise (as usual with the CI).

PR #2509, opened 2023-04-27
update: improve control and UX of ignored events

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind design

/area engine

This PR segregates the logic behind the `-i`/`-A` ignored events in a single place. The set of ignored events is now set to be the one of the high-throughput I/O syscalls (that can be impactful for performance), but excluding those that are part of the base set of events required by libsinsp to collect its internal state.

Accordingly, this removes all mentions to the I/O syscalls set from the CLI helpers and comments. The fact that we chose that set as ignored should be an implementation detail that we don't leak as part of the tool's UX. This way, the output of `-i` is the only contract we have with users about which events are forcely ignored without `-A`, which is also easier to test too.

This fixes an issue where Falco "ignores" some syscalls that were automatically added as part of the default base set, which ended up being written in the logs as ignored events even when not specified by users (e.g. `recvfrom`, `sendto`, ...).

cc @incertum, @Andreagit97, @FedeDP 

```release-note
update: improve control and UX of ignored events
```

### DIFF
--- a/unit_tests/falco/app/actions/test_configure_interesting_sets.cpp
+++ b/unit_tests/falco/app/actions/test_configure_interesting_sets.cpp
@@ -17,6 +17,7 @@ limitations under the License.
 
 #include <falco_engine.h>
 
+#include <falco/app/app.h>
 #include <falco/app/state.h>
 #include <falco/app/actions/actions.h>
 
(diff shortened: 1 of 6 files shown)


FedeDP (approved):
/approve

incertum:
While we already merged it, just leaving a record LGTM as well. As we discussed many times I am still hesitant if our new default option (while improved) works well for many end users. The good news however is that we have `base_syscalls` which we might want to promote more going forward, especially for end users who have to be very careful with the performance impact associated with running Falco.

Issue #2181, opened 2022-08-29
Migrate Falco image from debian to distroless


CashWilliams:
**Motivation**

To reduce attack surface and patching maintenance, we should use the most minimal base image for the Falco image as possible.

Currently debian:11-slim is used, however Google's distroless image would be more lightweight.

See https://github.com/GoogleContainerTools/distroless/blob/main/base/README.md 

**Feature**

Migrate to distroless as the base image.

**Alternatives**

TBD

**Additional context**

A quick test showed that at least some libraries were missing from distroless. There are methods to add these libraries, but I'm not sure what all would need to be included.

leogr:
/remove-lifecycle stale

any news?

leogr:
cc @LucaGuerra

LucaGuerra:
I think we can use this issue to discuss which image we want and why, and I really agree with the sentiment of using a light image.

The hard requirements:
- Needs to have core system libraries that are compatible with the binaries we ship for Falco. We try to build Falco as statically as possible but in order to run plugins at least a little bit of compatibility with glibc is necessary.
- Needs to work on both ARM and x86_64

Recently @dlorenc suggested to try the [Wolfi](https://github.com/wolfi-dev/) image that he develops with his team at Chainguard. I think it's a good idea to try it out, also considering that he'll be able to answer our questions :)

dlorenc:
Yes! Let's help out here! cc @amouat who should be able to jump in from our side when he gets back from vacation.

hpvd:
Since traditional, this approach is somehow strenuous to implement and associated with restrictions...
there are 2 tools that seems to make it pretty easy and straight forward
Haven't tested it on my own, but it looks promising...

Maybe this is useful:

**good overview on distroless containers**
https://dev.to/dansiviter/distroless-alpine-ci8
and
https://blog.chainguard.dev/minimal-container-images-towards-a-more-secure-future/

**the tools:**
source to abk:
https://github.com/chainguard-dev/melange

abk to oci:
https://github.com/chainguard-dev/apko

**to debug distroless containers:**
official: https://kubernetes.io/docs/concepts/workloads/pods/ephemeral-containers/
detail flow: https://iximiuz.com/en/posts/kubernetes-ephemeral-containers/

MukulKolpe:
Hey, I'm new in the Falco community. Once all the requirements along with the base image are discussed and finalized, I would be really interested to work on this one.

amouat:
[...]

PR #2510, opened 2023-04-27
new: Add distroless Dockerfiles

amouat:
**What type of PR is this?**

/kind feature

/area build

Adds distroless dockerfiles for Falco. These should be smaller and less likely to include vulnerabilities when scanned (due to simply having less packages in them).

Note there are two Dockerfiles, one of which builds from source, the other uses precompiled binaries.

Fixes: https://github.com/falcosecurity/falco/issues/2181

This a draft currently, just wanted to put something up for discussion.

YES

```release-note

BREAKING CHANGE: the Docker image no longer includes Debian utilities or a shell - meaning some debug workflows may be impacted.
```

### DIFF
--- a/docker/no-driver/Dockerfile
+++ b/docker/no-driver/Dockerfile
@@ -20,9 +20,7 @@ RUN curl -L -o falco.tar.gz \
 RUN sed -e 's/time_format_iso_8601: false/time_format_iso_8601: true/' < /falco/etc/falco/falco.yaml > /falco/etc/falco/falco.yaml.new \
     && mv /falco/etc/falco/falco.yaml.new /falco/etc/falco/falco.yaml
 
-# Our default glibc-dynamic image doesn't include libstdc++ so temporarily using this image
-# Working on an alternative (e.g. new image or adding the lib to glibc-dynamic)
-FROM cgr.dev/chainguard/cc-dynamic
+FROM cgr.dev/chainguard/glibc-dynamic
 
 LABEL maintainer="cncf-falco-dev@lists.cncf.io"
 LABEL org.opencontainers.image.source="https://github.com/falcosecurity/falco"
@@ -33,9 +31,7 @@ LABEL usage="docker run -i -t --privileged -v /var/run/docker.sock:/host/var/run
 ENV HOST_ROOT /host
 ENV HOME /root
 
-# Ideally these libs would be part of base image and not copied
-COPY --from=builder /usr/lib/libstdc++.so.6 /usr/lib/libstdc++.so.6 
-
 COPY --from=builder /falco /
+
 
 CMD ["/usr/bin/falco", "-o", "time_format_iso_8601=true"]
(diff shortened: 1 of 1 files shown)


amouat:
There are a few things to be solved:
 - There seems to be a difference between this and the old image that I haven't isolated yet; I get a different error when running `--modern-bpf`
 - `libstdc++.so` should be provided by the base image, not copied over from the build image. I'll try to get a new base image for this.
 - I'm not sure if we want the src version, but we could clean up the COPY statements.

leogr:
I'd love to have this soon, however, since we are approaching Falco 0.35 release (in about 4 weeks), I assume we are too late to include this. 

/milestone 0.36.0

amouat:
We added the libstdc++ and libgcc libaries to the cgr.dev/chainguard/glibc-dynamic image. This means we have a clean build and don't need to copy any libraries over to run the binary.

I tested the image in an Ubuntu VM - it worked identically to the official image for my simple tests and is roughly 1/2 the size of the official image.

amouat:
I've also removed the source Dockerfile as I don't think we're interested in it for the moment.

There is a final issue - the original script wraps the Falco version in a `urlencode` call, presumably to take care of spaces or similar. This function is provided by an old and somewhat obscure package `gridsite-clients`. We don't have this in Wolfi, but I could do something similar with `jq` if it is required functionality.

amouat:
Last commit uses `jq` to urlencode the variable.

leogr:
We still want this, tentatively for 0.36

/remove-lifecycle stale
cc @LucaGuerra

LucaGuerra:
I will also try with the latest nightly version, if it works there we should be good.

LucaGuerra:
It looks like the `bpf` syscall is failing with `-EPERM` which makes me wonder what's happening there, the image is running with the right set of permission.

amouat:
Could be capabilities on the file or similar, will take a look as soon as I can. A few things came up today.

amouat:
Had a quick look but not sorted it yet, looking at it tomorrow.

amouat:
It was fairly simple - it wants to run as root.

I've updated the Dockerfiles to set the user to root. We could try to figure out if we can use a non-root user by modifying privileges or capabilities, but I'm not sure it's worth it.

amouat:
@LucaGuerra I think you're working in a new PR, let me know if you need me to update this one or anything.

LucaGuerra:
Thank you so much @amouat ! I will port your changes to the other PR (with a commit that has your name of course) that is already rebased and has CI/CD, test it and close this one. We can then continue to work on this directly in the repo!

LucaGuerra:
The modifications have been ported to https://github.com/falcosecurity/falco/pull/2768 :rocket: . @amouat feel free to review the readme changes in #2768 if you have time. Thanks!

amouat:
Great!

I think the table is broken in the README, but otherwise this looks fantastic :)

LucaGuerra:
Thanks, fixed :)

PR #2519, opened 2023-04-28
fix(ci): fetch version step, download rpms/debs, minor change

LucaGuerra:
**What type of PR is this?**

/kind bug

/area CI

* A new job `fetch-version` was added to detect the Falco version, but not all dependent jobs were updated.
* Fixed downloading of artifacts since we were trying to use a wildcard which isn't really supported 😅
* Minor fix (security best practice) with handling the secret

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/reusable_build_packages.yaml
+++ b/.github/workflows/reusable_build_packages.yaml
@@ -132,21 +132,21 @@ jobs:
       - name: Upload Falco tar.gz package
         uses: actions/upload-artifact@v3
         with:
-          name: falco-${{ steps.store_version.outputs.version }}-${{ inputs.arch }}.tar.gz
+          name: falco-${{ needs.fetch-version.outputs.version }}-${{ inputs.arch }}.tar.gz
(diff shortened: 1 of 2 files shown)


FedeDP (dismissed):
/approve

FedeDP:
/milestone 0.35.0

Refs #2501

LucaGuerra:
I think this change is right but not enough to fix the latest error though

LucaGuerra:
Added a couple more things:
* fixed downloading of artifacts since we were trying to use a wildcard which isn't really supported 😅
* minor change with the `GPG_KEY` env variable use, instead of using it in all jobs it's only imported only in the two steps that actually use it. This is a security best practice but does not change the functionality of the jobs.

Andreagit97 (approved):
/approve

FedeDP (approved):
/approve
Thank you!

leogr (approved):
:+1:

PR #1719, opened 2021-09-01
Adding libssl-dev, fixing upstream image reference.

michalschott:
**What type of PR is this?**
/kind bug

**Any specific area of the project related to this PR?**
/area build

**What this PR does / why we need it**:
This PR updates base image with libssl-dev package required to compile kernel module.
It also updates upstream container reference, since recently Buster becomes new stable and container refuses to build.

**Which issue(s) this PR fixes**:

Fixes https://github.com/falcosecurity/falco/issues/1718

**Special notes for your reviewer**:

Since Debian Bullseye become new `stable`, it is also required to update base image reference.

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #2523, opened 2023-04-29
unpin binutils

lichao127:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind release

/area build

> /area engine

> /area tests

> /area proposals

> /area CI

`binutils` was pinned to version `2.30-22` in https://github.com/falcosecurity/falco/pull/397 because `binutils-2.31` requires kernel > 4.16. Current based image (`debian:buster`) is on kernel 5.10. There's no need to pin to the lower version. This will keep it patched.

Fixes #

NO

```release-note
NONE
```

### DIFF
--- a/docker/falco/Dockerfile
+++ b/docker/falco/Dockerfile
@@ -21,6 +21,7 @@ RUN apt-get update \
 	bash-completion \
 	bc \
 	bison \
+	binutils \
 	clang-7 \
 	ca-certificates \
 	curl \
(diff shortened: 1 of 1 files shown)


leogr:
cc @FedeDP @Andreagit97

FedeDP:
Hi! Thanks for this PR!
> binutils was pinned to version 2.30-22 in https://github.com/falcosecurity/falco/pull/397 because binutils-2.31 requires kernel > 4.16

I don't think that's true; the issue is that drivers built using newer binutils are not going to work on older kernels; that's what the comment says after all:
> debian:stable head contains binutils 2.31, which generates
> binaries that are incompatible with kernels < 4.16. So manually
> forcibly install binutils 2.30-22 instead.

Basically, we need to use old binutils version to have a wider compat matrix!

lichao127:
> I don't think that's true; the issue is that drivers built using newer binutils are not going to work on older kernels; that's what the comment says after all:
Basically, we need to use old binutils version to have a wider compat matrix!

Pinning binutils to an old version means it will not be patched for CVEs like CVE-2018-10372, CVE-2018-10534 etc. It also blocks other packages from security patching as a lot of packages are has dependencies on it.

Also, the base image has changed from `debian:stable` to `debian:buster` in https://github.com/falcosecurity/falco/pull/1719.
The comments added in https://github.com/falcosecurity/falco/pull/397 should either be updated to reflect the reason to pin binutils, or they should be cleaned up.

FedeDP:
> Pinning binutils to an old version means it will not be patched for CVEs like https://github.com/advisories/GHSA-pw2g-xg7w-mccx, https://github.com/advisories/GHSA-672m-f2rr-8wvq etc. It also blocks other packages from security patching as a lot of packages have dependencies on it.

Yep that's an issue, for sure. Still, we need the widest possible compatibility unfortunately.
> Also, the base image has changed from debian:stable to debian:buster in https://github.com/falcosecurity/falco/pull/1719.

Yes because `debian:stable` switched to bullseye  but we needed buster.

cc also @LucaGuerra

leogr:
cc @LucaGuerra

FedeDP:
Ehy, #2718 is updating the default Falco image to Debian 12 (keeping the current image as `legacy`). Do you think this would still be needed?

FedeDP:
Moving to TBD until further user input.
/milestone TBD

Andreagit97:
To be honest I would close it since the issue should be solved by https://github.com/falcosecurity/falco/pull/2718... Of course, if this is not the case feel free to re-open it! :)

PR #2533, opened 2023-05-04
new(ci): add RC/prerelease support

LucaGuerra:
**What type of PR is this?**

/kind feature
/area CI

In this new pipeline we need RC support for several reasons:
* We want to actually publish RCs before release for testing purposes
* Once we're in the release branch there was no way to test the latest version or to get a working image unless it is already released, which is painful, inconvenient and slows the release process
* We actually want to test the release flow and any new feature we add before we do an actual release :)

*How does it work?*

Once a maintainer clicks the release button on GitHub a tag is created and a GH release added as well. If the tag name is formatted like `x.y.z` the release is official, also if the "make latest" button is pressed the release is also latest (meaning that the `:latest` tag of the container image will be overwritten)

If, instead, the tag is formatted as `x.y.z-k` the release is a prerelease. Artifacts will go in the `-dev` bucket and the `:latest` tag will not be overwritten.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/master.yaml
+++ b/.github/workflows/master.yaml
@@ -9,46 +9,86 @@ concurrency:
   cancel-in-progress: true  
 
 jobs:
+  # We need to use an ubuntu-latest to fetch Falco version because
+  # Falco version is computed by some cmake scripts that do git sorceries
+  # to get the current version.
+  # But centos7 jobs have a git version too old and actions/checkout does not 
+  # fully clone the repo, but uses http rest api instead.
+  fetch-version:
+    runs-on: ubuntu-latest
+    # Map the job outputs to step outputs
+    outputs:
+      version: ${{ steps.store_version.outputs.version }}  
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          
+      - name: Install build dependencies
+        run: |
+          sudo apt update 
+          sudo apt install -y cmake build-essential
+      
+      - name: Configure project
+        run: |
+          mkdir build && cd build
+          cmake -DUSE_BUNDLED_DEPS=On ..
+          
+      - name: Load and store Falco version output
+        id: store_version
+        run: |
+          FALCO_VERSION=$(cat build/userspace/falco/config_falco.h | grep 'FALCO_VERSION ' | cut -d' ' -f3 | sed -e 's/^"//' -e 's/"$//')
+          echo "version=${FALCO_VERSION}" >> $GITHUB_OUTPUT     
+
   build-dev-packages:
+    needs: [fetch-version]
     uses: falcosecurity/falco/.github/workflows/reusable_build_packages.yaml@master
     with:
       arch: x86_64
+      version: ${{ needs.fetch-version.outputs.version }}
     secrets: inherit
   
   build-dev-packages-arm64:
+    needs: [fetch-version]
     uses: falcosecurity/falco/.github/workflows/reusable_build_packages.yaml@master
     with:
       arch: aarch64
+      version: ${{ needs.fetch-version.outputs.version }}
     secrets: inherit
     
   publish-dev-packages:
-    needs: [build-dev-packages, build-dev-packages-arm64]
+    needs: [fetch-version, build-dev-packages, build-dev-packages-arm64]
     uses: falcosecurity/falco/.github/workflows/reusable_publish_packages.yaml@master
     with:
       bucket_suffix: '-dev'
-      version: ${{ needs.build-dev-packages.outputs.version }}
+      version: ${{ needs.fetch-version.outputs.version }}
     secrets: inherit
   
   # Both build-dev-docker and its arm64 counterpart require build-dev-packages because they use its output
   build-dev-docker:
-    needs: [build-dev-packages, publish-dev-packages]
+    needs: [fetch-version, build-dev-packages, publish-dev-packages]
     uses: falcosecurity/falco/.github/workflows/reusable_build_docker.yaml@master
     with:
       arch: x86_64
       bucket_suffix: '-dev'
-      version: ${{ needs.build-dev-packages.outputs.version }}
+      version: ${{ needs.fetch-version.outputs.version }}
+      tag: ${{ github.ref_name }}
     secrets: inherit
     
   build-dev-docker-arm64:
-    needs: [build-dev-packages, publish-dev-packages]
+    needs: [fetch-version, build-dev-packages, publish-dev-packages]
     uses: falcosecurity/falco/.github/workflows/reusable_build_docker.yaml@master
     with:
       arch: aarch64
       bucket_suffix: '-dev'
-      version: ${{ needs.build-dev-packages.outputs.version }}
+      version: ${{ needs.fetch-version.outputs.version }}
+      tag: ${{ github.ref_name }}
     secrets: inherit
     
   publish-dev-docker:
-    needs: [build-dev-docker, build-dev-docker-arm64]
+    needs: [fetch-version, build-dev-docker, build-dev-docker-arm64]
     uses: falcosecurity/falco/.github/workflows/reusable_publish_docker.yaml@master
+    with:
+      tag: ${{ github.ref_name }}
     secrets: inherit
--- a/.github/workflows/master.yaml
+++ b/.github/workflows/master.yaml
@@ -65,9 +65,8 @@ jobs:
       version: ${{ needs.fetch-version.outputs.version }}
     secrets: inherit
   
-  # Both build-dev-docker and its arm64 counterpart require build-dev-packages because they use its output
   build-dev-docker:
-    needs: [fetch-version, build-dev-packages, publish-dev-packages]
+    needs: [fetch-version, publish-dev-packages]
     uses: falcosecurity/falco/.github/workflows/reusable_build_docker.yaml@master
     with:
       arch: x86_64
@@ -77,7 +76,7 @@ jobs:
     secrets: inherit
     
   build-dev-docker-arm64:
-    needs: [fetch-version, build-dev-packages, publish-dev-packages]
+    needs: [fetch-version, publish-dev-packages]
     uses: falcosecurity/falco/.github/workflows/reusable_build_docker.yaml@master
     with:
       arch: aarch64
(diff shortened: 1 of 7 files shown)


.github/workflows/release.yaml
```diff
@@ -2,14 +2,45 @@ name: Release Packages and Docker images
...
+
+          # Safeguard: you need to both set "latest" in GH and not have suffixes to overwrite latest
+          is_latest = '${{ steps.latest_release.outputs.release }}' == '${{ github.ref_name }}' and not is_prerelease
+
+          bucket_suffix = '-dev' if is_prerelease else ''
+
+          with open(os.environ['GITHUB_OUTPUT'], 'a') as ofp:
+            print(f'is_latest={is_latest}'.lower(), file=ofp)
+            print(f'bucket_suffix={bucket_suffix}', file=ofp)
+    outputs:
```
FedeDP, line 40: I'd move `outputs` right after the `runs-on` at the start of the job, so that it gets more highlighted.

.github/workflows/release.yaml
```diff
@@ -26,6 +57,7 @@ jobs:
     needs: [build-packages, build-packages-arm64]
     uses: falcosecurity/falco/.github/workflows/reusable_publish_packages.yaml@master
     with:
+      bucket_suffix: ${{ steps.get_settings.outputs.bucket_suffix }}
```
FedeDP, line 60: We need to add the `needs` on the `release-settings` job so that it is always performed before this job; then we need to use
```
${{ needs.release-settings.outputs.bucket_suffix }}
```
And same for other usages of `steps.get_settings` :)

.github/workflows/release.yaml
```diff
@@ -2,14 +2,45 @@ name: Release Packages and Docker images
...
+          # Safeguard: you need to both set "latest" in GH and not have suffixes to overwrite latest
+          is_latest = '${{ steps.latest_release.outputs.release }}' == '${{ github.ref_name }}' and not is_prerelease
+
+          bucket_suffix = '-dev' if is_prerelease else ''
+
+          with open(os.environ['GITHUB_OUTPUT'], 'a') as ofp:
+            print(f'is_latest={is_latest}'.lower(), file=ofp)
+            print(f'bucket_suffix={bucket_suffix}', file=ofp)
+    outputs:
+      is_latest: ${{ steps.get_settings.outputs.is_latest }}
```
FedeDP, line 41: I'd expose `is_latest` directly as a bool:
```
is_latest: ${{ steps.get_settings.outputs.is_latest == 'true' }} 
```

So that we can use it later without the string comparison.

.github/workflows/release.yaml
```diff
@@ -2,14 +2,45 @@ name: Release Packages and Docker images
...
+        id: latest_release
+        env:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          repository: ${{ github.repository }}
+          type: "stable"
+
+      - name: Get settings for this release
+        id: get_settings
+        shell: python
+        run: |
```
FedeDP, line 28: I tested the python code; it LGTM.

.github/workflows/release.yaml
```diff
@@ -2,14 +2,45 @@ name: Release Packages and Docker images
...
 concurrency:
   group: ci-release
   cancel-in-progress: true  
 
 jobs:
+  release-settings:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get latest release
+        uses: rez0n/actions-github-release@v2.0
```
FedeDP, line 18: I have one concern: our release process states that we push a tag and then open the draft release.
I think this step won't find the release when it gets triggered on tag push; eg:
* `git tag 0.35.0 && git push origin 0.35.0`
* the release workflow is started
* we haven't still opened the new release (and we would open it in draft btw: https://github.com/falcosecurity/falco/blob/master/RELEASE.md#2-update-the-github-release)
* the action will find `0.34.1` as latest release
* we are doomed

LucaGuerra: I forgot about the possibility of a draft release! That's a great catch, thank you!

.github/workflows/release.yaml
LucaGuerra: That's because it's not bash :)

.github/workflows/release.yaml
LucaGuerra: I have an honest question here, does GHA support boolean outputs at all? From here https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idoutputs it looked to me that they can only be strings :/

.github/workflows/release.yaml
LucaGuerra: Wait a second though, do we want to start this on manual tag or release creation? We could switch to start on GH release creation so that whether a tag already existed or not the process will be the exact same.

FedeDP: The current release workflow starts on tags pushing, that what CircleCI does and what our release document states!

.github/workflows/release.yaml
LucaGuerra: Done!

.github/workflows/release.yaml
LucaGuerra: That was a great catch.

.github/workflows/release.yaml
leogr: Hey @LucaGuerra 
I recommend updating https://github.com/falcosecurity/falco/blob/master/RELEASE.md in this PR directly.

LucaGuerra: Done!

RELEASE.md
```diff
@@ -176,8 +177,9 @@ Assume `M.m.p` is the new version.
     ```
 
 - Finally, publish the release!
+- The release pipeline will start automatically upon publication and all packages and container images will be uploaded to the stable repositories.
```
leogr, line 180: @LucaGuerra 
How can the release manager get that artifacts have been published? Generally, what should the release manager check before moving forward?

leogr (commented):
Just one suggestion (see my comment below). Otherwise SGMT.

RELEASE.md
LucaGuerra: Just to make sure I understand, is your suggestion to add the link or instructions (i.e. click on the actions button) to take a look at the github actions pipeline to check in on the progress? In that case I agree!

leogr: Yes! Link + instructions. :pray:

.github/workflows/release.yaml
```diff
@@ -1,53 +1,110 @@
...
+          bucket_suffix = '-dev' if is_prerelease else ''
+
+          with open(os.environ['GITHUB_OUTPUT'], 'a') as ofp:
+            print(f'is_latest={is_latest}'.lower(), file=ofp)
+            print(f'should_publish={should_publish}'.lower(), file=ofp)
+            print(f'bucket_suffix={bucket_suffix}', file=ofp)
+
   build-packages:
+    needs: [release-settings]
+    if: ${{ needs.release-settings.outputs.should_publish == 'true' }}
```
FedeDP, line 53: I'd still build artifacts even if we are not going to publish them (they can still be accessed as github action run artifact).

RELEASE.md
LucaGuerra: Done!

.github/workflows/master.yaml
```diff
@@ -9,46 +9,86 @@ concurrency:
...
   build-dev-docker-arm64:
-    needs: [build-dev-packages, publish-dev-packages]
+    needs: [fetch-version, build-dev-packages, publish-dev-packages]
     uses: falcosecurity/falco/.github/workflows/reusable_build_docker.yaml@master
     with:
       arch: aarch64
       bucket_suffix: '-dev'
-      version: ${{ needs.build-dev-packages.outputs.version }}
+      version: ${{ needs.fetch-version.outputs.version }}
+      tag: ${{ github.ref_name }}
```
FedeDP, line 86: I'd use `master` directly here in place of `github.ref_name`; being more explicit is fine and desired given how complex the actions are ;)

.github/workflows/release.yaml
LucaGuerra: So, should_publish means that this is a release with a tag name we understand (e.g. `0.1.2-rc1` or `3.4.5`) and will be false for anything that is not that (a release tagged `foo`). I'm not sure what should we do in those cases so I picked nothing. Do you have suggestions? A better variable name?

.github/workflows/master.yaml
```diff
@@ -9,46 +9,86 @@ concurrency:
...
       arch: x86_64
       bucket_suffix: '-dev'
-      version: ${{ needs.build-dev-packages.outputs.version }}
+      version: ${{ needs.fetch-version.outputs.version }}
+      tag: ${{ github.ref_name }}
     secrets: inherit
     
   build-dev-docker-arm64:
-    needs: [build-dev-packages, publish-dev-packages]
+    needs: [fetch-version, build-dev-packages, publish-dev-packages]
```
FedeDP, line 80: We can skip the dep on `build-dev-packages` now that we have a `fetch-version` step.
THat was needed before to ensure we could use the `build-dev-packages` version output.

FedeDP: SAme for `build-dev-docker` ofc

LucaGuerra: Makes total sense, updated!

LucaGuerra: They both depend on the same jobs, `fetch-version` for the version and `build-dev-packages` which will complete successfully once all arch packages have been uploaded.

.github/workflows/release.yaml
LucaGuerra: fail the pipeline upon a `git tag`? would that make sense? honest question

FedeDP: IMHO yep since that's a bug for us (ie: we don't want to support non semversioned tags i think).

FedeDP: If `should_publish` is false the whole workflow won't do nothing, right? At least if we fail immediately, we will have the nice red cross telling us something went wrong.

leogr (approved):
LGTM

FedeDP (approved):
Lgtm! 
/approve

PR #2542, opened 2023-05-11
fix(ci): simplify and fix multi-arch image publishing process

LucaGuerra:
**What type of PR is this?**

/kind bug
/kind cleanup

/area CI

This is a refactor and fix of the container image publishing process. There are two major changes:
1. Images with arch-specific tags are pushed to Docker Hub (e.g. `:x86_64-0.35.0`). This is done because in order to create a multiarch (`index`, `manifest.list`) manifest easily we need manifests with the referenced digests already published. It is probably possible to do it without (buildx can do something similar) but it may involve a local registry or some more complicated process. I think this is fine for now.
2. Instead of building images multiple times and then publishing all of them, this one only builds one version (`docker.io/falcosecurity/falco`) and then adds any additional tags or registries (ECR, latest tags...). This simplifies things especially when the next PR will be published, which is related to signatures with cosign.

Fixes #

Note that the arch specific tags are not pushed to ECR. We can change this behavior

```release-note
NONE
```

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -80,7 +80,6 @@ jobs:
     uses: falcosecurity/falco/.github/workflows/reusable_build_docker.yaml@master
     with:
       arch: x86_64
-      is_latest: ${{ needs.release-settings.outputs.is_latest == 'true' }}
       bucket_suffix: ${{ needs.release-settings.outputs.bucket_suffix }}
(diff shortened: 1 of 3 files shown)


FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

LucaGuerra:
Refs https://github.com/falcosecurity/falco/pull/2501

Issue #2372, opened 2023-01-25
[New Feature]: Dry run option that outputs the syscalls used by rules


incertum:
**Motivation + Feature**

As extension to https://github.com/falcosecurity/falco/issues/2371, "Dry run option that outputs the syscalls used by rules" proposed by @Happy-Dude. 

Quote @Happy-Dude from https://github.com/falcosecurity/falco/pull/2361#issuecomment-1399402369

> is it possible to have a "dry-run" option that reads in the ruleset (rules.yaml) and outputs the set of syscalls that will be inspected
> - as part of the Falco state engine
> - from the dynamic set (rules.yaml)
> - and from the additional set?
>This can help for debugging by making it clear what is causing a particular syscall to be inspected.

Quote @jasondellaluce from https://github.com/falcosecurity/falco/pull/2361#issuecomment-1401652172

> Dry run option that outputs the syscalls used by rules: I see this request falling into the umbrella of "printing descriptive information about one or more rules (up until the whole ruleset)". We have other feature requests like this (see: https://github.com/falcosecurity/falco/issues/1814), and I think we should make this happen. In my opinion, enhancing --list-syscalls might not be the right call because it's only supposed to print static data. Instead, we also have -L and -l that are meant to print informations about rules specifically. I think we could enhance the output of those commands to include the set of matching events and other info as well. Plus, I'm recently advocating the idea of making all Falco outputs optionally machine-readable in JSON format in the future (see special notes for your reviewer in: https://github.com/falcosecurity/falco/pull/2351), and in which case we could include extra information in the JSON output compared to the plain-text counterpart.

FedeDP:
@incertum should we close this?

Happy-Dude:
Works for me! I'm wondering if something like that is necessary now after having gained some experience writing rules and with the utility of `base_syscalls` 🤔

But I digress! Feel free to close and I'll open another issue ticket if the need rises 👍

PR #2544, opened 2023-05-12
new: add possibility to retrieve details about rules with filter_details_visitor

loresuso:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind release

> /area build

/area engine

> /area tests

> /area proposals

> /area CI

This PR introduces a specialized visitor that can be used to construct a normalized form of a rule, containing all the macros, lists, operators, and fields it uses in JSON format. The functionality is embedded in the `-l`/`-L` flags and can be triggered when the output is set to JSON. 
This functionality can be used later in CI tools for rules to e.g understand how to automatically handle and increase rules version.

Example:
```
$ sudo ./userspace/falco/falco -c ../falco.yaml -l "Launch Privileged Container" -o "json_output=true" | jq 
Fri May 12 10:23:13 2023: Falco version: 0.34.1-194+f5c7574 (x86_64)
Fri May 12 10:23:13 2023: Falco initialized with configuration file: ../falco.yaml
Fri May 12 10:23:13 2023: Loading rules from file /etc/falco/falco_rules.yaml
Fri May 12 10:23:13 2023: Loading rules from file /etc/falco/falco_rules.local.yaml
{
  "conditionFields": [
    "evt.num",
    "container.id",
    "container.privileged",
    "container.image.repository",
    "proc.vpid",
    "evt.dir",
    "evt.type"
  ],
  "description": "Detect the initial process started in a privileged container. Exceptions are made for known trusted images.",
  "exceptionFields": [],
  "lists": [
    "redhat_io_images_privileged",
    "falco_privileged_images",
    "trusted_images"
  ],
  "macros": [
    "redhat_image",
    "aws_eks_core_images",
    "user_trusted_containers",
    "allowed_openshift_registry_root",
    "openshift_image",
    "never_true",
    "falco_privileged_containers",
    "user_privileged_containers",
    "container",
    "spawned_process",
    "allowed_aws_ecr_registry_root_for_eks",
    "container_started"
  ],
  "name": "Launch Privileged Container",
  "operators": [
    "endswith",
    "startswith",
    "in",
    "!=",
    "="
  ],
  "outputFields": [
    "user.name",
    "user.loginuid",
    "proc.cmdline",
    "proc.pid",
    "container.name",
    "container.id",
    "container.image.repository",
    "container.image.tag"
  ],
  "priority": "Informational"
}
```

Fixes #

```release-note
update: get rules details with `-l` or `-L` flags when json output format is specified
```

### DIFF
--- a/userspace/engine/falco_engine.cpp
+++ b/userspace/engine/falco_engine.cpp
@@ -527,6 +527,8 @@ Json::Value falco_engine::get_json_rule_details(const falco_rule& r, filter_deta
 	output["name"] = r.name;
 	output["description"] = r.description;
 	output["priority"] = format_priority(r.priority, false);
+	output["output"] = r.output;
+	output["enabled"] = rule_info->enabled;
 
 	Json::Value macros = Json::arrayValue;
 	for(const auto &m : details.macros)
--- a/userspace/engine/falco_engine.cpp
+++ b/userspace/engine/falco_engine.cpp
@@ -462,29 +462,23 @@ void falco_engine::describe_rule(std::string *rule, bool json) const
 
 	Json::FastWriter writer;
 	std::string json_str;
-	filter_details details;
-	for(const auto &m : m_rule_collector.macros())
-	{
-		// Assumption: no exception because rules have already been loaded.
-		auto cond_ast = libsinsp::filter::parser(m.cond).parse();
-		std::shared_ptr<libsinsp::filter::ast::expr> cond_ast_ptr = std::move(cond_ast);
-		details.known_macros[m.name] = cond_ast_ptr;
-	}
-
-	for(const auto &l : m_rule_collector.lists())
-	{
-		details.known_lists.insert(l.name);
-	}
 
 	if(!rule)
 	{
+		// In this case we build json information about
+		// all rules, macros and lists
 		Json::Value output;
 
+		// Store information about rules
 		Json::Value rules_array = Json::arrayValue;
 		for(const auto& r : m_rules)
 		{
-			auto json_details = get_json_rule_details(r, details);
-			rules_array.append(json_details);
+			auto ri = m_rule_collector.rules().at(r.name);
+			Json::Value rule;
+			get_json_details(r, *ri, rule);
+
+			// Append to rule array
+			rules_array.append(rule);
 		}
 		output["rules"] = rules_array;
 		
@@ -492,7 +486,9 @@ void falco_engine::describe_rule(std::string *rule, bool json) const
 		Json::Value macros_array;
 		for(const auto &m : m_rule_collector.macros())
 		{
-			macros_array.append(m.name);
+			Json::Value macro;
+			get_json_details(m, macro);
+			macros_array.append(macro);
 		}
 		output["macros"] = macros_array;
 
@@ -501,16 +497,8 @@ void falco_engine::describe_rule(std::string *rule, bool json) const
 		for(const auto &l : m_rule_collector.lists())
 		{
 			Json::Value list;
-			list["name"] = l.name;
-
-			Json::Value items = Json::arrayValue;
-			for(const auto& i : l.items)
-			{
-				items.append(i);
-			}
-			list["items"] = items;
-
-			lists_array.append(list);
+			get_json_details(l, list);
+			lists_array.append(list);			
 		}
 		output["lists"] = lists_array;
 
@@ -518,30 +506,49 @@ void falco_engine::describe_rule(std::string *rule, bool json) const
 	} 
 	else
 	{
-		auto r = m_rules.at(*rule);
-		if(r == nullptr)
+		// build json information for just the specified rule
+		auto ri = m_rule_collector.rules().at(*rule);
+		if(ri == nullptr)
 		{
 			throw falco_exception("Rule \"" + *rule + "\" is not loaded");
 		}
-
-		auto json_details = get_json_rule_details(*r, details);
-		json_str = writer.write(json_details);
+		auto r = m_rules.at(ri->name);
+		Json::Value rule; 
+		get_json_details(*r, *ri, rule);
+		json_str = writer.write(rule);
 	}
 
 	fprintf(stdout, "%s", json_str.c_str());
 }
 
-Json::Value falco_engine::get_json_rule_details(const falco_rule& r, filter_details& details) const
+void falco_engine::get_json_details(const falco_rule &r,
+	const rule_loader::rule_info &ri,
+	Json::Value &rule) const
 {
-	// Parse rule condition and build the AST
-	// Assumption: the parsing will not throw an exception because
-	// rules have already been loaded.
-	auto rule_info = m_rule_collector.rules().at(r.name);
-	auto ast = libsinsp::filter::parser(rule_info->cond).parse();
+	Json::Value rule_info;
 
-	// Resolve the AST details
-	filter_details_resolver resolver;
-	resolver.run(ast.get(), details);
+	// Fill general rule information
+	rule_info["name"] = r.name;
+	rule_info["condition"] = ri.cond;
+	rule_info["priority"] = format_priority(r.priority, false);
+	rule_info["output"] = r.output;
+	rule_info["description"] = r.description;
+	rule_info["enabled"] = ri.enabled;
+	rule_info["source"] = r.source;
+	Json::Value tags = Json::arrayValue;
+	for(const auto &t : ri.tags)
+	{
+		tags.append(t);
+	}
+	rule_info["tags"] = tags;
+	rule["info"] = rule_info;
+
+	// Parse rule condition and build the AST
+	// Assumption: no exception because rules have already been loaded.
+	auto ast = libsinsp::filter::parser(ri.cond).parse();
+	Json::Value json_details;
+	get_json_details(ast.get(), json_details);
+	rule["details"] = json_details;
 
 	// Get fields from output string
 	auto insp = new sinsp;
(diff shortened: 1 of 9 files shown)


loresuso:
cc @jasondellaluce

FedeDP:
This is super cool and highly requested! Thank you @loresuso !

jasondellaluce:
Links to https://github.com/falcosecurity/falco/issues/2372 and potentially fixes it, if we just add the information about the `evt.type`s matched by each rule.

loresuso:
Last two commits address the following:
- retrieving information about enabled rules
- information about lists and macros when the global -L option is specified
This will allow to check for instance:
- if a rule used to be disabled and then enabled, patch number of version of the rule must be increased
- adding, removing elements from rules, same as above

Currently working on adding `evt.type` information

userspace/engine/rule_loader_compiler.cpp
```diff
@@ -386,7 +386,11 @@ void rule_loader::compiler::compile_rule_infos(
 	std::string err, condition;
 	std::set<falco::load_result::load_result::warning_code> warn_codes;
 	filter_warning_resolver warn_resolver;
-	for (auto &r : col.rules())
+
+	// note: cast away the const qualifier in the for loop
+	// this is needed because we want to store information about evttypes
+	// used by any rules, which might come in handy when describing rules.
+	for (auto &r : const_cast<indexed_vector<rule_info>&>(col.rules()))
```
loresuso, line 393: Added information about evttypes as well. However, there is a little bit of a hack.
The collector returns rule infos in a const indexed vector. The only part of code that has information about evt codes (or the complete ast with all macros resolved) is this one. So we may want to add this information to rule_info, but we can't given the const qualifier. For this reason we a couple of options:
- stop using const. This is not considerable as a good option, since const correctness is correctly throughout this part of code, and I don't like remove it just for this use case. Removing it here would touch a chain of const object, so it would have effect also in other parts of the code
- use `const_cast` to cast away the const qualifier only in this case, and to not break the clean interface we have with these classes. This is the cleanest solution I found to this problem, happy to discuss alternatives if any!

userspace/falco/app/actions/load_rules_files.cpp
```diff
@@ -126,13 +126,13 @@ falco::app::run_result falco::app::actions::load_rules_files(falco::app::state&
 
 	if (s.options.describe_all_rules)
```
Andreagit97, line 127: I would initialize both `describe_all_rules` and `describe_rule` variables here

```cpp
options::options()
	: event_buffer_format(sinsp_evt::PF_NORMAL),
	  gvisor_config(""),	
	  list_fields(false),
	  list_plugins(false),
	  list_syscall_events(false),
	  markdown(false),
	  modern_bpf(false)
{
}
```

Andreagit97 (commented):
Left some minor comments, LGTM!

userspace/engine/filter_details_resolver.h
```diff
@@ -0,0 +1,75 @@
...
+	std::unordered_set<std::string> operators;
+	std::unordered_set<std::string> lists;
+
+	void reset();
+};
+
+/*!
+	\brief Helper class for getting details about rules' filters.
+*/
+class filter_details_resolver
```
Andreagit97, line 42: maybe we could add some tests to this class

userspace/engine/falco_engine.cpp
```diff
@@ -427,25 +431,204 @@ std::size_t falco_engine::add_source(const std::string &source,
...
+
+	Json::FastWriter writer;
+	std::string json_str;
+	filter_details details;
+	for(const auto &m : m_rule_collector.macros())
+	{
+		// Assumption: no exception because rules have already been loaded.
+		auto cond_ast = libsinsp::filter::parser(m.cond).parse();
+		std::shared_ptr<libsinsp::filter::ast::expr> cond_ast_ptr = std::move(cond_ast);
+		details.known_macros[m.name] = cond_ast_ptr;
```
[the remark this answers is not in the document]
loresuso: Slightly reworked the PR, not needed anymore :)

userspace/engine/filter_details_resolver.h
loresuso: Agreed, I am adding this

userspace/falco/app/actions/load_rules_files.cpp
loresuso: They should already have their default I guess in options.cpp

userspace/engine/falco_engine.cpp
```diff
@@ -427,25 +431,204 @@ std::size_t falco_engine::add_source(const std::string &source,
 	return m_sources.insert(src, source);
 }
 
-void falco_engine::describe_rule(std::string *rule) const
+void falco_engine::describe_rule(std::string *rule, bool json) const
```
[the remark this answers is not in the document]
loresuso: Tried to address this, but figured that is better like it is already. Normally I agree with you, but here we have different behaviors when listing all rules in json or all rules in the old format, that is with the first one we have also to append all macros and lists information. However, with last commits duplication of code is reduced to minimum :)

userspace/engine/falco_engine.cpp
```diff
@@ -427,27 +431,288 @@ std::size_t falco_engine::add_source(const std::string &source,
...
+	{
+		if(m_rule_collector.lists().at(i) != nullptr)
+		{
+			lists.append(i);
+			continue;
+		}
+		items.append(i);
+	}
+
+	list["details"]["items"] = items;
```
jasondellaluce, line 642: Shouldn't `items` be inside `info` alongside `name`?

userspace/engine/falco_engine.cpp
```diff
@@ -427,27 +431,288 @@ std::size_t falco_engine::add_source(const std::string &source,
...
+
+	// Parse rule condition and build the AST
+	// Assumption: no exception because rules have already been loaded.
+	auto ast = libsinsp::filter::parser(ri.cond).parse();
+	Json::Value json_details;
+	get_json_details(ast.get(), json_details);
+	rule["details"] = json_details;
+
+	// Get fields from output string
+	auto insp = new sinsp;
```
jasondellaluce, line 554: I'd suggest passing an inspector as a parameter to the function, so that we don't have to re-alloc a new one for every rule. Moreover, i suggest wrapping it into a unique_ptr to make sure it is properly deallocated without having to deal with manual deletion. Alternatively, you could do a stack allocation and just pass the reference.

userspace/engine/falco_engine.cpp
```diff
@@ -427,27 +431,288 @@ std::size_t falco_engine::add_source(const std::string &source,
...
+
+	// Get operators from exceptions
+	Json::Value exception_operators = Json::arrayValue;
+	for(const auto &e : ri.exceptions)
+	{
+		if(e.comps.is_list)
+		{
+			for(const auto& c : e.comps.items)
+			{
+				exception_operators.append(c.item);
```
jasondellaluce, line 582: I truly don't remember if `c` can be a list here. Just to be sure, I would do a `c.is_list` check here.

userspace/engine/falco_engine.cpp
```diff
@@ -427,27 +431,288 @@ std::size_t falco_engine::add_source(const std::string &source,
...
+	sinsp_evt_formatter fmt(insp, r.output);
+	std::vector<std::string> out_fields;
+	fmt.get_field_names(out_fields);
+	delete insp;
+	Json::Value outputFields = Json::arrayValue;
+	for(const auto &of : out_fields)
+	{
+		outputFields.append(of);
+	}
+	rule["details"]["outputFields"] = outputFields;
```
jasondellaluce, line 564: Question here and for the ones below: should we use camel case for the json fields? I personally like it, but let's try to be consistent with the other Falco json outputs. If we usually adopt snake_case, then I'd suggest using it here too.

userspace/engine/falco_engine.cpp
```diff
@@ -427,27 +431,288 @@ std::size_t falco_engine::add_source(const std::string &source,
...
+		{
+			auto evt_info = libsinsp::events::info(e);
+			auto res = evts.insert(std::string(evt_info->name));
+			if(res.second)
+			{
+				output.append(evt_info->name);
+			}
+		}
+	}
+}
```
jasondellaluce, line 713:
```suggestion
	auto evtcodes = libsinsp::filter::ast::ppm_event_codes(ast);
	if(evtcodes.size() != libsinsp::events::all_event_set().size())
	{
		auto syscodes = libsinsp::filter::ast::ppm_sc_codes(ast);
		for (const auto& n : libsinsp::events::sc_set_to_event_names(syscodes))
		{
			output.append(n);
		}
		for (const auto& n : libsinsp::events::event_set_to_names(evtcodes, false))
		{
			output.append(n);
		}
	}
}
```

I know it's not intuitive... but I think that's what we want (this is how evt.type works under the hood):
- names from the event table for all non-generic events
- names from the syscall table for all generic events

userspace/engine/filter_details_resolver.cpp
```diff
@@ -0,0 +1,96 @@
...
+{
+
+}
+
+void filter_details_resolver::visitor::visit(ast::binary_check_expr* e)
+{
+	m_details.fields.insert(e->field);
+	m_details.operators.insert(e->op);
+	
+	auto list = dynamic_cast<ast::list_expr*>(e->value.get());
```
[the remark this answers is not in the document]
loresuso: Thanks for this comment! I tried to address it in the last commit, mind to take a look again?

userspace/engine/filter_details_resolver.cpp
```diff
@@ -0,0 +1,96 @@
...
+}
+
+void filter_details_resolver::visitor::visit(ast::unary_check_expr* e)
+{
+	m_details.fields.insert(e->field);
+	m_details.operators.insert(e->op);
+}
+
+void filter_details_resolver::visitor::visit(ast::value_expr* e)
+{
```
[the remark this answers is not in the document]
loresuso: Same as above

userspace/engine/falco_engine.cpp
loresuso: Yes, agreed!

userspace/engine/falco_engine.cpp
loresuso: Agreed as well and addressed

userspace/engine/falco_engine.cpp
loresuso: Thank you, I totally missed this. Added this suggestion plus the merging of the two sets so that we don't have duplicates

userspace/engine/falco_engine.cpp
loresuso: In case it can be a list, what would you do? We should use a recursive approach or just check that it is not a list?
I would do something like:
```
if(!c.is_list)
  exception_operators.append(c.item)
```
but that alone wouldn't address the case of a list

userspace/engine/filter_details_resolver.cpp
```diff
@@ -0,0 +1,100 @@
...
+}
+
+void filter_details_resolver::visitor::visit(ast::and_expr* e)
+{
+	for(size_t i = 0; i < e->children.size(); i++)
+	{
+		m_expect_macro = true;
+		e->children[i]->accept(this);
+	}
+}
```
jasondellaluce, line 42:
```suggestion
		m_expect_macro = false;
	}
}
```

userspace/engine/filter_details_resolver.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+void filter_details_resolver::visitor::visit(ast::or_expr* e)
+{
+	for(size_t i = 0; i < e->children.size(); i++)
+	{
+		m_expect_macro = true;
+		e->children[i]->accept(this);
+	}
+}
```
jasondellaluce, line 51:
```suggestion
		m_expect_macro = false;
	}
}
```

userspace/engine/filter_details_resolver.cpp
```diff
@@ -0,0 +1,100 @@
...
+void filter_details_resolver::visitor::visit(ast::not_expr* e)
+{
+	e->child->accept(this);
+}
+
+void filter_details_resolver::visitor::visit(ast::list_expr* e)
+{
+	if(m_expect_list)
+	{
+		m_expect_list = false;
```
jasondellaluce, line 62:
```suggestion
```

userspace/engine/filter_details_resolver.cpp
```diff
@@ -0,0 +1,100 @@
...
+		}
+	}
+}
+
+void filter_details_resolver::visitor::visit(ast::binary_check_expr* e)
+{
+	m_details.fields.insert(e->field);
+	m_details.operators.insert(e->op);
+	m_expect_list = true;
+	e->value->accept(this);
```
jasondellaluce, line 78:
```suggestion
	m_expect_list = true;
	e->value->accept(this);
	m_expect_list = false;
```

userspace/engine/filter_details_resolver.cpp
```diff
@@ -0,0 +1,100 @@
...
+{
+	m_details.fields.insert(e->field);
+	m_details.operators.insert(e->op);
+}
+
+void filter_details_resolver::visitor::visit(ast::value_expr* e)
+{
+	if(m_expect_macro)
+	{
+		m_expect_macro = false;
```
jasondellaluce, line 91:
```suggestion
```

userspace/engine/filter_details_resolver.cpp
```diff
@@ -0,0 +1,100 @@
...
+			{
+				m_details.lists.insert(item);
+			}
+		}
+	}
+}
+
+void filter_details_resolver::visitor::visit(ast::binary_check_expr* e)
+{
+	m_details.fields.insert(e->field);
```
jasondellaluce, line 75:
```suggestion
	m_expect_macro = false;
	m_details.fields.insert(e->field);
```

userspace/engine/filter_details_resolver.cpp
```diff
@@ -0,0 +1,100 @@
...
+	m_details.operators.insert(e->op);
+	m_expect_list = true;
+	e->value->accept(this);
+}
+
+void filter_details_resolver::visitor::visit(ast::unary_check_expr* e)
+{
+	m_details.fields.insert(e->field);
```
jasondellaluce, line 83:
```suggestion
	m_expect_macro = false;
	m_details.fields.insert(e->field);
```

userspace/engine/falco_engine.cpp
jasondellaluce: I think you could just do another level of `if(!c.is_list) { ... } else { ...}`. Recursion would be the safest option but I don't think exceptions go beyond 2 levels of lists.

jasondellaluce (commented):
Looks good Lore:
- Left you a simple comment on that thing about exceptions and list of operators
- Left you suggestions to cover corner cases in the macro/list visitor search
- Once all's done, we'll need to rebase on top of https://github.com/falcosecurity/falco/pull/2552, that should bump the libs and hopefully fix the broken tests

loresuso:
Everything's green, I think we are there now!

Andreagit97 (approved):
/approve

PR #2552, opened 2023-05-17
update!: bump libs version, and support latest plugin features, add --nodriver option

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind feature

/area build

/area engine

This PR updates the libs and driver dependencies to the latest mainline of falcosecurity/libs. This brings in all the latest changes of the plugin framework, including the new plugin API major version change 3.0.0 (which can be seen as a breaking changes, old plugins based on the API 2.X.X will be rejected by future Falco versions).

This also supports plugins with the new event parsing and async events capabilities. Accordingly, a new `--nodriver` option has been introduced to open live captures without drivers. This can be used when loading with plugins generating system events to instruct Falco about using the plugins for the system events capture instead of one of the drivers.

Also, this carries a couple of fixes in Falco, such as properly constructing the config object (which caused spurious test failures), and making all the different open options mutually exclusive (e.g. `-u`, `--moder-bpf`, etc...).

```release-note
update!: bump libs version, and support latest plugin features, add --nodriver option
```


userspace/falco/app/options.cpp
```diff
@@ -147,6 +149,19 @@ bool options::parse(int argc, char **argv, std::string &errstr)
 
 	list_fields = m_cmdline_parsed.count("list") > 0 ? true : false;
 
+	int open_modes = 0;
+	open_modes += !trace_filename.empty();
```
Andreagit97, line 153: it seems like a joke that every engine is opened in a different way but that's the sad truth :joy: maybe one day we will us an enum

userspace/falco/app/actions/init_falco_engine.cpp
```diff
@@ -117,7 +117,7 @@ falco::app::run_result falco::app::actions::init_falco_engine(falco::app::state&
 		auto manager = s.offline_inspector->get_plugin_manager();
 		for (const auto &p : manager->plugins())
 		{
-			if (p->caps() & CAP_SOURCING)
+			if (p->caps() & CAP_SOURCING && p->id() != 0)
```
Andreagit97, line 120: this check is used in more than one place probably at a certain point we could create an helper for it

Andreagit97 (approved):
/approve

cmake/modules/driver.cmake
```diff
@@ -26,8 +26,8 @@ else()
   # In case you want to test against another driver version (or branch, or commit) just pass the variable -
   # ie., `cmake -DDRIVER_VERSION=dev ..`
   if(NOT DRIVER_VERSION)
-    set(DRIVER_VERSION "ffcd702cf22e99d4d999c278be0cc3d713c6375c")
-    set(DRIVER_CHECKSUM "SHA256=7ed19cd8e13887b02c823f5167aac3d1e997b0a60e305979848dfb339e1b774d")
+    set(DRIVER_VERSION "0b9ca98fee2453a16f4538db55dcfa34bc8f5aef")
```
FedeDP, line 29: I'd directly use `5.0.0+driver` here.

userspace/falco/app/actions/init_falco_engine.cpp
FedeDP: I agree, i think a small method would help making the code more readable.

leogr (approved):

userspace/engine/rule_loader_compiler.cpp
```diff
@@ -495,12 +495,11 @@ void rule_loader::compiler::compile_rule_infos(
...
 		// populate set of event types and emit an special warning
-		libsinsp::events::set<ppm_event_code> evttypes = { ppm_event_code::PPME_PLUGINEVENT_E };
 		if(rule.source == falco_common::syscall_source)
 		{
-			evttypes = libsinsp::filter::ast::ppm_event_codes(ast.get());
-			if ((evttypes.empty() || evttypes.size() > 100)
-			    && r.warn_evttypes)
+			auto evttypes = libsinsp::filter::ast::ppm_event_codes(ast.get());
+			evttypes.insert(ppm_event_code::PPME_ASYNCEVENT_E);
+			if ((evttypes.empty() || evttypes.size() > 100) && r.warn_evttypes)
```
gnosek, line 502: evvtypes.empty() can't ever be true here since we've just added PPME_ASYNCEVENT_E, right?

jasondellaluce: Good catch, thanks!

FedeDP (approved):
All green!
/approve

Great job @jasondellaluce !

Andreagit97 (approved):
/approve

PR #2562, opened 2023-05-22
feat: support parsing of system environment variables in yaml

### LINKED ISSUE #2561: Support Environment Variable Values in Configuration File
**Motivation**

When setting up Falco in my cluster, I could not see a way to use environment variables in the config file to store sensitive values such as secrets.

**Feature**

When a sure specifies an environment variable using standard techniques such as `$SOME_ENV_VAR` or `${SOME_ENV_VAR}`, the configuration is parsed in the environment config.

**Alternatives**

Default environment variables for certain configuration options would be another option, I could find this on the docs after a cursory look.

**Additional context**

N/A

therealdwright:
**What type of PR is this?**

/kind feature

/area engine
/area tests

In order to allow the user to supply environment variables in standard
ways performed in other applications the get_scalar function has been
extended to support defining an environment variable in the format
`${FOO}`. Environment variables can be escaped via defining as `$${FOO}`. 
As this handles some additional complexity, a unit test has been  added 
to cover this new functionality

Fixes  #2561

```release-note
feat: support parsing of system environment variables in yaml
```

### DIFF
--- a/unit_tests/falco/test_configuration.cpp
+++ b/unit_tests/falco/test_configuration.cpp
@@ -88,16 +88,88 @@ TEST(Configuration, modify_yaml_fields)
 {
 	std::string key = "base_value.subvalue.subvalue2.boolean";
 	yaml_helper conf;
-	
+
     /* Get original value */
     conf.load_from_string(sample_yaml);
 	ASSERT_EQ(conf.get_scalar<bool>(key, false), true);
-	
+
     /* Modify the original value */
     conf.set_scalar<bool>(key, false);
 	ASSERT_EQ(conf.get_scalar<bool>(key, true), false);
-	
+
     /* Modify it again */
     conf.set_scalar<bool>(key, true);
 	ASSERT_EQ(conf.get_scalar<bool>(key, false), true);
+}
+
+TEST(Configuration, configuration_environment_variables)
+{
+    // Set an environment variable for testing purposes
+    std::string env_var_value = "envVarValue";
+    std::string env_var_name = "ENV_VAR";
+    std::string default_value = "default";
+    setenv(env_var_name.c_str(), env_var_value.c_str(), 1);
+    yaml_helper conf;
+
+    std::string sample_yaml =
+        "base_value:\n"
+        "    id: $ENV_VAR\n"
+        "    name: '${ENV_VAR}'\n"
+        "    string: my_string\n"
+        "    invalid: $${ENV_VAR}\n"
+		"    invalid_env: $$ENV_VAR\n"
+        "    escaped: \"${ENV_VAR}\"\n"
+        "    subvalue:\n"
+        "        subvalue2:\n"
+        "            boolean: ${UNSED_XX_X_X_VAR}\n"
+        "base_value_2:\n"
+        "    sample_list:\n"
+        "        - ${ENV_VAR}\n"
+        "        - ' ${ENV_VAR}'\n"
+        "        - $UNSED_XX_X_X_VAR\n";
+    conf.load_from_string(sample_yaml);
+
+    /* Check if the base values are defined */
+    ASSERT_TRUE(conf.is_defined("base_value"));
+    ASSERT_TRUE(conf.is_defined("base_value_2"));
+    ASSERT_FALSE(conf.is_defined("unknown_base_value"));
+
+    /* Test fetching of a regular string without any environment variable */
+    std::string base_value_string = conf.get_scalar<std::string>("base_value.string", default_value);
+    ASSERT_EQ(base_value_string, "my_string");
+
+    /* Test fetching of escaped environment variable format. Should return the string as-is after stripping the leading `$` */
+    std::string base_value_invalid = conf.get_scalar<std::string>("base_value.invalid", default_value);
+    ASSERT_EQ(base_value_invalid, "${ENV_VAR}");
+
+	/* Test fetching of invalid escaped environment variable format. Should return the string as-is */
+    std::string base_value_invalid_env = conf.get_scalar<std::string>("base_value.invalid_env", default_value);
+    ASSERT_EQ(base_value_invalid_env, "$$ENV_VAR");
+
+    /* Test fetching of strings that contain environment variables */
+    std::string base_value_id = conf.get_scalar<std::string>("base_value.id", default_value);
+    ASSERT_EQ(base_value_id, "$ENV_VAR"); // Does not follow the `${VAR}` format, so it should be treated as a regular string
+
+    std::string base_value_name = conf.get_scalar<std::string>("base_value.name", default_value);
+    ASSERT_EQ(base_value_name, env_var_value); // Proper environment variable format
+
+    std::string base_value_escaped = conf.get_scalar<std::string>("base_value.escaped", default_value);
+    ASSERT_EQ(base_value_escaped, env_var_value); // Environment variable within quotes
+
+    /* Test fetching of an undefined environment variable. Expected to return the default value.*/
+    std::string unknown_boolean = conf.get_scalar<std::string>("base_value.subvalue.subvalue2.boolean", default_value);
+    ASSERT_EQ(unknown_boolean, default_value);
+
+    /* Test fetching of environment variables from a list */
+    std::string base_value_2_list_0 = conf.get_scalar<std::string>("base_value_2.sample_list[0]", default_value);
+    ASSERT_EQ(base_value_2_list_0, env_var_value); // Proper environment variable format
+
+    std::string base_value_2_list_1 = conf.get_scalar<std::string>("base_value_2.sample_list[1]", default_value);
+    ASSERT_EQ(base_value_2_list_1, " ${ENV_VAR}"); // Environment variable preceded by a space, hence treated as a regular string
+
+    std::string base_value_2_list_2 = conf.get_scalar<std::string>("base_value_2.sample_list[2]", default_value);
+    ASSERT_EQ(base_value_2_list_2, "$UNSED_XX_X_X_VAR"); // Does not follow the `${VAR}` format, so should be treated as a regular string
+
+    /* Clear the set environment variable after testing */
+    unsetenv(env_var_name.c_str());
 }
--- a/userspace/falco/yaml_helper.h
+++ b/userspace/falco/yaml_helper.h
@@ -72,6 +72,41 @@ public:
 		get_node(node, key);
 		if(node.IsDefined())
 		{
+			std::string value = node.as<std::string>();
+
+			// Helper function to convert string to the desired type T
+			auto convert_str_to_t = [&default_value](const std::string& str) -> T {
+				std::stringstream ss(str);
+				T result;
+				if (ss >> result) return result;
+				return default_value;
+			};
+
+			// If the value starts with `$$`, check for a subsequent `{...}`
+			if (value.size() >= 3 && value[0] == '$' && value[1] == '$')
+			{
+				// If after stripping the first `$`, the string format is like `${VAR}`, treat it as a plain string and don't resolve.
+				if (value[2] == '{' && value[value.size() - 1] == '}')
+				{
+					value = value.substr(1);
+					return convert_str_to_t(value);
+				}
+				else return convert_str_to_t(value);
+			}
+
+			// Check if the value is an environment variable reference
+			if(value.size() >= 2 && value[0] == '$' && value[1] == '{' && value[value.size() - 1] == '}')
+			{
+				// Format: ${ENV_VAR_NAME}
+				std::string env_var = value.substr(2, value.size() - 3);
+
+				const char* env_value = std::getenv(env_var.c_str()); // Get the environment variable value
+				if(env_value) return convert_str_to_t(env_value);
+
+				return default_value;
+			}
+
+			// If it's not an environment variable reference, return the value as is
 			return node.as<T>();
 		}
 
@@ -118,10 +153,10 @@ private:
 	 * The provided key string can navigate the document in its
 	 * nested nodes, with arbitrary depth. The key string follows
 	 * this regular language:
-	 * 
+	 *
 	 * Key 		:= NodeKey ('.' NodeKey)*
 	 * NodeKey	:= (any)+ ('[' (integer)+ ']')*
-	 * 
+	 *
 	 * Some examples of accepted key strings:
 	 * - NodeName
 	 * - ListValue[3].subvalue
@@ -146,7 +181,7 @@ private:
 					if (i > 0 && nodeKey.empty() && key[i - 1] != '.')
 					{
 						throw std::runtime_error(
-							"Parsing error: expected '.' character at pos " 
+							"Parsing error: expected '.' character at pos "
 							+ std::to_string(i - 1));
 					}
 					nodeKey += c;
@@ -157,7 +192,7 @@ private:
 					if (nodeKey.empty())
 					{
 						throw std::runtime_error(
-							"Parsing error: unexpected character at pos " 
+							"Parsing error: unexpected character at pos "
 							+ std::to_string(i));
 					}
 					ret.reset(ret[nodeKey]);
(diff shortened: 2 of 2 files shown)


jasondellaluce:
This is interesting! Thanks a lot. I'm gonna set this to the current release milestone just to not lose track of it, however keep in mind that we're very close to a Falco release and getting this into it will require a decision among the Falco maintainers and the release managers.

/milestone 0.35.0

FedeDP (approved):
/approve

leogr (commented):
I love this feature. However, I would like more time for testing and thinking about corner cases. For example:

- How can I escape from it if I wanted the literal `$SOMETHING` as a string (instead of the env var value)? :thinking: 

- How this feature interact with the [program_output](https://github.com/falcosecurity/falco/blob/master/falco.yaml#L404) which is a shell code? :thinking: 

- Are any other possible side effects? :thinking: 

For the reasons above and since we are a few days from releasing Falco 0.35, I propose pushing this to Falco 0.36. Let's discuss it!

therealdwright:
> * How can I escape from it if I wanted the literal `$SOMETHING` as a string (instead of the env var value)? 🤔
In most other applications, you would quote the string e.g. `"$SOMETHING"` .  However,  This has not something I have come across in practice. This implementation looks for a `{` at position 1, and a `}` at position  n - 1, so it only handles unquoted environment strings, consistent with other OSS implementations (e.g. filebeat).

> * How this feature interact with the [program_output](https://github.com/falcosecurity/falco/blob/master/falco.yaml#L404) which is a shell code? 🤔
The above response covers this somewhat, but I also tested the original behaviour and ran regression tests while developing the change. Does the regression suite cover such cases?

In terms of timeline, I'll leave it to the maintainers, I selfishly want this sooner rather than later as not being able to define environment variables leads to one of two consequences for my organisation.

1. We have to encrypt the entire configuration file with something like sops if we want to use the current implementation>
2. We must write some custom tooling to extend the upstream Falco image to iterate through the YAML and eval all environment variables.

Neither of these are the direction we _want_ to take, but if we have to, so be it.

leogr:
> > * How can I escape from it if I wanted the literal `$SOMETHING` as a string (instead of the env var value)? thinking
> 
> In most other applications, you would quote the string e.g. `"$SOMETHING"` . However, This has not something I have come across in practice. This implementation looks for a `{` at position 1, and a `}` at position n - 1, so it only handles unquoted environment strings, consistent with other OSS implementations (e.g. filebeat).
> 
> > * How this feature interact with the [program_output](https://github.com/falcosecurity/falco/blob/master/falco.yaml#L404) which is a shell code? thinking
> 
> The above response covers this somewhat, but I also tested the original behaviour and ran regression tests while developing the change. Does the regression suite cover such cases?

AFAIK, no.

The [filebeat](https://www.elastic.co/guide/en/beats/filebeat/current/using-environ-vars.html) example is interesting. I have never used it so, but their docs report:
> If you need to use a special character in your configuration file, use $ to escape the expansion. For example, you can escape ${ or } with $${ or $}.


Anyway, I'm not saying this is necessarily an issue - I just don't know yet and I couldn't find enough time to iterate on that  :sweat_smile: 


> In terms of timeline, I'll leave it to the maintainers, I selfishly want this sooner rather than later as not being able to define environment variables leads to one of two consequences for my organisation.

:+1:  I'm open to any decisions, just the timeline is unfortunate. Let's see other maintainers' opinions.


> 1. We have to encrypt the entire configuration file with something like sops if we want to use the current implementation as is.
> 2. We must write some custom tooling to extend the upstream Falco image to iterate through the YAML and eval all environment variables.
> 
> Neither of these are the direction we _want_ to take, but if we have to, so be it.

Out of curiosity: have you ever tried using `--option` as a workaround? For example:
```
falco ... --option someopt=$SOMEVAR
```
N.B. I'm not saying this is a clean solution.

therealdwright:
Let me know if you would like this PR extended to handle the special case scenario similarly to filebeat. I'm happy to update the PR.

> Out of curiosity: have you ever tried using --option as a workaround? For example:

I wasn't aware this was an option; I'll take a look tomorrow and give it a try.

therealdwright:
Unfortunately, moving forward with http_output configured as an option did the job of getting the config in but resulted in a new bug which has been raised as #2579

FedeDP:
I'd move this to 0.36.0 since we have not reach an agreement for this to get merged in time for 0.35.
I am sorry! We will review it asap right after 0.36.0

/milestone 0.36.0

therealdwright:
@FedeDP - shall I abandon this PR? It has been some time now and there's not really been any engagement.

FedeDP:
> shall I abandon this PR? It has been some time now and there's not really been any engagement.

I think it is good indeed! Let me ask for a review :)
Sorry for the long time no see, we were all very busy with cncf graduation related stuff!

/cc @leogr @jasondellaluce

Andreagit97:
we are reviewing this, sorry for the delay!

userspace/falco/yaml_helper.h
```diff
@@ -72,6 +72,42 @@ class yaml_helper
 		get_node(node, key);
 		if(node.IsDefined())
 		{
+			// Check if the value is an environment variable reference
+			std::string value = node.as<std::string>();
+			if(!value.empty() && value[0] == '$')
+			{
+				std::string env_var;
+				if(value.size() > 1 && value[1] == '{' && value[value.size() - 1] == '}')
```
Andreagit97, line 80: what would be the value to resolve both `${VAR}` and `$VAR` format? If there is no difference I would go with just one possible way of specifying the env variable in the config -> `${VAR}` 
WDYT?

unit_tests/falco/test_configuration.cpp
```diff
@@ -88,16 +88,71 @@ TEST(Configuration, modify_yaml_fields)
...
+
+    std::string base_value_2_list_1 = conf.get_scalar<std::string>("base_value_2.sample_list[1]", env_var_value);
+    ASSERT_EQ(base_value_2_list_1, env_var_value);
+
+    std::string base_value_2_list_2 = conf.get_scalar<std::string>("base_value_2.sample_list[2]", "");
+    ASSERT_EQ(base_value_2_list_2, "");
+
+    /* Clear environment variable */
+    unsetenv(env_var_name.c_str());
+}
```
Andreagit97, line 158: Proposing some little changes in the tests

```suggestion
TEST(Configuration, configuration_environment_variables)
{
	// Set environment variable for testing
	std::string env_var_value = "envVarValue";
	std::string env_var_name = "ENV_VAR";
	std::string default_value = "default";

	setenv(env_var_name.c_str(), env_var_value.c_str(), 1);

	yaml_helper conf;
	std::string sample_yaml =
		"base_value:\n"
		"    id: $ENV_VAR\n"
		"    name:    '${ENV_VAR}'\n"
		"    string: my_string\n"
		"    invalid: $$ENV_VAR\n"
		"    escaped: \"$ENV_VAR\"\n"
		"    subvalue:\n"
		"        subvalue2:\n"
		"            boolean: $UNSED_XX_X_X_VAR\n"
		"base_value_2:\n"
		"    sample_list:\n"
		"        - $ENV_VAR\n"
		"        - ' ${ENV_VAR}'\n"
		"        - $UNSED_XX_X_X_VAR\n";
	conf.load_from_string(sample_yaml);

	/* is_defined */
	ASSERT_TRUE(conf.is_defined("base_value"));
	ASSERT_TRUE(conf.is_defined("base_value_2"));
	ASSERT_FALSE(conf.is_defined("unknown_base_value"));

	/* get_scalar - no environment variables */
	std::string base_value_string = conf.get_scalar<std::string>("base_value.string", default_value);
	ASSERT_EQ(base_value_string, "my_string");

	/* Invalid value for an env var `$$MY_STRING`, is invalid.
	 * The first `$` is not taken into consideration but the second one
	 * is kept, so we will try to resolve the env var `$MY_STRING`.
	 * We won't find any match so the default value is returned.
	 */
	std::string base_value_invalid = conf.get_scalar<std::string>("base_value.invalid", default_value);
	ASSERT_EQ(base_value_invalid, default_value);

	/* get_scalar - environment variables */
	std::string base_value_id = conf.get_scalar<std::string>("base_value.id", default_value);
	ASSERT_EQ(base_value_id, env_var_value);

	std::string base_value_name = conf.get_scalar<std::string>("base_value.name", default_value);
	ASSERT_EQ(base_value_name, env_var_value);

	std::string base_value_escaped = conf.get_scalar<std::string>("base_value.escaped", default_value);
	ASSERT_EQ(base_value_escaped, env_var_value);

	/* get_scalar - unknown environment variable, we expect the default value */
	std::string unknown_boolean = conf.get_scalar<std::string>("base_value.subvalue.subvalue2.boolean", default_value);
	ASSERT_EQ(unknown_boolean, default_value);

	/* get_scalar - environment variables in list */
	std::string base_value_2_list_0 = conf.get_scalar<std::string>("base_value_2.sample_list[0]", default_value);
	ASSERT_EQ(base_value_2_list_0, env_var_value);

	/* Note that there is a space `' ${ENV_VAR}'` between `'` and `$` so `$` is not the first character
	 * and the value is not recognized as an env variable
	 */
	std::string base_value_2_list_1 = conf.get_scalar<std::string>("base_value_2.sample_list[1]", default_value);
	ASSERT_EQ(base_value_2_list_1, " ${ENV_VAR}");

	std::string base_value_2_list_2 = conf.get_scalar<std::string>("base_value_2.sample_list[2]", default_value);
	ASSERT_EQ(base_value_2_list_2, default_value);

	/* Clear environment variable */
	unsetenv(env_var_name.c_str());
}
```

Andreagit97 (commented):
I find this feature really useful, thank you for this! and sorry for the delay in reviewing it!
This implementation is fine, I have only some points:

1. Using just one format for env variables, like `${VAR}` and not also `$VAR`. No strong opinion but IMO having just one way to use env var would be easier for end users.
2. With these changes if a user specifies `${VAR}` in the config it will be resolved to an env variable but maybe this is not what the user want, so I would propose a possible escaping method like adding an extra `$`, not so smart but it should be effective and easy to implement:

**Example**
* Setting `VAR=hello`
* if in the config we use `${VAR}` we will resolve it to "hello", while if we use `$${VAR}` we will keep only `${VAR}` without resolving it


As a final implementation, it would be nice to parse the whole yaml searching for `$` characters and expanding them when needed, so before the yaml is parsed, in this way we could expand also env variable inside a string like `"my var is $VAR"`. But this is probably a future improvement at the moment this PR is fine apart from the 2 points above.

**Example possible final solution**

Setting `VAR=hello`

```yaml
base:
  1: $VAR
  2: "my var is $VAR"
  3: $$VAR
```

Before parsing the yaml will become

```yaml
base:
  1: hello
  2: "my var is hello"
  3: $VAR
```
WDYT @therealdwright @FedeDP @leogr ?

leogr:
> 1. Using just one format for env variables, like `${VAR}` and not also `$VAR`. No strong opinion but IMO having just one way to use env var would be easier for end users

No strong opinion.

> 2. With these changes if a user specifies `${VAR}` in the config it will be resolved to an env variable but maybe this is not what the user want, so I would propose a possible escaping method like adding an extra `$`, not so smart but it should be effective and easy to implement:
> 
> **Example**
> 
> * Setting `VAR=hello`
> * if in the config we use `${VAR}` we will resolve it to "hello", while if we use `$${VAR}` we will keep only `${VAR}` without resolving it

I'm totally on board with this.
However, since this has been waiting for a long time, I understand there are compelling use cases, I'm also ok let merging this PR as-is (maybe just removing the `$VAR` support) and improving it in a followup PR.

leogr:
cc @Andreagit97 @FedeDP ?

FedeDP:
> However, since this has been waiting for a long time, I understand there are compelling use cases, I'm also ok let merging this PR as-is (maybe just removing the $VAR support) and improving it in a followup PR.

I'd ask the PR author whether he is willing to quickly fix this up @therealdwright , otherwise i am on board with merging this as is, and then rework it later.

therealdwright:
I'll take a look shortly

therealdwright:
> if in the config we use ${VAR} we will resolve it to "hello", while if we use $${VAR} we will keep only ${VAR} without resolving it

This will take a bit of extra care to handle, for now I have kept the PR to only handle the `${VAR}` case to keep things simple and if there is a strong desire to add variable esscaping please let me know and I will push up a follow-up. I think this is a vary rare edge case that we are unlikely to see in the real world and implementing will require some thought and user-facing documentation.

userspace/falco/yaml_helper.h
therealdwright: I have removed the logic to process both ${VAR} and $VAR in favour of using just ${VAR} this conflicts with some of your test suggestions which I have updated accordingly.

therealdwright:
OK @Andreagit97 @leogr and @FedeDP

This has taken me a bit to get around as I have not thought about or touched this codebase for three months, and each piece of feedback in isolation was reasonable, but there were some conflicts to resolve to appease it all. I think I have struck a good balance here.

1. Only one format of an environment variable is now supported, being `${ENV_VAR}`
2. YAML values following a `$ENV_VAR` format will be processed as a regular string
3. YAML Values following a `$${ENV_VAR}` format will be processed as a regular string with the first $ stripped
4. YAML values following a `$$ENV_VAR` format will be processed as a regular string

I think I have adequately covered the various edge cases in the unit test, as I hope you can appreciate there are a lot of edge cases to check for here. Please re-review and let me know what you think. I would appreciate it if the team could come back to me soon as it's very hard to step away from this project and C++ for so long and be productive.

If you would like me to write some user-facing docs, LMK, I'd be happy to do so.

Andreagit97:
> 1. Only one format of an environment variable is now supported, being `${ENV_VAR}`
> 2. YAML values following a `$ENV_VAR` format will be processed as a regular string
> 3. YAML Values following a `$${ENV_VAR}` format will be processed as a regular string with the first $ stripped
> 4. YAML values following a `$$ENV_VAR` format will be processed as a regular string

Agree with the design, thanks!

FedeDP (commented):
LGTM! Agree with @Andreagit97 comments, then we are good to go! 
Thanks for the effort @therealdwright !

unit_tests/falco/test_configuration.cpp
```diff
@@ -88,16 +88,88 @@ TEST(Configuration, modify_yaml_fields)
...
+
+    /* Test fetching of a regular string without any environment variable */
+    std::string base_value_string = conf.get_scalar<std::string>("base_value.string", default_value);
+    ASSERT_EQ(base_value_string, "my_string");
+
+    /* Test fetching of escaped environment variable format. Should return the string as-is after stripping the leading `$` */
+    std::string base_value_invalid = conf.get_scalar<std::string>("base_value.invalid", default_value);
+    ASSERT_EQ(base_value_invalid, "${ENV_VAR}");
+
+	/* Test fetching of escaped environment variable ub ubvakud format. Should return the string as-is */
```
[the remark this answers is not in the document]
therealdwright: Thanks, accepted this change.

userspace/falco/yaml_helper.h
```diff
@@ -72,6 +72,41 @@ class yaml_helper
...
+				if (value[2] == '{' && value[value.size() - 1] == '}')
+				{
+					value = value.substr(1);
+					return convert_str_to_t(value);
+				}
+				else return convert_str_to_t(value);
+			}
+
+			// Check if the value is an environment variable reference
+			if(!value.empty() && value[0] == '$' && value.size() > 2 && value[1] == '{' && value[value.size() - 1] == '}')
```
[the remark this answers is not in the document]
therealdwright: I agree this is a more efficient/elegant way so I've accepted this change.

FedeDP (approved):
/approve

Andreagit97 (approved):
Thanks!

/approve

FedeDP:
Let's go with this one! Again, thank you very much for your effort :rocket:

PR #2568, opened 2023-05-23
update(ci): use repo instead of master branch for reusable workflows

LucaGuerra:
**What type of PR is this?**

/kind cleanup
/area CI

Originally, the reusable workflows were pinned on the `master` branch. This is ok-ish but since we have a release model based on release branches this means that if a PR gets into master that modifies the release process (e.g. by changing some compilation option) we may break releases and have issues when we try to issue a patch release.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/master.yaml
+++ b/.github/workflows/master.yaml
@@ -43,7 +43,7 @@ jobs:
 
   build-dev-packages:
     needs: [fetch-version]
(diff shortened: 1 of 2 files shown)


FedeDP (approved):
Makes totally sense.
/approve

Andreagit97 (approved):
/approve

PR #2569, opened 2023-05-23
fix(userspace/falco): properly format numeric values in metrics

jasondellaluce:
**What type of PR is this?**

/kind bug

/kind cleanup

/area engine

Following up #2333, this makes sure that numeric values are formatted as numbers in the JSON outputs of the new metrics system.

cc @incertum

```release-note
fix(userspace/falco): properly format numeric values in metrics
```

### DIFF
--- a/userspace/falco/stats_writer.cpp
+++ b/userspace/falco/stats_writer.cpp
@@ -250,7 +250,7 @@ void stats_writer::collector::get_metrics_output_fields_additional(
 			for(uint32_t stat = 0; stat < nstats; stat++)
 			{
 				char metric_name[STATS_NAME_MAX] = "falco.";
-				strncat(metric_name, utilization[stat].name, sizeof(metric_name) - strlen(metric_name));
+				strncat(metric_name, utilization[stat].name, sizeof(metric_name) - strlen(metric_name) - 1);
 				switch(utilization[stat].type)
 				{
 				case STATS_VALUE_TYPE_U64:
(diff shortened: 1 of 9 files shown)


leogr (dismissed):
I don't like too much using `nlohmann::json`, but I can't think of a better solution right now.

So, LGTM.

jasondellaluce:
> I don't like too much using nlohmann::json, but I can't think of a better solution right now.

Same 😸

leogr (dismissed):

userspace/falco/stats_writer.cpp
FedeDP, line 1: @jasondellaluce 
please consider fixing https://github.com/falcosecurity/falco/pull/2333#discussion_r1201717285 here :)

FedeDP (commented):
LGTM; left a comment about `strncat` issue.
/hold

jasondellaluce:
> Wouldn't we need a ! as we change the data types for all Falco rules not just the internal stats / metrics related rules?

I don't think so, this changes the types only for the new metrics internal alerts, which is something we haven't released yet.

incertum:
> > Wouldn't we need a ! as we change the data types for all Falco rules not just the internal stats / metrics related rules?
> 
> I don't think so, this changes the types only for the new metrics internal alerts, which is something we haven't released yet.

Wait so `handle_msg` is just used for the internal rules and not for the regular Falco rules? In that case just out of curiosity could you point me to the place where regular Falco rules are formatted?

leogr (approved):

PR #2575, opened 2023-05-24
fix(userspace): check the supported number of online CPUs with modern bpf

Andreagit97:
**What type of PR is this?**

/kind bug

/kind cleanup

/area engine

/area tests

Right now the modern bpf config `cpus_for_each_syscall_buffer` can be used only with a number of CPUs lower than the actual number of online CPUs in the system. The real fix should be to handle this situation in a better way directly in the modern bpf engine but since we are short in time this patch on the fly in Falco seems the right way to go!

Example:
- in the system, there is only 1 online CPU
- by default, the modern bpf opens one buffer every 2 CPUs
- in the modern bpf engine there is a check that compares the required CPUs with the online ones: so in this case, 2>1 -> the modern bpf engine will generate an error
- here we have seen exactly what I've just described: https://kubernetes.slack.com/archives/CMWH3EH32/p1684561641886179

Fix:

If the number of required CPUs exceeds the online CPUs number, we set the required number to the number of online CPU
 
```release-note
fix(userspace): check the supported number of online CPUs with modern bpf
```


FedeDP (approved):
/approve

jasondellaluce (approved):
/approve

/milestone 0.35.0

PR #2576, opened 2023-05-24
fix(userspace/falco): don't hang on terminating error when multi sourcing

jasondellaluce:
**What type of PR is this?**

/kind bug

/area engine

When in multi-source mode, any failure of one event source should cause the others to be forcefully terminated too. This PR solves a bug that caused Falco to hang on a semaphore when the `syscall` source was the one terminating with error.

Fixes #

```release-note
fix(userspace/falco): don't hang on terminating error when multi sourcing
```


Andreagit97 (approved):
/approve

leogr (approved):

PR #2578, opened 2023-05-24
chore(userspace/falco): always print invalid syscalls from custom set

jasondellaluce:
Co-authored-by: Roberto Scolaro <roberto.scolaro21@gmail.com>

**What type of PR is this?**

/kind bug

/area engine

When we insert only one bad entry in the `base_syscall.custom_set` sequence (or a sequence of bad entries), Falco fails to output the "invalid syscalls" warning because it sees the custom set as empty, given that it crops the invalid values. This PR makes sure that the warning is always printed.

```release-note
chore(userspace/falco): always print invalid syscalls from custom set
```


FedeDP (approved):
/approve
Nice UX improvement!

Andreagit97 (approved):
/approve

PR #2602, opened 2023-06-01
cleanup(userspace/falco): improvements to the http output perf.

FedeDP:
**What type of PR is this?**

/kind cleanup

/area engine

Improves http_output performance by creating the curl handle just once with all its options; then it's just a matter of performing the call with the correct post fields.

Moreover, add an option to disable server answer echoing to stdout.

Fixes #

```release-note
cleanup(userspace/falco): improved http output performance
new(userspace/falco): added option to disable echoing of server answeer to stdout
```


FedeDP:
We also need to support output reloading! TODO.

FedeDP:
> We also need to support output reloading! TODO.

No problem, really. We don't need to support http output reloading since it is just a "close file open file" behavior (it is not linked to any config update).

FedeDP:
Rebased on top of master.

falco.yaml
```diff
@@ -345,6 +345,8 @@ http_output:
   # Path to a folder that will be used as the CA certificate store. CA certificate need to be
   # stored as indivitual PEM files in this directory.
   ca_path: "/etc/ssl/certs"
+  # Whether to echo server answers to stdout
+  echo: true
```
FedeDP, line 349: Keep old behavior.

userspace/falco/outputs_http.cpp
```diff
@@ -47,62 +56,83 @@ void falco::outputs::output_http::output(const message *msg)
...
+	return true;
+}
+
+void falco::outputs::output_http::output(const message *msg)
+{
+	CURLcode res = curl_easy_setopt(m_curl, CURLOPT_POSTFIELDS, msg->msg.c_str());
+	if (res == CURLE_OK)
+	{
+		res = curl_easy_perform(m_curl);
+	}
```
Andreagit97, line 125: Here is not so relevant but in the `init` function it could cleanup a little bit
```suggestion
	res = res == CURLE_OK? curl_easy_perform(m_curl) : res;
```

userspace/falco/outputs_http.cpp
```diff
@@ -47,62 +56,83 @@ void falco::outputs::output_http::output(const message *msg)
...
-				res = curl_easy_setopt(curl, CURLOPT_CAPATH, m_oc.options["ca_path"].c_str());
+				res = curl_easy_setopt(m_curl, CURLOPT_CAPATH, m_oc.options["ca_path"].c_str());
 			}
 		}
 
-  		if(res == CURLE_OK)
+		if(m_oc.options["echo"] == std::string("false"))
 		{
-			res = curl_easy_perform(curl);
+			// If unset, libcurl defaults to fwrite to stdout, ie: echoing
```
Andreagit97, line 103:
```suggestion
			// If `echo=="true"`, libcurl defaults to fwrite to stdout, ie: echoing
```

Andreagit97 (commented):
Thank you for this! Some comments about the style and one more important comment about what we want to do with "echo" config

Andreagit97:
Still asking myself why we initialized curl at every iteration... are we missing something? @leogr

falco.yaml
FedeDP: I agree. I will move it to `false` by default :)

FedeDP:
Rebased and fixed up @Andreagit97 suggestions!

submodules/falcosecurity-rules
Andreagit97, line 1: i don't think we need this :/

userspace/falco/outputs_http.cpp
```diff
@@ -18,104 +18,104 @@ limitations under the License.
...
 	}
+	if(m_json_output)
+	{
+		m_http_headers = curl_slist_append(m_http_headers, "Content-Type: application/json");
+	}
+	else
+	{
+		m_http_headers = curl_slist_append(m_http_headers, "Content-Type: text/plain");
+	}
+	CHECK_RES(curl_easy_setopt(m_curl, CURLOPT_HTTPHEADER, m_http_headers));
```
Andreagit97, line 51: Uhm I think this would fail all the times, `res` is still `CURLE_FAILED_INIT` probably we need to set it to `CURLE_OK` and set `CURLE_FAILED_INIT` directly here:

```cpp
if(!m_curl)
	{
		falco_logger::log(LOG_ERR, "libcurl failed to initialize the handle: " + std::string(curl_easy_strerror(CURLE_FAILED_INIT)));
```

submodules/falcosecurity-rules
FedeDP: Ouch :/

userspace/falco/outputs_http.cpp
FedeDP: Nice! Fixed, sorry

Andreagit97:
IMO this is ready, could you rebase please? @FedeDP

FedeDP:
:( i hate falcosecurity-rules submodule @jasondellaluce :laughing:

FedeDP:
Bumped rules to latest master version.

Andreagit97 (approved):
/approve

leogr:
> Still asking myself why we initialized curl at every iteration... are we missing something? @leogr

@Andreagit97 
The [original design](https://github.com/falcosecurity/falco/pull/1412/files#diff-12b7abc7ed9bc0a2615ebffe08941792a0d500891e47e58e693173a987d94fe9L270) of the http_out likely assumed that dispose curl at every iteration was safer to avoid memory leak and the performance cost was acceptable since alerts should usually have a low frequency. This is just a speculation since, when I refactored the whole output system, I just ported the same approach (the refactor was already too big also to introduce too many improvements).

Thus, we should have everything ok.

Andreagit97:
cool! thank you for the explanation :)

userspace/falco/outputs_http.cpp
```diff
@@ -18,104 +18,105 @@ limitations under the License.
...
+static size_t noop_write_callback(void *contents, size_t size, size_t nmemb, void *userp)
+{
+	// We don't want to echo anything. Just return size of bytes ignored
+	return size * nmemb;
+}
+
+bool falco::outputs::output_http::init(const config& oc, bool buffered, const std::string& hostname, bool json_output, std::string &err)
 {
-	CURL *curl = NULL;
+	falco::outputs::abstract_output::init(oc, buffered, hostname, json_output, err);
```
leogr, line 31: Should the return value of the parent `init` be checked?

I believe yes since `falco::outputs::abstract_output::init` can change in the future and start returning errors in some cases.

leogr (commented):
Just left a comment, otherwise SGTM!

userspace/falco/outputs_http.cpp
FedeDP: Makes sense :) thanks!

FedeDP: Fixed!

leogr (approved):

Andreagit97 (approved):
/approve

PR #2642, opened 2023-06-15
fix(ci): fixed publish deb packages with new debian bookworm

FedeDP:
**What type of PR is this?**

/kind bug

/area CI

New debian bookworm, release on 10th of June, dropped python2 package.
Instead, we need to install `python-is-python3` package that creates the symlink for `/usr/bin/python3` to `/usr/bin/python`.

Fixes CI.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/reusable_publish_packages.yaml
+++ b/.github/workflows/reusable_publish_packages.yaml
@@ -117,7 +117,7 @@ jobs:
       - name: Install dependencies
         run: |
           apt update -y
-          apt-get install apt-utils bzip2 gpg python python3-pip -y
+          apt-get install apt-utils bzip2 gpg python-is-python3 python3-pip -y
           pip install awscli
       
       # Configure AWS role; see https://github.com/falcosecurity/test-infra/pull/1102
--- a/.github/workflows/reusable_publish_packages.yaml
+++ b/.github/workflows/reusable_publish_packages.yaml
@@ -117,8 +117,7 @@ jobs:
       - name: Install dependencies
         run: |
           apt update -y
-          apt-get install apt-utils bzip2 gpg python-is-python3 python3-pip -y
-          pip install awscli
+          apt-get install apt-utils bzip2 gpg awscli -y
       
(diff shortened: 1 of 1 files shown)


Andreagit97 (approved):
/approve

LucaGuerra:
@FedeDP Are you sure that works? It fails on my machine:

```
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.11/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

LucaGuerra:
Also, do we actually need Python2 or is it only to use `pip`? In which case, can't we use `pip3` instead?

FedeDP:
We do install pip3, right? @LucaGuerra
`apt-get install apt-utils bzip2 gpg python-is-python3 python3-pip -y`
The new `python-is-python3` package just installs the `usr/bin/python -> /usr/bin/python3` link in the system.

Mmh the error is weird, i can confirm i got the same. I don't get what it means though.
I think we might just want to install `awscli` manually from debian repo in the end.

jasondellaluce (dismissed):
/approve

LucaGuerra (commented):
Thanks! LGTM

Andreagit97 (approved):
/approve

jasondellaluce (approved):
/approve

PR #2705, opened 2023-07-28
new: allow falco to match multiple rules on same event

loresuso:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind release

> /area build

> /area engine

> /area tests

> /area proposals

> /area CI

This PR adds a configuration key to Falco to let it match multiple rules on the same event, avoiding a well-known shadowing problem between rules. To not introduce breaking changes, the old behavior is kept and can still be used (expressed trough config). To implement this, Falco will simply continue trying to evaluate the condition of rules in the bucket of same evt.type rules, without stopping at the first match. Thinking about it, this should not cause any performance degradation, since the rules in the bucket are always evaluated in their entirety whenever there's no matching rule, and the cardinality of events not matching rules compared to the ones that do is orders of magnitude higher (meaning that evaluating all conditions of rules in the same bucket is something that we do all the time, so we can afford to do it even when a rule in a bucket is matching). 

Putting this on wip while figuring out how to introduce some tests, but open to discuss! 

Fixes #

```release-note
new: allow falco to match multiple rules on same event 
```

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -139,23 +139,6 @@ rules_file:
   - /etc/falco/falco_rules.local.yaml
   - /etc/falco/rules.d
 
-# [Experimental] `rule_matching`
-#
-# Falco has to be performant when evaluating rules against events. To quickly 
-# understand which rules could trigger on a specific event, Falco maintains 
-# buckets of rules sharing the same event type in a map. Then, the lookup 
-# in each bucket is performed through linear search. The `rule_matching`
-# configuration key's values are:
-#  - "first": when evaluating conditions of rules in a bucket, Falco will stop
-#    to evaluate rules if it finds a matching rules. Since rules are stored 
-#    in buckets in the order they are defined in the rules files, this option 
-#    could prevent other rules to trigger even if their condition is met, causing
-#    a shadowing problem.
-#  - "all": with this value Falco will continue evaluating all the rules 
-#    stored in the bucket, so that multiple rules could be triggered upon one
-#    event.
-rule_matching: first
-
 #################
 # Falco plugins #
 #################
@@ -305,6 +288,21 @@ buffered_outputs: false
 outputs:
   rate: 0
   max_burst: 1000
+
+# [Experimental] `rule_matching`
+#
+# The `rule_matching` configuration key's values are:
+#  - `first`: Falco stops checking conditions of rules against upcoming event
+#    at the first matching rule
+#  - `all`: Falco will continue checking conditions of rules even if a matching 
+#    one was already found
+#
+# Rules conditions are evaluated in the order they are defined in the rulesfiles.
+# For this reason, when using `first` as value, only the first defined rule will 
+# trigger, possibly shadowing other rules.
+# In case `all` is used as value, rules still trigger in the order they were
+# defined.
+rule_matching: first
 
 
 ##########################
--- a/falco.yaml
+++ b/falco.yaml
@@ -139,6 +139,22 @@ rules_file:
   - /etc/falco/falco_rules.local.yaml
   - /etc/falco/rules.d
 
+# [Experimental] `rule_matching`
+#
+# Falco has to be performant when evaluating rules against events. To quickly 
+# understand which rules could trigger on a specific event, Falco maintains 
+# buckets of rules sharing the same event type in a map. Then, the lookup 
+# in each bucket is performed through linear search. The `rule_matching`
+# configuration key's values are:
+#  - "first": when evaluating conditions of rules in a bucket, Falco will stop
+#    to evaluate rules if it finds a matching rules. Since rules are stored 
+#    in buckets in the order they are defined in the rules files, this option 
+#    could prevent other rules to trigger even if their condition is met, causing
+#    a shadowing problem.
+#  - "all": with this value Falco will continue evaluating all the rules 
+#    stored in the bucket, so that multiple rules could be triggered upon one
+#    event.
+rule_matching: first
 
 #################
 # Falco plugins #
--- a/userspace/engine/falco_common.cpp
+++ b/userspace/engine/falco_common.cpp
@@ -27,6 +27,11 @@ static std::vector<std::string> priority_names = {
 	"Debug"
 };
 
+static std::vector<std::string> rule_matching_names = {
+	"first",
+	"all"
+};
+
 bool falco_common::parse_priority(std::string v, priority_type& out)
 {
 	for (size_t i = 0; i < priority_names.size(); i++)
@@ -79,4 +84,17 @@ std::string falco_common::format_priority(priority_type v, bool shortfmt)
 		throw falco_exception("Unknown priority enum value: " + std::to_string(v));
 	}
 	return out;
+}
+
+bool falco_common::parse_rule_matching(std::string v, rule_matching& out)
+{
+	for (size_t i = 0; i < rule_matching_names.size(); i++)
+	{
+		if (!strcasecmp(v.c_str(), rule_matching_names[i].c_str()))
+		{
+			out = (rule_matching) i;
+			return true;
+		}
+	}
+	return false;
 }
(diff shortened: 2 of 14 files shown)


falco.yaml
```diff
@@ -139,6 +139,22 @@ rules_file:
   - /etc/falco/falco_rules.local.yaml
   - /etc/falco/rules.d
 
+# [Experimental] `rule_matching`
```
loresuso, line 142: Maybe we should find a proper place for this new config, don't know if here is where we want it

userspace/engine/falco_engine.cpp
```diff
@@ -378,24 +385,47 @@ std::unique_ptr<falco_engine::rule_result> falco_engine::process_event(std::size
...
-	std::unique_ptr<struct rule_result> res(new rule_result());
-	res->evt = ev;
-	res->rule = source->m_rule.name;
-	res->source = source->m_rule.source;
-	res->format = source->m_rule.output;
-	res->priority_num = source->m_rule.priority;
-	res->tags = source->m_rule.tags;
-	res->exception_fields = source->m_rule.exception_fields;
-	m_rule_stats_manager.on_event(source->m_rule);
+	source->m_rules.clear();
```
jasondellaluce, line 424: Maybe let's clear only if we had at least one match.

userspace/engine/falco_engine.cpp
```diff
@@ -354,7 +360,7 @@ std::shared_ptr<gen_event_formatter> falco_engine::create_formatter(const std::s
 	return find_source(source)->formatter_factory->create_formatter(output);
 }
 
-std::unique_ptr<falco_engine::rule_result> falco_engine::process_event(std::size_t source_idx, gen_event *ev, uint16_t ruleset_id)
+std::vector<falco_engine::rule_result> falco_engine::process_event(std::size_t source_idx, gen_event *ev, uint16_t ruleset_id)
```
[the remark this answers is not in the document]
loresuso: Yes, it makes a lot of sense, thanks @jasondellaluce!

falco.yaml
jasondellaluce: I would suggest moving it in one of the sections below, perhaps around/before the rate limiter options.

Also, I would avoid leaking knowledge about the evttype bucket indexing here in this comment. I think that for end users what matters is just understanding that Falco stops or not at the first rule match. The rest is an internal implementation detail on which we need to honor this interface.

userspace/engine/falco_engine.cpp
```diff
@@ -378,24 +384,51 @@ std::unique_ptr<falco_engine::rule_result> falco_engine::process_event(std::size
...
+	if (m_rule_matching == falco_common::rule_matching::ALL)
+	{
+		if (!source->ruleset->run(ev, source->m_rules, ruleset_id))
+		{
+			return nullptr;
+		}
+	}
+	else if (m_rule_matching == falco_common::rule_matching::FIRST)
+	{
+		falco_rule rule;
```
jasondellaluce, line 401: Let's retain this one too in falco_source. Alternatively, let's make sure that `source->m_rules` always retains 1 element when `m_rule_matching == falco_common::rule_matching::FIRST` and always use that one. Allocating this on the stack at each round is costly, as we found in he past.

userspace/falco/configuration.cpp
```diff
@@ -244,6 +244,12 @@ void falco_configuration::load_yaml(const std::string& config_name, const yaml_h
 	m_notifications_rate = config.get_scalar<uint32_t>("outputs.rate", 0);
 	m_notifications_max_burst = config.get_scalar<uint32_t>("outputs.max_burst", 1000);
 
+	std::string rule_matching = config.get_scalar<std::string>("rule_matching", "first");
+	if (!falco_common::parse_rule_matching(rule_matching, m_rule_matching))
```
jasondellaluce, line 248: Let's make sure `m_rule_matching` is initialized in the constructor as well.

userspace/engine/falco_engine.cpp
```diff
@@ -378,24 +384,51 @@ std::unique_ptr<falco_engine::rule_result> falco_engine::process_event(std::size
...
-	std::unique_ptr<struct rule_result> res(new rule_result());
-	res->evt = ev;
-	res->rule = source->m_rule.name;
-	res->source = source->m_rule.source;
-	res->format = source->m_rule.output;
-	res->priority_num = source->m_rule.priority;
-	res->tags = source->m_rule.tags;
-	res->exception_fields = source->m_rule.exception_fields;
-	m_rule_stats_manager.on_event(source->m_rule);
+	if (m_rule_matching == falco_common::rule_matching::ALL)
```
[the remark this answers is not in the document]
loresuso: Agreed

userspace/engine/falco_engine.cpp
loresuso: Makes perfect sense, I am now always retaining 1 element in case of `falco_common::rule_matching::FIRST`

falco.yaml
loresuso: Addressed, let me know what you think now

jasondellaluce:
> 
> 

This is right, we're effectively adding a new feature to the engine and this should have a version bump.

userspace/engine/falco_engine.cpp
```diff
@@ -377,24 +383,54 @@ std::unique_ptr<falco_engine::rule_result> falco_engine::process_event(std::size
 		source = find_source(source_idx);
 	}
 
-	if(should_drop_evt() || !source || !source->ruleset->run(ev, source->m_rule, ruleset_id))
+	if(should_drop_evt() || !source)
+	{
+		return nullptr;
```
Andreagit97, line 388: before this patch we sent `return std::unique_ptr<struct rule_result>();` now we send `nullptr` is this intentional?

userspace/engine/falco_engine.h
```diff
@@ -189,14 +194,14 @@ class falco_engine
...
-	std::unique_ptr<rule_result> process_event(std::size_t source_idx, gen_event *ev, uint16_t ruleset_id);
+	std::unique_ptr<std::vector<rule_result>> process_event(std::size_t source_idx, gen_event *ev, uint16_t ruleset_id);
 
 	//
 	// Wrapper assuming the default ruleset.
 	//
 	// This inherits the same thread-safety guarantees.
 	//
-	std::unique_ptr<rule_result> process_event(std::size_t source_idx, gen_event *ev);
+	std::unique_ptr<std::vector<rule_result>> process_event(std::size_t source_idx, gen_event *ev);
```
Andreagit97, line 204: These 2 methods seem to do the same thing, the second one has just the third parameter defaulted to `m_default_ruleset_id`

Andreagit97 (commented):
do we want some unit tests on it or do we think e2e tests are enough @jasondellaluce?

userspace/engine/falco_common.cpp
```diff
@@ -79,4 +84,17 @@ std::string falco_common::format_priority(priority_type v, bool shortfmt)
 		throw falco_exception("Unknown priority enum value: " + std::to_string(v));
 	}
 	return out;
+}
+
+bool falco_common::parse_rule_matching(std::string v, rule_matching& out)
+{
+	for (size_t i = 0; i < rule_matching_names.size(); i++)
+	{
```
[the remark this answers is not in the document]
loresuso: The signature of the function is pretty the same as `parse_priority` right above, I think it's pretty self-explanatory from the types (string in, rule_matching enum out) and the name of the function

userspace/engine/falco_engine.cpp
loresuso: I think there's no need to construct an empty `rule_result`, `nullptr` will do what we need, and the result is handled in the `do_inspect` function

userspace/engine/falco_engine.h
loresuso: Yes, this one is just a utility function. It was already there, I'd keep it, but I can remove it if you want!

Andreagit97 (approved):
/approve

jasondellaluce (approved):

PR #2663, opened 2023-06-28
update : support build for wasm 

Rohith-Raju:
**What type of PR is this?**

/kind feature

**Any specific area of the project related to this PR?**

/area build

/area tests

/area CI

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
This PR adds support to WASM target. This is a preliminary work for Google Summer of Code
https://github.com/falcosecurity/evolution/issues/262
https://github.com/falcosecurity/libs/pull/1156

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
update : support build for wasm 
```

PR #2711, opened 2023-08-01
fix(outputs): expose `outputs_queue` config for memory control

incertum:
**What type of PR is this?**

/kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind release

> /area build

/area engine

> /area tests

> /area proposals

> /area CI

The tbb output queue is currently unbounded and is a very likely the cause for memory leak issues in production when the pipe is not holding up, see some simulated results https://github.com/falcosecurity/falco/issues/2495#issuecomment-1643215114.

We discussed that exposing new configs is best practices offering more options for adopters to control consumption and limits beyond the external cgroups limits. It's also in line with the kernel buffer being bounded and its size exposed as config.

After all no one wants SRE s or system admins calling out Falco may be leaking memory.

There could still be other issues. Therefore, the best next step would be to merge an acceptable solution and then test deploy and see and then go from there.

@leogr @jasondellaluce 

https://github.com/falcosecurity/falco/issues/2495#issuecomment-1643215114

Fixes #

```release-note
fix(outputs): expose queue_capacity_outputs config for memory control
```


falco.yaml
```diff
@@ -290,6 +291,34 @@ outputs:
...
+# the queue is unbounded.
+# You can experiment with values greater or smaller than the anchor value 1000000.
+# 
+# `recovery`: the strategy to follow when the queue becomes filled up. This also applies when
+# the queue is unbounded, and all available memory on the system is consumed.
+# recovery: 0 means continue.
+# recovery: 1 means simply exit (default behavior).
+# recovery: 2 means empty the queue and then continue.
+queue_capacity_outputs:
+  items: 0
```
incertum, line 319: @leogr as per our chat preserving the old default (unbounded and exiting). Obviously we can change the encoding etc.

incertum: Furthermore, @leogr I would like to better understand the interrelationships with for instance `output_timeout`, `buffered_outputs` and `outputs.rate` and `outputs.max_burst`. Their descriptions are also still more cryptic especially for the outputs one. aka I am looking for a scaling factor recommendation aka if you change let's 

```
outputs:
  rate: 100000
  max_burst: 100000000
``` 

what capacity of the queue should you choose in relation?

Maybe the timeout default is not right? we block for too long? How was the current default derived?

incertum: Also reading this raises some question marks ... because not having a recovery strategy as well here appears suboptimal:

```
# It's important to note that Falco outputs will not be discarded from the
# output queue. This means that if an output channel becomes blocked
# indefinitely, it indicates a potential issue that needs to be addressed by the
# user.
output_timeout: 2000
```

leogr: **Why the `output_timeout` is  so important?**

 When output consumer blocks (ie. imagine the `program` output or any I/O related op that happens downstream to Falco) and the alert can be delivered within the given deadline, Falco uses `stderr` to notify the user of a not addressable problem. 
 
 Why `stderr` ? Because output channels are blocked. 
 
Why the problem is not addressable by Falco? Because such a situation indicates an I/O problem downstream to Falco (for example, a network slowdown) or, more rarely, a misconfiguration (for instance, imagine configuring the `program` output with a program that blocks for more than 2 seconds, you can simulate this using `sleep 10`).

So `output_timeout` is just an emergency mechanism to notify the user that something (outside Falco) is blocking Falco and Falco is giving up because it can't handle the problem. It has no direct with Falco's performance or with its behavior during normal operations.

leogr, line 319:
```suggestion
outputs_queue:
  capacity: 0
```

I'd prefer this way just because it's more consistent with other options (I know - we are not consistent in all cases), and it can handle future queue options.

Alternatively, we can put it under `outputs` (since the purpose is similar).
```
outputs:
  rate: 0
  max_burst: 1000
  queue_capacity: 0
```

incertum: We can put it under outputs, where to put the recovery then and what wording?

leogr: > We can put it under outputs, where to put the recovery then and what wording?

Sorry, I did not notice that. 
:thinking: At this point, I believe this :point_down: would be preferable

```
outputs_queue:
	  capacity: 0
	  recovery: exit
```	  

PS
As we do with `log_level` , I'd prefer strings for enums, not just numbers
https://github.com/falcosecurity/falco/blob/e0c6c9de8463f2ab8fc1de50cfec8710e646b439/falco.yaml#L488

incertum: Ok will do!

incertum: Let me read more carefully and check later. Meanwhile would you have scaling / relationship recommendations for the rate and bursts wrt queue capacity? https://github.com/falcosecurity/falco/pull/2711#discussion_r1281348771

userspace/falco/falco_outputs.cpp
```diff
@@ -268,8 +276,21 @@ inline void falco_outputs::push(const ctrl_msg& cmsg)
...
+		{
+		case falco_common::RECOVERY_EXIT:
+			fprintf(stderr, "Fatal error: Output queue out of memory. Exiting ... \n");
+			exit(EXIT_FAILURE);
+		case falco_common::RECOVERY_EMPTY:
+			m_outputs_queue_num_drops += m_queue.size();
+			fprintf(stderr, "Output queue out of memory. Empty queue and continue ... \n");
+			m_queue.empty();
+			break;
+		default:
```
Andreagit97, line 289: I would use the `case falco_common::RECOVERY_CONTINUE:` here and the `default:` could become just an `exit(EXIT_FAILURE);` with a log message like `"strategy unknown"`. In this way we can check it will never happen, it's more future proof

userspace/falco/falco_outputs.cpp
```diff
@@ -268,8 +276,21 @@ inline void falco_outputs::push(const ctrl_msg& cmsg)
 {
 	if (!m_queue.try_push(cmsg))
 	{
-		fprintf(stderr, "Fatal error: Output queue reached maximum capacity. Exiting.\n");
-		exit(EXIT_FAILURE);
+		switch (m_recovery)
```
Andreagit97, line 279: Not sure if we are 100% thread safe here :thinking:

userspace/falco/stats_writer.cpp
```diff
@@ -84,6 +84,12 @@ stats_writer::stats_writer(
 	: m_initialized(false), m_total_samples(0)
 {
 	m_config = config;
+	// capacity and controls should not be relevant for stats outputs, adopt capacity
+	// for completeness, but do not implement config recovery strategies.
+	if (config->m_outputs_queue_capacity > 0)
```
Andreagit97, line 89: same as before if we use `-1` as default

userspace/falco/stats_writer.cpp
```diff
@@ -84,6 +84,12 @@ stats_writer::stats_writer(
...
+	{
+		m_queue.set_capacity(config->m_outputs_queue_capacity);
```
Andreagit97, line 91: just a general question, Why did we choose to have a separate queue with respect to the outputs one?

falco.yaml
Andreagit97: thank you Leo for the good explanation! 
What about deprecating the `rate_limiter` (just put a deprecation warning in Falco and a `[DEPRACATED]` label in the config) in this release `0.36` and then remove it in `0.37`? WDYT?

falco.yaml
```diff
@@ -304,6 +306,33 @@ outputs:
...
+# killer stops the Falco process. We provide recovery actions to self-limit or self-kill
+# in order to handle this situation earlier, similar to how we expose the kernel buffer size
+# as a parameter.
+# However, it will not address the root cause of the event pipe not keeping up.
+# 
+# `capacity`: the maximum number of items allowed in the queue, defaulting to 0. This means that
+# the queue remains unbounded aka this setting is disabled.
+# You can experiment with values greater or smaller than the anchor value 1000000.
+# 
+# `recovery`: the strategy to follow when the queue becomes filled up. This also applies when
```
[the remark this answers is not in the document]
leogr: I agree with @Andreagit97 

However, the message says:
> This also applies when the queue is unbounded, and all available memory on the system is consumed.

@incertum Can you elaborate more on how this recovery strategy works when the queue is unbounded?
:thinking:

userspace/falco/falco_outputs.cpp
```diff
@@ -66,6 +67,13 @@ falco_outputs::falco_outputs(
 	}
 
 	m_worker_thread = std::thread(&falco_outputs::worker, this);
+	if (outputs_queue_capacity > 0)
+	{
+		m_queue.set_capacity(outputs_queue_capacity);
+	}
+
+	m_recovery = outputs_queue_recovery;
+	m_outputs_queue_num_drops = 0UL;
```
leogr, line 76: Is there any particular reason to do this after 
```
m_worker_thread = std::thread(&falco_outputs::worker, this);
```
and not before? :thinking:

userspace/falco/falco_outputs.cpp
```diff
@@ -268,8 +276,21 @@ inline void falco_outputs::push(const ctrl_msg& cmsg)
...
-		fprintf(stderr, "Fatal error: Output queue reached maximum capacity. Exiting.\n");
-		exit(EXIT_FAILURE);
+		switch (m_recovery)
+		{
+		case falco_common::RECOVERY_EXIT:
+			fprintf(stderr, "Fatal error: Output queue out of memory. Exiting ... \n");
+			exit(EXIT_FAILURE);
+		case falco_common::RECOVERY_EMPTY:
+			m_outputs_queue_num_drops += m_queue.size();
+			fprintf(stderr, "Output queue out of memory. Empty queue and continue ... \n");
```
leogr, line 286: I believe it is worth informing the user that we are dropping alerts here.

userspace/falco/falco_outputs.cpp
```diff
@@ -268,8 +276,21 @@ inline void falco_outputs::push(const ctrl_msg& cmsg)
...
+			m_queue.empty();
+			break;
+		default:
+			m_outputs_queue_num_drops++;
+			fprintf(stderr, "Output queue out of memory. Continue on ... \n");
```
leogr, line 291: I believe it is worth informing the user that we are dropping alerts here.

leogr (commented):
I just left a few comments.

Also, please keep in mind that this PR should be rebased, and we must make sure the changes introduced by https://github.com/falcosecurity/falco/pull/2663 are preserved (i.e. the WASM build support)


Finally, a question. Since we are dropping a message from the queue, what will happen when a control message is lost (for example, `ctrl_msg_type::CTRL_MSG_STOP` or 
`ctrl_msg_type::CTRL_MSG_REOPEN`)?

I wonder if it can be a problem in some case :thinking:

userspace/falco/stats_writer.cpp
incertum: Also wondered about that when I added the metrics ... perhaps potential for future consolidation?

userspace/falco/stats_writer.cpp
incertum: will do

userspace/falco/falco_outputs.cpp
incertum: Probably not, maybe @jasondellaluce once back may have additional ideas that we could also address in a follow up PR since Jason worked a lot on the thread safety aspects all around.

falco.yaml
```diff
@@ -304,6 +306,33 @@ outputs:
...
+# the queue is unbounded, and all available memory on the system is consumed.
+# `exit` is default, `continue` does nothing special and `empty` empties the queue and then
+# continues.
+outputs_queue:
+  capacity: 0
```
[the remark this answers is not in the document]
leogr: @Andreagit97 @incertum 

Before deciding to use `-1`, please ensure this is consistent with the rest of the Falco config. IIRC, our convention is to use `0` in such cases. For example

> # When the `threadiness` value is set to 0, Falco will automatically determine
>  # the appropriate number of threads based on the number of online cores in the system.
>  threadiness: 0

I don't have strong preferences, but I would like to use the same convention across the whole config. Users should not be aware of internals.

Do you agree?

falco.yaml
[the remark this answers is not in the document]
leogr: btw,

> try_push(item) pushes item only if it would not exceed the queue’s capacity.

So, in case of an unbound queue, `try_push` will never return false, if I understood correctly.

incertum:
> Also, please keep in mind that this PR should be rebased, and we must make sure the changes introduced by #2663 are preserved (i.e. the WASM build support)

Let me rebase very carefully and then implement each change request very carefully later on.

Also @Andreagit97 great suggestions all around!

> Finally, a question. Since we are dropping a message from the queue, what will happen when a control message is lost (for example, `ctrl_msg_type::CTRL_MSG_STOP` or `ctrl_msg_type::CTRL_MSG_REOPEN`)?
> 
> I wonder if it can be a problem in some case 🤔

Good question, we need to check.

falco.yaml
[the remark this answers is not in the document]
incertum: yes until OOM and at that point all was lost anyways in the past.

falco.yaml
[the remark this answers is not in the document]
incertum: I left the default value 0, but internally I followed @Andreagit97 suggestion.

falco.yaml
incertum: +1 on this @Andreagit97: 

Don't wanna come across as always being in favor of just getting rid of things, but at the same time, we do need to improve technical clarity and clean things up. Referring to it as "technical debt" without any negative connotation -- it's normal in every project.

@Andreagit97 would you be willing to run point on this? Asking because you know I will be out most of Sep and will limit code contributions now for this release, but happy to help as reviewer!

leogr: I don't have a strong opinion on the rate limiter. I wonder if it can still be helpful with some use cases or not.
I vaguely recall I had a conversation with @jasondellaluce a while ago.

Anyway, I'm okay with removing it if we get a consensus. In such a case, I recommend putting a deprecation warning for at least 1 release before completely removing it.

incertum: Hmm I am just confused that we then silently drop events without even counting these drops.  Is there another way to do it? If we keep it can we at least add drop counters in?

I guess it comes back to Falco being a bit of a different type of a service, where the end goal would be to not produce too many alerts and try to never drop an event.

userspace/falco/configuration.cpp
```diff
@@ -258,6 +260,18 @@ void falco_configuration::load_yaml(const std::string& config_name, const yaml_h
 	}
 
 	m_buffered_outputs = config.get_scalar<bool>("buffered_outputs", false);
+	m_outputs_queue_capacity = config.get_scalar<size_t>("outputs_queue.capacity", DEFAULT_OUTPUTS_QUEUE_CAPACITY_UNBOUNDED_MAX_LONG_VALUE);
```
Andreagit97, line 263: can we use an `int64_t` instead of `size_t`? in this way, the `DEFAULT_OUTPUTS_QUEUE_CAPACITY` could be simply be `-1`

falco.yaml
Andreagit97: yeah I will open an issue to deprecate it and see if we reach consensus

Andreagit97 (commented):
The PR LGTM! I still have an open point that I would like to understand unrelated to the PR,
why do we have 2 queues, one for outputs and another for metrics? @jasondellaluce @leogr

userspace/falco/falco_outputs.cpp
Andreagit97: what would happen in multi-source scenarios?

Andreagit97:
> Finally, a question. Since we are dropping a message from the queue, what will happen when a control message is lost (for example, ctrl_msg_type::CTRL_MSG_STOP or
ctrl_msg_type::CTRL_MSG_REOPEN)?

This is a good question, actually, the worker thread won't stop because it doesn't receive the stop message... My real question is, should a user use Falco with alert drops? I mean, if you drop alerts it means that Falco is no longer useful, so yes I get the point of having a bounded queue but if the user faces some alert drops or there is a bug or he has to configure Falco with a great queue capacity :thinking:

userspace/falco/configuration.cpp
incertum: We can, at the same time always preferring to stay close to the framework, here TBB, and they use `size_t`. Let me know if its a [nit] or if we significantly gain from this change, then I'll do it.

Andreagit97:
on a second thought, I'm asking myself if we really want the `continue` and `empty` policies :thinking: 

IMO what we need to achieve is to prevent the OOM killer from killing us and provide users with a graceful shutdown. If we drop alerts this is a symptom of a bug because we have a dedicated thread that just outputs logs all the time so it's almost impossible that it is not able to keep up with the main thread. It could happen if the tool is misconfigured (too noise, so rule misconfiguration, too outputs, so tool misconfiguration). In my opinion in this case we want to report the problem to the user and stop Falco instead of "hiding" the issue keeping up with drops. Moreover as pointed out by @leogr we could face issues if we drop CTRL messages... WDYT? @incertum @jasondellaluce @leogr 

Please let's try to reach a consensus before Melissa go away in this way we can close this workstream :+1:

incertum:
> > Finally, a question. Since we are dropping a message from the queue, what will happen when a control message is lost (for example, ctrl_msg_type::CTRL_MSG_STOP or
> > ctrl_msg_type::CTRL_MSG_REOPEN)?
> 
> This is a good question, actually, the worker thread won't stop because it doesn't receive the stop message... 

Currently there can also be issues with the control messages being possibly stuck a long time. Personally think the control messages should not be in the alerts output queue and rather directly be processes or on a separate hot path. @leogr and I chatted and concluded it's not a high priority. Would vouch for cleaning it up in a separate PR regardless.

> My real question is, should a user use Falco with alert drops? I mean, if you drop alerts it means that Falco is no longer useful, so yes I get the point of having a bounded queue but if the user faces some alert drops or there is a bug or he has to configure Falco with a great queue capacity 🤔

I do agree that Falco shall not be used with random alert sampling / dropping or dropping based the queue filling up. Here is how I think about this initial change: It's a first step to perform data-driven decisions vs speculations to see how much of an issue this can be in production and help determine if this could cause the memory leaks on some servers or it shows it's not a problem and we have to look elsewhere. Based on some field experience we can come back and optimize this. For example, it would be interesting to see if there could be few kernel side drops, but because Falco is used for too verbose auditing / alerting we fail to keep up in userspace and such.

incertum:
> on a second thought, I'm asking myself if we really want the `continue` and `empty` policies 🤔
> 

Let's experiment with it. Maybe `empty` can help handling short bursts vs restarting is a more heavy undesired recovery. If it's not useful, we remove the recovery strategies again, since we tag this as "Experimental" we can do this easily.

> IMO what we need to achieve is to prevent the OOM killer from killing us and provide users with a graceful shutdown. If we drop alerts this is a symptom of a bug because we have a dedicated thread that just outputs logs all the time so it's almost impossible that it is not able to keep up with the main thread. It could happen if the tool is misconfigured (too noise, so rule misconfiguration, too outputs, so tool misconfiguration). In my opinion in this case we want to report the problem to the user and stop Falco instead of "hiding" the issue keeping up with drops. Moreover as pointed out by @leogr we could face issues if we drop CTRL messages... WDYT? @incertum @jasondellaluce @leogr
> 
> Please let's try to reach a consensus before Melissa go away in this way we can close this workstream 👍

ACK, few more thoughts: I believe the new Falco metrics counters will be helpful to gain a better understanding about what could be going on on troublesome servers, hence we do count the new drops just like the kernel side drops. In addition, adding the capacity config is also good software engineering practices and effectively will make the kernel buffer and output queue equivalent in terms of end user exposed pre-defined bounds.

Re your last point what the end user should do after possibly finding out that the output queue is not working ... I would defer to this discussion once we have some data.

Andreagit97:
> Let's experiment with it. Maybe empty can help handling short bursts vs restarting is a more undesired recovery. If it's not useful, we remove the recovery strategies again, since we tag this as "Experimental" we can do this easily

Yep since they are experimental we could also introduce them for now but i would consider it again before promoting them as stable

leogr (approved):

userspace/falco/falco_outputs.cpp
```diff
@@ -274,8 +278,25 @@ inline void falco_outputs::push(const ctrl_msg& cmsg)
 #ifndef __EMSCRIPTEN__
 	if (!m_queue.try_push(cmsg))
 	{
-		fprintf(stderr, "Fatal error: Output queue reached maximum capacity. Exiting.\n");
-		exit(EXIT_FAILURE);
+		switch (m_outputs_queue_recovery)
+		{
+		case falco_common::RECOVERY_EXIT:
+			falco_logger::log(LOG_ERR, "Fatal error: Output queue out of memory. Exiting ...");
```
jasondellaluce, line 284: Why not throwing an exception here? It would cause one open event source thread to be stopped (and in cascade, all the others open in parallel), and Falco to terminate gracefully.

userspace/falco/falco_outputs.cpp
```diff
@@ -274,8 +278,25 @@ inline void falco_outputs::push(const ctrl_msg& cmsg)
...
+			exit(EXIT_FAILURE);
+		case falco_common::RECOVERY_EMPTY:
+			m_outputs_queue_num_drops += m_queue.size();
```
jasondellaluce, line 287: Stating the docs of `falco_outputs`:
```
All methods in this class are thread-safe. The output framework supports a multi-producer model where messages are stored in a queue and consumed by each configured output asynchrounously.
```

Since `m_queue` is a thread-safe object as guaranteed by TBB, I'd say `m_outputs_queue_num_drops` should be atomic here.

jasondellaluce (commented):
/hold

Found a thread safety hazard. Overall looks good!

userspace/falco/falco_outputs.cpp
leogr: good catch!

incertum: Also realized I think when counting the number of drops for the empty queue case I need to add  +1 will do this later.

userspace/falco/falco_outputs.cpp
incertum: Sounds good will do!

Andreagit97 (approved):
/approve

leogr (approved):

PR #2718, opened 2023-08-04
new(falco): update the Falco base image to debian 12

### LINKED ISSUE #2489: Discussion: Updating the default falco-driver-loader image
As we know we have three main avenues install the system instrumentation to get live syscall info into Falco:

- The kernel module, for most kernel versions starting from selected 2.6.xx
- The classic eBPF probe, for selected versions of kernels 4.14 and above
- The modern eBPF probe, for kernel versions 5.8 and above with BTF support (if you use this, you have none of the issues that are described below, but relatively few distros support that as of today)

The first two are the most compatible but require building the drivers or downloading a prebuilt for your kernel. While we have [thousands](https://download.falco.org/?prefix=driver/4.0.0%2Bdriver/x86_64/) of prebuilts it is sometimes necessary to build your own driver. If you're using a container-based install, this is what the [falco-driver-loader image](https://github.com/falcosecurity/falco/blob/master/docker/falco/Dockerfile) attempts to do.

That image has a noble purpose, to try and build the driver for the live system it is on provided that kernel headers are present, but it's not without issues. For historical reasons, a ton of old compilers are still installed in a relatively old image (Debian Buster) to retain compatibility with older kernels (3.x and below). However, that image is unable to build the driver for newer versions (kernel > 5.0 and perhaps some 4.xx as well). In addition, if you try to run the image through a SCA / vulnerability scanner tool it will alert about maybe hundreds of vulnerabilities which are present in very old compilers and accompanying libs that are necessary for older distributions; for all practical security purposes those are all false positives but they would show up flashing on a security status dashboard and would make compliance experts feel uneasy.

The sad fact is that it's not even possible to build an image that works for everyone. In fact, even [driverkit](https://github.com/falcosecurity/driverkit) has resorted to using several different images to to this job.

This led me to try and look into the compiler versions we actually use when creating our prebuilt and trying to understand if we want to upgrade the default driver loader image. To learn which compilers we may want to put there I ran a little experiment: took all the current driverkit configs we use and looked at all the image selections. The results are as follows (after removing duplicate kernel versions):

```
Count Name
----- ----
    1 falcosecurity/driverkit-builder-any-x86_64_gcc10.0.0_gcc9.0.0:latest
 1457 falcosecurity/driverkit-builder-any-x86_64_gcc12.0.0_gcc11.0.0:latest
  399 falcosecurity/driverkit-builder-any-x86_64_gcc8.0.0_gcc6.0.0_gcc5.0.0_gcc4.9.0_gcc4.8.0:latest  <<-- this is the current Falco image
    1 falcosecurity/driverkit-builder-centos-x86_64_gcc4.8.5:latest
```

Basically, we use the gcc 11-12 image most of the times. In fact, as a rule of thumb, we use the **gcc11-12 image for 5.x and 6.x** while we use the old image for 4.x and below. So, we're recently picking up many variations of kernel 5.x because it seems like it's the most widely supported at the moment (Ubuntu, Amazon Linux, some recent RHEL and CentOS...) while 4.x versions are not seeing many updates.

I believe it may be time to upgrade the base image to something like this: https://github.com/falcosecurity/driverkit/blob/master/docker/builders/builder-any-x86_64_gcc12.0.0_gcc11.0.0.Dockerfile since building on 5.x and 6.x seems to be becoming more important than the previous versions. If we could also add another gcc version that is compatible with some 4.x I believe it would work even better. 

This issue acts as a place of discussion about what we want to do. There may be more ideas, and even more things we want to do for people that cannot use this base image, but I did want to open this discussion.

/kind design

LucaGuerra:
**What type of PR is this?**

/kind feature
/area build
/area CI

As per our discussion in https://github.com/falcosecurity/falco/issues/2489 and during yesterday's core maintainer call, I propose to:
* Upgrade the Falco base image with the one that seems to build the most kernels at the moment
* Still distribute the old one, marked as legacy, adding the necessary CI/CD job steps

Fixes #2489

```release-note
update!: the Falco base image is now based on Debian 12 with gcc 11-12
new: the legacy falco image is available as driver-loader-legacy
```

### DIFF
--- a/docker/falco/Dockerfile
+++ b/docker/falco/Dockerfile
@@ -19,37 +19,17 @@ RUN cp /etc/skel/.bashrc /root && cp /etc/skel/.profile /root
 
 RUN apt-get update \
 	&& apt-get install -y --no-install-recommends \
-	bash-completion \
-	bc \
-	bison \
 	ca-certificates \
 	clang \
-	cpio \
 	curl \
 	dkms \
-	dwarves \
-	flex \
 	gcc \
 	gcc-11 \
 	gnupg2 \
-	gpg \
 	jq \
-	libc6-dev \
-	libelf-dev \
-	libiberty-dev \
-	libncurses-dev \
-	libpci-dev \
-	libssl-dev \
-	libudev-dev \
+	libelf1 \
 	llvm \
-	lsb-release \
-	netcat-openbsd \
-	openssl \
-	rpm2cpio \
-	software-properties-common \
-	wget \
-	xz-utils \
-	zstd \
+	make \
 	&& rm -rf /var/lib/apt/lists/*
 
 RUN curl -s https://falco.org/repo/falcosecurity-packages.asc | apt-key add - \
--- a/docker/falco/Dockerfile
+++ b/docker/falco/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:buster
+FROM debian:bookworm
 
 LABEL maintainer="cncf-falco-dev@lists.cncf.io"
 LABEL org.opencontainers.image.source="https://github.com/falcosecurity/falco"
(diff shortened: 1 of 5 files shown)


docker/falco/Dockerfile
```diff
@@ -22,72 +22,39 @@ RUN apt-get update \
...
-# Since our base Debian image ships with GCC 7 which breaks older kernels, revert the
-# default to gcc-5.
-RUN rm -rf /usr/bin/gcc && ln -s /usr/bin/gcc-5 /usr/bin/gcc
-
-RUN rm -rf /usr/bin/clang \
-	&& rm -rf /usr/bin/llc \
-	&& ln -s /usr/bin/clang-7 /usr/bin/clang \
-	&& ln -s /usr/bin/llc-7 /usr/bin/llc
+# Properly create soft links
+RUN ln -s /usr/bin/gcc-11 /usr/bin/gcc-11.0.0
```
FedeDP, line 56: We don't need these soft links here :) 
They are only needed by driverkit builder images selection.

FedeDP (commented):
Left a minor cleanup comment.
Everything else LGTM!

docker/falco/Dockerfile
LucaGuerra: Thank you @FedeDP ! Removed those two lines.

FedeDP (dismissed):
/approve

docker/falco/Dockerfile
```diff
@@ -22,73 +22,36 @@ RUN apt-get update \
...
+	libudev-dev \
+	llvm \
+	lsb-release \
+	netcat-openbsd \
+	openssl \
+	rpm2cpio \
+	software-properties-common \
+	wget \
 	xz-utils \
+	zstd \
```
leogr, line 52: This list seems a bit too long :angel: 

Perhaps, some of them are not needed. Could we double check, please?

leogr (commented):
Thank you! :star_struck: 

SGTM, just left one comment :point_down:

docker/falco/Dockerfile
LucaGuerra: Great catch! I'll check with `falco-driver-loader` to see what we actually need. :+1:

LucaGuerra: Done! I tested falco-driver-loader with download and compilation of the kmod and ebpf probe.

LucaGuerra:
Thanks @leogr ! I commend your support for improving our supply chain security 🫡 (I admit I should have thought about it first)

leogr (approved):
great :partying_face:

PR #2728, opened 2023-08-14
Allow specifying explicit kernel release and version for `falco-driver-loader`

### LINKED ISSUE #2658: Allow specifying explicit kernel release and version for `falco-driver-loader`
**Motivation**

I'm installing Falco using Ansible as part of a Packer build which results in a VM image. As part of the build process I'm updating the kernel to a newer version. When running `falco-driver-loader --compile module`, the module is compiled and installed for the **running kernel**:

https://github.com/falcosecurity/falco/blob/fbcb971751060a7f73fbb93ff9d5ac5bc80c7517/scripts/falco-driver-loader#L679
https://github.com/falcosecurity/falco/blob/fbcb971751060a7f73fbb93ff9d5ac5bc80c7517/scripts/falco-driver-loader#L685

However, since the kernel version was updated but the VM hadn't been rebooted since the update, the result is that the kmod isn't compiled for the kernel version that is used in practice when creating VMs from the image.

One way to solve the problem is to explicitly specify the kernel release and kernel version when running `falco-driver-loader`. However, the script doesn't currently support this.

**Feature**

Maybe it makes sense to add the following two argument to `falco-driver-loader` and default to the current behavior if either of the vars isn't specified:

* `KERNEL_RELEASE`
* `KERNEL_VERSION`

I'm happy to open PR if this sounds like a good idea.

**Alternatives**

I could try to insert a reboot to my image building process, however this would be suboptimal. In addition, I imagine many Falco users rely on automation for installing Falco and therefore may benefit from being able to explicitly specify the kernel for which to compile the driver. As my use case demonstrates, the assumption that Falco is used with the currently-running kernel isn't always true.

**Additional context**

Whatever gets decided, we should ensure the functionality works for both kmod and BPF modes.

johananl:
**What type of PR is this?**

/kind feature

> /area build

> /area engine

> /area tests

> /area proposals

> /area CI

This PR adds the ability to specify explicit kernel release and version when running `falco-driver-loader`. This is important in cases where the Falco driver needs to be compiled for a kernel other than the currently-running kernel (such as when the kernel had been updated but not yet booted into).

Fixes #2658 

None

```release-note
feat: Allow specifying explicit kernel release and version for falco-driver-loader
```

### DIFF
--- a/scripts/falco-driver-loader
+++ b/scripts/falco-driver-loader
@@ -163,7 +163,7 @@ get_target_id() {
 		then
 			ARCH_extra="-${BASH_REMATCH[1]}${BASH_REMATCH[2]}"
 		fi
-		if [[ $(uname -v) =~ ([0-9]+\.[0-9]+\.[0-9]+\-[0-9]+) ]];
+		if [[ ${DRIVER_KERNEL_VERSION} =~ ([0-9]+\.[0-9]+\.[0-9]+\-[0-9]+) ]];
 		then
 			KERNEL_RELEASE="${BASH_REMATCH[1]}${ARCH_extra}"
 		fi
@@ -187,9 +187,9 @@ get_target_id() {
 		# from the following `uname -v` result
 		# `#26~22.04.1-Ubuntu SMP Mon Apr 24 01:58:15 UTC 2023`
 		# we obtain the kernelversion`26~22.04.1`
-		if [[ $(uname -v) =~ (^\#[0-9]+\~[^-]*-Ubuntu .*$) ]];
+		if [[ ${DRIVER_KERNEL_VERSION} =~ (^\#[0-9]+\~[^-]*-Ubuntu .*$) ]];
 		then
-			KERNEL_VERSION=$(uname -v | sed 's/#\([^-\\ ]*\).*/\1/g')
+			KERNEL_VERSION=$(echo "${DRIVER_KERNEL_VERSION}" | sed 's/#\([^-\\ ]*\).*/\1/g')
 		fi
 		;;
 	("flatcar")
@@ -258,7 +258,7 @@ flatcar_relocate_tools() {
 
 load_kernel_module_compile() {
 	# Skip dkms on UEK hosts because it will always fail
-	if [[ $(uname -r) == *uek* ]]; then
+	if [[ ${DRIVER_KERNEL_RELEASE} == *uek* ]]; then
 		>&2 echo "Skipping because the dkms install always fail (on UEK hosts)"
 		return
 	fi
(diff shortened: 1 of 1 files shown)


FedeDP:
/milestone 0.36.0
Thanks for taking the time to tackle this one! We are going to look into it asap!

Andreagit97 (approved):
Thank you for this!
/approve

scripts/falco-driver-loader
```diff
@@ -535,9 +535,9 @@ load_bpf_probe_compile() {
 	fi
 
 	if [ -n "${BPF_USE_LOCAL_KERNEL_SOURCES}" ]; then
-		local -r kernel_version_major=$(uname -r | cut -d. -f1)
-		local -r kernel_version=$(uname -r | cut -d- -f1)
-		KERNEL_EXTRA_VERSION="-$(uname -r | cut -d- -f2)"
+		local -r kernel_version_major=$(echo "${DRIVER_KERNEL_RELEASE}" | cut -d. -f1)
```
FedeDP, line 538: Wouldn't it be better to use `KERNEL_RELEASE` here?
I mean, i think we should use `KERNEL_RELEASE` and `KERNEL_VERSION` everywhere, not their `DRIVER_` counterparts.

leogr: Good catch!

johananl: The places where I used e.g. `DRIVER_KERNEL_RELEASE` are places where there was an explicit call to `uname -r`. Note that the values of `KERNEL_RELEASE` and `KERNEL_VERSION` change throughout the script, which can be tricky to follow as there is lots of distro-specific logic and special cases.

My idea was to keep existing functionality as is by storing the result of `uname -r` and `uname -v` **once** at the start of the execution, then reuse these values. This allows specifying explicit release and version values as arguments without altering the entire script.

Does that make sense? Did I miss anything?

leogr: Oh, I see. It makes sense to me now. 

This also adheres to our implicit convention that the `DRIVER_` prefix represents input values (those that may be set using env vars), whereas other vars like `KERNEL_RELEASE` are just for internal use.

:+1: 

Thank you!

N.B.: I've checked the whole PR. Just giving my 2 cents (specifically related to the above comment).

johananl:
Forgot to mention that I tested my changes by executing the script manually in both modes (kmod and BPF) with default values as well as with explicit kernel release and version on the following distros:

* Amazon Linux
* Debian (kmod compilation didn't work for me, but that's true even before my changes)
* Ubuntu
* Flatcar (download doesn't work because the driver repo doesn't contain files for Flatcar, bpf compilation didn't work due to https://github.com/falcosecurity/falco/pull/2729 but should work now)

If we refactor the script to be functional, we could use something like https://github.com/bats-core/bats-core to unit-test the file.

scripts/falco-driver-loader
johananl: Yes, that makes sense to me.

FedeDP (approved):
/approve

PR #2739, opened 2023-08-23
feat(userspace): deprecate stats command args option in favor of metrics configs in falco.yaml

incertum:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind release

> /area build

/area engine

> /area tests

> /area proposals

> /area CI

As scheduled deprecate stats command args option in favor of metrics configs in falco.yaml for Falco 0.36 release.

@jasondellaluce 

Fixes #

```release-note
feat(userspace): deprecate stats command args option in favor of metrics configs in falco.yaml
```


falco.yaml
```diff
@@ -613,7 +613,7 @@ syscall_event_drops:
   max_burst: 1
   simulate_drops: false
 
-# [Experimental] `metrics`
+# [Stable] `metrics`
```
incertum, line 616: @FedeDP @leogr WDYT? Other than that one odd behavior where the timer did not work when using http output (which Federico fixed) I have not heard of more issues when using the metrics feature, have you? It is working well for me in production.

Proposing to bump to Stable, which does not mean that there may be no additional bugs, it rather signals we are supporting `metrics` now as stable feature going forward.

FedeDP: I think this is good 2 go! Agree with you.
@leogr more thoughts?

Andreagit97 (commented):
Just one minor cleanup otherwise LGTM!
Talking with Fede, he correctly pointed out that removing `-s` would break all current Falco settings that are running with this flag enabled by default, but in the end, we depracted it in Falco 0.35 so it should be fine

userspace/falco/app/options.cpp
```diff
@@ -219,8 +219,6 @@ void options::define(cxxopts::Options& opts)
 		("p,print",                       "Add additional information to each falco notification's output.\nWith -pc or -pcontainer will use a container-friendly format.\nWith -pk or -pkubernetes will use a kubernetes-friendly format.\nAdditionally, specifying -pc/-pk will change the interpretation of %container.info in rule output fields.", cxxopts::value(print_additional), "<output_format>")
 		("P,pidfile",                     "When run as a daemon, write pid to specified file", cxxopts::value(pidfilename)->default_value("/var/run/falco.pid"), "<pid_file>")
 		("r",                             "Rules file/directory (defaults to value set in configuration file, or /etc/falco_rules.yaml). This option can be passed multiple times to read from multiple files/directories.", cxxopts::value<std::vector<std::string>>(), "<rules_file>")
-		("s",                             "If specified, append statistics related to Falco's reading/processing of events to this file (only useful in live mode).", cxxopts::value(stats_output_file), "<stats_file>")
```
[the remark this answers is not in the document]
incertum: Thank you! Also rebased since we adjusted the build setup a bit.

Andreagit97 (approved):
/approve
Thanks

FedeDP (approved):
/approve

FedeDP:
Per https://github.com/falcosecurity/falco/pull/2739#issuecomment-1690755613, do we need an engine bump?

Andreagit97:
> Per https://github.com/falcosecurity/falco/pull/2739#issuecomment-1690755613, do we need an engine bump?

uhm the message was printed because this PR touches a file under `/userspace/engine/*` but not sure we need an engine version bump in the end the logic is almost the same we removed a fallback check :thinking:  cc @leogr

userspace/falco/app/actions/process_events.cpp
```diff
@@ -435,7 +435,6 @@ static falco::app::run_result init_stats_writer(
 		falco_logger::log(LOG_WARNING, "Metrics are enabled with no output configured, no snapshot will be collected");
 	}
```
Andreagit97, line 437: I cannot comment on the right line but above (https://github.com/falcosecurity/falco/pull/2739/files#diff-3a3a5ebb7d4bf516a4c732cadcd83feabe05a5cc76a9ca9e1f55802a94b3c4a7R428), there is a check on the prometheus stats format that say 
```cpp
falco_logger::log(LOG_WARNING, "Metrics interval was passed as numeric value without Prometheus time unit, this option will be deprecated in the future");
``` 
Probably we want to turn it into an `falco::app::run_result::fatal` , WDYT?
It could be a future cleanup if we want merge this :)

leogr:
> > Per [#2739 (comment)](https://github.com/falcosecurity/falco/pull/2739#issuecomment-1690755613), do we need an engine bump?
> 
> uhm the message was printed because this PR touches a file under `/userspace/engine/*` but not sure we need an engine version bump in the end the logic is almost the same we removed a fallback check 🤔 cc @leogr

Correct, it was a false positive. We can merge this without bumping the engine version.
cc @jasondellaluce

leogr (approved):
/hold cancel

PR #2743, opened 2023-08-28
update(ci): minimize retention days for build-only CI artifacts

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/area CI

Github Actions have a default retention period of 90 days. However, some of our artifacts are used at build-time only to pass around artifacts between different CI jobs, thus we don't need them after the workflow is finished. In particular, `falco-images` weights ~4GB, which is a big waste of storage.

```release-note
NONE
```


Andreagit97 (approved):
/approve

FedeDP (approved):
/approve

PR #2747, opened 2023-08-30
update(cmake): support building libs and driver from forks

jasondellaluce:
**What type of PR is this?**

/kind feature

/area build

Whoever of us ever tried to build Falco with their own libs fork when testing changes in libs, had to hack a patch like this one. So for sake of completeness I'll propose supporting this natively in our cmake files. Example of how this could be used:

```
cmake -DFALCOSECURITY_LIBS_REPO=jasondellaluce/falco-libs ..
make falco -j4
...
```

```release-note
update(cmake): support building libs and driver from forks
```


Andreagit97 (approved):
/approve

PR #2753, opened 2023-08-31
fix(userspace/engine): support appending to unknown sources

jasondellaluce:
**What type of PR is this?**

/kind bug

/area engine

In case we have a rule using an unknown source (e.g. when its plugin is not leaded), we have a bug that prevents appending to that rule. For example:

```yaml
- rule: Rule1
  desc: NoDesc
  condition: evt.type=open
  priority: INFO
  output: Never
  source: mysource

- rule: Rule1
  append: true
  condition: or evt.type=openat
```

```release-note
fix(userspace/engine): support appending to unknown sources
```

### DIFF
--- a/userspace/engine/falco_engine.cpp
+++ b/userspace/engine/falco_engine.cpp
@@ -568,12 +568,12 @@ void falco_engine::describe_rule(std::string *rule, bool json) const
 		output["lists"] = lists_array;
 
 		json_str = writer.write(output);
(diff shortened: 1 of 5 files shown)


jasondellaluce:
> 
> 

/unhold

This is a just a bug fix, so this is a false positive.

incertum (approved):
/approve

PR #2762, opened 2023-09-01
cleanup: deprecate rate limiter mechanism

Andreagit97:
**What type of PR is this?**

/kind cleanup

/area engine

This PR deprecates the rate limiter mechanism since it seems to be no longer used and moreover, it discards events, so one can potentially lose critical alerts. For this reason, it's not enabled by default.

See this discussion for more details https://github.com/falcosecurity/falco/pull/2711#discussion_r1281346149

```release-note
cleanup: deprecate rate limiter mechanism
```

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -266,6 +266,9 @@ buffered_outputs: false
 
 # [Stable] `outputs`
 #
+# [DEPRECATED]
+# This config is deprecated and it will be removed in Falco 0.37
+#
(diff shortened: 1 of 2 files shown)


incertum (approved):
I support this because the current implementation of rate limiting silently drops alerts, which is undesirable for Falco.

/approve

FedeDP (approved):
/approve

PR #2771, opened 2023-09-05
new(falco): print all events with flags

LucaGuerra:
**What type of PR is this?**

/kind feature
/area engine

This patch changes the way events are printed, especially in markdown format. See this page: https://falco.org/docs/reference/rules/supported-events/ . There are significant changes:
* All events are displayed, not just syscalls
* Flags are also displayed
* Generic events are also included
* The "default" column now reflects whether or not the `-A` flag is required

Updating the event list is just a matter of running `./userspace/falco/falco --list-syscall-events --markdown` and saving the output in the `supported-events.md`  in the website.

Fixes #

```release-note
update!: the --list-syscall-events flag is now called --list-events and lists all events
```


userspace/engine/falco_engine_version.h
```diff
@@ -15,12 +15,12 @@ limitations under the License.
...
+#define FALCO_ENGINE_VERSION (26)
 
 // This is the result of running the following command:
 //   FALCO="falco -c ./falco.yaml"
 //   echo $($FALCO --version | grep 'Engine:' | awk '{print $2}') $(echo $($FALCO --version | grep 'Schema version:' | awk '{print $3}') $($FALCO --list --markdown | grep '^`' | sort) $($FALCO --list-syscall-events | sort) | sha256sum)
 // It represents the fields supported by this version of Falco,
 // the event types, and the underlying driverevent schema. It's used to
 // detetect changes in engine version in our CI jobs.
-#define FALCO_ENGINE_CHECKSUM "41b5dc700216b243d294b40c46264d4e89d0ee00098fdc1c21bb4b1e7639da06"
+#define FALCO_ENGINE_CHECKSUM "98c6e665031b95c666a9ab02d5470e7008e8636bf02f4cc410912005b90dff5c"
```
LucaGuerra, line 26: Note for the reviewer: the list of events did not change, but the format did so we need to update the checksum. Updating the engine version _should_ be OK but if I forgot something and it's not please let me know and I'll revert

userspace/falco/app/actions/print_syscall_events.cpp
```diff
@@ -55,60 +99,104 @@ static std::vector<event_entry> get_event_entries(bool include_generics, const l
...
+		const struct events_by_category events_bc = get_event_entries_by_category(true, available);
+
+		printf("## Syscall events\n\n");
+		print_events(events_bc.syscalls, s.options.markdown);
+
+		printf("\n\n## Tracepoint events\n\n");
+		print_events(events_bc.tracepoints, s.options.markdown);
+
+		printf("\n\n## Plugin events\n\n");
+		print_events(events_bc.pluginevents, s.options.markdown);
```
leogr, line 196: Do we intend to print all events (regardless of the data source) here? If so, I would change the option name from `--list-syscall-events` to `--list-events`.

leogr:
> * The "default" column now reflects whether or not the `-A` flag is required

Why do we need this? :thinking: 
It may be confusing for users.

userspace/falco/app/actions/print_syscall_events.cpp
Andreagit97: yes agree with leo i would change it to `--list-events`

LucaGuerra:
> Why do we need this? 🤔
> It may be confusing for users.

Even after our recent changes and optimizations, some events _require_ the `-A` flag to be considered.

We need to convey this information to users somehow otherwise people would (rightly) expect all events to work in all cases! If you have ideas about how to make it clearer I'd be happy to hear it but that's the one idea I found right now.

userspace/falco/app/actions/print_syscall_events.cpp
LucaGuerra: I noticed that this flag never printed syscall events, but always printed all events. Yes, I introduced it and decided the name 😅 . So I agree with you, let me change it to `--list-events`.

Andreagit97 (approved):
/approve

I'm okay with keeping the `-A` distinction

FedeDP (approved):
/approve

PR #2781, opened 2023-09-06
new(docker): allow passing options to falco-driver-loader from the cointainer image

LucaGuerra:
**What type of PR is this?**

/kind feature
/area build

Sometimes we may want to specify options for the driver loader script, for example when we may decide to only download or compile the eBPF probe rather than the kmod. This is not strictly require, more like something that is nice to have.

Fixes #

```release-note
new(docker): allow passing options to falco-driver-loader from the driver loader cointainer
```


Andreagit97 (approved):
/approve

I would document it somewhere!

docker/falco/docker-entrypoint.sh
```diff
@@ -27,7 +27,8 @@ if [[ -z "${SKIP_DRIVER_LOADER}" ]]; then
         ln -s "$i" "/usr/src/$base"
     done
 
-    /usr/bin/falco-driver-loader
+    falco_driver_loader_option_arr=($FALCO_DRIVER_LOADER_OPTIONS)
+    /usr/bin/falco-driver-loader "${falco_driver_loader_option_arr[@]}"
```
leogr, line 31:
```suggestion
    /usr/bin/falco-driver-loader ${falco_driver_loader_option_arr[@]}
```

why do we need `"` here? :thinking: 
what will happen if `$FALCO_DRIVER_LOADER_OPTIONS` contains `"`? :thinking: 

/hold

LucaGuerra: It's not a good PR review if nobody mentions either mystifying Bash rules or the C standard.

The `"` is there to prevent variable globbing and expansion, meaning that if you pass a `*` in the array it will not expand to all files in this directory. Then I tried and I noticed that there is a mistake in the line _above_, which should be `read -a args_arr <<< $FALCO_DRIVER_LOADER_OPTIONS` because otherwise ... variable expansion will happen anyways. So I'm going to fix that.

To answer the second part of the question, arguments are simply space-delimited.

You can verify by copying these two scripts:

print-args.sh
```
#!/bin/bash

for arg
do
    printf '%s\n' "$arg"
done
```

args-from-env.sh
```
#!/bin/bash

read -a args_arr <<< $MY_ARGS

echo "--- with quotes ---"
./print-args.sh "${args_arr[@]}"

echo "--- without quotes ---"
./print-args.sh ${args_arr[@]}

echo "--- without quotes and array conversion ---"
./print-args.sh ${MY_ARGS}
```

and running `MY_ARGS='hello bash * old not friend' ./args-from-env.sh`

leogr: :exploding_head:

leogr: /hold cancel

Andreagit97 (approved):
/approve

FedeDP (approved):
/approve

PR #2782, opened 2023-09-06
update(docker): use debian 12 slim for falco no driver

LucaGuerra:
**What type of PR is this?**

/kind feature
/area build

Since we're updating the large image to Debian 12 I believe it makes sense to update the nodriver one as well. It's also slightly smaller (about ~30MB smaller).

I tested it with the *modern eBPF* probe and the (previously installed) kernel module.

Fixes #

```release-note
update(docker): the Falco no-driver image is now based on Debian 12
```

### DIFF
--- a/docker/no-driver/Dockerfile
+++ b/docker/no-driver/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04 as ubuntu
+FROM debian:12 as builder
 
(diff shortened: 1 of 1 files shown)


FedeDP (approved):
/approve
Love this, thank you Luca!

Andreagit97 (approved):
/approve

PR #2789, opened 2023-09-07
fix(stats): always initialize m_output field

Andreagit97:
**What type of PR is this?**

/kind bug

/area engine

When running Falco with these changes in the default config:

```yaml
# todo: syscall_counters_enabled option
metrics:
  enabled: true
  interval: 5s
  # Typically, in production, you only use `output_rule` or `output_file`, but not both. 
  # However, if you have a very unique use case, you can use both together.
  output_rule: false
  output_file: ./falco_stats.json
  resource_utilization_enabled: true
  kernel_event_counters_enabled: true
  libbpf_stats_enabled: true
  convert_memory_to_mb: true
  include_empty_values: false
```

it crashes with a segfault. The reason was that the `m_output` field of the stats writer was not initialized, but we tried to use it anyway in the collection stats phase

```release-note
NONE
```

### DIFF
--- a/userspace/falco/stats_writer.cpp
+++ b/userspace/falco/stats_writer.cpp
@@ -91,6 +91,11 @@ stats_writer::stats_writer(
 	m_config = config;
 	if (config->m_metrics_enabled)
 	{
+		/* m_outputs should always be initialized because we use it
+		 * to extract output-queue stats in both cases: rule output and file output.
+		 */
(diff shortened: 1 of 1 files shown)


FedeDP (approved):
/approve

Makes sense, great catch!

incertum (approved):
Thanks!

/approve

PR #2800, opened 2023-09-12
fix(build): set the right bucket and version for driver legacy

LucaGuerra:
**What type of PR is this?**

/kind bug
/area CI

Download the appropriate version of Falco when building the legacy driver loader image.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/reusable_build_docker.yaml
+++ b/.github/workflows/reusable_build_docker.yaml
@@ -80,6 +80,8 @@ jobs:
         run: |
           cd ${{ github.workspace }}/docker/driver-loader/
           docker build -t docker.io/falcosecurity/falco-driver-loader-legacy:${{ inputs.arch }}-${{ inputs.tag }} \
+            --build-arg VERSION_BUCKET=deb${{ inputs.bucket_suffix }} \
+            --build-arg FALCO_VERSION=${{ inputs.version }} \
             --build-arg TARGETARCH=${TARGETARCH} \
             .
             docker save docker.io/falcosecurity/falco-driver-loader-legacy:${{ inputs.arch }}-${{ inputs.tag }} --output /tmp/falco-driver-loader-legacy-${{ inputs.arch }}.tar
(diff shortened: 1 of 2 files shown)


leogr (approved):
:+1:

docker/driver-loader-legacy/docker-entrypoint.sh
```diff
@@ -16,18 +16,13 @@
 # limitations under the License.
 #
 
-# Set the SKIP_DRIVER_LOADER variable to skip loading the driver
 
-if [[ -z "${SKIP_DRIVER_LOADER}" ]]; then
```
FedeDP, line 21: Why this change?

.github/workflows/reusable_build_docker.yaml
```diff
@@ -80,6 +80,8 @@ jobs:
         run: |
           cd ${{ github.workspace }}/docker/driver-loader/
           docker build -t docker.io/falcosecurity/falco-driver-loader-legacy:${{ inputs.arch }}-${{ inputs.tag }} \
+            --build-arg VERSION_BUCKET=deb${{ inputs.bucket_suffix }} \
```
FedeDP, line 83: Why this was not required before #2718 ?

LucaGuerra: /hold I found another mistake :facepalm:

LucaGuerra: These are optional arguments, by default it was somehow reusing the old `0.35.1` image instead of building the legacy one from master.

docker/driver-loader-legacy/docker-entrypoint.sh
LucaGuerra: the `driver-loader` image is different from `falco`. In fact, it has a different entrypoint! This makes sure that the change is reflected, so that the `driver-loader-legacy` image actually behaves like a `driver-loader` (as in, it loads the driver) and not like a `falco` image (which loads the driver AND starts Falco)

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #2812, opened 2023-09-15
new(ci): autogenerate release body.

FedeDP:
**What type of PR is this?**

/kind feature

/area CI

This PR autogenerates release body with:
* LIBS and DRIVER versions badges
* Tables with packages and docker images
* release notes using `rn2md` tool (https://github.com/leodido/rn2md)

Moreover, the RELEASE.md file is updated.

Fixes #

~~Keeping this as `wip` because we need https://github.com/leodido/rn2md/pull/11 first (the composite action that the release workflow is now using).~~

```release-note
NONE
```


FedeDP:
This is not needed for Falco 0.36.0, even if it would be a nice to have.
Let's see if we can put this in.

.github/workflows/release.yaml
```diff
@@ -122,3 +122,39 @@ jobs:
       is_latest: ${{ needs.release-settings.outputs.is_latest == 'true' }}
       tag: ${{ github.event.release.tag_name }}
       sign: true
+      
+  release-body:
+    needs: [release-settings, publish-docker]
```
FedeDP, line 127: Only at the end of the publish!

submodules/falcosecurity-rules
FedeDP, line 1: THIS ALWAYS SLIPS IN...

FedeDP:
This is now ready. See the composite action `leodido/rn2md` in action: https://github.com/FedeDP/rn2md/actions/runs/6245532470/job/16954354425

I think we might give this a shot for 0.36.0 to replace some manual tasks for the release, but given the time constraint we might skip this. (note however that even if this fails, nothing wrong happens because the new job only pushes the release body; worst case scenario is that we need to fill in the body manually, just like we used to do; best case, it will be automatically done by this. I see no risk involved with merging this.
Also, as i shared above, i tested the composite action just fine; everything else is the part of the code we use in the libs automation to upload a release body.)

/cc @leogr @LucaGuerra

Andreagit97 (dismissed):
/approve

.github/workflows/release.yaml
```diff
@@ -122,3 +122,38 @@ jobs:
...
+          output: ./notes.md
+        
+      - name: Merge release notes to pre existent body
+        run: cat notes.md >> release-body.md
+        
+      - name: Release
+        uses: softprops/action-gh-release@v1
+        with:
+          body_path: ./release-body.md
+          tag_name: ${{ github.event.release.tag_name }}
```
LucaGuerra, line 159: Remember that this entire workflow reacts to the _creation_ of a new release. Wouldn't this create another one?

FedeDP: I think it will just update the existing one; perhaps i need to add `name:` input though! Thanks for the suggestion btw

FedeDP: But yep i totally forgot about prereleases and draft; this should only run on final releases.
I will update the workflow.

.github/workflows/release.yaml
```diff
@@ -122,3 +122,40 @@ jobs:
       is_latest: ${{ needs.release-settings.outputs.is_latest == 'true' }}
       tag: ${{ github.event.release.tag_name }}
       sign: true
+      
+  release-body:
+    needs: [release-settings, publish-docker]
+    if: ${{ needs.release-settings.outputs.is_latest == 'true' }} # only for latest releases
```
FedeDP, line 128: @LucaGuerra this should only run for latest releases.

.github/workflows/release.yaml
FedeDP: Ok i should've fixed it. PTAL :)

LucaGuerra: My question is more like: is https://github.com/softprops/action-gh-release able to modify an existing release? So, have you tried creating an empty repository with a pipeline which reacts to this

```
on:
  release:
    types: [published]
```

and check if it is able to update the release body that you added? It is possible that you did or that the manual states it somewhere but from a quick look I didn't find it

FedeDP: So, this is the same action used by libs; on libs, you create a release (not a tag) BUT the workflow triggers on tag. Anyway, the action updates the existing github release.
Also, the action is able to append body to an existing release (this is what we do in libs actually):  

> append_body | Boolean | Append to existing body instead of overwriting it

FedeDP: So, yes the action CAN update an existing github release.

FedeDP: > 💡 When the release info keys (such as name, body, draft, prerelease, etc.) are not explicitly set and there is already an existing release for the tag, the release will retain its original info.

FedeDP: (added correct permissions to the job btw)

leogr (commented):
Will it run for RC, too? :thinking:

FedeDP:
Only for releases `set as latest` , so no RCs.

FedeDP:
Also, thanks to @leodido who was super fast at reviewing and merging my PRs against [rn2md](https://github.com/leodido/rn2md)!

leogr (approved):

jasondellaluce (approved):
/approve

FedeDP:
This does not need to be backported on the release branch!

PR #2856, opened 2023-10-05
fix(gha): update rpmsign

LucaGuerra:
**What type of PR is this?**

/kind bug

/area build

Seems like rpmsign does not work in the same way for a new distribution. This is a simpler way of running it (and uses v4 signatures) . Not sure if this would break anything but it seems to validate the signature.

Fixes #

```release-note
NONE
```


FedeDP (approved):
Thanks!
/approve

jasondellaluce (approved):
/approve

PR #2881, opened 2023-10-19
cleanup: replace banned.h with semgrep

LucaGuerra:
**What type of PR is this?**

/kind cleanup

/area engine

/area CI

We used to have a `banned.h` header file that prevents you from using specific functions. The problem with it is that we have to manually add it to any `.cpp` file that we ever write. This does not happen in practice unless there is a specific header that always needs to be included (`falco_common.h` seems to be it but really is not). In libs we have decided to use semgrep in CI instead.

The advantage of using Semgrep is that we don't have to modify the code or remember to include any extra header. The disadvantage is that you won't get compile time errors but only CI time errors. Let's do the same in Falco and in libs.

In addition, a file was using `strncat`. Replace it with `strlcat`.

Fixes #

```release-note
NONE
```


FedeDP (approved):
/approve

Issue #2672, opened 2023-07-06
Enable build CI for Windows/MacOS to improve multi-platform compatibility


LucaGuerra:
In https://github.com/falcosecurity/libs/ we are able to perform some basic builds on platforms that are not Linux, namely Windows and MacOS. While we don't provide specific artifacts for those platforms there's no reason why the main code shouldn't be multiplatform, especially since we can now build cross platform plugins. Of course kmod or ebpf are strictly Linux-only but the generic code should work. Given the fact that we're also working on WASM this effort might be much less than expected.

therealbobo:
I think we can close this. cc @LucaGuerra

PR #2889, opened 2023-10-25
build: enable compilation on win32 and macOS

therealbobo:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind release

/area build

> /area engine

> /area tests

> /area proposals

/area CI

This enables falco to be build on Windows and MacOS.

https://github.com/falcosecurity/falco/issues/2672

Fixes #

```release-note
new(build): enable compilation on win32 and macOS.
```


userspace/falco/app/actions/print_page_size.cpp
```diff
@@ -24,7 +24,15 @@ falco::app::run_result falco::app::actions::print_page_size(falco::app::state& s
 {
 	if(s.options.print_page_size)
 	{
+#ifndef _WIN32
 		long page_size = getpagesize();
+#else
+		SYSTEM_INFO sysInfo;
+
+		GetSystemInfo(&sysInfo);
```
LucaGuerra, line 32: Looks like there is no header file included for this function :thinking:

userspace/falco/app/actions/print_support.cpp
```diff
@@ -37,23 +39,61 @@ falco::app::run_result falco::app::actions::print_support(falco::app::state& s)
...
 		support["engine_info"] = infos.as_json();
 
+#ifndef _WIN32
 		support["system_info"]["sysname"] = sysinfo.sysname;
 		support["system_info"]["nodename"] = sysinfo.nodename;
 		support["system_info"]["release"] = sysinfo.release;
 		support["system_info"]["version"] = sysinfo.version;
 		support["system_info"]["machine"] = sysinfo.machine;
+#else
+		support["system_info"]["sysname"] = "Windows";
```
LucaGuerra, line 75: Wouldn't it be a little cleaner to create a function that is able to output the `system_info`, have one for linux/mac and one for Windows so we can better separate the two implementations?

userspace/falco/app/actions/create_requested_paths.cpp
```diff
@@ -93,23 +94,15 @@ falco::app::run_result falco::app::actions::create_requested_paths(falco::app::s
...
+
+    std::filesystem::path dirPath(path);
+
+    try {
+        std::filesystem::create_directories(dirPath);
+    } catch (const std::exception& ex) {
+		return -1;
+    }
+
+    return 0;
```
LucaGuerra, line 106: <3

userspace/falco/stats_writer.cpp
```diff
@@ -48,7 +54,10 @@ static void timer_handler(int signum)
 
 bool stats_writer::init_ticker(uint32_t interval_msec, std::string &err)
 {
+#if !defined(_WIN32)
```
LucaGuerra, line 57: This is getting a bit hard to follow

userspace/falco/logger.h
```diff
@@ -18,7 +18,18 @@ limitations under the License.
...
+#ifdef _WIN32
+#define	LOG_EMERG	0
+#define	LOG_ALERT	1
+#define	LOG_CRIT	2
+#define	LOG_ERR		3
+#define	LOG_WARNING	4
+#define	LOG_NOTICE	5
+#define	LOG_INFO	6
+#define	LOG_DEBUG	7
+#else
```
LucaGuerra, line 30: can we make an enum for `falco_logger::level` instead?

userspace/falco/app/actions/print_page_size.cpp
LucaGuerra: I have seen this in other files as well, a bit confused as to where this is imported from

LucaGuerra:
Thanks a lot @therealbobo !! This is very useful to make Falco more cross-platform compatible <3 

My main issues here is mostly about readability because sometimes it's easy to get lost in a bunch of `ifdef`s. Would also love to hear @FedeDP 's opinion about this as he usually has good ideas :)

userspace/falco/event_drops.h
```diff
@@ -29,7 +29,7 @@ limitations under the License.
 // detecting a syscall event drop.
 enum class syscall_evt_drop_action : uint8_t
 {
-	IGNORE = 0,
+	DISREGARD = 0,
```
LucaGuerra, line 32: It's not really clear to me why we need this. Is it because just having anything that is called `IGNORE` (even if it's a scoped enum) ends up being a compilation error? Seems a little weird but not impossible!

userspace/falco/app/actions/print_support.cpp
therealbobo: That makes sense! 👍

userspace/falco/event_drops.h
therealbobo: The problem is this define: `winbase.h:479:#define IGNORE 0` :/

userspace/falco/app/actions/print_support.cpp
```diff
@@ -32,28 +38,84 @@ static std::string read_file(const std::string &filename)
...
+	support["system_info"]["nodename"] = sysinfo.nodename;
+	support["system_info"]["release"] = sysinfo.release;
+	support["system_info"]["version"] = sysinfo.version;
+	support["system_info"]["machine"] = sysinfo.machine;
+	return 0;
+}
+#endif
+
+#ifdef _WIN32
+static int get_win32_sysinfo(nlohmann::json &support)
```
jasondellaluce, line 60: Why not call both of these with the same name? That way we don't need any ifdef when invoking them.

.github/workflows/reusable_build_packages.yaml
```diff
@@ -216,3 +216,63 @@ jobs:
...
+
+      - name: Run unit Tests
+        run: |
+          build/unit_tests/Release/falco_unit_tests.exe
+
+      - name: Upload Falco win32 package
+        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+        with:
+          name: falco-${{ inputs.version }}-win32.exe
+          path: |
```
jasondellaluce, line 247: Question -- can't we update cpack options and just use `make package` and grab the package generated by cmake?

FedeDP: Agree with Jason!

.github/workflows/reusable_build_packages.yaml
```diff
@@ -216,3 +216,69 @@ jobs:
           name: falco-${{ inputs.version }}-wasm.tar.gz
           path: |
             ${{ github.workspace }}/build/falco-${{ inputs.version }}-wasm.tar.gz
+
+  build-win32-package:
```
FedeDP, line 220: Can't we use a matrix `build-others-packages` like we do eg: on libs?

CMakeLists.txt
```diff
@@ -109,30 +113,55 @@ endif()
...
-if(NOT EMSCRIPTEN)
+if(LINUX)
   set(FALCO_SECURITY_FLAGS "${FALCO_SECURITY_FLAGS} -Wl,-z,relro,-z,now -fstack-protector-strong")
 endif()
-if(CMAKE_BUILD_TYPE STREQUAL "release")
-  set(FALCO_SECURITY_FLAGS "${FALCO_SECURITY_FLAGS} -D_FORTIFY_SOURCE=2")
-endif()
 
-set(CMAKE_COMMON_FLAGS "${FALCO_SECURITY_FLAGS} -Wall -ggdb ${FALCO_EXTRA_FEATURE_FLAGS} ${MINIMAL_BUILD_FLAGS} ${MUSL_FLAGS}")
+if(NOT WIN32)
```
FedeDP, line 120: Can we move all of this in a `CompilerFlags.cmake` like we do for example in libs?

cmake/modules/falcosecurity-libs.cmake
```diff
@@ -28,15 +28,15 @@ else()
   # In case you want to test against a fork of falcosecurity/libs just pass the variable -
   # ie., `cmake -DFALCOSECURITY_LIBS_REPO=<your-gh-handle>/libs ..` 
   if (NOT FALCOSECURITY_LIBS_REPO)
-    set(FALCOSECURITY_LIBS_REPO "falcosecurity/libs")
+    set(FALCOSECURITY_LIBS_REPO "therealbobo/libs")
```
FedeDP, line 31: I assume you are just testing things; in case please add back the `wip` status :)

unit_tests/falco/test_configuration.cpp
```diff
@@ -109,7 +109,11 @@ TEST(Configuration, configuration_environment_variables)
     std::string env_var_value = "envVarValue";
     std::string env_var_name = "ENV_VAR";
     std::string default_value = "default";
+#ifndef _WIN32
     setenv(env_var_name.c_str(), env_var_value.c_str(), 1);
+#else
+    _putenv_s(env_var_name.c_str(), env_var_value.c_str());
```
FedeDP, line 115: We could use a macro on top of this file perhaps?

userspace/falco/outputs_program.cpp
FedeDP, line 1: Since we now won't support all outputs on all OSes, we could build just the list of supported outputs on each OS from CMakeLists. Would this be feasible?
Then we only need a change on falco_outputs.cpp to only include the supported outputs.

userspace/falco/stats_writer.cpp
FedeDP: I would just have different versions of init_ticker for each OS.

cmake/modules/falcosecurity-libs.cmake
therealbobo: Yeap! Exactly! Just for testing purposes :)

userspace/falco/app/actions/print_page_size.cpp
FedeDP: Cc @therealbobo

jasondellaluce (approved):
/approve

FedeDP:
Updated the release note label to allow the PR to get merged.

PR #2890, opened 2023-10-26
Implement configuration of webserver listening address

### LINKED ISSUE #2774: Adding the opportunity to select Falco webserver port in configuration file
**Motivation**

When I deploy Falco on our Linux infrastructure, I see that it starts listening on port: 8765/TCP on all host IPs (it also includes the external IP if there is one).
And we can get the Falco web server port available on the entire Internet on our servers if we do not prepare firewall rules. Perhaps now there are only two available endpoints, access to which will not give a potential attacker the opportunity to hack us, but we at least reveal to the world security tools we use, which is not very good.
I think I found responsible [line](https://github.com/falcosecurity/falco/blob/f7c628f6237d5a6f7a33e2ed053ced70386193e3/userspace/falco/webserver.cpp#L83) of code.

**Feature**

To add an option in **falco.yaml** configuration file to select IP address for Falco web server between 0.0.0.0 and 127.0.0.1 (with default 0.0.0.0)

**Alternative**

Give a user an opportunity to select an arbitrary IP address for the Falco web server and validate it through regular expression or compare with available network interfaces on host.

sgaist:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind release

> /area build

/area engine

> /area tests

> /area proposals

> /area CI

Currently the webserver is listening on the hard coded 0.0.0.0 address.
This patch keeps this default but allows the administrator to change it through the `falco.yaml` file.

Fixes #2774

```release-note
feat(userspace/falco): falco administrators can now configure the address on which the webserver listen using the new listen_address field in the webserver block of the `falco.yaml` file.
```


falco.yaml
```diff
@@ -492,6 +492,7 @@ webserver:
   # the appropriate number of threads based on the number of online cores in the system.
   threadiness: 0
   listen_port: 8765
+  listen_address: 0.0.0.0
```
incertum, line 495: Should we explicitly mention that only TCP/IPv4 is supported? Compare to grpc bind_address implementation (which btw perhaps could be cleaned up as well by someone in the future).

incertum (commented):
Thank you @sgaist!

Totally makes sense to me and implementation LGTM.
Since I am not using this functionality myself, let's have another maintainer take a closer look.

CC @LucaGuerra @FedeDP thanks!

userspace/falco/configuration.cpp
```diff
@@ -31,6 +31,11 @@ limitations under the License.
 #include "configuration.h"
 #include "logger.h"
 
+#include <re2/re2.h>
+
+static re2::RE2 ipv4_address_re("^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$");
```
FedeDP, line 36: Can we add some tests around this? Eg here: https://github.com/falcosecurity/falco/tree/master/unit_tests/falco/app/actions

FedeDP (commented):
LGTM, i  really like it, thank you!
Left only a comment!

userspace/falco/configuration.cpp
incertum: +1 on some unit tests.

We more recently refactored a few things to make unit testing easier, however you will notice we still have a lot of work ahead, especially in terms of coverage.

Therefore, as minimum test requirement testing the regex only seems fair, for example checkout the prometheus interval test https://github.com/falcosecurity/falco/blob/1609ee85596cd4e57c15423e17a54fe1eb329092/unit_tests/engine/test_falco_utils.cpp#L41

We also have a newer repo here https://github.com/falcosecurity/testing.

As said additional tests and more future contribution in this regard would be appreciated, but only a regex test should be the requirement to get this PR merged.

Thanks in advance!

falco.yaml
sgaist: Do you mean TCP/IP VS unix socket ?

userspace/falco/configuration.cpp
incertum: Feel free to implement the best test. Happy to provide more feedback once it is up. Just wanted to communicate the absolute minimum test requirement beforehand as you will see that currently not all configurations are tested in this repo's unit test suite :)

incertum: EDIT: I see it's up, one sec!

falco.yaml
incertum: And Ipv4 only.

unit_tests/falco/test_configuration.cpp
```diff
@@ -174,3 +174,34 @@ TEST(Configuration, configuration_environment_variables)
     /* Clear the set environment variable after testing */
     unsetenv(env_var_name.c_str());
 }
+
+TEST(Configuration, configuration_webserver_ip)
```
incertum, line 178: Since we are now accepting user input, lets also check for some typos such as accidental white spaces and such for completeness? WDYT?

falco.yaml
sgaist: On the TCP side, shouldn't IPV6 also work with httplib ?

unit_tests/falco/test_configuration.cpp
sgaist: Sure thing

unit_tests/falco/test_configuration.cpp
```diff
@@ -174,3 +174,45 @@ TEST(Configuration, configuration_environment_variables)
...
+
+    std::vector<std::string> invalid_addresses = {"327.0.0.1",
+                                                  "1.327.0.1",
+                                                  "1.1.327.1",
+                                                  "1.1.1.327",
+                                                  "12 7.0.0.1",
+                                                  "127. 0.0.1",
+                                                  "127.0. 0.1",
+                                                  "127.0.0. 1",
+                                                  "!27.0.0.1"};
```
incertum, line 207: :smile:

falco.yaml
incertum: You wanna try? If we easily can, we should. Perhaps we just need the ipv6 regex parser then?

FedeDP: This LGTM; re ipv6, i think it is just a mattter of updating the regex (and tests) as Melissa said.
We can do that in a follow up PR anyway!

sgaist: Sorry for the late reply, I did a small test to see if the webserver starts when listening to an IPV6 address other than `::` (which is working and thus already a good sign) and it does for pure IPV6 addresses. From the looks of it, literal IPV6 addresses are not supported. I currently don't know whether IPV4 mapped address work either.

Related to that topic, we might want to discuss things a bit further with regard to the IPV6 support design. Do we just reuse the same setting but with a regex that can validate both IPV4 and IPV6 values ? Or should there be a separation between both through the use of `listen_address` and `listen_ip6_address` ?

Andreagit97: > Do we just reuse the same setting but with a regex that can validate both IPV4 and IPV6 values ?

I would say yes. I would reuse the same config allowing users to specify either an ipv4 or an ipv6

userspace/falco/configuration.cpp
```diff
@@ -33,9 +33,8 @@ limitations under the License.
...
...|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]))\\s*$)|(^\\s*((([0-9A-Fa-f]{1,4}:){7}([0-9A-Fa-f]{1,4}|:))|(([0-9A-Fa-f]{1,4}:){6}(:[0-9A-Fa-f]{1,4}|((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(\\.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3})|:))|(([0-9A-Fa-f]{1,4}:){5}(((:[0-9A-Fa-f]{1,4}){1,2})|:((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(\\.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3})|:))|(([0-9A-Fa-f]{1,4}:){4}(((:[0-9A-Fa-f]{1,4}){1,3})|((:[0-9A-Fa-f]{1,4})?:((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(\\.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){3}(((:[0-9A-Fa-f]{1,4}){1,4})|((:[0-9A-Fa-f]{1,4}){0,2}:((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(\\.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){2}(((:[0-9A-Fa-f]{1,4}){1,5})|((:[0-9A-Fa-f]{1,4}){0,3}:((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(\\.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){1}(((:[0-9A-Fa-f]{1,4}){1,6})|((:[0-9A-Fa-f]{1,4}){0,4}:((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(\\.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3}))|:))|(:(((:[0-9A-Fa-f]{1,4}){1,7})|((:[0-9A-Fa-f]{1,4}){0,5}:((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(\\.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3}))|:)))(%.+)?\\s*$))");
```
incertum, line 37: 😅

incertum (commented):
LGTM, will approve after you re-push the last commit with signed off, ty!

userspace/falco/configuration.cpp
sgaist: "savage" might be a good description 😁

sgaist:
> LGTM, will approve after you re-push the last commit with signed off, ty!

sign-off added and re-based as well :-)

incertum (approved):
/approve

@sgaist 🎉  thanks so much for this contributions, hopefully we see more in the future 🙃 !

userspace/falco/configuration.cpp
FedeDP: Why did you share my password manager passphrase here?? :rofl:

FedeDP (approved):
/approve

Great job, thank you very much!

PR #2899, opened 2023-11-02
update(userspace/falco): introduce new engine_version_semver key in v…

loresuso:
…ersions endpoint

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind release

> /area build

/area engine

> /area tests

> /area proposals

> /area CI

With this PR we introduce a new key in the Falco `/versions` endpoint. This key will be used by `falcoctl` to match new artifact configs containing a full semver string for representing the required engine version.

The rationale for this change is the following:
- Falco is already able to read `engine_version` both as numeric valuer or semver string
- Rules repositories must be updated to use a full semver string for `required_engine_version` (already did it for Falco rules, will need this also for plugin rules). This will require a ***major*** bump for each artifact
- New artifact config layers will contains the new `engine_version_semver` key
- Let's also point out that `falcoctl` is already "keys agnostic". It just starts processing and matching keys starting from the config layer, and as long as it finds the same key in the `/versions` endpoint, we are fine.

With this being said, we end up in this situation (example for falco rules, will be the same for plugin rules):

| Falco version      | Rules and CI versions | Outcome |
| ----------- | ----------- | --------------|
| < 0.37| <= 2.0.0|  Nothing changed, will work as before |
| < 0.37| > 2.0.0|  If you upgrade only the rules, falcoctl will find two different keys and will not be able to match them, causing an error in falcoctl, not upgrading the rules. Also, if you manually start Falco with the new rules (on host, maybe) Falco will not start. I think this is perfectly fine, cause the rule will have a new major, so you must update Falco as well (would be the same if the engine version was only bumped, even if it was numeric)|
| >= 0.37| <= 2.0.0| Falco is updated. You can continue to follow 2.x.y rules, because it is able to read both a numeric engine version and a semver one. If you want, you can also follow the new rules (3.0.0)(see below) |
| >= 0.37| >2.0.0| Everything is new, so it will work as expected by matching the new key|

PRs for updating the config layers are coming soon!

Fixes #

```release-note
update add `engine_version_semver` key in `/versions` endpoint
```

### DIFF
--- a/userspace/falco/versions_info.cpp
+++ b/userspace/falco/versions_info.cpp
@@ -72,7 +72,12 @@ nlohmann::json falco::versions_info::as_json() const
     version_info["driver_api_version"] = driver_api_version;
     version_info["driver_schema_version"] = driver_schema_version;
     version_info["default_driver_version"] = default_driver_version;
-    version_info["engine_version"] = engine_version;
+    // note: the 'engine_version' key below must be removed in the next major bump (0.x.y -> 1.0.0)
+    // the two keys are kept for existing tooling that relies on the old key
+    // (falcoctl will match old rules artifacts configs by using this key, and the new ones using 
+    // the engine_version_semver key)
+    version_info["engine_version"] = std::to_string(FALCO_ENGINE_VERSION_MINOR);
+    version_info["engine_version_semver"] = engine_version;
     for (const auto& pv : plugin_versions)
     {
         version_info["plugin_versions"][pv.first] = pv.second;


loresuso:
cc @alacuku, @LucaGuerra for visibility

alacuku:
Hey @loresuso, thanks for the PR!

I think we should write some integrations tests in falcoctl to cover all the cases described above. To make sure that we are not going to break running instances of falco when upgrading them. I wuold be happy to help writing those tests.

loresuso:
@alacuku Agreed, we should somehow mock the version endpoint for the falcoctl test and try the out the function that checks the requirments. Happy to review if you are up for doing it!

FedeDP (approved):
/approve

leogr (approved):

PR #2900, opened 2023-11-02
fix(userspace/engine): cache latest rules compilation output

jasondellaluce:
**What type of PR is this?**

/kind bug

/area build

This fixes a performance regression introduced in `falco -L` since the latest release, caused by the fact that we re-compile the whole set of loaded rules twice. 

```release-note
fix(userspace/engine): cache latest rules compilation output
```


FedeDP (approved):
/approve
Great catch :tm: !

incertum (approved):
/approve

PR #2917, opened 2023-11-20
cleanup(falco): remove `init` in the configuration constructor

Andreagit97:
**What type of PR is this?**

/kind cleanup

/area engine

This `init({});` in the falco_configuration constructor is unnecessary since when we call the action `load_config`, if we don't have a config file, we will call the same `init` we have just removed. This cleanup avoids calling `falco_configuration::init` 2 times

```release-note
NONE
```


FedeDP:
LGTM btw but i want to hear the expert's voice eheh

FedeDP (approved):
/approve

PR #2918, opened 2023-11-20
chore(unit_test,userspace): allow env var to get expanded in yaml even when part of a string

FedeDP:
**What type of PR is this?**

/kind feature

/area engine

Moreover, support env variable embedding another env variable.

Basically, this PR would allow us to use eg:
```
engine:
  ebpf:
    probe: /${HOME}/.falco/falco-bpf.o
```
in #2413.

Fixes #

```release-note
new(userspace): support env variable expansion in all yaml, even inside strings.
```

### DIFF
--- a/unit_tests/falco/test_configuration.cpp
+++ b/unit_tests/falco/test_configuration.cpp
@@ -128,6 +128,10 @@ TEST(Configuration, configuration_environment_variables)
     std::string int_env_var_name = "ENV_VAR_INT";
     SET_ENV_VAR(int_env_var_name.c_str(), int_env_var_value.c_str());
 
+    std::string empty_env_var_value = "";
+    std::string empty_env_var_name = "ENV_VAR_EMPTY";
+    SET_ENV_VAR(empty_env_var_name.c_str(), empty_env_var_value.c_str());
+
     std::string default_value = "default";
     std::string env_var_sample_yaml =
         "base_value:\n"
@@ -153,7 +157,8 @@ TEST(Configuration, configuration_environment_variables)
 	"    - /${ENV_VAR}/${ENV_VAR}${ENV_VAR}/foo\n"
 	"    - ${ENV_VAR_EMBEDDED}/foo\n"
 	"is_test: ${ENV_VAR_BOOL}\n"
-	"num_test: ${ENV_VAR_INT}\n";
+	"num_test: ${ENV_VAR_INT}\n"
+	"empty_test: ${ENV_VAR_EMPTY}\n";
 
     yaml_helper conf;
     conf.load_from_string(env_var_sample_yaml);
@@ -198,7 +203,7 @@ TEST(Configuration, configuration_environment_variables)
     ASSERT_EQ(base_value_2_list_1, " " + env_var_value); // Environment variable preceded by a space, still extracted env var with leading space
 
     auto base_value_2_list_2 = conf.get_scalar<std::string>("base_value_2.sample_list[2]", default_value);
-    ASSERT_EQ(base_value_2_list_2, env_var_value + " "); // Environment variable proceeded by a space, still extracted env var with trailing space
+    ASSERT_EQ(base_value_2_list_2, env_var_value + " "); // Environment variable followed by a space, still extracted env var with trailing space
 
     auto base_value_2_list_3 = conf.get_scalar<std::string>("base_value_2.sample_list[3]", default_value);
     ASSERT_EQ(base_value_2_list_3, "$UNSED_XX_X_X_VAR"); // Does not follow the `${VAR}` format, so should be treated as a regular string
@@ -232,11 +237,16 @@ TEST(Configuration, configuration_environment_variables)
     auto integer = conf.get_scalar<int32_t>("num_test", -1);
     ASSERT_EQ(integer, 12);
 
+    // An env var that resolves to an empty string returns default value
+    auto empty_default_str = conf.get_scalar<std::string>("empty_test", "test");
+    ASSERT_EQ(empty_default_str, "test");
+
     /* Clear the set environment variables after testing */
     SET_ENV_VAR(env_var_name.c_str(), "");
     SET_ENV_VAR(embedded_env_var_name.c_str(), "");
     SET_ENV_VAR(bool_env_var_name.c_str(), "");
     SET_ENV_VAR(int_env_var_name.c_str(), "");
+    SET_ENV_VAR(empty_env_var_name.c_str(), "");
 }
 
 TEST(Configuration, configuration_webserver_ip)
--- a/unit_tests/falco/test_configuration.cpp
+++ b/unit_tests/falco/test_configuration.cpp
@@ -114,11 +114,14 @@ TEST(Configuration, configuration_environment_variables)
     // Set an environment variable for testing purposes
     std::string env_var_value = "envVarValue";
     std::string env_var_name = "ENV_VAR";
-    std::string default_value = "default";
     SET_ENV_VAR(env_var_name.c_str(), env_var_value.c_str());
-    yaml_helper conf;
 
-    std::string sample_yaml =
+    std::string embedded_env_var_value = "${ENV_VAR}";
+    std::string embedded_env_var_name = "ENV_VAR_EMBEDDED";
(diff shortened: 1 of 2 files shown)


FedeDP:
Expands https://github.com/falcosecurity/falco/pull/2562.

unit_tests/falco/test_configuration.cpp
```diff
@@ -133,52 +136,81 @@ TEST(Configuration, configuration_environment_variables)
         "    sample_list:\n"
         "        - ${ENV_VAR}\n"
         "        - ' ${ENV_VAR}'\n"
-        "        - $UNSED_XX_X_X_VAR\n";
-    conf.load_from_string(sample_yaml);
+	"        - '${ENV_VAR} '\n"
```
FedeDP, line 139: Tests were expanded:
* new case with whitespace AFTER the expanded string. Moreover, since we now support env variables embedded inside bigger strings, both ` ${ENV_VAR}` and `${ENV_VAR} ` are now expanded
* we do now tests that env variables found in the middle of strings are properly expanded (`paths[0,1,2,3]` cases)
* `paths[4]` checks an even weirder case, ie: when the env variable gets expanded to another env variable. It must get resolved twice.

userspace/falco/yaml_helper.h
```diff
@@ -77,42 +78,77 @@ class yaml_helper
 		get_node(node, key);
 		if(node.IsDefined())
 		{
-			std::string value = node.as<std::string>();
+			auto value = node.as<std::string>();
 
 			// Helper function to convert string to the desired type T
 			auto convert_str_to_t = [&default_value](const std::string& str) -> T {
+				if (str.empty())
```
FedeDP, line 85: Short-circuite, when empty value is passed, just return default.

userspace/falco/yaml_helper.h
```diff
@@ -77,42 +78,77 @@ class yaml_helper
...
+				{
+					return default_value;
+				}
+
+				if constexpr (std::is_same_v<T, std::string>)
+				{
+					return str;
```
FedeDP, line 92: For strings, just return it.
This workarounds the fact that `stringstream` splits values by whitespaces, and this is something we don't want to be done when converting string->string.

userspace/falco/yaml_helper.h
```diff
@@ -77,42 +78,77 @@ class yaml_helper
...
+				}
 				std::stringstream ss(str);
 				T result;
-				if (ss >> result) return result;
+				if (ss >> std::boolalpha >> result) return result;
```
FedeDP, line 96: Enforce `std::boolalpha` so that `true` and `false` are correctly parsed to booleans.

FedeDP:
/cc @jasondellaluce 

I am not sure whether we want this for 0.37.0, leave it to other maintainers to decide ;) 
Also cc @therealdwright who first opened the original PR!

userspace/falco/yaml_helper.h
incertum: Do we have a test for it? I may have overlooked it.

incertum (approved):
/approve

unit_tests/falco/test_configuration.cpp
```diff
@@ -133,52 +144,99 @@ TEST(Configuration, configuration_environment_variables)
...
+    auto boolean_as_str = conf.get_scalar<std::string>("is_test", "false");
+    ASSERT_EQ(boolean_as_str, "true"); // `true` can be parsed to string.
+
+    auto boolean_as_int = conf.get_scalar<int32_t>("is_test", 0);
+    ASSERT_EQ(boolean_as_int, 0); // `true` cannot be parsed to integer.
+
+    auto integer = conf.get_scalar<int32_t>("num_test", -1);
+    ASSERT_EQ(integer, 12);
+
+    /* Clear the set environment variables after testing */
```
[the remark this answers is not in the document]
FedeDP: On windows we should map `unsetenv` to  `_putenv(string,"")` anyway, so i decided to avoid adding a new macro and just use what i got :)

unit_tests/falco/test_configuration.cpp
```diff
@@ -133,52 +144,99 @@ TEST(Configuration, configuration_environment_variables)
...
 
-    std::string base_value_2_list_1 = conf.get_scalar<std::string>("base_value_2.sample_list[1]", default_value);
-    ASSERT_EQ(base_value_2_list_1, " ${ENV_VAR}"); // Environment variable preceded by a space, hence treated as a regular string
+    auto base_value_2_list_1 = conf.get_scalar<std::string>("base_value_2.sample_list[1]", default_value);
+    ASSERT_EQ(base_value_2_list_1, " " + env_var_value); // Environment variable preceded by a space, still extracted env var with leading space
 
-    std::string base_value_2_list_2 = conf.get_scalar<std::string>("base_value_2.sample_list[2]", default_value);
-    ASSERT_EQ(base_value_2_list_2, "$UNSED_XX_X_X_VAR"); // Does not follow the `${VAR}` format, so should be treated as a regular string
+    auto base_value_2_list_2 = conf.get_scalar<std::string>("base_value_2.sample_list[2]", default_value);
+    ASSERT_EQ(base_value_2_list_2, env_var_value + " "); // Environment variable proceeded by a space, still extracted env var with trailing space
```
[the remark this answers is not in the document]
FedeDP: Thanks :D

userspace/falco/yaml_helper.h
FedeDP: Will double check and eventually add one!

FedeDP:
> Do we have a test for it? I may have overlooked it.


Added in latest commit, thank you!

incertum (approved):
/approve

FedeDP:
Trying to put this in the 0.37.0 milestone; let's see if we can make this happen :D 
/milestone 0.37.0

unit_tests/falco/test_configuration.cpp
FedeDP: We check that `plugins` config (that is getted as a sequence)  is still correctly expanded.

userspace/falco/yaml_helper.h
```diff
@@ -26,17 +26,54 @@ limitations under the License.
...
+#include <regex>
 
 #include "config_falco.h"
 
 #include "event_drops.h"
 #include "falco_outputs.h"
 
+class yaml_helper;
+
+class yaml_visitor {
```
FedeDP, line 45: `yaml_visitor` is a small visitor API for yaml files, to call a callback function on each scalar value.
It is private to `yaml_helper` since it receives non-constant `YAML::Node`s (thus it can modify the parsed yaml structure).

userspace/falco/yaml_helper.h
```diff
@@ -77,44 +116,8 @@ class yaml_helper
...
-
-				const char* env_value = std::getenv(env_var.c_str()); // Get the environment variable value
-				if(env_value) return convert_str_to_t(env_value);
-
-				return default_value;
-			}
-
-			// If it's not an environment variable reference, return the value as is
-			return node.as<T>();
+			return node.as<T>(default_value);
```
FedeDP, line 119: No more custom logic needed.

FedeDP: BREAKING CHANGE: in https://github.com/falcosecurity/falco/pull/2562 env var that resolved to empty string (both non-set env vars and env vars set to "") returned default value.
This was in constrast with everything else in our yaml impl (ie: we only returned default value if node is not defined).

FedeDP: BREAKING CHANGE: `node.as<T>(default_value);` will return `default_value` when it is not able to parse the node to `T`.

FedeDP:
/unhold

Now we support env vars expansion in all getters, since we `pre-process` the yaml root to expand any scalar value as soon as the document is loaded.

leogr:
/assign

I'll take a look as soon as I can

leogr (approved):
As discussed with @FedeDP this PR introduces a minor behavior change that needs to be signaled and documented.


Anyway, the feature is great and the PR LGTM.

@incertum PTAL again

incertum (approved):
/approve

PR #2920, opened 2023-11-22
Allow enabling rules by ruleset id in addition to name

mstemm:
Add alternate enable_* methods that allow enabling rulesets by ruleset id in addition to name. This might be used by some filter_rulesets to enable/disable rules on the fly via the falco engine.

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind release

> /area build

> /area engine

> /area tests

> /area proposals

> /area CI

Allow enabling rules by ruleset id in addition to name. In cases where the ruleset id is already know, this avoids a redundant lookup.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/engine/falco_engine.cpp
+++ b/userspace/engine/falco_engine.cpp
@@ -298,6 +298,12 @@ std::unique_ptr<load_result> falco_engine::load_rules_file(const std::string &ru
 void falco_engine::enable_rule(const std::string &substring, bool enabled, const std::string &ruleset)
 {
 	uint16_t ruleset_id = find_ruleset_id(ruleset);
+
+	enable_rule(substring, enabled, ruleset_id);
+}
+
+void falco_engine::enable_rule(const std::string &substring, bool enabled, const uint16_t ruleset_id)
+{
 	bool match_exact = false;
 
 	for(const auto &it : m_sources)
@@ -316,6 +322,12 @@ void falco_engine::enable_rule(const std::string &substring, bool enabled, const
 void falco_engine::enable_rule_exact(const std::string &rule_name, bool enabled, const std::string &ruleset)
 {
 	uint16_t ruleset_id = find_ruleset_id(ruleset);
+
+	enable_rule_exact(rule_name, enabled, ruleset_id);
+}
+
+void falco_engine::enable_rule_exact(const std::string &rule_name, bool enabled, const uint16_t ruleset_id)
+{
 	bool match_exact = true;
 
 	for(const auto &it : m_sources)
@@ -335,6 +347,11 @@ void falco_engine::enable_rule_by_tag(const std::set<std::string> &tags, bool en
 {
 	uint16_t ruleset_id = find_ruleset_id(ruleset);
 
+	enable_rule_by_tag(tags, enabled, ruleset_id);
+}
+
+void falco_engine::enable_rule_by_tag(const std::set<std::string> &tags, bool enabled, const uint16_t ruleset_id)
+{
 	for(const auto &it : m_sources)
 	{
 		if(enabled)
--- a/unit_tests/engine/test_enable_rule.cpp
+++ b/unit_tests/engine/test_enable_rule.cpp
@@ -1,0 +1,251 @@
+// SPDX-License-Identifier: Apache-2.0
+/*
+Copyright (C) 2023 The Falco Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


userspace/engine/falco_engine.cpp
```diff
@@ -335,6 +347,11 @@ void falco_engine::enable_rule_by_tag(const std::set<std::string> &tags, bool en
 {
 	uint16_t ruleset_id = find_ruleset_id(ruleset);
 
+	enable_rule_by_tag(tags, enabled, ruleset_id);
+}
+
+void falco_engine::enable_rule_by_tag(const std::set<std::string> &tags, bool enabled, const uint16_t ruleset_id)
```
incertum, line 353: @mstemm could we add and/or expand unit tests? We recently have been more rigorous wrt always adding new tests when touching any part of the code base especially in libs.

Touching these parts of the code base reminded me of a closed PR https://github.com/falcosecurity/falco/pull/2671/files (will be reworked later by Jason). 

I wasn't sure then if enabling by tags was really fixed? What if we enable a rule by a tag, but the rule is disabled by `enabled: false` -> what takes precedence in such a case? @jasondellaluce do you remember the Falco issue around issues with enabling by tags? We might as well resolve this here in this PR especially as we expand test cases either here or in the falco testing Go repo.

mstemm: Hrm, there used to be tests related to loading rules with enabled/disabled flags, but they all were removed: https://github.com/falcosecurity/falco/commit/b1ae5d3dd9f0cc19abfc306aa4110b04d614be8e#diff-b0d96f583cb4d5e7f6ef87d9614e3826058db099ea1c6cff8a6af305c609ce4cL605. It doesn't cover this specific case of using tags, but a lot of other combinations were already tested. Did those tests get migrated to google test?

I can add unit tests for loading rules although I don't actually see any tests for that right now below unit_tests. Is there an example I can follow?

mstemm: I see that there's https://github.com/falcosecurity/testing. I guess I should add tests there?

incertum: @mstemm up to you re the How To. If we instead need some traditional unit tests here I would appreciate them especially for the case of enabling a rule by a tag when the rule was explicitly disabled via the `enabled` key. Also aware I am asking for more tests a bit outside of the scope of this PR, hope that is ok?

mstemm: I added unit tests for enabling/disabling rules: unit_tests/engine/test_enable_rule.cpp.

mstemm: Ok added test for tags as well as the specific case of enabling a disabled rule by tag for the default ruleset.

incertum: Thanks @mstemm! It's a start for unit testing. Few more thoughts: 

It would be nice to have more tests in this regard at the app actions level for `falco::app::run_result falco::app::actions::load_rules_files(falco::app::state& s)` for instance with some more realistic rules (that is: a few more rules that combine all the possibilities we can think of in terms of for example has `enabled: false` but also a tag that would enable the rule and also a tag that disables that rule and all sorts of combinations). That way we would clearly test what has precedence, meaning what enables or disables a rule at the end as that's what can really break things. WDYT? We could do that here (I would appreciate it) or would you prefer to defer this (create a ticket so we don't forget about it)? I am not intending to block this PR.

mstemm: I think a lot of those cases are already covered in https://github.com/falcosecurity/testing/blob/main/tests/falco/legacy_test.go, aren't they?

incertum: I see bunch of tests that indeed sound about right, however what are we asserting? Just checking if exit code is 0 and no errors? Not seeing any assertion(s) of whether the rule(s) were enabled or disabled in the end, do you?

By the way the docs probably could be updated as well with more clarity around the use of enabling/disabling via tags vs the enabled key https://falco.org/docs/rules/controlling-rules/#tags

mstemm: Depends on the test, but they do check specific rules. For example https://github.com/falcosecurity/testing/blob/main/tests/falco/legacy_test.go#L385 has a bunch of rules with specific tags, enables some tags, and verifies the rules that should match the sample event.

incertum: ACK, left a comment in the existing open issue https://github.com/falcosecurity/testing/issues/7#issuecomment-1832811140, worth checking in the future.

So we consider this PR done then right? Or anything you wanted to add?

incertum (approved):
/approve

userspace/engine/falco_engine.cpp
mstemm: I didn't have anything else to add.

jasondellaluce (approved):
/approve

PR #2926, opened 2023-11-27
fix(userspace/falco): use resolved plugin name when opening a plugin

FedeDP:
**What type of PR is this?**

/kind bug

/area engine

Previously, to enable a plugin users needed to know the plugin name, ie: the name exposed by the `plugin_get_name` plugin API.
Now, we automatically use the plugin name internally, while externally we satisfy the name given by the user in the `plugins` section.
Basically, user can now write:
```
load_plugins: [dummyFOO]

plugins:
  - name: dummyFOO
    library_path: libdummy.so
    init_config:
    jitter: 10
    open_params: "100"
```

And getting its dummyFOO plugin loaded correctly, while previously we would've failed, unless the `load_plugins` contained the real plugin name (in this case `dummy`).

From a UX PoV, i think the user shall know nothing about real plugin name; she/he should just have the `.so` of the plugin to be loaded and use the desired name in the config (like, in the example above, `dummyFOO`).

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/falco/app/actions/helpers_inspector.cpp
+++ b/userspace/falco/app/actions/helpers_inspector.cpp
@@ -63,7 +63,7 @@ falco::app::run_result falco::app::actions::open_live_inspector(
 				{
 					auto cfg = s.plugin_configs.at(p->name());
 					falco_logger::log(falco_logger::level::INFO, "Opening '" + source + "' source with plugin '" + cfg->m_name + "'");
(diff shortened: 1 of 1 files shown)


leogr:
Waiting for feedback, see https://kubernetes.slack.com/archives/CMWH3EH32/p1701179948861249

FedeDP:
Since there was no agreement among maintainers, this seems overkill and an unneeded behavioral change.
/close

PR #2932, opened 2023-11-29
chore(docker/falco): add back some deps to falco docker image.

FedeDP:
**What type of PR is this?**

/kind bug

/area build

The PR adds back some deps that were removed in https://github.com/falcosecurity/falco/commit/c2b940f8c4e0d23a6acf24394b99c1814cb59e64 from the Falco "fat" image.

Those deps are needed to build drivers on some distros/kernel configs. 
I added back the ones that are provided by the `falco-driver-loader-legacy` image.
See relevant slack thread: https://kubernetes.slack.com/archives/CMWH3EH32/p1701249107807099

Fixes #

```release-note
NONE
```

### DIFF
--- a/docker/falco/Dockerfile
+++ b/docker/falco/Dockerfile
@@ -19,17 +19,26 @@ RUN cp /etc/skel/.bashrc /root && cp /etc/skel/.profile /root
 
 RUN apt-get update \
 	&& apt-get install -y --no-install-recommends \
+	bc \
+	bison \
 	ca-certificates \
 	clang \
 	curl \
 	dkms \
+	dwarves \
+	flex \
 	gcc \
 	gcc-11 \
 	gnupg2 \
 	jq \
-	libelf1 \
+	libc6-dev \
+	libelf-dev \
+	libssl-dev \
 	llvm \
 	make \
+	netcat-openbsd \
+	patchelf \
+	xz-utils \
 	&& rm -rf /var/lib/apt/lists/*
 
 RUN curl -s https://falco.org/repo/falcosecurity-packages.asc | apt-key add - \


docker/falco/Dockerfile
```diff
@@ -19,17 +19,26 @@ RUN cp /etc/skel/.bashrc /root && cp /etc/skel/.profile /root
...
 	dkms \
+	dwarves \
+	flex \
 	gcc \
 	gcc-11 \
 	gnupg2 \
 	jq \
-	libelf1 \
+	libc6-dev \
+	libelf-dev \
```
LucaGuerra, line 35: I know `libelf-dev` is useful for OpenShift compatibility but can we understand why the others? This is because we may be tempted to remove them in the future and it'd be good to have any kind of "trail" we can follow to understand why something was there in the first place.

FedeDP: These were used in the `-legacy` image. 
My opinion is: using the ones used by `-legacy` image we are sure that we are not breaking an existing behavior; perhaps we miss some more, but at least the 2 images are equal now.

LucaGuerra: Let me take a bit of a closer look. I don't think we want a bloated image but at the same time we want to increase compatibility.

LucaGuerra:
Hi! I'm back here. I analyzed two things about what we are attempting to do:
1. The size difference of the two images
2. The potential impact of the vulnerabilities that we may see in scanners due to these added packages (because I had a quick discussion with @leogr in the past about this)

The size of the compressed image is not a lot, about 23 MBs, I think we can live with that.
Regarding the vulnerabilities, I won't bore you with how I conducted the analysis but I noticed which packages, dependencies and "source packages" (a dpkg concept) are introduced with this change. Out of those, how often vulnerabilities are detected and if it's likely that they would very negatively impact our scanning results.

The result is that they won't. libbpf had two vulnerabilities per year recently and bison had a couple in 2020, the rest are in the security DBs even less frequently. They are massively outweighted by the vulnerabilities (technically false positives but I digress) that are detected due to the dependencies of LLVM and the base image. Conclusion: I think we can add all of those! 

Thank you Fede.

LucaGuerra (approved):

FedeDP:
Thanks to you Luca for your great analysis!

leogr (approved):

PR #2948, opened 2023-12-05
cleanup: fix several warnings from a Clang build

federico-sysdig:
**What type of PR is this?**
/kind cleanup

A Clang build revealed a number of warnings whose fix should increase the quality of the code.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/engine/rule_loader.cpp
+++ b/userspace/engine/rule_loader.cpp
@@ -562,10 +562,8 @@ rule_loader::rule_load_exception::~rule_load_exception()
 {
 }
 
-const char* rule_loader::rule_load_exception::what()
+const char* rule_loader::rule_load_exception::what() const noexcept
 {
-	errstr = falco::load_result::error_code_str(ec) + ": "
-		+ msg.c_str();
-
-	return errstr.c_str();
+	// const + noexcept: can't use functions that change the object or throw
+	return msg.c_str();
 }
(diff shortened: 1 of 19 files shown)


federico-sysdig:
The fix is only a cleanup of warnings that introduces no interface of functionality change.

/unhold

LucaGuerra (approved):
Thanks! 🤩

userspace/engine/rule_loader.cpp
```diff
@@ -562,10 +562,8 @@ rule_loader::rule_load_exception::~rule_load_exception()
...
 
-const char* rule_loader::rule_load_exception::what()
+const char* rule_loader::rule_load_exception::what() const noexcept
 {
-	errstr = falco::load_result::error_code_str(ec) + ": "
-		+ msg.c_str();
-
-	return errstr.c_str();
+	// const + noexcept: can't use functions that change the object or throw
+	return msg.c_str();
```
FedeDP, line 568: This changes rule load exception errors, right?
cc @jasondellaluce i am not sure we want it.

federico-sysdig: I knew it was a deeper change and I was reluctant to do it, but in the end I decided to go ahead because
1. the function has the same name of a virtual function and should have the same signature (i.e. with const + noexcept), this is why a warning was raised by the compiler
2. being noexcept forbids invoking functions that might throw, such as string concatenation
3. I could not find any place in the code where the function is invoked.
If the new implementation is problematic as it changes the original message, let's talk about it.

PR #2971, opened 2023-12-13
chore(falco.yaml): use HOME env var for ebpf probe path.

FedeDP:
**What type of PR is this?**

/kind cleanup

Since #2918 is now merged, we can freely use env var.
This is more correct since eg: `falcoctl` will copy built drivers there: https://github.com/falcosecurity/falcoctl/blob/main/pkg/driver/type/bpf.go#L55.
Most of the time, both falcoctl and Falco will run as root, thus the env var will resolve to the current `/root/.falco/falco-bpf.o`.
But if user decides to eg: have a Falco user with reduced privileges, she/he will still be able to let falcoctl install its ebpf probe and then Falco would start without issues.

Example output without and with `sudo`:
```
Wed Dec 13 17:10:40 2023: Opening 'syscall' source with BPF probe. BPF probe path: /home/federico/.falco/falco-bpf.o
Wed Dec 13 17:10:52 2023: Opening 'syscall' source with BPF probe. BPF probe path: /root/.falco/falco-bpf.o
```

Also, this was already the default behavior when `FALCO_BPF_PROBE=""` env var was passed: https://github.com/falcosecurity/falco/blob/cbbcb611532e0439851616a19b1844628afd5a1d/userspace/falco/app/actions/helpers_inspector.cpp#L114

Fixes #

```release-note
NONE
```

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -318,7 +318,7 @@ engine:
     drop_failed_exit: false
   ebpf:
     # path to the elf file to load.
-    probe: /root/.falco/falco-bpf.o
+    probe: ${HOME}/.falco/falco-bpf.o
     buf_size_preset: 4
     drop_failed_exit: false
   modern_ebpf:


incertum (approved):
/approve

FedeDP:
Wow Melissa, quicker than me :laughing:

Andreagit97 (approved):
/approve

PR #2974, opened 2023-12-16
update(config): Improve http output, add compression and keep_alive

### LINKED ISSUE #2955: new http_output options (compression, batching, keep_alive)
http_output must have compression, batching and keep_alive options.
This is relevant in HighLoad installations and without a falcosidekick in each k8s cluster.

sgaist:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind release

> /area build

/area engine

> /area tests

> /area proposals

> /area CI

This PR adds two new configuration options for the http output:
- [X] Post compression
- [X] Connection keep alive
- [ ] Batching

Fixes #2955

```release-note
feat(userspace/falco): falco administrators can now configure the http output to compress the data sent as well as enable keep alive for the connection. Two new fields (compress_uploads and keep_alive) in the http_output block of the `falco.yaml` file can be used for that purpose. Both are disabled by default.
```

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -566,6 +566,7 @@ http_output:
   client_key: "/etc/ssl/certs/client.key"
   # Whether to echo server answers to stdout
   echo: false
+  compress_uploads: false
 
 # [Stable] `program_output`
 #
--- a/falco.yaml
+++ b/falco.yaml
@@ -567,6 +567,7 @@ http_output:
   # Whether to echo server answers to stdout
   echo: false
   compress_uploads: false
+  keep_alive: false
 
 # [Stable] `program_output`
 #
--- a/userspace/falco/configuration.cpp
+++ b/userspace/falco/configuration.cpp
@@ -309,6 +309,10 @@ void falco_configuration::load_yaml(const std::string& config_name, const yaml_h
 		client_key = config.get_scalar<std::string>("http_output.client_key", "/etc/ssl/certs/client.key");
 		http_output.options["client_key"] = client_key;
 
+		bool compress_uploads;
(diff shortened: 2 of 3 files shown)


incertum (commented):
Thanks a bunch @sgaist 🚀 !

Recommend not marking this as feat and instead as new feature that doesn't break the old default behavior. More thoughts?

CC @leogr @Issif 

/milestone 0.37.0

FedeDP:
> Recommend not marking this as feat and instead as new feature that doesn't break the old default behavior. More thoughts?

What do you mean?
A feat does not necessarily impose a breaking change. In this case, no breaking change in the default behavior is to be expected, right? 
I think we are already ok with this one.

sgaist:
@incertum I am currently waiting on an answer with regard to the batching part which will likely require a bit more work and design discussions.

What could be done is to split things and implement said batching in a follow up patch so the benefits of compression and keep alive are already made available. This would only require a bit of rewording of the pull request message.

Issif:
Don't if the batch will be really useful, very few targets accept that. On the other hands, having more control on the http output would be a really nice thing:
* allow to set the User-Agent
* allow to set the Headers (really useful for Authentications either)
* allow to set the Http Method (PUT/POST)
* allow to use mTLS for the authentication

sgaist:
@Issif based on the current code of the http output, the following are already available:
- User agent setting
- mTLS configuration

When you say authentication, are you thinking about setting a username/password or a token in the authorization header or something more involved like oauth ?

Issue #2541, opened 2023-05-10
Override Default Priority


ericjulien-okta:
**Motivation**
I'm not able to change the Priority of rules in falco_rules.yaml with 'append'.  To change the Priority I need to copy the entire rule, add it to my falco_rules.local.yaml file, and change just the Priority.

**Feature**
I would like to be able change just the Priority of a default rule, similar to the ability to enable a disabled default rule.

**Alternatives**
Either allow this with 'append' or make is similar to 'enable'.  For example:
- rule: some_default_rule
  priority: WARNING

atmosx:
Same here. Looks like I have to re-write the rule :-/

jasondellaluce:
Yeah, I can confirm that for now you need to re-write the whole rule. This is a known issue and I'm tracking this personally. Hopefully we'll find a way to solve this kind of issues. Just FYI, the reason why nobody gave all this a stab so far is that changing this behavior would be a breaking change for all existing rulesets, so when we'll tackle this problem we'll need to be cautious.

incertum:
Please see https://github.com/falcosecurity/falco/issues/1340#issuecomment-1622563824 we would need a re-designed override feature instead of extending `append` here.

incertum:
@ericjulien-okta I am closing this as we are tracking this effort in other issues already https://github.com/falcosecurity/falco/issues/1340. I will tag you in there, thanks!

Issue #2477, opened 2023-04-07
[New Feature] Support partial rule overrides (merge rules of the same name)


tspearconquest:
**Motivation**

I came across the [rule exceptions](https://falco.org/docs/rules/exceptions/) page this evening and tried it out.

Falco is printing output against the `Contact K8S API Server From Container` rule and I want to make it less spammy in my environment. I wanted to add the container name to the output, in order to better identify the containers triggering the spammy messages; so I added this block to a new values file and did helm upgrade with this values file appended in addition to my normal values:

```
customRules:
## Add the container name to the output
  falco_rules.local.yaml: |-
    - rule: Contact K8S API Server From Container
      output: Unexpected connection to K8s API Server from container (command=%proc.cmdline pid=%proc.pid %container.info container.name=%container.name image=%container.image.repository:%container.image.tag connection=%fd.name)
```

When loading Falco with this config:

```
Error: /etc/falco/rules.d/falco_rules.local.yaml: Invalid
1 Errors:
In rules content: (/etc/falco/rules.d/falco_rules.local.yaml:0:0)
    rule 'Contact K8S API Server From Container': (/etc/falco/rules.d/falco_rules.local.yaml:14:2)
------
- rule: Contact K8S API Server From Container
  ^
------
LOAD_ERR_YAML_VALIDATE (Error validating internal structure of YAML file): Item has no mapping for key 'enabled'
```

When I added `enabled: true` and redeployed falco, I got:

```
Error: /etc/falco/rules.d/falco_rules.local.yaml: Invalid
1 Errors:
In rules content: (/etc/falco/rules.d/falco_rules.local.yaml:0:0)
    rule 'Contact K8S API Server From Container': (/etc/falco/rules.d/falco_rules.local.yaml:14:2)
------
- rule: Contact K8S API Server From Container
  ^
------
LOAD_ERR_YAML_VALIDATE (Error validating internal structure of YAML file): Item has no mapping for key 'condition'
```

I expected that Falco would read this override config file and realize that its an override for the rule with the same name, and because it already has a definition for the rule from the original rules yaml, it would merge the two together. As a Kubernetes user, I don't want to have to copy the entire rule from upstream to override part of the rule. Details on why this is are provided near the bottom.

**Feature**

It would be far more preferable if falco would do a YAML merge between the rule named in this override and the original content of the rule in `falco_rules.yaml` to produce the following in memory:

```
- rule: Contact K8S API Server From Container
[...]

PR #2981, opened 2023-12-19
new(engine): add selective rule overrides

### LINKED ISSUE #1340: Support partial overrides / merging  for existing rule (e.g. `priority` field) to complement the append functionality

LucaGuerra:
**What type of PR is this?**

/kind feature

/area engine

/area tests

This is a rather straightforward implementation of https://github.com/falcosecurity/falco/issues/1340#issuecomment-1710135769 . See the comment for examples.

It is now possible to use the `override` key in rules, lists and macros. Some fields can be overridden: for lists and macros there is only one field that we care about, while for rules, the following fields have been selected:
* Available for `append`: `{"condition", "output", "desc", "tags", "exceptions"}`
* Available for `replace`: `{"condition", "output", "desc", "priority", "tags", "exceptions", "enabled", "warn_evttypes", "skip-if-unknown-filter"}`

As per the discussion linked above, it is an error to specify both `append: true` and any `override`. Given the syntax it is not possible to attempt to both replace and append the same field in the same entry.

Also, if you specify the `override` key you *must* specify all the fields that you wish to override and only those fields.

Fixes #1340

```release-note
new(engine): add selective overrides for Falco rules
```

Also, this needs to be documented ( cc @mikegcoleman )

### DIFF
--- a/unit_tests/engine/test_rule_loader.cpp
+++ b/unit_tests/engine/test_rule_loader.cpp
@@ -1,0 +1,235 @@
+#include <gtest/gtest.h>
+
+#include "falco_engine.h"
+#include "rule_loader_reader.h"
+#include "rule_loader_compiler.h"
+
+class engine_loader_test : public ::testing::Test {
+protected:
+	void SetUp() override
+	{
+		m_sample_ruleset = "sample-ruleset";
+		m_sample_source = falco_common::syscall_source;
+
+		// create a falco engine ready to load the ruleset
+		m_inspector.reset(new sinsp());
+		m_engine.reset(new falco_engine());
+		m_filter_factory = std::shared_ptr<gen_event_filter_factory>(
+			new sinsp_filter_factory(m_inspector.get(), m_filterlist));
+		m_formatter_factory = std::shared_ptr<gen_event_formatter_factory>(
+			new sinsp_evt_formatter_factory(m_inspector.get(), m_filterlist));
+		m_engine->add_source(m_sample_source, m_filter_factory, m_formatter_factory);
+	}
+
+	void TearDown() override
+	{
+
+	}
+
+	bool load_rules(std::string rules_content, std::string rules_filename)
+	{
+		bool ret = false;
+		falco::load_result::rules_contents_t rc = {{rules_filename, rules_content}};
+		m_load_result = m_engine->load_rules(rules_content, rules_filename);
+		m_load_result_string = m_load_result->as_string(true, rc);
+		ret = m_load_result->successful();
+
+		if (ret)
+		{
+			m_engine->enable_rule("", true, m_sample_ruleset);
+		}
+
+		return ret;
+	}
+
+	std::string m_sample_ruleset;
+	std::string m_sample_source;
+	sinsp_filter_check_list m_filterlist;
+	std::shared_ptr<gen_event_filter_factory> m_filter_factory;
+	std::shared_ptr<gen_event_formatter_factory> m_formatter_factory;
+	std::unique_ptr<falco_engine> m_engine;
+	std::unique_ptr<falco::load_result> m_load_result;
+	std::string m_load_result_string;
+	std::unique_ptr<sinsp> m_inspector;
+};
+
+std::string s_sample_ruleset = "sample-ruleset";
+std::string s_sample_source = falco_common::syscall_source;
+
+TEST_F(engine_loader_test, list_append)
+{
+    std::string rules_content = R"END(
+- list: shell_binaries
+  items: [ash, bash, csh, ksh, sh, tcsh, zsh, dash]
+
+- rule: legit_rule
+  desc: legit rule description
+  condition: evt.type=open and proc.name in (shell_binaries)
+  output: user=%user.name command=%proc.cmdline file=%fd.name
+  priority: INFO
+
+- list: shell_binaries
+  items: [pwsh]
+  override:
+    items: append
+)END";
+
+	std::string rule_name = "legit_rule";
+	ASSERT_TRUE(load_rules(rules_content, "legit_rules.yaml")) << m_load_result_string;
+
+	auto rule_description = m_engine->describe_rule(&rule_name, {});
+	ASSERT_EQ(rule_description["rules"][0]["details"]["condition_compiled"].template get<std::string>(),
+		"(evt.type = open and proc.name in (ash, bash, csh, ksh, sh, tcsh, zsh, dash, pwsh))");
+}
+
+TEST_F(engine_loader_test, condition_append)
+{
+    std::string rules_content = R"END(
+- macro: interactive
+  condition: >
+    ((proc.aname=sshd and proc.name != sshd) or
+    proc.name=systemd-logind or proc.name=login)
+
+- rule: legit_rule
+  desc: legit rule description
+  condition: evt.type=open and interactive
+  output: user=%user.name command=%proc.cmdline file=%fd.name
+  priority: INFO
+
+- macro: interactive
+  condition: or proc.name = ssh
+  override:
+    condition: append
+)END";
+
+	std::string rule_name = "legit_rule";
+	ASSERT_TRUE(load_rules(rules_content, "legit_rules.yaml")) << m_load_result_string;
+
+	auto rule_description = m_engine->describe_rule(&rule_name, {});
+	ASSERT_EQ(rule_description["rules"][0]["details"]["condition_compiled"].template get<std::string>(),
+		"(evt.type = open and (((proc.aname = sshd and proc.name != sshd) or proc.name = systemd-logind or proc.name = login) or proc.name = ssh))");
+}
+
+TEST_F(engine_loader_test, rule_override_append)
+{
+    std::string rules_content = R"END(
+- rule: legit_rule
+  desc: legit rule description
+  condition: evt.type=open
+  output: user=%user.name command=%proc.cmdline file=%fd.name
+  priority: INFO
+
+- rule: legit_rule
+  desc: with append
+  condition: and proc.name = cat
+  output: proc=%proc.name
+  override:
+    desc: append
+    condition: append
+    output: append
+)END";
+
+	std::string rule_name = "legit_rule";
+	ASSERT_TRUE(load_rules(rules_content, "legit_rules.yaml")) << m_load_result_string;
+
+	auto rule_description = m_engine->describe_rule(&rule_name, {});
+	ASSERT_EQ(rule_description["rules"][0]["info"]["condition"].template get<std::string>(),
+	 	"evt.type=open and proc.name = cat");
+
+	ASSERT_EQ(rule_description["rules"][0]["info"]["output"].template get<std::string>(),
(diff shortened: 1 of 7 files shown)


incertum:
Thank you @LucaGuerra could you let me know when it would be ready to test drive it?

userspace/engine/rule_loader_collector.cpp
```diff
@@ -185,7 +191,7 @@ void rule_loader::collector::append(configuration& cfg, list_info& info)
 {
 	auto prev = m_list_infos.at(info.name);
 	THROW(!prev,
-	       "List has 'append' key but no list by that name already exists",
+	       "List has 'append' key or an append override but no list by that name already exists",
```
LucaGuerra, line 194: Note for the reviewer: this string change is making our testing pipelines fail. No big deal, once we pick the right string I can update the pipelines.

userspace/engine/rule_loader_collector.cpp
```diff
@@ -48,8 +48,14 @@ static inline void define_info(indexed_vector<T>& infos, T& info, uint32_t id)
...
-template <typename T>
-static inline void append_info(T* prev, T& info, uint32_t id)
+template <typename T, typename U>
+static inline void append_info(T* prev, U& info, uint32_t id)
+{
+	prev->visibility = id;
+}
+
+template <typename T, typename U>
+static inline void replace_info(T* prev, U& info, uint32_t id)
```
LucaGuerra, line 58: Note for the reviewer: the addition of `replace_info` could be cargo cult programming, I am not sure if we need to do more here or if this is useful.

userspace/engine/rule_loader.h
```diff
@@ -451,4 +453,38 @@ namespace rule_loader
 		bool warn_evttypes;
 		bool skip_if_unknown_filter;
 	};
+
+	/*!
+		\brief Represents infos about a rule update (append or replace) request
+	*/
+
+	struct rule_update_info
```
LucaGuerra, line 461: Note for the reviewer: while we could add `std::optional`s to the existing structures, those are also used to evaluate and handle actual rules which must have those fields defined. I wouldn't want the rest of the engine to care about whether a mandatory field was defined or not (as it must be) so it sounds ok to keep those fields here.

LucaGuerra:
@incertum if you are able to test this it would be appreciated :)

userspace/engine/rule_loader_collector.cpp
LucaGuerra: After speaking with @jasondellaluce , this seems the right approach.

unit_tests/engine/test_rule_loader.cpp
```diff
@@ -0,0 +1,235 @@
+#include <gtest/gtest.h>
+
+#include "falco_engine.h"
+#include "rule_loader_reader.h"
+#include "rule_loader_compiler.h"
+
+class engine_loader_test : public ::testing::Test {
```
incertum, line 7: Very neat setup @LucaGuerra much more elegant :heart: than what I had tried in the past :woman_facepalming:

unit_tests/engine/test_rule_loader.cpp
```diff
@@ -0,0 +1,235 @@
...
+	 	"evt.type = close");
+
+	ASSERT_EQ(rule_description["rules"][0]["info"]["output"].template get<std::string>(),
+	 	"user=%user.name command=%proc.cmdline file=%fd.name");
+
+	ASSERT_EQ(rule_description["rules"][0]["info"]["description"].template get<std::string>(),
+	 	"a replaced legit description");
+}
+
+TEST_F(engine_loader_test, rule_override_append_replace)
```
incertum, line 201: Few ideas to add more tests to gain even more confidence:

Inline with original requirements:

> It is now possible to use the override key in rules, lists and macros. Some fields can be overridden: for lists and macros there is only one field that we care about, while for rules, the following fields have been selected:

Add a test to verify an error when using `append` for the field `priority`

> - Available for append: {"condition", "output", "desc", "tags", "exceptions"}

Add a test for an unsupported field, e.g. `enabled` in `append` mode and verify error message and no success.

> -Available for replace: {"condition", "output", "desc", "priority", "tags", "exceptions", "enabled", "warn_evttypes", "skip-if-unknown-filter"}

> As per the discussion linked above, it is an error to specify both append: true and any override. Given the syntax it is not possible to attempt to both replace and append the same field in the same entry.

Test the error when using both `append` and `override`

> Also, if you specify the override key you must specify all the fields that you wish to override and only those fields.

Create few tests with a mismatch, that is, you use a field in `override` that you have not provided a statement and vice versa add a statement for which there is no matching `override` definition. Actually what is the behavior in that case? Just keep the old and no error?


More ideas:

- Test successive overrides of the same rule just to be future proof and verify the final result.
- Plus I keep calling it out: What is the behavior when I disable a rule using this selective override but then enable all rules with a certain tag and unfortunately the rule I disabled with selective overrides happened to have that tag. Is that rule at the end enabled or disabled? 
- Test error when defining a rule with `override` feature, but the rule was not defined in the first place.

incertum (commented):
First pass review focused on the tests. Could we add some more? See comment.

Code implementation LGTM and @jasondellaluce already approved the approach as well :smile:.

unit_tests/engine/test_rule_loader.cpp
LucaGuerra: I noticed we actually do something similar in at least three different locations in the test suite (including this one). This is out of scope for this PR but we need to consolidate them and have one fixture to allow loading engine + rulesets.

userspace/engine/rule_loader_collector.cpp
LucaGuerra: I have updated the code to revert the error messages, I really want to be sure that our tests are happy before merging. Messages can be updated in a subsequent PR coordinated with testing.

unit_tests/engine/test_rule_loader.cpp
LucaGuerra: Good point. I have added error testcases for most cases you mentioned, those are not only useful for these specifics but are easy enough to understand so they can be extended for more cases we can think of

> Actually what is the behavior in that case? Just keep the old and no error?
The way that makes more sense to me is that is an error to ask for an override and not define it and vice versa. Actually, the opposite was not checked and I now added a check for it. Essentially, if you define a field and not the corresponding append/replace override it will result in an error; this is because if you defined the field in an entry with an override section you probably wanted to override that. I also added the corresponding test.

Also, for more complex cases (e.g. enabling/disabling with tags) I think they may be out of scope for this PR and if you can think about unit or integration tests for those cases it would be great to add them.

incertum:
> Also, for more complex cases (e.g. enabling/disabling with tags) I think they may be out of scope for this PR and if you can think about unit or integration tests for those cases it would be great to add them.

Agreed, just wanted to mention it once again :upside_down_face:

unit_tests/engine/test_rule_loader.cpp
incertum: Big +1, it would be super useful to do this cleanup!

unit_tests/engine/test_rule_loader.cpp
```diff
@@ -0,0 +1,351 @@
...
+  override:
+    desc: replace
+    condition: append
+    priority: append
+)END";
+
+	std::string rule_name = "failing_rule";
+
+	ASSERT_FALSE(load_rules(rules_content, "rules.yaml"));
+	ASSERT_EQ(m_load_result_json["errors"][0]["message"], "Key 'priority' cannot be appended to");
```
incertum, line 261: [nit] `Key 'priority' cannot be appended to, use 'replace' instead` -> same for `enabled`

unit_tests/engine/test_rule_loader.cpp
```diff
@@ -0,0 +1,351 @@
...
+  override:
+    desc: replace
+    condition: append
+)END";
+
+	std::string rule_name = "failing_rule";
+
+	ASSERT_FALSE(load_rules(rules_content, "rules.yaml"));
+	// std::cout << m_load_result_json.dump(4) << std::endl;
+	ASSERT_TRUE(std::string(m_load_result_json["errors"][0]["message"]).find("no rule by that name already exists") != std::string::npos);
```
incertum, line 304: [nit] `you are overriding a rule, but no rule by that name already exists` or similar

unit_tests/engine/test_rule_loader.cpp
```diff
@@ -0,0 +1,351 @@
...
+  priority: WARNING
+  override:
+    desc: replace
+    condition: append
+)END";
+
+	std::string rule_name = "failing_rule";
+
+	ASSERT_FALSE(load_rules(rules_content, "rules.yaml"));
+	ASSERT_TRUE(std::string(m_load_result_json["errors"][0]["message"]).find("Unexpected key 'priority'") != std::string::npos);
```
incertum, line 350: [nit] same can we make the error message more expressive?

unit_tests/engine/test_rule_loader.cpp
LucaGuerra: This is an "error message contains" not equals :)

incertum: :)

unit_tests/engine/test_rule_loader.cpp
LucaGuerra: Good point, updated!

incertum (approved):
/approve

Wonderful work @LucaGuerra, thank you!

jasondellaluce (approved):
/approve

/hold

Super job Luca! I'm approving and holding just so that you can decide when you feel confident in merging this, but it looks great to me.

incertum:
/unhold

Had spoken with Luca on slack. We are good to merge it.

incertum:
The following 2 issues were duplicates of the primary issue https://github.com/falcosecurity/falco/issues/1340:

- https://github.com/falcosecurity/falco/issues/2477
- https://github.com/falcosecurity/falco/issues/2541

Link them for tracking.

mikegcoleman:
Added a docs issue to track the documentation need

https://github.com/falcosecurity/falco-website/issues/1226

pmusa:
I think this part also needs to be updated:
https://falco.org/docs/rules/controlling-rules/#via-custom-rule-definition

PR #2986, opened 2023-12-20
docs(proposal): 20231220-features-adoption-and-deprecation.md

leogr:
**What type of PR is this?**

/kind design
/kind documentation

/area proposals

As discussed among maintainers and based on feedback, the need to balance quick project evolution and avoid frequent changes that can disrupt adopters' workflow has emerged. This proposal aims to find that balance in the long term. 

Although apparently well structured, the proposal must be considered as an RFC at the moment. The current goal is to ensure the proposed solution offers a good trade-off and will not introduce any unnecessary burden. Any aspect of the proposal can be changed at this stage. 

Tentatively for
/milestone 0.37.0

```release-note
NONE
```

### DIFF
--- a/proposals/20231220-features-adoption-and-deprecation.md
+++ b/proposals/20231220-features-adoption-and-deprecation.md
@@ -115,7 +115,7 @@ The adoption policy applies to any backward compatible user-facing changes which
 
 **Adoption rules**:
 1. A feature can be introduced at only one of the following levels:
-   - Sandbox: The feature must be opt-in (e.g., a feature flag), labeled as *Sandbox* and the user must be proactively informed by the experimental nature of the feature (i.e. emitting a notice when the feature is being enabled).
+   - Sandbox: The feature must be opt-in (e.g., using a feature flag), labeled as *Sandbox* and the user must be proactively informed by the experimental nature of the feature (i.e. emitting a notice when the feature is being enabled).
    - Incubating: The feature must be labeled as *Incubating*.
 2. A feature can be promoted *from Sandbox to Incubating* or *from Incubating to Stable* only after at least one release cycle has passed without user-facing changes to the feature.
 3. A feature cannot be demoted to a previous level.
--- a/proposals/20231220-features-adoption-and-deprecation.md
+++ b/proposals/20231220-features-adoption-and-deprecation.md
@@ -90,6 +90,14 @@ In the context of this proposal, only changes to **public APIs** are assumed to 
 
 Public APIs are defined as those supporting Falco functioning and explicitly intended for user usage. Internal APIs consumed by Falco or other tools are out of scope for this area. For instance, the driver APIs or libs APIs are intended to be mainly consumed by Falco and not by users.
 
+### Platform Support Area
+
+Platform support for Falco encompasses the range of platforms, systems, and environments it is designed to operate in. Platform support may significantly vary by Falco's data sources and use cases. For example, its compatibility differs when utilized for Kubernetes audit events versus system call events. Additionally, platform support can be influenced by deployment methods (e.g., directly on a host versus within Kubernetes) or configurations (e.g., running in privileged versus least privileged mode).
+
+Given the diversity of potential platforms and setups, only those explicitly listed in Falco's documentation are considered officially supported. While Falco may function on other platforms, official support is guaranteed solely for documented ones.
+
+Therefore, changes in platform compatibility or behavior that are documented explicitly assumed to be user-facing changes to the Platform Support area.
+
 ### Release Cycle
 
 In the context of this proposal, a release cycle is the period between two consecutive major or minor releases of Falco. Hotfix/Patch releases must not be counted.
--- a/proposals/20231220-features-adoption-and-deprecation.md
+++ b/proposals/20231220-features-adoption-and-deprecation.md
@@ -114,7 +114,7 @@ Maturation levels (inspired by those we already have in place for [repositories]
 The adoption policy applies to any backward compatible user-facing changes which add functionalities. For non-backward compatible changes see the [deprecation policy](#Deprecation-policy) below.
 
 **Adoption rules**:
-1. A feature can be introduced at any level at only one of the following levels:
+1. A feature can be introduced at only one of the following levels:
    - Sandbox: The feature must be opt-in (e.g., a feature flag), labeled as *Sandbox* and the user must be proactively informed by the experimental nature of the feature (i.e. emitting a notice when the feature is being enabled).
    - Incubating: The feature must be labeled as *Incubating*.
 2. A feature can be promoted *from Sandbox to Incubating* or *from Incubating to Stable* only after at least one release cycle has passed without user-facing changes to the feature.
--- a/proposals/20231220-features-adoption-and-deprecation.md
+++ b/proposals/20231220-features-adoption-and-deprecation.md
@@ -41,10 +41,6 @@ Undocumented behaviors may be included in this definition if there's strong evid
 
 User-facing changes refer to any feature changes, behavior changes, modifications, or additions that are directly noticeable and interactable by the end users. These changes affect how Falco operates from the user's perspective (notably any change that can lead to user disruption). Unlike internal changes (i.e., code refactoring, CI, maintenance-related changes), which are under-the-hood improvements not directly visible to the user, user-facing changes are evident in the Falco and its core components interface and functionality.
 
-### Feature
-
-TODO: Add a definition of feature.
-
 ### CLI/Config Area
 
 Falco is comprised of the Falco binary and other programs and tools cooperating (notably [falcoctl](https://github.com/falcosecurity/falcoctl)). These programs are the primary user interface for Falco. Any feature or behavior changes to the following elements of these programs are assumed to be user-facing changes to the CLI/Config area:
@@ -59,9 +55,9 @@ Falco is comprised of the Falco binary and other programs and tools cooperating 
 - Program inputs, excluding elements explicitly governed by other areas (e.g., [Falco rules](#rules-area)).
 - Program outputs excluding elements explicitly governed by other areas (e.g., [Falco outputs/alerts](#outputs-alerts-area)).
 
-### Rules Area
+### Rules System Area
 
-Rules are the primary input for Falco. Any feature or behavior changes to the following aspects or elements are assumed to be user-facing changes to the rules area:
+Rules are the primary input for Falco. Any feature or behavior changes to the following aspects or elements are assumed to be user-facing changes to the rules system area:
 
 - Syntax.
 - File format.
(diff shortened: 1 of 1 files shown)


proposals/20231220-features-adoption-and-deprecation.md
```diff
@@ -0,0 +1,207 @@
...
+### Adoption Policy
+
+Feature additions will follow a straightforward adoption model with defined status levels (inspired by those we already have in place for [repositories](https://github.com/falcosecurity/evolution/blob/main/REPOSITORIES.md#status)) aiming to characterize the current maturity level.
+
+Each feature will have an assigned status level at any specific time (i.e., a Falco release). Statuses are shown in the table below.
+
+| Status | Feature Gate | Intended for |
+| --- | --- | --- |
+| Sandbox | Disabled by default | Experimental/alpha features, not recommended for production use, can be removed at any time without further notice. |
+| Incubating | Enabled by default | Beta features, long-term support is not guaranteed. |
```
leogr, line 106: We should consider putting this off by default.

leogr:

proposals/20231220-features-adoption-and-deprecation.md
incertum: Features like `metrics` are disabled by default because maybe not everyone needs or wants them, but should still be considered a "Stable" core feature. Another example is the different outputs channels. Only because we enabled one output channel by default, does not mean the other ones are not stable. 

I believe we need more nuances here. Allow stable features to be disabled by default. And Incubating features to be enabled by default. This will align with the rules maturity framework where the enabled tag is separate from the maturity level. Instead we can use terms like "shipped with the default package / configuration" or similar to let users know what options are enabled by default regardless of Status.

leogr: Thanks to your comment, I've realized that what I wrote in the proposal is highly misleading :sweat_smile:  

I meant that we need to introduce a toggle to bulk turn all features on/off in a specific set (i.e., sandbox, incubating, etc.). This would be on top of a specific feature option that allows users to turn the specific feature on/off. The term "feature gate" probably is not accurate for the purpose I had in mind. Perhaps "maturity level toggle" works better. I will rework the text and fix that.

So, I totally agree that not all stable features must be enabled by default. I just meant that we don't need a "maturity level toggle" for stable features (because it wouldn't be helpful).

proposals/20231220-features-adoption-and-deprecation.md
```diff
@@ -0,0 +1,207 @@
...
+- Different parties may provide plugins, and each plugin may have a different maturity level. Only those plugins officially maintained by The Falco Project and identified as "core" for Falco are in scope for this policy; all others are excluded.
+- Any other exceptions to the rules provided by this policy require a formal core maintainer majority vote.
+
+### Versioning
+
+Regarding the above policies, component versioning must adhere to [Semantic Versioning 2.0.0](https://semver.org/). However, in the context of Falco core components, the scope extends beyond the strict API definition and includes any user-facing changes.
+
+Thus, given a version number `MAJOR.MINOR.PATCH` increment the:
+
+- *MAJOR* version when the deprecation period ends, thus introducing incompatible user-facing or API changes.
```
leogr, line 184:
```suggestion
- *MAJOR* version when the deprecation period of one or more _stable_ features ends, thus introducing incompatible user-facing or API changes.
```

This edit would prevent us from bumping the MAJOR when introducing a breaking change in incubating/stable features. Otherwise, the MAJOR version will increase very quickly and we probably don't want that.
wdyt?

proposals/20231220-features-adoption-and-deprecation.md
incertum: ACK do you happen to know software that uses such bulk feature activation? Was thinking and unsure yet if I would find it useful for adoption (never heard of such a concept before). Usually you are used to toggle individual config option on and off.

incertum:
@leogr some more feedback

- re "Define the number of previous releases that will receive patches or security updates and the duration of this support": I would be very careful with signing up for LTS as it is a huge overhead. For the foreseeable future I would like us to continue to only support patches for the current stable release
- In the table I would just keep "Status" and "Intended for" and as I said before I am not sure if we will benefit from a "Feature Gate". It may make things too complicated.
- Need formal criteria for each status plus for deciding what can be enabled by default (this would also help to audit some of the current defaults, some defaults may not be optimal atm) -> user-friendliness and performance should be on top of the criteria list
- Proposing to formalize an adopters check list for each release going forward aligning with the "adopter expectations on the operational cost of using Falco" you mentioned
- Since this is just the proposal where will the final doc live, under governance?
- For the next 3 releases we need to communicate that we are introducing more formality and as a result we are going to clean up more than usually. Such as acknowledging that some configs or settings are not working well anymore and are replaced by better options xyz
- Another suggestion is to also set some expectations wrt what areas of the project are more frequntly changing compared to areas or aspects that are more stable. For example, configs change often, but we never remove filter fields or rename them and never changed the minimum supported kernel versions per driver etc.
- We also need a plan for undesired changes ... what if a config is broken and can't be fixed? It happened before.
- One area under "Definitions" may be missing, e.g. "Platform support" which should include the kernel version support matrix, container runtimes etc

leogr:
> * Since this is just the proposal where will the final doc live, under governance?

The governance is intended for the overall project, not for components. So, I thought just a .MD in the root of the `falco` repo and a mirror on the website. wdyt?

leogr:
> * We also need a plan for undesired changes ... what if a config is broken and can't be fixed? It happened before.

Could you provide an example? Thanks!

incertum:
> > * Since this is just the proposal where will the final doc live, under governance?
> 
> The governance is intended for the overall project, not for components. So, I thought just a .MD in the root of the `falco` repo and a mirror on the website. wdyt?

Great idea yes!

incertum:
> > * We also need a plan for undesired changes ... what if a config is broken and can't be fixed? It happened before.
> 
> Could you provide an example? Thanks!

Remember the `-d` flag just broke and we simply could not fix it, hence we decided to just remove it. I agree it may be less likely to happen again, but if it's not a big deal to add a sentence around this why not?

leogr:
> Remember the `-d` flag just broke and we simply could not fix it, hence we decided to just remove it. I agree it may be less likely to happen again, but if it's not a big deal to add a sentence around this why not?

I would leave this edge case as a general exception :point_down: so we can decide on a case-by-case basis. wdyt?

> Any other exceptions to the rules provided by this policy require a formal core maintainer majority vote.

proposals/20231220-features-adoption-and-deprecation.md
leogr: Removed the feature gate concept in the end. The policies should be simpler now.

https://github.com/falcosecurity/falco/pull/2986/commits/ef240a5568fb26d2b84ee7a90fbbcae8c88c5849

proposals/20231220-features-adoption-and-deprecation.md
```diff
@@ -94,56 +98,45 @@ The actual duration of a release cycle can vary. Still, it's assumed to be about
...
-| Deprecated | Enabled by default | See the [deprecation policy](#Deprecation-policy) section below. |
+### Adoption Policy
+
+The adoption policy applies to any backward compatible user-facing changes which add functionalities. For non-backward compatible changes see the [deprecation policy](#Deprecation-policy) below.
 
 **Adoption rules**:
-1. Only Sandbox or Incubating status can be assigned (at the maintainers' discretion) when introducing a new feature. Incubating is the recommended path.
-2. A feature can be promoted to a higher status level only after at least one release cycle has passed without user-facing changes to the feature.
-3. Demoting a feature to a lower status is not allowed; use deprecation instead.
+1. A feature can be introduced at any level at only one of the following levels:
```
leogr, line 117:
```suggestion
1. A feature can be introduced at only one of the following levels:
```

proposals/20231220-features-adoption-and-deprecation.md
```diff
@@ -94,56 +98,45 @@ The actual duration of a release cycle can vary. Still, it's assumed to be about
...
+   - Sandbox: The feature must be opt-in (e.g., a feature flag), labeled as *Sandbox* and the user must be proactively informed by the experimental nature of the feature (i.e. emitting a notice when the feature is being enabled).
```
leogr, line 118:
```suggestion
   - Sandbox: The feature must be opt-in (e.g., using a feature flag), labeled as *Sandbox* and the user must be proactively informed by the experimental nature of the feature (i.e. emitting a notice when the feature is being enabled).
```

incertum:
Hi @leogr I think the most recent WIP commit is great.

Looking back at some of my initial feedback https://github.com/falcosecurity/falco/pull/2986#issuecomment-1867277929 I think that was a lot, maybe too much.


After a fresh look I would now just propose to add one more Area, namely `Platform Support Area` and outline all things OS, arch, kernel versions, container runtime sockets etc support. WDYT?

proposals/20231220-features-adoption-and-deprecation.md
```diff
@@ -0,0 +1,219 @@
...
+
+### Exceptions
+
+- Ruleset in the official distributions follow the [Rules Maturity Framework](https://github.com/falcosecurity/rules/blob/main/CONTRIBUTING.md#rules-maturity-framework) policies.
+- Subsystems or subcomponents may have additional criteria and exceptions. Stated other criteria and exceptions must not directly affect the main Falco distribution (e.g., *falcoctl* can have a different release cycle and different policies; however, if Falco relies on a specific *falcoctl* feature, that *falcoctl* feature adoption and deprecation must be strictly compatible with the rules described in this proposal).
+- Internal APIs are out of scope for this policy. Their adoption models and deprecation policies might be regulated separately.
+- Different parties may provide plugins, and each plugin may have a different maturity level. Only those plugins officially maintained by The Falco Project and identified as "core" for Falco are in scope for this policy; all others are excluded.
+- Any other exceptions to the rules provided by this policy require a formal core maintainer majority vote.
+
+### Versioning
```
incertum, line 190: @leogr proposal LGTM! One super minor comment: Do you think we could enumerate all core components here for which we plan to introduce semver versioning?

leogr: We can enumerate core repositories, but we haven't yet defined a consistent way to enumerate core components.

Anyway, the goals I had in mind are:
- Now we agree that all components affected by adoption/deprecation policies must be _SemVer_sioned
- During the transition phases we will identify all of them and make them explicit

If this is not already clear, I can try to improve to improve the text a bit. May this help?

incertum (approved):
/approve

I know it's still marked as WIP, but wanted to show support already that this LGTM. I can re-approve should you still want to make minor adjustments.

leogr:
@falcosecurity/falco-maintainers please review :pray:

cpanato (approved):
+1

/lgtm

incertum:
Bump @falcosecurity/falco-maintainers any additional feedback?

FedeDP (approved):
/approve
Great proposal and wording. Totally agree!

Issue #2763, opened 2023-09-01
[TRACKING] Breaking changes in Falco 0.37.0


Andreagit97:
This issue keeps track of all deprecated Falco features that will be removed in Falco 0.37

- [x] rate-limiter mechanism
   - deprecation PR: https://github.com/falcosecurity/falco/pull/2762
   - breaking change PR: https://github.com/falcosecurity/falco/pull/2841
- [x] `--userspace` CLI option
   - deprecation PR: https://github.com/falcosecurity/falco/pull/2761
   - breaking change PR: https://github.com/falcosecurity/falco/pull/2839
- [x] `falco-driver-loader` script will be removed and embedded into `falcoctl` 
    - deprecation issue: https://github.com/falcosecurity/falco/issues/2675
    - breaking change PR: https://github.com/falcosecurity/falco/pull/2905
    - About falco-driver-loader; the new `falcoctl` driven implementation will drop:
      * [`--source-only`](https://github.com/falcosecurity/falco/blob/master/scripts/falco-driver-loader#L685)
      * [`BPF_USE_LOCAL_KERNEL_SOURCES`](https://github.com/falcosecurity/falco/blob/master/scripts/falco-driver-loader#L559)
      * [`DRIVER_CURL_OPTIONS`](https://github.com/falcosecurity/falco/blob/master/scripts/falco-driver-loader#L692)
      * [`FALCO_BPF_PROBE`](https://github.com/falcosecurity/falco/blob/1b62b5ccd1c64cd972ef0252262075cbf42a130c/scripts/falco-driver-loader#L738C9-L738) environment variable won't be considered by new falcoctl driver loader, since it is already deprecated (above points, Falco 0.37.0) and going to be removed in Falco 0.38.0
    - Also, some env vars were renamed:  
      * [`DRIVERS_REPO`](https://github.com/falcosecurity/falco/blob/1b62b5ccd1c64cd972ef0252262075cbf42a130c/scripts/falco-driver-loader#L690) env variable has been replaced by `FALCOCTL_DRIVER_NAME`, or `--name` command line argument for falcoctl driver tool
      * [`DRIVERS_NAME`](https://github.com/falcosecurity/falco/blob/1b62b5ccd1c64cd972ef0252262075cbf42a130c/scripts/falco-driver-loader#L689) env variable has been replaced by `FALCOCTL_DRIVER_REPOS`, or `--repo` command line argument for falcoctl driver tool
[...]

PR #2992, opened 2024-01-03
update(rule_loader): deprecate the `append` flag in falco rules

Andreagit97:
**What type of PR is this?**

/kind cleanup

/kind design

/area engine

/area tests

In this PR I propose to deprecate the `append` flag in the Falco rules. With `deprecate` I mean that we should log a warning when the `append` key is used suggesting the new `override` key. The flag shouldn't be removed until Falco 1.0.0 since it could cause a big breaking change between users (I'm open to other ideas, maybe we can wait for some release cycle). In this PR there are also some cleanups.

Side note: I'm not fully convinced by the fact that warnings in the rule loading are suppressed by default...if we introduce a warning I expect that users will see it by default since probably they need to take some actions... To be more concrete, in this PR i added a warning that in my opinion should be shown by default because users should know how to update the rules with the most recent features to avoid breaking changes in the future. Example:

```
Wed Jan  3 15:46:27 2024: Falco version: 0.37.0-231+ad964c0 (x86_64)
Wed Jan  3 15:46:27 2024: Falco initialized with configuration file: ../falco.yaml
Wed Jan  3 15:46:27 2024: System info: Linux version 6.2.0-39-generic (buildd@lcy02-amd64-045) (x86_64-linux-gnu-gcc-11 (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0, GNU ld (GNU Binutils for Ubuntu) 2.38) #40~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Nov 16 10:53:04 UTC 2
Wed Jan  3 15:46:27 2024: Loading rules from file ../rules/falco-deprecated_rules.yaml
Wed Jan  3 15:46:27 2024: Loading rules from file ../rules/falco-incubating_rules.yaml
../rules/falco-incubating_rules.yaml: Ok, with warnings
1 Warnings:
In rules content: (../rules/falco-incubating_rules.yaml:0:0)
    rule 'Non sudo setuid': (../rules/falco-incubating_rules.yaml:1275:2)
------
- rule: Non sudo setuid
  ^
------
LOAD_DEPRECATED_ITEM (Used deprecated item): 'append' key is deprecated. Add an append entry (e.g. 'condition: append') under 'override' instead.

Wed Jan  3 15:46:27 2024: Loading rules from file ../rules/falco-sandbox_rules.yaml
Wed Jan  3 15:46:28 2024: Loading rules from file ../rules/falco_rules.yaml
Wed Jan  3 15:46:28 2024: The chosen syscall buffer dimension is: 8388608 bytes (8 MBs)
Wed Jan  3 15:46:28 2024: Starting health webserver with threadiness 8, listening on 0.0.0.0:8765
Wed Jan  3 15:46:28 2024: Loaded event sources: syscall
Wed Jan  3 15:46:28 2024: Enabled event sources: syscall
Wed Jan  3 15:46:28 2024: Opening 'syscall' source with modern BPF probe.
Wed Jan  3 15:46:28 2024: One ring buffer every '2' CPUs.
```

Today to obtain the above result users should run Falco with the `-v` which IMO is not ideal, WDYT @falcosecurity/falco-maintainers ?

```release-note
update(rule_loader): deprecate the `append` flag in Falco rules
```

### DIFF
--- a/unit_tests/engine/test_rule_loader.cpp
+++ b/unit_tests/engine/test_rule_loader.cpp
@@ -98,6 +98,12 @@ protected:
 		return false;
 	}
 
+	std::string get_compiled_rule_condition(std::string rule_name = "")
+	{
+		auto rule_description = m_engine->describe_rule(&rule_name, {});
+		return rule_description["rules"][0]["details"]["condition_compiled"].template get<std::string>();
+	}
+
 	std::string m_sample_ruleset;
 	std::string m_sample_source;
 	sinsp_filter_check_list m_filterlist;
@@ -131,12 +137,8 @@ TEST_F(engine_loader_test, list_append)
     items: append
 )END";
 
-	std::string rule_name = "legit_rule";
 	ASSERT_TRUE(load_rules(rules_content, "legit_rules.yaml")) << m_load_result_string;
-
-	auto rule_description = m_engine->describe_rule(&rule_name, {});
-	ASSERT_EQ(rule_description["rules"][0]["details"]["condition_compiled"].template get<std::string>(),
-		"(evt.type = open and proc.name in (ash, bash, csh, ksh, sh, tcsh, zsh, dash, pwsh))");
+	ASSERT_EQ(get_compiled_rule_condition("legit_rule"),"(evt.type = open and proc.name in (ash, bash, csh, ksh, sh, tcsh, zsh, dash, pwsh))");
 }
 
 TEST_F(engine_loader_test, condition_append)
@@ -159,12 +161,8 @@ TEST_F(engine_loader_test, condition_append)
     condition: append
 )END";
 
-	std::string rule_name = "legit_rule";
 	ASSERT_TRUE(load_rules(rules_content, "legit_rules.yaml")) << m_load_result_string;
-
-	auto rule_description = m_engine->describe_rule(&rule_name, {});
-	ASSERT_EQ(rule_description["rules"][0]["details"]["condition_compiled"].template get<std::string>(),
-		"(evt.type = open and (((proc.aname = sshd and proc.name != sshd) or proc.name = systemd-logind or proc.name = login) or proc.name = ssh))");
+	ASSERT_EQ(get_compiled_rule_condition("legit_rule"),"(evt.type = open and (((proc.aname = sshd and proc.name != sshd) or proc.name = systemd-logind or proc.name = login) or proc.name = ssh))");
 }
 
 TEST_F(engine_loader_test, rule_override_append)
@@ -217,15 +215,12 @@ TEST_F(engine_loader_test, rule_append)
   append: true
 )END";
 
-	std::string rule_name = "legit_rule";
 	ASSERT_TRUE(load_rules(rules_content, "legit_rules.yaml")) << m_load_result_string;
 
 	// We should have at least one warning because the 'append' flag is deprecated.
 	ASSERT_TRUE(check_warning_message(WARNING_APPEND));
 
-	auto rule_description = m_engine->describe_rule(&rule_name, {});
-	ASSERT_EQ(rule_description["rules"][0]["details"]["condition_compiled"].template get<std::string>(),
-	 	"(evt.type = open and proc.name = cat)");
+	ASSERT_EQ(get_compiled_rule_condition("legit_rule"),"(evt.type = open and proc.name = cat)");
 }
 
 TEST_F(engine_loader_test, rule_override_replace)
@@ -396,10 +391,7 @@ TEST_F(engine_loader_test, macro_override_replace_before_macro_definition)
 
 	// The first override defines a macro that is overridden by the second macro definition
 	ASSERT_TRUE(load_rules(rules_content, "rules.yaml"));
-	std::string rule_name = "test_rule";
-	auto rule_description = m_engine->describe_rule(&rule_name, {});
-	ASSERT_EQ(rule_description["rules"][0]["details"]["condition_compiled"].template get<std::string>(),
-	 	"evt.type in (open, openat)");
+	ASSERT_EQ(get_compiled_rule_condition("test_rule"),"evt.type in (open, openat)");	
 }
 
 TEST_F(engine_loader_test, macro_append_before_macro_definition)
@@ -448,10 +440,7 @@ TEST_F(engine_loader_test, macro_override_append_after_macro_definition)
 
 	// We cannot define a macro override before the macro definition.
 	ASSERT_TRUE(load_rules(rules_content, "rules.yaml"));
-	std::string rule_name = "test_rule";
-	auto rule_description = m_engine->describe_rule(&rule_name, {});
-	ASSERT_EQ(rule_description["rules"][0]["details"]["condition_compiled"].template get<std::string>(),
-	 	"(evt.type in (open, openat) or evt.type = openat2)");
+	ASSERT_EQ(get_compiled_rule_condition("test_rule"),"(evt.type in (open, openat) or evt.type = openat2)");
 }
 
 TEST_F(engine_loader_test, macro_append_after_macro_definition)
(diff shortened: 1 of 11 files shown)


jasondellaluce:
> What about the special snowflake `enabled` override?

+1 for this.

Andreagit97:
Update: 

This PR deprecates 3 things now: 
1. `append` usage without `override`
2. `enabled` usage without `override`
3. `required_engine_version` not SemVer compatible

Summary for Falco 0.37.0 here: https://github.com/falcosecurity/falco/issues/2763

I have enabled all rules warnings by default, let me know if it is OK for you. This is a possible output example:
```
Thu Jan  4 18:10:03 2024: Falco version: 0.37.0-231+ad964c0 (x86_64)
Thu Jan  4 18:10:03 2024: Falco initialized with configuration file: ../falco.yaml
Thu Jan  4 18:10:03 2024: System info: Linux version 6.2.0-39-generic (buildd@lcy02-amd64-045) (x86_64-linux-gnu-gcc-11 (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0, GNU ld (GNU Binutils for Ubuntu) 2.38) #40~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Nov 16 10:53:04 UTC 2
Thu Jan  4 18:10:03 2024: Loading rules from file ../rules/falco-deprecated_rules.yaml
Thu Jan  4 18:10:03 2024: ../rules/falco-deprecated_rules.yaml: Ok, with warnings
1 Warnings:
In rules content: (../rules/falco-deprecated_rules.yaml:0:0)
    required_engine_version: (../rules/falco-deprecated_rules.yaml:27:2)
------
- required_engine_version: 17
  ^
------
LOAD_DEPRECATED_ITEM (Used deprecated item): The 'required_engine_version' should be SemVer compatible. All non-SemVer compatible values are deprecated.
Thu Jan  4 18:10:03 2024: Loading rules from file ../rules/falco-incubating_rules.yaml
Thu Jan  4 18:10:03 2024: Loading rules from file ../rules/falco-sandbox_rules.yaml
Thu Jan  4 18:10:03 2024: Loading rules from file ../rules/falco_rules.yaml
Thu Jan  4 18:10:03 2024: ../rules/falco_rules.yaml: Ok, with warnings
1 Warnings:
In rules content: (../rules/falco_rules.yaml:0:0)
    rule 'Fileless execution via memfd_create': (../rules/falco_rules.yaml:1252:2)
------
- rule: Fileless execution via memfd_create
  ^
------
LOAD_DEPRECATED_ITEM (Used deprecated item): 'append' key is deprecated. Add an 'append' entry (e.g. 'condition: append') under 'override' instead.
Thu Jan  4 18:10:03 2024: The chosen syscall buffer dimension is: 8388608 bytes (8 MBs)
Thu Jan  4 18:10:03 2024: Starting health webserver with threadiness 8, listening on 0.0.0.0:8765
Thu Jan  4 18:10:03 2024: Loaded event sources: syscall
Thu Jan  4 18:10:03 2024: Enabled event sources: syscall
Thu Jan  4 18:10:03 2024: Opening 'syscall' source with modern BPF probe.
Thu Jan  4 18:10:03 2024: One ring buffer every '2' CPUs.
```

I don't love the multi-line log but at the moment this is what we have... @jasondellaluce there was a particular reason to use `fprintf(stderr,...)` instead of `falco::log`?

Andreagit97:
probably we need to change the engine version of `falco-deprecated_rules.yaml` from `17` to `0.17.0` to avoid the warning, WDYT?

userspace/engine/falco_engine.h
```diff
@@ -74,15 +74,8 @@ class falco_engine
 	void list_fields(std::string &source, bool verbose, bool names_only, bool markdown) const;
 
 	//
-	// Load rules either directly or from a filename.
+	// Load rules and returns a result object.
 	//
-	void load_rules_file(const std::string &rules_filename, bool verbose, bool all_events);
```
Andreagit97, line 79: This and all the other methods were unused :/

userspace/falco/app/actions/load_rules_files.cpp
```diff
@@ -79,10 +79,9 @@ falco::app::run_result falco::app::actions::load_rules_files(falco::app::state&
 			break;
 		}
 
-		// If verbose is true, also print any warnings
-		if(s.options.verbose && res->has_warnings())
+		if(res->has_warnings())
 		{
-			fprintf(stderr, "%s\n", res->as_string(true, rc).c_str());
+			falco_logger::log(falco_logger::level::WARNING,res->as_string(true, rc) + "\n");
```
Andreagit97, line 84: not sure if we want to keep `fprintf(stderr,...` or use the default logger instead

incertum: default logger would say.

leogr:
+1 for me :+1:

Andreagit97:
updated the engine version for the deprecated rule file https://github.com/falcosecurity/rules/pull/216/files

Andreagit97:
Now this should be ready

FedeDP (dismissed):
/approve

userspace/engine/rule_loader_reader.cpp
```diff
@@ -435,27 +435,28 @@ static void read_item(
 		rule_loader::context ctx(item, rule_loader::context::LIST, name, parent);
 		rule_loader::list_info v(ctx);
 
-		bool append = false;
+		bool has_append_flag = false;
```
LucaGuerra, line 438: Why the variable name change? The reason why it was called `append` is because it is true if either the `append` flag is true or there is an append override set (see line 457), hence the wording `append` instead of `has_append_flag`.

userspace/engine/rule_loader_reader.cpp
```diff
@@ -475,29 +476,30 @@ static void read_item(
 		rule_loader::macro_info v(ctx);
 		v.name = name;
 
-		bool append = false;
+		bool has_append_flag = false;
```
LucaGuerra, line 479: Same comment as above

userspace/engine/rule_loader_collector.cpp
```diff
@@ -244,10 +240,7 @@ void rule_loader::collector::append(configuration& cfg, rule_update_info& info)
 {
 	auto prev = m_rule_infos.at(info.name);
 
-	THROW(!prev,
-	       // "Rule has 'append' key or an append override but no rule by that name already exists", // TODO replace with this and update testing
```
LucaGuerra, line 248: Thank you for updating these errors <3

unit_tests/engine/test_rule_loader.cpp
```diff
@@ -43,6 +44,66 @@ class engine_loader_test : public ::testing::Test {
...
+	{
+		if(!m_load_result->has_warnings())
+		{
+			return false;
+		}
+
+		for(auto &warn : m_load_result_json["warnings"])
+		{
+			std::string msg = warn["message"];
+			// Debug:
```
LucaGuerra, line 68: leftover :)

LucaGuerra:
Great job Andrea! 🎉 

Just a couple minor comments and I can approve :)

userspace/engine/rule_loader_reader.cpp
Andreagit97: before my PR `has_append_flag` was already used in the 'rule` case. I adopted it also in the `list` and `macro` case to have consistency, but no strong opinion about the naming

unit_tests/engine/test_rule_loader.cpp
Andreagit97: actually, this was intentional, in this way when we want to debug the errors or warnings we can simply uncomment the line below. AFAIK gtest has no cool features to enable logs only when the test fails, while this is the default for other languages like Rust :(

Andreagit97: i can remove it if you want, again no strong opinion here

userspace/engine/rule_loader_reader.cpp
LucaGuerra: There's a difference, in the rule case that variable means that there is an append flag set, as the override append behavior is different (and does not change the flag). I think it's clearer with the `append` name. Thank you :)

unit_tests/engine/test_rule_loader.cpp
LucaGuerra: It's ok, we can leave it here!

userspace/engine/rule_loader_reader.cpp
Andreagit97: done, thanks!

LucaGuerra (approved):
LGTM

FedeDP (approved):
/approve

PR #2997, opened 2024-01-08
cleanup!(cmake): dropped bundled plugins since falcoctl takes care of everything

FedeDP:
**What type of PR is this?**

/kind cleanup

/area build

Since:
* falcoctl takes care of installing plugins
* rpm/deb host installation requires further configuration anyway when wishing to run with plugins
* charts make use of falcoctl to install plugins

Completely drop plugins shipped in Falco bundle.
This way, we will have a single way to install and enable plugins (using falcoctl + updating the Falco config) for any environment where Falco is going to be installed.

Fixes #

```release-note
cleanup!(cmake): drop bundled plugins in Falco
```


FedeDP:
Will update https://github.com/falcosecurity/falco/issues/2763 issue once we all agree.

Andreagit97:
agree with this! but this is a breaking change so in case I would add a `!` and track it here https://github.com/falcosecurity/falco/issues/2763

FedeDP:
It is already there (at least in the release notes); will add it to title too.

.github/workflows/reusable_test_packages.yaml
```diff
@@ -43,6 +43,14 @@ jobs:
           tar -xvf $(ls falco-*.tar.gz)
           cd falco-${{ inputs.version }}-${{ inputs.arch }}
           sudo cp -r * /
+      
+      # Note: most probably the plugin related tests should be moved to the plugin repo sooner or later.
+      - name: Install needed artifacts using falcoctl
+        if: ${{ inputs.static == false }}
+        run: |
+          sudo mkdir -p /usr/share/falco/plugins
```
Andreagit97, line 51: so the e2e tests now look at plugins + rules under `/usr/share/falco/plugins` instead of relying on the ones bundled in the `tar.gz`. This is possible without modification because `/usr/share/falco/` and `/etc/falco` are the default paths, isn't it?

FedeDP: Yes

Andreagit97 (approved):
/approve

incertum (approved):
Truly thanks for the cleanup!

/approve

PR #3012, opened 2024-01-15
cleanup(falco.yaml): rename `none` in `nodriver`

Andreagit97:
**What type of PR is this?**

/kind cleanup

/area engine

This PR renames the `none` engine config into `state_only`. The idea is that `none` should mean no engine enabled, but this is not the case. Even though the `nodriver` is a particular engine, it opens an inspector without generating events. **So the idea is to rename the `none` into `state_only` and leave the `none` config for future use cases**. 

For example `none` should be used when we want to run a source plugin like `k8saudit` without opening a syscall engine. Today we need to do something like this:

```bash
falco -c /etc/falco.yaml -r k8saudit.rules --disable-source=syscall
```

In the future, if we add the config `none` we could simply specify `none` in the config file and we have done

```bash
falco -c /etc/falco.yaml -r k8saudit.rules
```

I renamed the config into `state_only` because, in the end, this is what the `nodriver` does, it opens an inspector and the thread table allowing plugins with syscall source to run.

I would like to have it in this release because if we do it now this is not a breaking change, if we wait Falco 0.38.0 it will become a breaking change

```release-note
NONE
```

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -179,7 +179,8 @@ rules_file:
 # - `modern_ebpf`: Modern eBPF (CO-RE eBPF probe)
 # - `gvisor`: gVisor (gVisor sandbox)
 # - `replay`: Replay a scap trace file
-# - `none`: No event producer loaded, useful to run with plugins.
+# - `nodriver`: No driver is injected into the system.
+#   This is useful to debug and to run plugins with 'syscall' source.
 #
 # Only one engine can be specified in the `kind` key.
 # Moreover, for each engine multiple options might be available,
--- a/userspace/falco/app/actions/load_config.cpp
+++ b/userspace/falco/app/actions/load_config.cpp
@@ -125,8 +125,8 @@ static falco::app::run_result apply_deprecated_options(falco::app::state& s)
 	}
 	if (s.options.nodriver)
 	{
-		falco_logger::log(falco_logger::level::WARNING, "DEPRECATION NOTICE: the '--nodriver' command line option is deprecated and will be removed in Falco 0.38! Set 'engine.kind: none' instead in falco.yaml\n");
-		s.config->m_engine_mode =  engine_kind_t::NONE;
+		falco_logger::log(falco_logger::level::WARNING, "DEPRECATION NOTICE: the '--nodriver' command line option is deprecated and will be removed in Falco 0.38! Set 'engine.kind: nodriver' instead in falco.yaml\n");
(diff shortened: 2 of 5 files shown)


leogr:
> The idea is that `none` should mean no engine enabled, but this is not the case
I agree with this, stating the syscall source is still enabled.


> Even though the `nodriver` is a particular engine, it opens an inspector without generating events
I also agree with this statement. Indeed, [nodriver](https://github.com/falcosecurity/libs/blob/master/userspace/libscap/engine/nodriver/nodriver.h) is really an engine, which however, does not connect o any [driver](https://github.com/falcosecurity/falco/tree/master/brand#driver) so it can't produce any _syscall_ event. However, other than just the state, other events might come into the flow (from container engine? plugins? I can't think of others, but there may be). 

In other words, the `nodriver` implementation is kind of _Null Object Pattern_ and it has been introduced mainly for testing and CI purposes, AFAIK.

> So the idea is to rename the none into `state_only` and leave the `none` config for future use cases.'
Although I understand `none` config should be reserved for the _no syscall data source at all_ use case, `state_only` seems too specific to me for the `nodriver` case.

Thus, I'd suggest:
- just use `nodriver`. I know we had an initial discussion regarding this that went in a different direction, but since the config node is now called `engine.kind`, `nodriver` seems appropriate. Basically, _no driver_ would mean _no kernel instrumentation_ thus the engine will produce no syscall event. That's consistent.
- `nulldriver`, `driverless`, etc.. can be valid alternatives.

I've highlighted this because it would be nice to choose a name without risking the need to change it again.

> I would like to have it in this release because if we do it now this is not a breaking change, if we wait Falco 0.38.0 it will become a breaking change

+1 better to change it now.

Andreagit97:
so you are suggesting renaming the actual `none` into `nodriver` since it means that there is an engine but not a driver, and in the future use `none` (or something similar) to indicate that there is no engine at all, is this true?

FedeDP:
> just use nodriver. I know we had an initial discussion regarding this that went in a different direction, but since the config node is now called engine.kind, nodriver seems appropriate. Basically, no driver would mean no kernel instrumentation thus the engine will produce no syscall event. That's consistent.

Agree, `nodriver` is far better than `state_only` and 1:1 matches the libscap engine name.

Andreagit97:
done!

leogr (approved):
+1

FedeDP (approved):
/approve

Issue #2910, opened 2023-11-14
[DISCUSSION] deprecate / remove old internal drop stats?


incertum:
@Andreagit97 opening a dedicated issue to discuss https://github.com/falcosecurity/libs/pull/1433#issuecomment-1807694658

> I saw that in Falco (and I imagine also in other consumers) we use the get_capture_stats method to obtain the number of drops/events. On the other hand, with stats_v2 we are using an agnostic approach, where the final consumer receives a vector of metrics already populated by sinsp. My question here is, do we want scap_stats_v2 to replace the old scap_stats? If yes, how do we obtain the specific number of drops/events from this agnostic approach? Do we want to keep these specific numbers or the final goal is to expose a set of metrics with a Prometheus endpoint?

First and foremost we are talking about https://github.com/falcosecurity/falco/blob/master/userspace/falco/event_drops.cpp aka `Falco internal: syscall event drop` that we will call "old drop stats" versus the new `metrics` Falco option that is also capable of creating an internal rule `Falco internal: metrics snapshot` containing not just the drop counters but also more metrics.

@leogr starting to summarize a few shortcomings of the old stats from my perspective. At the same time I would be honoring that some adopters prefer to keep the old stats around for longer. Therfore I would be fine keeping it, but also willing to help work out a transition plan.

Cons old drop stats:

- Old stats report the number of drops within a 1s time frame window which may be a less transparent or intuitive metric versus being able to define your deltas based on what you think has the most meaning (the new metrics framework snapshots the current drop counts at an interval you choose and subsequently you can derive your own deltas the way you prefer)
- Can generate an unpredictable high amount of logs
- By default only reports drops when drop percentage is above 10%
- Enabled by default
- Not intuitive to customize or even turn off other than adjusting a logging level for Falco rules

Pros old drop stats:
- Not bound to a fixed metrics interval to send the first alert of drops occurring

Andreagit97:
yeah we need to explore a little bit all the usages of these old drops stats in Falco, to understand if we really need them or if we can just replace them with the new ones

incertum:
When I asked on slack no one seems to be urgently still needing this. 

In the last 3+ debugging sessions I have been involved, we always found the newer `metrics` feature to provide more actionable insights.

Proposing to introduce a deprecation warning for Falco 0.38 or Falco 0.37 and then follow the formal deprecation cycle?
WDYT @falcosecurity/falco-maintainers?

[...]

PR #3015, opened 2024-01-16
update!(config): soft deprecation of drop stats counters in `syscall_event_drops`

incertum:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind release

> /area build

/area engine

> /area tests

> /area proposals

> /area CI

Let's discuss deprecating `syscall_event_drops`.
@falcosecurity/falco-maintainers 

https://github.com/falcosecurity/falco/issues/2910

Fixes #

```release-note
update!(config): soft deprecation of drop stats counters in `syscall_event_drops`
```

### DIFF
--- a/userspace/falco/app/actions/load_config.cpp
+++ b/userspace/falco/app/actions/load_config.cpp
@@ -41,6 +41,12 @@ static falco::app::run_result apply_deprecated_options(falco::app::state& s)
 		return run_result::fatal("You can not specify more than one of -e, -g (--gvisor-config), --modern-bpf, --nodriver, and the FALCO_BPF_PROBE env var");
 	}
 
+	if(s.config->m_min_priority == falco_common::PRIORITY_DEBUG)
+	{
+		falco_logger::log(falco_logger::level::WARNING,
+				"DEPRECATION NOTICE: 'syscall_event_drops' config is deprecated and will be removed in Falco 0.38! Use 'metrics' config instead. Note that the 'syscall_event_drops' config is enabled by default when the 'priority' is set to 'debug'. You can turn it off by setting the 'priority' to any higher level\n");
+	}
+
 	// Please note: is not possible to mix command line options and configs to obtain a configuration
 	// we need to use only one method. For example, is not possible to set the gvisor-config through
 	// the command line and the gvisor-root through the config file. For this reason, if we detect
--- a/userspace/falco/app/actions/load_config.cpp
+++ b/userspace/falco/app/actions/load_config.cpp
@@ -44,7 +44,7 @@ static falco::app::run_result apply_deprecated_options(falco::app::state& s)
 	if(s.config->m_min_priority == falco_common::PRIORITY_DEBUG)
 	{
 		falco_logger::log(falco_logger::level::WARNING,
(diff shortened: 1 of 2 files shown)


userspace/falco/app/actions/load_config.cpp
```diff
@@ -41,6 +41,12 @@ static falco::app::run_result apply_deprecated_options(falco::app::state& s)
 		return run_result::fatal("You can not specify more than one of -e, -g (--gvisor-config), --modern-bpf, --nodriver, and the FALCO_BPF_PROBE env var");
 	}
 
+	if(s.config->m_min_priority == falco_common::PRIORITY_DEBUG)
```
Andreagit97, line 44: Personally i had no time to check if we can really drop it, anyone willing to help @falcosecurity/falco-maintainers ?

leogr: Yes, it still is worth investigating a bit since it seems not really used. However, if we are in doubt, we can still postpone this deprecation to 0.38.0 (and remove it in 0.39.0). Let's see if we find any info.

userspace/falco/app/actions/load_config.cpp
```diff
@@ -41,6 +41,12 @@ static falco::app::run_result apply_deprecated_options(falco::app::state& s)
...
+	{
+		falco_logger::log(falco_logger::level::WARNING,
+				"DEPRECATION NOTICE: 'syscall_event_drops' config is deprecated and will be removed in Falco 0.38! Use 'metrics' config instead. Note that the 'syscall_event_drops' config is enabled by default when the 'priority' is set to 'debug'. You can turn it off by setting the 'priority' to any higher level\n");
```
Andreagit97, line 47:
```suggestion
				"DEPRECATION NOTICE: 'syscall_event_drops' config is deprecated and will be removed in Falco 0.38! If you rely on this config use instead 'metrics.output_rule' to monitor the number of drops. Note that the 'syscall_event_drops' config is enabled by default when the 'priority' is set to 'debug'. You can turn it off by setting the 'priority' to any higher level\n");
```

Andreagit97: I would try to highlight that `syscall_event_drops` is enabled by default but it probably not many users rely on it, they probably don't even know that is config is doing something under the hood

userspace/falco/app/actions/load_config.cpp
Andreagit97: Ok, I took some time to check it, I'm ok with deprecating this! If we agree I would love to do it in Falco 0.37.0 so that in the next dev cycle we can clean up the code!

If we are not sure, I propose to deprecate it now, and if we face some blockers when we try to remove it in the next dev-cycle we can expand the depreciation also for Falco 0.38.0, WDYT? btw I don't think this is the case, I don't see big blockers in removing it

Andreagit97:
if we go for the deprecation we need to add it to https://github.com/falcosecurity/falco/issues/2763

incertum:
+1 re @Andreagit97 assessment, the old drop drop stats also add an avoidable overhead to the current main processing loop. We have an opportunity to clean up the underlying code a lot and I think this will be great as well. In addition, we have help guides and additional documentation around the new metrics currently under review for the website. This is why I believe it could be a good point in time to start the deprecation cycle, at minimum the notice as @Andreagit97 mentioned.

userspace/falco/app/actions/load_config.cpp
incertum: I like this suggestion! Added.

leogr (approved):

PR #3017, opened 2024-01-16
update(CI): Publish wasm package as wasm-dev

Rohith-Raju:
**What type of PR is this?**

/kind feature

/area CI

This `PR` will help add a new feature to [falco-playground ](https://github.com/falcosecurity/falco-playground) where a user can select multiple versions of `falco`.  Since we don't support `wasm` "officially" it is a `dev` only package for now.  

**Which issue(s) this PR fixes**:  https://github.com/falcosecurity/falco-playground/issues/13

Fixes #

```release-note
update(CI): publish wasm package as dev-wasm 
```

### DIFF
--- a/scripts/publish-wasm
+++ b/scripts/publish-wasm
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+set -e
+
+usage() {
+    echo "usage: $0 -f <package.tar.gz>"
+    exit 1
+}
+
+# parse options
+while getopts ":f:" opt; do
+    case "${opt}" in
+        f )
+          file=${OPTARG}
+          ;;
+        \?)
+          echo "invalid option: ${OPTARG}" >&2
+          exit 1
+          ;;
+    esac
+done
+shift $((OPTIND-1))
+
+if [ -z "${file}" ]; then
+    usage
+fi
+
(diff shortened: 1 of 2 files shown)


leogr:
/assign @LucaGuerra 

as discussed :pray:

scripts/publish-wasm
```diff
@@ -0,0 +1,38 @@
...
+          ;;
+    esac
+done
+shift $((OPTIND-1))
+
+if [ -z "${file}" ]; then
+    usage
+fi
+
+repo="dev-wasm"
```
LucaGuerra, line 27: `wasm-dev` for consistency (sorry, I also always get that mixed up): https://download.falco.org/?prefix=packages/

Rohith-Raju:
> PR is still in draft, do you think there is anything missing?

Made it a draft to check my approach and permissions, I'll disable it now 

>Also just to double check, WASM does not have any other architecture right, it's like an architecture of its own with its own format? In that case I think the directory is right.

Correct!! Wasm is platform-independent. 

>Also permissions should already be OK on the S3 bucket to publish this.

Great!!

FedeDP (approved):
/approve

PR #3024, opened 2024-01-18
new(userspace): support split config files

### LINKED ISSUE #2880: Support a falco.local.yaml configuration file with local changes to support unattended upgrades
**Motivation**

I'm using the debian/ubuntu repository to install falco and upgrading falco on my servers. My `falco.yaml` contains local changes like the log level / priority and other things. Upgrading falco via apt always result in configuration file conflicts, as it contains local changes, and requires a manual merge.

**Feature**

There is already a `falco_rules.local.yaml`, what about having a `falco.local.yaml` that contains local configuration changes? This way, the upgrade of falco can be automated using unattended-upgrades without solving any configuration file conflicts.

**Alternatives**

Configuration options can also be provided using command-line arguments, but this means right now an additional systemd configuration or overwrite on systems using systemd. Not sure if there is a way to add custom command-line arguments to this.

**Additional context**

Not sure if this is already everything required to do automated upgrades, but it should at least support something like this, without any manual fixes:
`FALCO_FRONTEND=noninteractive apt-get install -y falco`

FedeDP:
**What type of PR is this?**

/kind feature

/area engine

This PR introduces support to split Falco config file into multiple files.
Given the length of the current Falco configuration, i think it would be great to be able to split the main Falco config file into multiple pieces (ie: perhaps one for each "logical section").

The implementation is as follows:
* only supports 1-level of depth (ie: included config files cannot have `configs_files` directives), to avoid recursion hells
* if specified filepath does not exist, it is silently skipped, just like we do for `rules_file`
* same as for `rules_file` directive, user can specify a folder in the `configs_files` list. The folder will be then processed in lexicographic order, same way as `rules_file`.
* Included config files CAN override existent key-value pairs
* Since `configs_files` directive is parsed right after parsing the main yaml node (ie: the Falco main config), its position does not matter; eg: if you have
a `foo.yaml` with
```
foo: foo
```
then, `foo` node is overridden even in this case:
```
configs_files: foo.yaml
foo: bar
```

* to fix #2880, we now install a `/etc/falco/config.d/` folder and that folder was added to default config file `configs_files` directive.

Fixes #2880 

```release-note
new(userspace): support split config files
```


FedeDP:
> This would require the aforementioned change to the current behavior, plus the need to avoid triggering an exception when an included file is missing (since the default Falco config would always have the includes on the local config, even when not present).

This is now done in latest commit, that also adds some more tests.

FedeDP:
WDYT @falcosecurity/falco-maintainers ?
I'd love to gather some feedback around this, this might be a good time since we are just past the 0.37.0 release.

incertum:
Thanks for the bump, I'll take a look!

incertum (commented):
LGTM, great work @FedeDP. Couldn't think of any missing tests really.

Q: Could we include the new "includes" key in the `falco.yaml` with a description? Else end user may not be aware of this new option. In addition, would it be possible to allow an empty `includes:` or would we need to fully comment it out?

incertum:
What if I pass includes over the cmd options `-o`, should be all fine right? 
Plus could we not allow self referencing the original `falco.yaml` in the includes. Maybe all of this is already addressed, just double-checking.

FedeDP:
> Could we include the new "includes" key in the falco.yaml with a description

Yes! I'd love to do that in a follow up PR, where i also split up main config file into many smaller sections eg:
- engine.yml
- outputs.yml
- ...

WDYT? Didn't want to do that in this PR to avoid increasing its size and reduce readability.

> In addition, would it be possible to allow an empty includes:

It should be already possible, but a test on this would be great indeed! I will add it!

> What if I pass includes over the cmd options -o, should be all fine right?

Neat! Didn't think of that, will test and report back!

> Plus could we not allow self referencing the original falco.yaml in the includes.

Neat too! I will add a test on this!

FedeDP:
> In addition, would it be possible to allow an empty includes:

Added a test; it was already working fine.

> Plus could we not allow self referencing the original falco.yaml in the includes.

Needed a fix; added a test too!
Thanks for the valuable input!

leogr:
> EDIT: i was wrong, we actually support overriding values like: `-o "load_plugins[0]=k8saudit" -o "load_plugins[1]=json"` but the `includes` are loaded too early (ie: as soon as the yaml config is loaded) and therefore `-o` is not taken into account.

So, what happens when the user tries `-o includes....` ? Does it get an error or silently fail?

FedeDP:
> Does it get an error or silently fail?

Silently fail. I think it must be documented in a follow-up PR (ie: when i split main Falco config into different section-specific configs), above the `includes:` node in the main Falco config.
After all it makes sense if you see it as a "preprocessor directive".

leogr:
> > Does it get an error or silently fail?
> 
> Silently fail. I think it must be documented in a follow-up PR (ie: when i split main Falco config into different section-specific configs), above the `includes:` node in the main Falco config. After all it makes sense if you see it as a "preprocessor directive".

I agree it makes sense. However, it should not be _silent_, in my opinion.

FedeDP:
We can add a warning perhaps?

leogr:
> We can add a warning perhaps?

Yes, that's exactly what I meant.

FedeDP:
Implemented.

incertum:
Agree with entire discussion!

incertum (dismissed):
/approve

Agree with proceeding in another PR @FedeDP, this is a great kick off of config files cleanups!

Andreagit97 (commented):
Some comments about the test. I will come back at this after 0.37.1 with some ideas, for example, a possible directory like the rules one `/etc/falco/config.d`

FedeDP:
Reached an agreement on the final design for the feature:
- [ ] split test in multiple tests
- [ ] support including directories; configs contained will be parsed in lexicographically order

leogr:
> * [ ]  split test in multiple tests

:+1: 

> * [ ]  support including directories; configs contained will be parsed in lexicographically order

:+1: :+1: :+1: (pls make sure loading ordering work as same as `rules_file`, for UX consistency :pray: )

> * [ ]  rename `includes` to `configs_file` to be consistent with https://github.com/falcosecurity/falco/blob/master/falco.yaml#L154 (? WDYT @incertum @Andreagit97 ? my 2c, this is different; `rules_file` is specifying a list of rule files to be loaded, while `includes` is specifying a list of config files to be included)

I like it, but not totally :innocent: 

 `rules_file` has always been wrong (for clear reasons), but it has been preserved to avoid breaking change. I believe this is a good chance to fix it, so I have a counter-proposal:

- (in another PR) add `rules_files` config option, and make `rules_file` (the old one) an alias to `rules_files`. This will improve name consistency and avoid any breaking changes here (I've strong opinions on avoiding breaking changes on this - we may just think about removing the alias when reaching Falco 1.0, but not before for sure).
- (this PR) just use `configs_files`


Also, an optional point: I think it would be a good idea to add `/etc/falco/config.d` by default as well.

FedeDP:
> Also, an optional point: I think it would be a good idea to add /etc/falco/config.d by default as well.

Agree! Adding it to the TODO.
> (this PR) just use configs_files

I have no strong opinions, if @Andreagit97 and @incertum all agree, we can go with this.

incertum:
> > Also, an optional point: I think it would be a good idea to add /etc/falco/config.d by default as well.
> 
> Agree! Adding it to the TODO.
> 
> > (this PR) just use configs_files
> 
> I have no strong opinions, if @Andreagit97 and @incertum all agree, we can go with this.

`configs_files` is good, also like the idea of changing `rules_file` to `rules_files` and keeping an alias for backward compatibility.

unit_tests/falco/test_configuration.cpp
```diff
@@ -109,6 +109,216 @@ TEST(Configuration, modify_yaml_fields)
...
+
+	/*
+	 * Test that when main config file is in a different folder,
+	 * other config files are searched relative to its folder,
+	 * when they are specified as relative paths
+	 */
+	const auto temp_main = std::filesystem::temp_directory_path() / "main.yaml";
+	const std::string main_conf_yaml_relative_absolute =
+		"includes:\n"
+		"  - conf_2.yaml\n"
```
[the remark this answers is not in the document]
FedeDP: Indeed, there is a warning:
```
if (std::filesystem::exists(include_file_path) && std::filesystem::is_regular_file(include_file_path))
			{
				...
                        }
			else
			{
				falco_logger::log(falco_logger::level::WARNING, "Included config file unexistent or wrong type: " + include_file_path.string());
			}
```

unit_tests/falco/test_configuration.cpp
```diff
@@ -109,6 +109,216 @@ TEST(Configuration, modify_yaml_fields)
...
+		"    id: 1\n"
+		"    name: foo\n";
+	outfile.open("main.yaml");
+	outfile << main_conf_yaml_include_itself;
+	outfile.close();
+
+	ASSERT_ANY_THROW(conf.load_from_file("main.yaml"));
+
+	// Cleanup everything
+	std::filesystem::remove("main.yaml");
```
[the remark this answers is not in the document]
FedeDP: Unfortunately no :( There is https://en.cppreference.com/w/cpp/io/c/tmpfile, but:
> If the program terminates abnormally, it is implementation-defined if these temporary files are deleted.

unit_tests/falco/test_configuration.cpp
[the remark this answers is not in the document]
FedeDP: This test triggers:
> Fri Apr  5 16:36:34 2024: Included config file unexistent or wrong type: /tmp/conf_2.yaml

falco.yaml
```diff
@@ -114,6 +114,25 @@
 # disabling the automatic artifacts followed by falcoctl.
 
 
+#######################
```
FedeDP, line 117: Please take a moment to deeply review this comment section and eventually fix / improve it.

userspace/falco/yaml_helper.h
```diff
@@ -77,22 +78,83 @@ class yaml_visitor {
...
+					for (auto const& dir_entry : std::filesystem::directory_iterator(include_file_path, it_options))
+					{
+						if (std::filesystem::is_regular_file(dir_entry.path()))
+						{
+							v.push_back(dir_entry.path().string());
+						}
+						// We don't support nested directories
+						else
+						{
+							falco_logger::log(falco_logger::level::WARNING, "Included config file has wrong type: " + dir_entry.path().string());
```
FedeDP, line 139: Am not sure whether this should be warning or debug messages, actually.

userspace/falco/yaml_helper.h
Andreagit97, line 1: Probably I would add 2 things:
1. When Falco starts it can log all the config files it has evaluated
2. A command line option or a debug log in which we print the final resulting configuration after all the overrides + the files considered in the process (if we implement step 1, we should be fine)

FedeDP: I agree with 1! Going to implement it asap.

FedeDP: > When Falco starts it can log all the config files it has evaluated

I'd say this should be a debug log btw

FedeDP: Done!

unit_tests/falco/test_configuration.cpp
```diff
@@ -136,7 +138,7 @@ TEST(Configuration, configuration_config_files_secondary_fail)
 	outfile.close();
 
 	yaml_helper conf;
-	ASSERT_ANY_THROW(conf.load_from_file("main.yaml"));
+	ASSERT_ANY_THROW(conf.load_from_file("main.yaml", loaded_conf_files));
```
FedeDP, line 141: `conf.load_from_file` takes care of emptying the vector.

userspace/falco/yaml_helper.h
```diff
@@ -77,22 +78,84 @@ class yaml_visitor {
...
```
FedeDP, line 140: NOTE: since we are logging BEFORE the config is actually parsed, we will always log at WARNING level even if user sets log level to `error`.
Otherwise we could return a list of warnings and let the caller manage them. WDYT?

userspace/falco/yaml_helper.h
FedeDP: Indeed, i decided to modify the existing INFO log to highlight all loaded config files:
```
Mon Apr  8 15:22:09 2024: Falco initialized with configuration files:
Mon Apr  8 15:22:09 2024:    ../falco.yaml
```

userspace/falco/app/actions/load_config.cpp
```diff
@@ -29,11 +29,15 @@ static falco::app::run_result apply_deprecated_options(const falco::app::state&
 
 falco::app::run_result falco::app::actions::load_config(const falco::app::state& s)
 {
+	// List of loaded conf files, ie: s.options.conf_filename
+	// plus all the `configs_files` expanded list of configs.
+	std::vector<std::string> loaded_conf_files;
```
FedeDP, line 34: Capture expanded list of loaded conf files + any warning, then print all of them.

userspace/falco/yaml_helper.h
FedeDP: >  Otherwise we could return a list of warnings and let the caller manage them.

Done in latest commit. This is the same thing we do with the ruleset so i guessed that's ok!

unit_tests/falco/test_configuration.cpp
```diff
@@ -37,6 +37,9 @@ static std::string sample_yaml =
 	"    - elem2\n"
 	"    - elem3\n";
 
+static std::vector<std::string> loaded_conf_files;
+static std::vector<std::string> loaded_conf_warnings;
```
Andreagit97, line 41: cool, now that we have a warnings vector we can assert that these tests correctly throw the right warning where needed, WDYT?

userspace/falco/yaml_helper.h
Andreagit97: Thank you! I still think that a command line option/debug log with the merged config would be really useful for us and for our users to debug possible issues

unit_tests/falco/test_configuration.cpp
FedeDP: I'd just assert the expected length (not the actual strings), so that they are more resilient to string changes.

FedeDP:
> Thank you! I still think that a command line option/debug log with the merged config would be really useful for us and for our users to debug possible issues

@Andreagit97 done in latest commit! I discovered that we already had a `--support` cmdline option that already printed the config file. I extended it to:
* print the expanded config file
* print the parsed yaml structure instead of the whole (as in, read from file) file

Example output:
```
./userspace/falco/falco --support -c ../falco.yaml  -r ../rules/falco_rules.yaml  | jq -r '.config' 
Tue Apr  9 11:10:08 2024: Falco version: 0.38.0-176+aa02401 (x86_64)
Tue Apr  9 11:10:08 2024: Falco initialized with configuration files:
Tue Apr  9 11:10:08 2024:    ../falco.yaml
Tue Apr  9 11:10:08 2024: Included config entry unexistent: /etc/falco/config.d
Tue Apr  9 11:10:08 2024: System info: Linux version 6.5.0-26-generic (buildd@lcy02-amd64-070) (x86_64-linux-gnu-gcc-13 (Ubuntu 13.2.0-4ubuntu3) 13.2.0, GNU ld (GNU Binutils for Ubuntu) 2.41) #26-Ubuntu SMP PREEMPT_DYNAMIC Tue Mar  5 21:19:28 UTC 2024
Tue Apr  9 11:10:08 2024: Loading rules from file ../rules/falco_rules.yaml
{"configs_files": ["/etc/falco/config.d"], "rules_file": ["/etc/falco/falco_rules.yaml", "/etc/falco/falco_rules.local.yaml", "/etc/falco/rules.d"], "engine": {"kind": "kmod", "kmod": {"buf_size_preset": "4", "drop_failed_exit": "false"}, "ebpf": {"probe": "/home/federico/.falco/falco-bpf.o", "buf_size_preset": "4", "drop_failed_exit": "false"}, "modern_ebpf": {"cpus_for_each_buffer": "2", "buf_size_preset": "4", "drop_failed_exit": "false"}, "replay": {"capture_file": ""}, "gvisor": {"config": "", "root": ""}}, "load_plugins": [], "plugins": [{"name": "k8saudit", "library_path": "libk8saudit.so", "init_config": null, "open_params": "http://:9765/k8s-audit"}, {"name": "cloudtrail", "library_path": "libcloudtrail.so"}, {"name": "json", "library_path": "libjson.so"}], "watch_config_files": "true", "time_format_iso_8601": "false", "priority": "debug", "json_output": "false", "json_include_output_property": "true", "json_include_tags_property": "true", "buffered_outputs": "false", "rule_matching": "first", "outputs_queue": {"capacity": "0"}, "stdout_output": {"enabled": "true"}, "syslog_output": {"enabled": "true"}, "file_output": {"enabled": "false", "keep_alive": "false", "filename": "./events.txt"}, "http_output": {"enabled": "false", "url": "http://some.url", "user_agent": "falcosecurity/falco", "insecure": "false", "ca_cert": "", "ca_bundle": "", "ca_path": "/etc/ssl/certs", "mtls": "false", "client_cert": "/etc/ssl/certs/client.crt", "client_key": "/etc/ssl/certs/client.key", "echo": "false", "compress_uploads": "false", "keep_alive": "false"}, "program_output": {"enabled": "false", "keep_alive": "false", "program": "jq '{text: .output}' | curl -d @- -X POST https://hooks.slack.com/services/XXX"}, "grpc_output": {"enabled": "false"}, "grpc": {"enabled": "false", "bind_address": "unix:///run/falco/falco.sock", "threadiness": "0"}, "webserver": {"enabled": "true", "threadiness": "0", "listen_port": "8765", "listen_address": "0.0.0.0", "k8s_healthz_endpoint": "/healthz", "ssl_enabled": "false", "ssl_certificate": "/etc/falco/falco.pem"}, "log_stderr": "true", "log_syslog": "true", "log_level": "info", "libs_logger": {"enabled": "false", "severity": "debug"}, "output_timeout": "2000", "syscall_event_timeouts": {"max_consecutives": "1000"}, "syscall_event_drops": {"threshold": ".1", "actions": ["log", "alert"], "rate": ".03333", "max_burst": "1", "simulate_drops": "false"}, "metrics": {"enabled": "false", "interval": "1h", "output_rule": "true", "resource_utilization_enabled": "true", "state_counters_enabled": "true", "kernel_event_counters_enabled": "true", "libbpf_stats_enabled": "true", "convert_memory_to_mb": "true", "include_empty_values": "false"}, "base_syscalls": {"custom_set": [], "repair": "false"}, "falco_libs": {"thread_table_size": "262144"}}
```
Adding another `jq` to parse the `.config` json string:
```
./userspace/falco/falco --support -c ../falco.yaml  -r ../rules/falco_rules.yaml  | jq -r '.config' | jq 
Tue Apr  9 11:10:33 2024: Falco version: 0.38.0-176+aa02401 (x86_64)
Tue Apr  9 11:10:33 2024: Falco initialized with configuration files:
Tue Apr  9 11:10:33 2024:    ../falco.yaml
Tue Apr  9 11:10:33 2024: Included config entry unexistent: /etc/falco/config.d
Tue Apr  9 11:10:33 2024: System info: Linux version 6.5.0-26-generic (buildd@lcy02-amd64-070) (x86_64-linux-gnu-gcc-13 (Ubuntu 13.2.0-4ubuntu3) 13.2.0, GNU ld (GNU Binutils for Ubuntu) 2.41) #26-Ubuntu SMP PREEMPT_DYNAMIC Tue Mar  5 21:19:28 UTC 2024
Tue Apr  9 11:10:33 2024: Loading rules from file ../rules/falco_rules.yaml
{
  "configs_files": [
    "/etc/falco/config.d"
  ],
  "rules_file": [
    "/etc/falco/falco_rules.yaml",
    "/etc/falco/falco_rules.local.yaml",
    "/etc/falco/rules.d"
  ],
  "engine": {
    "kind": "kmod",
    "kmod": {
      "buf_size_preset": "4",
      "drop_failed_exit": "false"
    },
    "ebpf": {
      "probe": "/home/federico/.falco/falco-bpf.o",
      "buf_size_preset": "4",
      "drop_failed_exit": "false"
    },
    "modern_ebpf": {
      "cpus_for_each_buffer": "2",
      "buf_size_preset": "4",
      "drop_failed_exit": "false"
    },
    "replay": {
      "capture_file": ""
    },
    "gvisor": {
      "config": "",
      "root": ""
    }
  },
  "load_plugins": [],
  "plugins": [
    {
      "name": "k8saudit",
      "library_path": "libk8saudit.so",
      "init_config": null,
      "open_params": "http://:9765/k8s-audit"
    },
    {
      "name": "cloudtrail",
      "library_path": "libcloudtrail.so"
    },
    {
      "name": "json",
      "library_path": "libjson.so"
    }
  ],
  "watch_config_files": "true",
  "time_format_iso_8601": "false",
  "priority": "debug",
  "json_output": "false",
  "json_include_output_property": "true",
  "json_include_tags_property": "true",
  "buffered_outputs": "false",
  "rule_matching": "first",
  "outputs_queue": {
    "capacity": "0"
  },
  "stdout_output": {
    "enabled": "true"
  },
  "syslog_output": {
    "enabled": "true"
  },
  "file_output": {
    "enabled": "false",
    "keep_alive": "false",
    "filename": "./events.txt"
  },
  "http_output": {
    "enabled": "false",
    "url": "http://some.url",
    "user_agent": "falcosecurity/falco",
    "insecure": "false",
    "ca_cert": "",
    "ca_bundle": "",
    "ca_path": "/etc/ssl/certs",
    "mtls": "false",
    "client_cert": "/etc/ssl/certs/client.crt",
    "client_key": "/etc/ssl/certs/client.key",
    "echo": "false",
    "compress_uploads": "false",
    "keep_alive": "false"
  },
  "program_output": {
    "enabled": "false",
    "keep_alive": "false",
    "program": "jq '{text: .output}' | curl -d @- -X POST https://hooks.slack.com/services/XXX"
  },
  "grpc_output": {
    "enabled": "false"
  },
  "grpc": {
    "enabled": "false",
    "bind_address": "unix:///run/falco/falco.sock",
    "threadiness": "0"
  },
  "webserver": {
    "enabled": "true",
    "threadiness": "0",
    "listen_port": "8765",
    "listen_address": "0.0.0.0",
    "k8s_healthz_endpoint": "/healthz",
    "ssl_enabled": "false",
    "ssl_certificate": "/etc/falco/falco.pem"
  },
  "log_stderr": "true",
  "log_syslog": "true",
  "log_level": "info",
  "libs_logger": {
    "enabled": "false",
    "severity": "debug"
  },
  "output_timeout": "2000",
  "syscall_event_timeouts": {
    "max_consecutives": "1000"
  },
  "syscall_event_drops": {
    "threshold": ".1",
    "actions": [
      "log",
      "alert"
    ],
    "rate": ".03333",
    "max_burst": "1",
    "simulate_drops": "false"
  },
  "metrics": {
    "enabled": "false",
    "interval": "1h",
    "output_rule": "true",
    "resource_utilization_enabled": "true",
    "state_counters_enabled": "true",
    "kernel_event_counters_enabled": "true",
    "libbpf_stats_enabled": "true",
    "convert_memory_to_mb": "true",
    "include_empty_values": "false"
  },
  "base_syscalls": {
    "custom_set": [],
    "repair": "false"
  },
  "falco_libs": {
    "thread_table_size": "262144"
  }
}
```

falco.yaml
```diff
@@ -114,6 +114,25 @@
...
+
+# [Stable] `configs_files`
+#
+# Falco will load additional configs files specified here.
+# Their loading is assumed to be made *after* main config file has been processed,
+# exactly in the order they are specified.
+# Therefore, loaded config files *can* override values from main config file.
+# Also, nested include is not allowed, ie: included config files won't be able to include other config files.
+# When relative path is used, the config file is assumed to be in the same folder as the main config file.
```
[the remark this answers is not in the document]
FedeDP: I don't think it is the same behavior since for rules files there is no a "main" rule file concept.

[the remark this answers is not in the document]
FedeDP: No check is done while reading rules files:
```
const std::string &filename = *it;
		std::ifstream is;
		is.open(filename);
		if (!is.is_open())
		{
			throw falco_exception("Could not open file " + filename + " for reading");
		}
```

Still, i think it makes much sense for configs files considering that here we actually have a `main` rule file concept.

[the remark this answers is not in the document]
FedeDP: Also, `rules_file` throws an exception when a file cannot be found. Instead, `configs_files` just gives a warning. Shall we throw an exception for configs files too?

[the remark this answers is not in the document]
leogr: 1. I agree it must throw an exception when a config file is not found.
2. So, using a relative path in `rules_file` will make Falco use the current working dir, on the other hand, using a relative path in `configs_files` will make Falco use the main config path as the base path. Is my understanding correct?
  - TBH, I would have completely disallowed the ability to use relative path in such context. Anyway, I don't have strong opinions.
  - At least, please document the actual behavior for both cases :pray:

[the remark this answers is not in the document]
FedeDP: > So, using a relative path in rules_file will make Falco use the current working dir, on the other hand, using a relative path in configs_files will make Falco use the main config path as the base path. Is my understanding correct?

Yep, otherwise we can also let configs files manage relative paths as just that, relative paths to cwd and we don't need to document anything because that's the expected behavior indeed :)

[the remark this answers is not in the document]
FedeDP: Ok doing so 
> I agree it must throw an exception when a config file is not found.

Allows us to also drop the `warnings` parameter, which leads to a cleaner API :) 
I pushed a commit that:
* throws an exception when file is not found
* silently ignores non-regular files (and non directory)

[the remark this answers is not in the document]
FedeDP: Updated PR body with final design: https://github.com/falcosecurity/falco/pull/3024#issue-2088412184

leogr (approved):

FedeDP:
>  CI Build / check-engine-checksum (pull_request) Failing after 6s 

What

leogr:
> > CI Build / check-engine-checksum (pull_request) Failing after 6s
> 
> What

```
encoded checksum: 12cb01b5d2379547819a37a4656f1cef2952893ea1225de79614c673f386f645
current checksum: -
current engine checksum differs from the one encoded in userspace/engine/falco_engine_version.h
```
:thinking:

FedeDP:
Rebased on top of new master :)

leogr (approved):

FedeDP:
Still broken :/ 
https://github.com/falcosecurity/falco/actions/runs/8618044652/job/23619985439?pr=3024

Given that it is not a required job, i'd avoid fighting with it rn; let's see if it disappears once in master.

FedeDP:
Found the mistery:
```
echo $(build/userspace/falco/falco -c ./falco.yaml --version | grep 'Engine:' | awk '{print $2}') $(echo $(build/userspace/falco/falco -c ./falco.yaml --version | grep 'Schema version:' | awk '{print $3}') $(build/userspace/falco/falco -c ./falco.yaml --list --markdown | grep '^`' | sort) $(build/userspace/falco/falco -c ./falco.yaml --list-events | sort) | sha256sum)
Cannot read config file (./falco.yaml): Included config entry not existent: /etc/falco/config.d
Error: std::exception
Cannot read config file (./falco.yaml): Included config entry not existent: /etc/falco/config.d
Error: std::exception
Cannot read config file (./falco.yaml): Included config entry not existent: /etc/falco/config.d
Error: std::exception
Cannot read config file (./falco.yaml): Included config entry not existent: /etc/falco/config.d
Error: std::exception
01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b -
```
:D

FedeDP:
> I think i'll adapt the code to do the same for configs_files.

Adapted! Also, moved core logic from `yaml_helper` to `falco_configuration` class, and updated tests accordingly.
This allowed to enable falco configs files parsing and merging from cmdline option.

FedeDP:
This is now finally ready, also added a test to check that we are able to include configs from cmdline.

leogr (approved):

Andreagit97 (commented):
LGTM!

Just one thing we need to document is that the reload doesn't work on secondary config files so if we update a secondary config file Falco won't restart with the new configuration

Andreagit97 (approved):

PR #3028, opened 2024-01-22
fix(userspace/engine): avoid storing escaped strings in engine

jasondellaluce:
**What type of PR is this?**

/kind bug

/area engine

When the engine compiles list infos, we have a small bug for which we store forcefully-quoted strings in the engine's data structures instead of quoting them only at the time of their usage (at list refs resolution time). After investigating, this doesn't seem to affect the runtime-evaluated rules data structures, but it's clearly visible with the `-L` functionality and reproduced on list items that are YAML-escaped or contain spaces.

Example:

```yaml
# rules_sample.yaml
- list: my_list
  items: [non_escaped_val, "escaped val"]
```

```
$ falco -L -o json_output=true -r rules_sample | jq .
{
  "lists": [
    {
      "details": {
        "items_compiled": [
          "non_escaped_val",
          "\"escaped val\""
        ],
        "lists": [
          "escaped val"
        ],
        "plugins": [],
        "used": false
      },
      "info": {
        "items": [
          "non_escaped_val"
        ],
        "name": "my_list"
      }
    }
  ],
  "macros": [],
  "required_engine_version": "0.0.0",
  "required_plugin_versions": [],
  "rules": []
}
```

Here, `"escaped val"` is interpreted as a list reference and is not properly escaped when printed as a value.

With these changes, the same example now correctly prints as:
```
{
  "lists": [
    {
      "details": {
        "items_compiled": [
          "non_escaped_val",
          "escaped val"
        ],
        "lists": [],
        "plugins": [],
        "used": false
      },
      "info": {
        "items": [
          "non_escaped_val",
          "escaped val"
        ],
        "name": "my_list"
      }
    }
  ],
  "macros": [],
  "required_engine_version": "0.0.0",
  "required_plugin_versions": [],
  "rules": []
}
```

```release-note
fix(userspace/engine): avoid storing escaped strings in engine defs
```

### DIFF
--- a/unit_tests/engine/test_rule_loader.cpp
+++ b/unit_tests/engine/test_rule_loader.cpp
@@ -941,3 +941,30 @@ TEST_F(engine_loader_test, required_engine_version_invalid)
 	ASSERT_FALSE(load_rules(rules_content, "rules.yaml"));
 	ASSERT_TRUE(check_error_message("Unable to parse engine version"));
 }
+
+// checks for issue described in https://github.com/falcosecurity/falco/pull/3028
+TEST_F(engine_loader_test, list_value_with_escaping)
+{
+    std::string rules_content = R"END(
+- list: my_list
+  items: [non_escaped_val, "escaped val"]
+)END";
+
+	ASSERT_TRUE(load_rules(rules_content, "rules.yaml"));
+	ASSERT_TRUE(m_load_result->successful());
+  ASSERT_TRUE(m_load_result->has_warnings()); // a warning for the unused list
+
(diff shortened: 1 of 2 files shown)


Andreagit97 (commented):
LGTM! Can we add a simple unit test like the one you used in the PR description?

jasondellaluce:
> LGTM! Can we add a simple unit test like the one you used in the PR description?

Definitely! Just added one.

Andreagit97 (approved):
/approve

LucaGuerra (approved):

PR #3071, opened 2024-02-10
new(config): add `falco_libs.thread_table_size`

incertum:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind release

> /area build

/area engine

> /area tests

> /area proposals

> /area CI

We recently increased the default threadtable size to accommodate modern infrastructures better as we have observed drops related to a full internal thread table. This PR makes this setting configurable by the end user.

Fixes #

```release-note
new(config): add `falco_libs.thread_table_size`
```

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -65,6 +65,8 @@
 #     metrics
 # Falco performance tuning (advanced)
 #     base_syscalls
+# Falco libs
+#     falco_libs
 
 ################################
 # Falco command-line arguments #
@@ -1099,6 +1101,29 @@ metrics:
 base_syscalls:
   custom_set: []
   repair: false
+
+##############
+# Falco libs #
+##############
+
+# [Experimental] `falco_libs` - Potentially subject to more frequent changes
+#
+# `thread_table_size`
+#
+# Set the maximum number of entries (the absolute maximum value can only be MAX UINT32) 
+# for Falco's internal threadtable (process cache). Please note that Falco operates at a 
+# granular level, focusing on individual threads. Falco rules reference the thread leader 
+# as the process. The size of the threadtable should typically be much higher than the 
+# number of currently alive processes. The default value should work well on modern 
+# infrastructures and be sufficient to absorb bursts.
+#
+# Reducing its size can help in better memory management, but as a consequence, your 
+# process tree may be more frequently disrupted due to missing threads. You can explore 
+# `metrics.state_counters_enabled` to measure how the internal state handling is performing, 
+# and the fields called `n_drops_full_threadtable` or `n_store_evts_drops` will inform you
+# if you should increase this value for optimal performance.
+falco_libs:
+  thread_table_size: 262144
 
 # [Stable] Guidance for Kubernetes container engine command-line args settings
 #
--- a/userspace/engine/falco_common.h
+++ b/userspace/engine/falco_common.h
@@ -28,6 +28,8 @@ limitations under the License.
 //
 #define DEFAULT_OUTPUTS_QUEUE_CAPACITY_UNBOUNDED_MAX_LONG_VALUE std::ptrdiff_t(~size_t(0) / 2)
 
+#define DEFAULT_FALCO_LIBS_THREAD_TABLE_SIZE 262144
+
 //
 // Most falco_* classes can throw exceptions. Unless directly related
 // to low-level failures like inability to open file, etc, they will
--- a/userspace/falco/configuration.cpp
+++ b/userspace/falco/configuration.cpp
@@ -70,6 +70,7 @@ falco_configuration::falco_configuration():
 	m_syscall_evt_drop_max_burst(1),
 	m_syscall_evt_simulate_drops(false),
 	m_syscall_evt_timeout_max_consecutives(1000),
+	m_falco_libs_thread_table_size(DEFAULT_FALCO_LIBS_THREAD_TABLE_SIZE),
 	m_base_syscalls_repair(false),
 	m_metrics_enabled(false),
 	m_metrics_interval_str("5000"),
@@ -442,6 +443,8 @@ void falco_configuration::load_yaml(const std::string& config_name, const yaml_h
 	{
 		throw std::logic_error("Error reading config file(" + config_name + "): the maximum consecutive timeouts without an event must be an unsigned integer > 0");
 	}
+
+	m_falco_libs_thread_table_size = config.get_scalar<std::uint32_t>("falco_libs.thread_table_size", DEFAULT_FALCO_LIBS_THREAD_TABLE_SIZE);
 
 	m_base_syscalls_custom_set.clear();
 	config.get_sequence<std::unordered_set<std::string>>(m_base_syscalls_custom_set, std::string("base_syscalls.custom_set"));
--- a/userspace/falco/configuration.h
+++ b/userspace/falco/configuration.h
@@ -136,6 +136,8 @@ public:
 
 	uint32_t m_syscall_evt_timeout_max_consecutives;
 
+	uint32_t m_falco_libs_thread_table_size;
+
 	// User supplied base_syscalls, overrides any Falco state engine enforcement.
 	std::unordered_set<std::string> m_base_syscalls_custom_set;
 	bool m_base_syscalls_repair;
--- a/userspace/falco/app/actions/helpers_inspector.cpp
+++ b/userspace/falco/app/actions/helpers_inspector.cpp
@@ -53,6 +53,12 @@ falco::app::run_result falco::app::actions::open_live_inspector(
 			inspector->set_sinsp_stats_v2_enabled();
 		}
 
+		if(s.config->m_falco_libs_thread_table_size > 0)
+		{
+			// Default value is set in libs as part of the sinsp_thread_manager setup
+			inspector->m_thread_manager->set_max_thread_table_size(s.config->m_falco_libs_thread_table_size);
+		}
+
 		if (source != falco_common::syscall_source) /* Plugin engine */
 		{
 			for (const auto& p: inspector->get_plugin_manager()->plugins())


falco.yaml
```diff
@@ -1262,6 +1263,23 @@ base_syscalls:
...
+# number of currently alive processes. The default value should work well on modern 
+# infrastructures and be sufficient to absorb bursts.
+#
+# Reducing its size can help in better memory management, but as a consequence, your 
+# process tree may be more frequently disrupted due to missing threads. You can explore 
+# `metrics.state_counters_enabled` to measure how the internal state handling is performing, 
+# and the fields called `n_drops_full_threadtable` or `n_store_evts_drops` will inform you
+# if you should increase this value for optimal performance.
+state_engine:
+  thread_table_size: 262144
```
incertum, line 1281: @FedeDP and @jasondellaluce since you reviewed the libs changes. Thank you!

userspace/falco/configuration.cpp
```diff
@@ -63,6 +63,7 @@ falco_configuration::falco_configuration():
 	m_syscall_evt_drop_max_burst(1),
 	m_syscall_evt_simulate_drops(false),
 	m_syscall_evt_timeout_max_consecutives(1000),
+	m_state_engine_thread_table_size(262144),
```
FedeDP, line 66: Can we have a macro for the default value? Given it is a "weird" number used twice :)

incertum: Agreed, funny enough if you search that number in the Linux kernel it is not so odd.

Before I re-push, was wondering about the Windows build failure. is there something in libs that needs to be adjusted?

leogr (commented):
One thought after a quick look.

I would like to group some top-level config keys instead of continuously adding new ones.

I mean, the top-level config keys are growing exponentially, and this may not be ideal for users approaching the configuration.

incertum:
> One thought after a quick look.
> 
> I would like to group some top-level config keys instead of continuously adding new ones.
> 
> I mean, the top-level config keys are growing exponentially, and this may not be ideal for users approaching the configuration.

What would it be? It seems that we don't have a new top level category for `state_engine` related configs that may grow over time?

Andreagit97:
maybe we could add a key called `libsinsp` or `internals` that in the future will collect all the libs configurations... for example, we already have the `libs_logger` config that we could move behind this new config (not now of course) WDYT?

FedeDP:
@Andreagit97 that's a great idea actually, even if it is exposing some "internal knowledge" about who owns what.
I'd rather call it just `falco-libs:`.

incertum:
> @Andreagit97 that's a great idea actually, even if it is exposing some "internal knowledge" about who owns what. I'd rather call it just `falco-libs:`.

`falco_libs` with underscore instead? Perhaps in the future the `libs_logger` can be moved under `falco_libs` as well. It should be less of a breaking change as likely few folks use that setting.

Additional question: Do we know what's wrong with the Windows build?

userspace/falco/configuration.cpp
FedeDP: > was wondering about the Windows build failure. is there something in libs that needs to be adjusted?

Don't really know why it is complaining :/ 
```
jsoncpp.lib(json_value.obj) : error LNK2019: unresolved external symbol __imp_modf referenced in function "public: bool __cdecl Json::Value::isInt64(void)const " (?isInt64@Value@Json@@QEBA_NXZ) [D:\a\falco\falco\build\libsinsp\examples\sinsp-example.vcxproj]
jsoncpp.lib(json_writer.obj) : error LNK2019: unresolved external symbol __imp__finite referenced in function "class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > __cdecl Json::`anonymous namespace'::valueToString(double,bool,unsigned int,enum Json::PrecisionType)" (?valueToString@?A0x7fc2ca38@Json@@YA?AV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@N_NIW4PrecisionType@2@@Z) [D:\a\falco\falco\build\libsinsp\examples\sinsp-example.vcxproj]
jsoncpp.lib(json_writer.obj) : error LNK2019: unresolved external symbol __imp__isnan referenced in function "class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > __cdecl Json::`anonymous namespace'::valueToString(double,bool,unsigned int,enum Json::PrecisionType)" (?valueToString@?A0x7fc2ca38@Json@@YA?AV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@N_NIW4PrecisionType@2@@Z) [D:\a\falco\falco\build\libsinsp\examples\sinsp-example.vcxproj]
```

Seems a libs issue on jsoncpp; perhaps triggered by: https://github.com/falcosecurity/libs/commit/a89fd5d510fac5b5f2418f263f8d2c4ec80410ef

cc @geraldcombs can you take a look? :pray:  thanks!!

Andreagit97:
> falco_libs with underscore instead?

Yep, we need to use the underscore

userspace/falco/configuration.cpp
LucaGuerra: Perhaps you could write it as `0x40000` so it's easier to understand. Not sure if you can do the same in yaml as well.

leogr:
In general, I wouldn't say I like the idea of exposing internal names to the users (because internals can change, but the external function remains the same). 

So perhaps something like `libs_internals` (to make it explicitly, these are internal settings and may potentially change very often) 

However, `falco_libs` or just `libs` is still for me if the consensus goes toward this.

userspace/falco/configuration.cpp
geraldcombs: > Seems a libs issue on jsoncpp

It might be because libs is linking with the dynamic CRT (MultiThreadedDLL) instead of the static one (MultiThreaded). If that's the case we should be able to fix it here by passing `-DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreaded` to CMake in the `build-win32-package` job, similar to the `build-libs-others-amd64` job in libs.

Alternatively we could try to fix it in libs, e.g. by trying to override CMAKE_MSVC_RUNTIME_LIBRARY in CMakeLists.txt.

incertum: @LucaGuerra internally yes we can default to 0x40000 in the macro, but don't you think for adopters that is even more weird than a normal number? 🙃

geraldcombs: > @geraldcombs whichever you think is the best approach. I can bump libs here to the new latest after we address it in libs if we choose that path.

I opened a libs PR (1695) to link to the static CRT by default. That seemed to make the most sense, since the project favors static builds.

geraldcombs:
> @geraldcombs using the latest libs commit and also rebased w/ falco master, still having the windows failure. Could you take another look? Thank you!

I'm able to replicate the issue here. If I make the following change here
```
--- a/.github/workflows/reusable_build_packages.yaml
+++ b/.github/workflows/reusable_build_packages.yaml
@@ -233,7 +233,7 @@ jobs:
         run: |
           mkdir build
           cd build
-          cmake -DCMAKE_BUILD_TYPE=Release -DMINIMAL_BUILD=On -DUSE_BUNDLED_DEPS=On -DBUILD_FALCO_UNIT_TESTS=On -DFALCO_VERSION=${{ inputs.version }} ..
+          cmake -DCMAKE_BUILD_TYPE=Release -DMINIMAL_BUILD=On -DUSE_BUNDLED_DEPS=On -DBUILD_FALCO_UNIT_TESTS=On -DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreaded -DFALCO_VERSION=${{ inputs.version }} ..
```
I can build and run `falco_unit_tests.exe`, although I'm not sure why that's necessary. If I build a standalone version of libs here without setting `CMAKE_MSVC_RUNTIME_LIBRARY` I get a static build.

incertum:
Thanks @geraldcombs I'll then rebase, thanks!

incertum:
> > @geraldcombs using the latest libs commit and also rebased w/ falco master, still having the windows failure. Could you take another look? Thank you!
> 
> I'm able to replicate the issue here. If I make the following change here
> 
> ```
> --- a/.github/workflows/reusable_build_packages.yaml
> +++ b/.github/workflows/reusable_build_packages.yaml
> @@ -233,7 +233,7 @@ jobs:
>          run: |
>            mkdir build
>            cd build
> -          cmake -DCMAKE_BUILD_TYPE=Release -DMINIMAL_BUILD=On -DUSE_BUNDLED_DEPS=On -DBUILD_FALCO_UNIT_TESTS=On -DFALCO_VERSION=${{ inputs.version }} ..
> +          cmake -DCMAKE_BUILD_TYPE=Release -DMINIMAL_BUILD=On -DUSE_BUNDLED_DEPS=On -DBUILD_FALCO_UNIT_TESTS=On -DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreaded -DFALCO_VERSION=${{ inputs.version }} ..
> ```
> 
> I can build and run `falco_unit_tests.exe`, although I'm not sure why that's necessary. If I build a standalone version of libs here without setting `CMAKE_MSVC_RUNTIME_LIBRARY` I get a static build.

@geraldcombs just double-checking if I interpret this correctly. You or someone else will open a PR and I shall rebase afterwards? Thanks in advance for clarifying!

incertum:
@jasondellaluce already bumped libs before hence I could remove it here. However, the CI still seems to be in a not good state (unrelated to these changes).

userspace/engine/falco_common.h
```diff
@@ -28,6 +28,8 @@ limitations under the License.
 //
 #define DEFAULT_OUTPUTS_QUEUE_CAPACITY_UNBOUNDED_MAX_LONG_VALUE std::ptrdiff_t(~size_t(0) / 2)
 
+#define DEFAULT_FALCO_LIBS_INTERNALS_THREAD_TABLE_SIZE 0x4000
```
jasondellaluce, line 31: Shouldn't this be `0x40000`? Seems like this is a quite smaller number

jasondellaluce (commented):
@incertum I have no major concerns here if not the naming decision, as pointed out above. I'm pretty much onboard with the suggestions of the other reviewers.

incertum:
> @incertum I have no major concerns here if not the naming decision, as pointed out above. I'm pretty much onboard with the suggestions of the other reviewers.

@jasondellaluce Could you elaborate? I tried to pick a name that addresses all feedback. If this should not be it, what should the final name be? No strong opinion.

leogr:
> > > @incertum I have no major concerns here if not the naming decision, as pointed out above. I'm pretty much onboard with the suggestions of the other reviewers.
> > 
> > 
> > @jasondellaluce Could you elaborate? I tried to pick a name that addresses all feedback. If this should not be it, what should the final name be? No strong opinion.
> 
> I think I'd go with either `falco_libs` or `libs_internals`. But again, just my personal taste!

After thinking about this again, I believe `falco_libs` can be nice, especially if we start distributing Falco's libs at some point in the future.

incertum:
SGTM, just changed it. Thanks.

FedeDP (approved):
/approve

userspace/falco/configuration.h
```diff
@@ -136,6 +136,8 @@ class falco_configuration
 
 	uint32_t m_syscall_evt_timeout_max_consecutives;
 
+	std::uint32_t m_falco_libs_thread_table_size;
```
Andreagit97, line 139:
```suggestion
	uint32_t m_falco_libs_thread_table_size;
```

Andreagit97 (commented):
a minor nit and then we can go! thank you!

leogr (approved):

Andreagit97 (approved):
/approve

PR #3083, opened 2024-02-13
Improve const correctness

### LINKED ISSUE #3082: Fix const correctness

sgaist:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind release

> /area build

/area engine

> /area tests

> /area proposals

> /area CI

This PR refactors the code to make use of const references rather than just reference or copy when applicable.
It also applies similar recommendations for pointers.

They were detected by cppcheck.

Fixes #3082

```release-note
NONE
```


userspace/engine/falco_common.cpp
```diff
@@ -33,7 +33,7 @@ static std::vector<std::string> rule_matching_names = {
 	"all"
 };
 
-bool falco_common::parse_priority(std::string v, priority_type& out)
+bool falco_common::parse_priority(const std::string& v, priority_type& out)
```
incertum, line 36: @federico-sysdig may we summon you again to help decide on a style more uniformly? For example have gotten feedback both ways, pass by value vs const ref ... and lately we transitioned to using std::string_view in libs. Thank you 🙏 .

userspace/engine/formats.cpp
```diff
@@ -128,7 +128,7 @@ std::string falco_formats::format_event(sinsp_evt *evt, const std::string &rule,
 		line = full_line;
 	}
 
-	return line.c_str();
+	return line;
```
incertum, line 131: ohhh

userspace/falco/falco_outputs.h
```diff
@@ -58,17 +58,17 @@ class falco_outputs
...
 
 	/*!
 		\brief Format then send a generic message to all outputs.
 		Not necessarily associated with any event.
 	*/
 	void handle_msg(uint64_t now,
 			falco_common::priority_type priority,
-			std::string &msg,
-			std::string &rule,
+			const std::string &msg,
```
incertum, line 70: Perhaps worth double-checking on msg

userspace/engine/falco_common.cpp
federico-sysdig: TL;DR:
Pass by value all types that are cheap to copy (e.g. built-in types, but not only...) and pass by const reference the other ones.

There are other cases where you'll here advice to pass-by-value and then move or pass-by-rvalue-reference. I won't discuss them here. I think the matter gets more confused and difficult and a simple rule that is good enough is a superior choice. So, these cases apart a, `std::string` should not be passed by value as it might not be "cheap" to copy, meaning it might require a heap allocation (unless it's very short, but small-string-optimization is another topic I'd leave out of the discussion).

Regarding `std::string_view`, the type is not built-in, clearly, but it's implemented as a pair of pointers (or a pointer and an integer, which is the same), thus it is cheap to copy and should be passed by value. See it as a more structured way to pass `(char* str, int strlen)`, like we often do in C.

"Style" might come into play on the choice between `const std::string&` or `std::string_view`, but the latter came only recently with the library moving to C++17, so you won't see it very much. It is more efficient at times when the caller of the function uses a string literal as no temporary string gets created, so it is a better choice which I fully endorse, but, beware, it cannot be always used for passing the string buffer to a C function as it gives no guarantee of being null-terminated and it has some gotchas if you use it with an associative container whose key is a `std::string` (e.g. `std::map<std::string, widget> C;` and then, later on, `C.find(mystringview)`).

Briefly put, passing strings by const reference is fine, passing them through a `std::string_view` is even better, if you can afford it.

LucaGuerra (approved):
Thank you!

FedeDP:
Uh needs a rebase :)

FedeDP (approved):
/approve

incertum (dismissed):
/approve

incertum:
CI seems to be in a bad state right now all over, we will check tomorrow.

sgaist:
@incertum it's innocent (in this case), I broke the test with too much constness.

sgaist:
I may have found something that could be refactored in a followup PR.

In [falco_engine::describe_rule](https://github.com/falcosecurity/falco/blob/792bcdca18ab422cc70a6dc20822239814910b60/userspace/engine/falco_engine.cpp#L489), does an empty pointer and an empty string represent the same thing ? I would say yes since it would not make sense to have a rule indexed at an empty string but it also does not seem to be forbidden.

incertum:
> I may have found something that could be refactored in a followup PR.
> 
> In [falco_engine::describe_rule](https://github.com/falcosecurity/falco/blob/792bcdca18ab422cc70a6dc20822239814910b60/userspace/engine/falco_engine.cpp#L489), does an empty pointer and an empty string represent the same thing ? I would say yes since it would not make sense to have a rule indexed at an empty string but it also does not seem to be forbidden.

@jasondellaluce and @mstemm, would you have any insights since you worked a lot on those parts?

incertum (approved):
/approve

FedeDP (approved):
/approve

PR #3092, opened 2024-02-15
new(build): add options to use bundled dependencies

mrgian:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind release

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

/area build

> /area engine

> /area tests

> /area proposals

/area CI

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #3107, opened 2024-02-21
refactor: Use FetchContent for integrating three bundled libs

federico-sysdig:
**What type of PR is this?**
/kind cleanup

/area build

Libraries `nlohmann-json`, `yaml-cpp`, and `cpp-httplib` in their "bundled" configuration are now integrated through the CMake construct `FetchContent` rather than `ExternalProject_Add`. This has the benefit of being treated in a way similar to integrating the library with `find_package`, thus making the bundled/non-bundled approach similar in the rest of the code.

One important beneficial consequence is the simplification in client code: new targets can reference `falco_engine` without having to specify these libs in their include path or explicitly link against it, because the target `falco_engine` brings all the needed properties on board through the usage of aliases such as `httplib::httplib` or `nlohmann_json::nlohmann_json`.
The libraries have also been upgraded as newer versions provide the target aliases and allow integration with `FetchContent` without building the library's test targets, which obviously Falco doesn't need.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/falco/webserver.h
+++ b/userspace/falco/webserver.h
@@ -16,13 +16,14 @@ limitations under the License.
 */
 #pragma once
 
-#define CPPHTTPLIB_OPENSSL_SUPPORT
-#define CPPHTTPLIB_ZLIB_SUPPORT
-#include <httplib.h>
-#include <thread>
-#include <memory>
-#include <libsinsp/sinsp.h>
 #include "configuration.h"
+
+#include <libsinsp/sinsp.h>
+
+#include <httplib.h>
+
+#include <memory>
+#include <thread>
 
 class falco_webserver
 {
--- a/cmake/modules/cpp-httplib.cmake
+++ b/cmake/modules/cpp-httplib.cmake
@@ -12,32 +12,15 @@
 # specific language governing permissions and limitations under the License.
 #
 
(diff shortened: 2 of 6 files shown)


FedeDP (dismissed):
/approve

leogr:
This will likely conflict with https://github.com/falcosecurity/falco/pull/3092

I suggest to merge #3092 first, then this. 

/hold

userspace/falco/webserver.h
```diff
@@ -16,13 +16,14 @@ limitations under the License.
 */
 #pragma once
 
-#define CPPHTTPLIB_OPENSSL_SUPPORT
-#define CPPHTTPLIB_ZLIB_SUPPORT
```
federico-sysdig, line 20: These two defines were resulting in a warning that the definition was already present.

FedeDP:
Ehy @federico-sysdig can you rebase? It should fix CI issues!

federico-sysdig:
> Ehy @federico-sysdig can you rebase? It should fix CI issues!

@FedeDP Sure. Done!

leogr (approved):
cc @jasondellaluce

jasondellaluce (approved):

PR #3108, opened 2024-02-22
fix(userspace/engine): make sure exception fields are not optional in replace mode

jasondellaluce:
**What type of PR is this?**

/kind bug

/area engine

When replaced with the new `override` YAML construct, `exceptions` must not have any of their sub-fields as optional, just like when first defined in regular rules definitions.

```release-note
fix(userspace/engine): make sure exception fields are not optional in replace mode
```

### DIFF
--- a/unit_tests/engine/test_rule_loader.cpp
+++ b/unit_tests/engine/test_rule_loader.cpp
@@ -712,7 +712,7 @@ TEST_F(test_falco_engine, rule_override_exceptions_required_fields)
 )END";
 
 	ASSERT_FALSE(load_rules(rules_content, "rules.yaml"));
-  ASSERT_FALSE(has_warnings());
+	ASSERT_FALSE(has_warnings());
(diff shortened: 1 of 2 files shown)


LucaGuerra:
Could you please add a test for this behavior? This way we won't break it in the future.

unit_tests/engine/test_rule_loader.cpp
```diff
@@ -680,6 +680,42 @@ TEST_F(test_falco_engine, rule_override_with_enabled)
...
+  exceptions:
+    - name: test_exception
+      values: [id]
+  override:
+    exceptions: replace
+)END";
+
+	ASSERT_FALSE(load_rules(rules_content, "rules.yaml"));
+  ASSERT_FALSE(has_warnings());
+	ASSERT_TRUE(check_error_message("Item has no mapping for key 'fields'")) << m_load_result_json.dump();
```
LucaGuerra, line 716: lint? otherwise looks good, thanks!

LucaGuerra (approved):

FedeDP (approved):
/approve

PR #3118, opened 2024-02-27
fix(ci): enforce bundled deps OFF in build-dev CI

FedeDP:
**What type of PR is this?**

/kind bug

/area CI

Fixes #

```release-note
NONE
```


leogr (approved):
As privately discussed with @FedeDP this is workaround is needed otherwise the CI will fail (with all bundled deps, it would take more than 1 hour, then fail for no space left on the disk).

LucaGuerra (approved):

PR #3133, opened 2024-03-08
new(engine): add warning when appending an exception with no values

### LINKED ISSUE #2354: Falco validation does not catch an exception that is not defined correctly
**Describe the bug**

If a falco rule is appending an exception and defines the exception with a property of `value`, it passes falco validation even though it is not valid.

**How to reproduce it**
The following test file can be used:
testCase.yaml
```
- rule: Test rule with exception append
  desc: A shell was used as the entrypoint/exec point into a container with an attached terminal.
  condition: >
    evt.type=setuid
  output: >
    Test output
  exceptions:
  - name: image_suffix
    fields: [container.image.repository]
    comps: [endswith]
    values:
    - [test]
  priority: NOTICE
  tags: [container]
  
- rule: Test rule with exception append
  exceptions:
    - name: image_suffix
      value: secure-inline-scan
  append: true
```

Run `falco -V testCase.yaml`

Falco responds that this file is ok.

**Expected behaviour**
An error message should be returned that the rule is not valid.

- Falco version:
Falco version: 0.33.1
Libs version:  0.9.2
Plugin API:    2.0.0
Driver:
  API version:    2.0.0
  Schema version: 2.0.0
  Default driver: 3.0.1+driver

- System info:
Falco version: 0.33.1
Libs version:  0.9.2
Plugin API:    2.0.0
Driver:
  API version:    2.0.0
  Schema version: 2.0.0
  Default driver: 3.0.1+driver

- OS:
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"

- Kernel:
Linux 678de6959a21 5.15.79.1-microsoft-standard-WSL2 #1 SMP Wed Nov 23 01:01:46 UTC 2022 x86_64 GNU/Linux
- Installation method:
used docker image `falcosecurity/falco;latest`

**Additional context**

mrgian:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind release

> /area build

/area engine

/area tests

> /area proposals

> /area CI

Adds a warning when a rule appends exceptions with no values.
Can be useful when an appended rule has a miswritten field.

**Which issue(s) this PR fixes**:  #2354 

Fixes #2354 

```release-note
new(engine): add warning when appending an exception with no values
```


jasondellaluce (dismissed):

unit_tests/engine/test_rule_loader.cpp
```diff
@@ -903,4 +903,32 @@ TEST_F(test_falco_engine, list_name_invalid)
 
   ASSERT_FALSE(load_rules(rules_content, "rules.yaml"));
   ASSERT_TRUE(check_error_message("List has an invalid name. List names must match a regular expression"));
+}
+
+// The appended exception has a purposely miswritten field (value),
+// simulating a typo or an incorrect usage.
+TEST_F(test_falco_engine, exceptions_append_no_values)
```
LucaGuerra, line 910: Should this behavior be the same when using `overrides`? Can we test it?

mrgian: The warning is emitted also when using `override`.
Even though the underlying logic is the same I agree on adding an additional test case using `override`.

We should also change the text of the warning, since the current one can be misleading when using `override: replace`.
(`values` field is not appended but replaced with an empty one)

WDYT?

LucaGuerra: Yes please, that additional test is appreciated

LucaGuerra (approved):

FedeDP (approved):
/approve

PR #3140, opened 2024-03-15
feat(webserver): implement metrics endpoint

### LINKED ISSUE #1772: Brainstorming: integrate Prometheus /metrics handler in Falco
**Motivation**

Since [falco-exporter]() is a great tool to expose prometheus metrics through gRPC, I think there are some caveats to use it. Our (w/ @developer-guy @f9n) some motivations to create this issue is that the following ones:

* **To reduce unnecessary gRPC communication**: we have to enable [gRPC output feature](https://falco.org/docs/grpc/#configuration) in the configuration. [^1]
* **To do not manage mTLS certs**: to connect gRPC over unix socket, we have to manage additionally [CA Certs](https://falco.org/docs/grpc/#certificates) as described under [connection options](https://github.com/falcosecurity/falco-exporter#connection-options)
* **To manage one single helm chart**: there are two different helm charts that we have to manage and maintain in our infrastructure, which are [falco](https://github.com/falcosecurity/charts/tree/master/falco) and [falco-exporter](https://github.com/falcosecurity/charts/tree/master/falco-exporter)
* **To reduce complexity**: simplicity is always better, to easily see the metrics in the Grafana, requires additional work to implement in current design
* **To make monitoring easier on local**: we can instantly check the metrics using `/metrics` endpoint in some troubleshooting scenarios 
```bash
$ kubectl port-forward svc/falco 8756
$ curl localhost:8756/metrics
```

_I think that implementing a Prometheus metrics in C++ would not be as easy as it looks like. Here is an example [metric server](https://github.com/fluent/fluent-bit/blob/master/src/http_server/api/v1/metrics.c) that currently using in [fluentbit](https://github.com/fluent/fluent-bit)._

**Feature**

This feature already clearly proposed in #421 and #530 and implemented in the projects.

**Alternatives**

* https://github.com/falcosecurity/falco-exporter
* https://github.com/falcosecurity/falcosidekick#prometheus

**Additional context**

By throwing this issue actually does NOT mean that _we should achieve the falco-exporter and provide this metrics in only falco by built-in_. What I want to say mostly are the following ones to clarify and understand the design:
* How we can make this process simpler by removing unnecessary dependencies?
* What is the correct way to do?
* Why did not falco provide `/metrics` endpoint in the first place?
* What is the motivation of behind this design: separating metrics module into another project?

Waiting your feedback!

[^1]: https://github.com/falcosecurity/falco-exporter#prerequisites

### LINKED ISSUE #3131: [TRACKING] Expand existing `webserver` to serve Prometheus `/metrics` endpoint
Expand existing `webserver` to serve Prometheus `/metrics` endpoint. 

The initial Prometheus tracking ticket lives here: https://github.com/falcosecurity/cncf-green-review-testing/issues/12.

---

Work Items:

- [x] `libs` metrics refactor supporting text-based Prometheus exposition format https://github.com/falcosecurity/libs/pull/1652 @incertum 
- [x] Sync `libs` referencing the new `libs_metrics_collector`, defer any other refactors (read below) https://github.com/falcosecurity/falco/pull/3129 @incertum 
- [ ] Expand existing `webserver` to serve Prometheus `/metrics` endpoint. TODO @sgaist 
- [ ] Given the new expanded metrics scope in falco, the current approach within `stats_writer::collector::collect` is no good anymore and we shall refactor it, also in anticipation to add yet another metrics category for rules counters. TODO @sgaist

sgaist:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind release

> /area build

/area engine

> /area tests

> /area proposals

> /area CI

This PR implements the `/metrics` endpoint in the webserver providing currently prometheus metrics.

Fixes #3131
Fixes #1772

```release-note
feat(webserver): a metrics endpoint has been added providing prometheus metrics.

It can be optionally enabled using the new `metrics.prometheus_enabled` configuration option.
It will only be activated if the `metrics.enabled` is true as well.
```

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -982,6 +982,7 @@ metrics:
   libbpf_stats_enabled: true
   convert_memory_to_mb: true
   include_empty_values: false
+  prometheus_enabled: false
 
 #######################################
 # Falco performance tuning (advanced) #
--- a/falco.yaml
+++ b/falco.yaml
@@ -695,6 +695,7 @@ webserver:
   # Can be an IPV4 or IPV6 address, defaults to IPV4
   listen_address: 0.0.0.0
   k8s_healthz_endpoint: /healthz
+  metrics_enabled: false
   ssl_enabled: false
   ssl_certificate: /etc/falco/falco.pem
 
@@ -982,7 +983,6 @@ metrics:
   libbpf_stats_enabled: true
   convert_memory_to_mb: true
   include_empty_values: false
-  prometheus_enabled: false
 
 #######################################
 # Falco performance tuning (advanced) #
--- a/falco.yaml
+++ b/falco.yaml
@@ -695,7 +695,9 @@ webserver:
   # Can be an IPV4 or IPV6 address, defaults to IPV4
   listen_address: 0.0.0.0
   k8s_healthz_endpoint: /healthz
-  metrics_enabled: false
+  # Enable the metrics endpoint providing Prometheus values
+  # It will only have an effect if metrics.enabled is set to true as well.
+  prometheus_metrics_enabled: false
   ssl_enabled: false
   ssl_certificate: /etc/falco/falco.pem
 
--- a/falco.yaml
+++ b/falco.yaml
@@ -970,7 +970,9 @@ syscall_event_drops:
 # beneficial for exploring the data schema and ensuring that fields with empty
 # values are included in the output.
 #
-# todo: prometheus export option
+# If metrics are enabled, the web server can be configured to activate the
+# corresponding Prometheus endpoint using webserver.prometheus_metrics_enabled.
+#
 # todo: syscall_counters_enabled option
 metrics:
   enabled: false
--- a/userspace/falco/app/actions/start_webserver.cpp
+++ b/userspace/falco/app/actions/start_webserver.cpp
@@ -45,7 +45,7 @@ falco::app::run_result falco::app::actions::start_webserver(falco::app::state& s
 			+ ssl_option + "\n");
 
 		std::vector<libs::metrics::libs_metrics_collector> metrics_collectors;
-		if (s.config->m_metrics_prometheus_enabled && s.config->m_metrics_prometheus_enabled)
+		if (s.config->m_metrics_enabled && s.config->m_metrics_prometheus_enabled)
 		{
 			for (const auto& source_info: s.source_infos)
 			{
--- a/userspace/falco/app/actions/start_webserver.cpp
+++ b/userspace/falco/app/actions/start_webserver.cpp
@@ -45,7 +45,7 @@ falco::app::run_result falco::app::actions::start_webserver(falco::app::state& s
 			+ ssl_option + "\n");
 
 		std::vector<libs::metrics::libs_metrics_collector> metrics_collectors;
-		if (s.config->m_metrics_enabled && s.config->m_metrics_prometheus_enabled)
+		if (s.config->m_metrics_enabled && s.config->m_webserver_metrics_enabled)
 		{
 			for (const auto& source_info: s.source_infos)
 			{
--- a/userspace/falco/app/actions/start_webserver.cpp
+++ b/userspace/falco/app/actions/start_webserver.cpp
@@ -19,6 +19,7 @@ limitations under the License.
 
 #if !defined(_WIN32) && !defined(__EMSCRIPTEN__) && !defined(MINIMAL_BUILD)
 #include "webserver.h"
+#include "falco_metrics.h"
 #endif
 
 using namespace falco::app;
@@ -45,19 +46,12 @@ falco::app::run_result falco::app::actions::start_webserver(falco::app::state& s
 			+ std::to_string(webserver_config.m_listen_port)
 			+ ssl_option + "\n");
 
-		std::vector<libs::metrics::libs_metrics_collector> metrics_collectors;
-		if (state.config->m_metrics_enabled && webserver_config.m_metrics_enabled)
-		{
-			for (const auto& source_info: state.source_infos)
-			{
-				metrics_collectors.push_back(libs::metrics::libs_metrics_collector(source_info.inspector.get(), state.config->m_metrics_flags));
-			}
-		}
+		falco_metrics metrics(state);
 
 		state.webserver.start(
 			state.offline_inspector,
-			metrics_collectors,
-			webserver_config);
+			webserver_config,
+			metrics);
 	}
 #endif
 	return run_result::ok();
--- a/userspace/falco/app/actions/start_webserver.cpp
+++ b/userspace/falco/app/actions/start_webserver.cpp
@@ -44,8 +44,18 @@ falco::app::run_result falco::app::actions::start_webserver(falco::app::state& s
 			+ std::to_string(s.config->m_webserver_listen_port)
 			+ ssl_option + "\n");
 
+		std::vector<libs::metrics::libs_metrics_collector> metrics_collectors;
+		if (s.config->m_metrics_prometheus_enabled && s.config->m_metrics_prometheus_enabled)
+		{
+			for (const auto& source_info: s.source_infos)
+			{
+				metrics_collectors.push_back(libs::metrics::libs_metrics_collector(source_info.inspector.get(), s.config->m_metrics_flags));
+			}
+		}
+
 		s.webserver.start(
 			s.offline_inspector,
+			metrics_collectors,
 			s.config->m_webserver_threadiness,
 			s.config->m_webserver_listen_port, 
 			s.config->m_webserver_listen_address,
--- a/userspace/falco/app/actions/start_webserver.cpp
+++ b/userspace/falco/app/actions/start_webserver.cpp
@@ -24,61 +24,57 @@ limitations under the License.
 using namespace falco::app;
 using namespace falco::app::actions;
 
-falco::app::run_result falco::app::actions::start_webserver(falco::app::state& s)
+falco::app::run_result falco::app::actions::start_webserver(falco::app::state& state)
 {
 #if !defined(_WIN32) && !defined(__EMSCRIPTEN__) && !defined(MINIMAL_BUILD)
-	if(!s.is_capture_mode() && s.config->m_webserver_enabled)
+	if(!state.is_capture_mode() && state.config->m_webserver_enabled)
 	{
-		if (s.options.dry_run)
+		if (state.options.dry_run)
 		{
 			falco_logger::log(falco_logger::level::DEBUG, "Skipping starting webserver in dry-run\n");
 			return run_result::ok();
 		}
 	
-		std::string ssl_option = (s.config->m_webserver_ssl_enabled ? " (SSL)" : "");
+		falco_configuration::webserver_config webserver_config = state.config->m_webserver_config;
+		std::string ssl_option = (webserver_config.m_ssl_enabled ? " (SSL)" : "");
 		falco_logger::log(falco_logger::level::INFO, "Starting health webserver with threadiness "
-			+ std::to_string(s.config->m_webserver_threadiness)
+			+ std::to_string(webserver_config.m_threadiness)
 			+ ", listening on "
-			+ s.config->m_webserver_listen_address
+			+ webserver_config.m_listen_address
 			+ ":"
-			+ std::to_string(s.config->m_webserver_listen_port)
+			+ std::to_string(webserver_config.m_listen_port)
 			+ ssl_option + "\n");
 
 		std::vector<libs::metrics::libs_metrics_collector> metrics_collectors;
-		if (s.config->m_metrics_enabled && s.config->m_webserver_metrics_enabled)
+		if (state.config->m_metrics_enabled && webserver_config.m_metrics_enabled)
 		{
-			for (const auto& source_info: s.source_infos)
+			for (const auto& source_info: state.source_infos)
 			{
-				metrics_collectors.push_back(libs::metrics::libs_metrics_collector(source_info.inspector.get(), s.config->m_metrics_flags));
+				metrics_collectors.push_back(libs::metrics::libs_metrics_collector(source_info.inspector.get(), state.config->m_metrics_flags));
 			}
 		}
 
-		s.webserver.start(
-			s.offline_inspector,
+		state.webserver.start(
+			state.offline_inspector,
 			metrics_collectors,
-			s.config->m_webserver_threadiness,
-			s.config->m_webserver_listen_port, 
-			s.config->m_webserver_listen_address,
-			s.config->m_webserver_k8s_healthz_endpoint,
-			s.config->m_webserver_ssl_certificate, 
-			s.config->m_webserver_ssl_enabled);
+			webserver_config);
 	}
 #endif
 	return run_result::ok();
 }
 
-falco::app::run_result falco::app::actions::stop_webserver(falco::app::state& s)
+falco::app::run_result falco::app::actions::stop_webserver(falco::app::state& state)
 {
 #if !defined(_WIN32) && !defined(__EMSCRIPTEN__) && !defined(MINIMAL_BUILD)
-	if(!s.is_capture_mode() && s.config->m_webserver_enabled)
+	if(!state.is_capture_mode() && state.config->m_webserver_enabled)
 	{
-		if (s.options.dry_run)
+		if (state.options.dry_run)
 		{
 			falco_logger::log(falco_logger::level::DEBUG, "Skipping stopping webserver in dry-run\n");
 			return run_result::ok();
 		}
 
-		s.webserver.stop();
+		state.webserver.stop();
 	}
 #endif
 	return run_result::ok();
--- a/userspace/falco/app/actions/start_webserver.cpp
+++ b/userspace/falco/app/actions/start_webserver.cpp
@@ -19,7 +19,6 @@ limitations under the License.
 
 #if !defined(_WIN32) && !defined(__EMSCRIPTEN__) && !defined(MINIMAL_BUILD)
 #include "webserver.h"
-#include "falco_metrics.h"
 #endif
 
 using namespace falco::app;
@@ -35,7 +34,7 @@ falco::app::run_result falco::app::actions::start_webserver(falco::app::state& s
 			falco_logger::log(falco_logger::level::DEBUG, "Skipping starting webserver in dry-run\n");
 			return run_result::ok();
 		}
-	
+
 		falco_configuration::webserver_config webserver_config = state.config->m_webserver_config;
 		std::string ssl_option = (webserver_config.m_ssl_enabled ? " (SSL)" : "");
 		falco_logger::log(falco_logger::level::INFO, "Starting health webserver with threadiness "
@@ -46,12 +45,9 @@ falco::app::run_result falco::app::actions::start_webserver(falco::app::state& s
 			+ std::to_string(webserver_config.m_listen_port)
 			+ ssl_option + "\n");
 
-		falco_metrics metrics(state);
-
 		state.webserver.start(
-			state.offline_inspector,
-			webserver_config,
-			metrics);
+			state,
+			webserver_config);
 	}
 #endif
 	return run_result::ok();
@@ -73,4 +69,3 @@ falco::app::run_result falco::app::actions::stop_webserver(falco::app::state& st
 #endif
 	return run_result::ok();
 }
-
--- a/userspace/falco/falco_metrics.cpp
+++ b/userspace/falco/falco_metrics.cpp
@@ -82,6 +82,10 @@ std::string falco_metrics::to_text(const falco::app::state& state)
 		prometheus_text += prometheus_metrics_converter.convert_metric_to_text_prometheus("kernel_release", "falcosecurity", "falco", {{"kernel_release", agent_info->uname_r}});
 		prometheus_text += prometheus_metrics_converter.convert_metric_to_text_prometheus("hostname", "falcosecurity", "evt", {{"hostname", machine_info->hostname}});
 
+		for (const std::string& source: inspector->event_sources())
+		{
+			prometheus_text += prometheus_metrics_converter.convert_metric_to_text_prometheus("evt_source", "falcosecurity", "falco", {{"evt_source", source}});
+		}
 		std::vector<metrics_v2> static_metrics;
 		static_metrics.push_back(libs_metrics_collector.new_metric("start_ts",
 																	METRICS_V2_MISC,
--- a/userspace/falco/falco_metrics.cpp
+++ b/userspace/falco/falco_metrics.cpp
@@ -21,6 +21,15 @@ limitations under the License.
 
 #include <libsinsp/sinsp.h>
 
+/*!
+	\class falco_metrics
+	\brief This class is used to convert the metrics provided by the application
+	and falco libs into a string to be return by the metrics endpoint.
+*/
+
+/*!
+	\brief Constructor that takes a \c state object to build its internal state
+*/
 falco_metrics::falco_metrics(falco::app::state& state)
 {
 	falco_configuration::webserver_config webserver_config = state.config->m_webserver_config;
@@ -37,6 +46,11 @@ falco_metrics::falco_metrics(falco::app::state& state)
 	}
 }
 
+/*!
+	\brief This method returns a textual representation of the metrics configured.
+
+	The current implementation returns a Prometheus exposition formatted string.
+*/
 std::string falco_metrics::to_text() const
 {
 	if (!m_metrics_enabled)
--- a/userspace/falco/falco_metrics.cpp
+++ b/userspace/falco/falco_metrics.cpp
@@ -28,6 +28,15 @@ limitations under the License.
 */
 
 /*!
+	\brief content_type to be returned by the webserver's metrics endpoint.
+
+	Currently it is the default Prometheus exposition format
+
+	https://prometheus.io/docs/instrumenting/exposition_formats/#text-based-format
+*/
+const std::string falco_metrics::content_type = "text/plain; version=0.0.4";
+
+/*!
 	\brief Constructor that takes a \c state object to build its internal state
 */
 falco_metrics::falco_metrics(falco::app::state& state)
--- a/userspace/falco/falco_metrics.cpp
+++ b/userspace/falco/falco_metrics.cpp
@@ -74,23 +74,23 @@ std::string falco_metrics::to_text() const
 
 		std::vector<metrics_v2> static_metrics;
 		static_metrics.push_back(libs_metrics_collector.new_metric("start_ts",
-											METRICS_V2_LIBBPF_STATS,
-											METRIC_VALUE_TYPE_U64,
-											METRIC_VALUE_UNIT_TIME_NS_COUNT,
-											METRIC_VALUE_METRIC_TYPE_MONOTONIC,
-											agent_info->start_ts_epoch));
-		static_metrics.push_back(libs_metrics_collector.new_metric("falco_host_boot_ts",
-											METRICS_V2_LIBBPF_STATS,
-											METRIC_VALUE_TYPE_U64,
-											METRIC_VALUE_UNIT_TIME_NS_COUNT,
-											METRIC_VALUE_METRIC_TYPE_MONOTONIC,
-											machine_info->boot_ts_epoch));
-		static_metrics.push_back(libs_metrics_collector.new_metric("falco_host_num_cpus",
-											METRICS_V2_LIBBPF_STATS,
-											METRIC_VALUE_TYPE_U64,
-											METRIC_VALUE_UNIT_TIME_NS_COUNT,
-											METRIC_VALUE_METRIC_TYPE_MONOTONIC,
-											machine_info->num_cpus));
+																	METRICS_V2_MISC,
+																	METRIC_VALUE_TYPE_U64,
+																	METRIC_VALUE_UNIT_TIME_TIMESTAMP_NS,
+																	METRIC_VALUE_METRIC_TYPE_NON_MONOTONIC_CURRENT,
+																	agent_info->start_ts_epoch));
+		static_metrics.push_back(libs_metrics_collector.new_metric("host_boot_ts",
+																	METRICS_V2_MISC,
+																	METRIC_VALUE_TYPE_U64,
+																	METRIC_VALUE_UNIT_TIME_TIMESTAMP_NS,
+																	METRIC_VALUE_METRIC_TYPE_NON_MONOTONIC_CURRENT,
+																	machine_info->boot_ts_epoch));
+		static_metrics.push_back(libs_metrics_collector.new_metric("host_num_cpus",
+																	METRICS_V2_MISC,
+																	METRIC_VALUE_TYPE_U32,
+																	METRIC_VALUE_UNIT_COUNT,
+																	METRIC_VALUE_METRIC_TYPE_NON_MONOTONIC_CURRENT,
+																	machine_info->num_cpus));
 
 		for (auto metrics: static_metrics)
 		{
--- a/userspace/falco/falco_metrics.cpp
+++ b/userspace/falco/falco_metrics.cpp
@@ -108,6 +108,15 @@ std::string falco_metrics::to_text(const falco::app::state& state)
 																	METRIC_VALUE_METRIC_TYPE_NON_MONOTONIC_CURRENT,
 																	state.outputs->get_outputs_queue_num_drops()));
 
+		auto now = std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
+
+		static_metrics.push_back(libs_metrics_collector.new_metric("duration_sec",
+																	METRICS_V2_MISC,
+																	METRIC_VALUE_TYPE_U64,
+																	METRIC_VALUE_UNIT_TIME_TIMESTAMP_NS,
+																	METRIC_VALUE_METRIC_TYPE_NON_MONOTONIC_CURRENT,
+																	(uint64_t)((now - agent_info->start_ts_epoch) / ONE_SECOND_IN_NS)));
+
 		for (auto metrics: static_metrics)
 		{
 			prometheus_metrics_converter.convert_metric_to_unit_convention(metrics);
--- a/userspace/falco/falco_metrics.cpp
+++ b/userspace/falco/falco_metrics.cpp
@@ -36,46 +36,33 @@ limitations under the License.
 */
 const std::string falco_metrics::content_type = "text/plain; version=0.0.4";
 
-/*!
-	\brief Constructor that takes a \c state object to build its internal state
-*/
-falco_metrics::falco_metrics(falco::app::state& state)
-{
-	falco_configuration::webserver_config webserver_config = state.config->m_webserver_config;
-	m_metrics_enabled = state.config->m_metrics_enabled && webserver_config.m_metrics_enabled;
-
-	if (m_metrics_enabled)
-	{
-		for (const auto& source_info: state.source_infos)
-		{
-			sinsp *source_inspector = source_info.inspector.get();
-			m_inspectors.push_back(source_inspector);
-			m_metrics_collectors.push_back(libs::metrics::libs_metrics_collector(source_inspector, state.config->m_metrics_flags));	
-		}
-	}
-}
 
 /*!
-	\brief This method returns a textual representation of the metrics configured.
+	\brief this method takes an application \c state and returns a textual representation of
+	its configured metrics.
 
 	The current implementation returns a Prometheus exposition formatted string.
 */
-std::string falco_metrics::to_text() const
+std::string falco_metrics::to_text(const falco::app::state& state)
 {
-	if (!m_metrics_enabled)
-	{
-		return "";
-	}
-
 	static const char* all_driver_engines[] = {
 		BPF_ENGINE, KMOD_ENGINE, MODERN_BPF_ENGINE,
 		SOURCE_PLUGIN_ENGINE, NODRIVER_ENGINE, GVISOR_ENGINE };
 
+	std::vector<sinsp*> inspectors;
+	std::vector<libs::metrics::libs_metrics_collector> metrics_collectors;
+
+	for (const auto& source_info: state.source_infos)
+	{
+		sinsp *source_inspector = source_info.inspector.get();
+		inspectors.push_back(source_inspector);
+		metrics_collectors.push_back(libs::metrics::libs_metrics_collector(source_inspector, state.config->m_metrics_flags));
+	}
 
 	libs::metrics::prometheus_metrics_converter prometheus_metrics_converter;
 	std::string prometheus_text;
 
-	for (auto* inspector: m_inspectors)
+	for (auto* inspector: inspectors)
 	{
 		for (size_t i = 0; i < sizeof(all_driver_engines) / sizeof(const char*); i++)
 		{
@@ -122,7 +109,7 @@ std::string falco_metrics::to_text() const
 		}
 	}
 
-	for (auto metrics_collector: m_metrics_collectors)
+	for (auto metrics_collector: metrics_collectors)
 	{
 		metrics_collector.snapshot();
 		auto metrics_snapshot = metrics_collector.get_metrics();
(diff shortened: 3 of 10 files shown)


userspace/falco/webserver.cpp
```diff
@@ -68,6 +69,28 @@ void falco_webserver::start(
...
+
+        m_server->Get("/metrics",
+            [metrics_collectors, prometheus_metrics_converter](const httplib::Request &, httplib::Response &res) {
+                std::string prometheus_text;
+
+                for (auto metrics_collector: metrics_collectors) {
+                    metrics_collector.snapshot();
+                    auto metrics_snapshot = metrics_collector.get_metrics();
+
+                    for (auto& metric: metrics_snapshot)
```
incertum, line 84: Could we also add all the wrapper metrics already in, see for example https://github.com/falcosecurity/falco/issues/3131#issuecomment-1984818730 (see the libs unit tests) or the website https://falco.org/docs/metrics/falco-metrics/ or run Falco w/ the output rule metrics.

Or would you just want to do that in another PR? Also ok for me.

Some more notes:

For Prometheus we do not need the `_prev` metrics values and even the drop percentage is not needed as it can be calculated via Prometheus queries, hence not a 1:1 translation in terms of the wrapper or additional metrics fields.

In addition, we distinguish between `falco.` and `scap.` prefixes -- probably best to preserve that and as namespace I personally would prefer `falcosecurity`.

incertum: From a code organization point of view we can also iterate as it would be nice to not add metrics logic into the webserver file itself and rather maybe create a more formal `falco_metrics_collector` class or something else.

Or maybe you planned that already in a follow up PR.

userspace/falco/webserver.cpp
```diff
@@ -68,6 +69,28 @@ void falco_webserver::start(
...
+                    {
+                       prometheus_metrics_converter.convert_metric_to_unit_convention(metric);
+                       prometheus_text += prometheus_metrics_converter.convert_metric_to_text_prometheus(metric, "namespace", "falco") + "\n";
```
incertum, line 87: Do we need the additional `+ "\n"` as the returned `prometheus_text` should already end with a new line.

incertum (commented):
🚀 nice thanks for getting this up so quickly!

userspace/falco/webserver.cpp
incertum: Additional note I forgot: It appears we need to move the code logic to the `stats_writer::collector::collector` so that the metrics get refreshed and reposted to `/metrics` every x interval.

userspace/falco/app/actions/start_webserver.cpp
```diff
@@ -44,8 +44,18 @@ falco::app::run_result falco::app::actions::start_webserver(falco::app::state& s
 			+ std::to_string(s.config->m_webserver_listen_port)
 			+ ssl_option + "\n");
 
+		std::vector<libs::metrics::libs_metrics_collector> metrics_collectors;
+		if (s.config->m_metrics_prometheus_enabled && s.config->m_metrics_prometheus_enabled)
```
FedeDP, line 48:
```suggestion
		if (s.config->m_metrics_enabled && s.config->m_metrics_prometheus_enabled)
```

userspace/falco/webserver.cpp
FedeDP: > In addition, we distinguish between falco. and scap. prefixes -- probably best to preserve that and as namespace I personally would prefer falcosecurity.

Agree with `falcosecurity`.

From what i see, we are refreshing the stats when the `/metrics` endpoint is called, instead of passing through the `stats_writer::collector::collector`.
This ensures that we follow Prometheus settings (ie: if one setups prometheus to scrape every 2s, we honor that even if stats collector interval is higher).
Don't know if this is the behavior we want though.

userspace/falco/webserver.cpp
sgaist: Indeed not, old string building muscle memory.

userspace/falco/webserver.cpp
incertum: I can see the dilemma. At the same time I also don't see a problem with just recommending to having the same Falco metrics interval as the Prometheus scrape interval? How is this solved in other projects?  At the end of the day metrics time series are always discrete ... I would also be fine to refresh the metrics in the Prometheus case whenever `/metrics` was called as @FedeDP suggested.

sgaist: One thing that comes to mind, doesn't the stats_writer class contain the required metrics generation logic ?

incertum: The metrics wrapper fields (the ones not part of the `libs_metrics_collector`) are easy to duplicate if needed, or we create some sort of more formal class `falco-metrics_collector` ...

FedeDP: > At the same time I also don't see a problem with just recommending to having the same Falco metrics interval as the Prometheus scrape interval?

Yep, but:
* `/metrics` endpoint is the only "interactive" metrics generator we have, since `output_rule` and `output_file` are passive. Therefore, it makes sense that is up to the caller to decide the frequency.
* also, consider the case where prometheus is setup to scrape every 5seconds; consider also that `output_file` is enabled too. This has a cost in terms of IO perf if user has to set 5s `metrics.interval` in the config file

At the same time, i quite don't like having a metric that does not obey to `metrics.interval` though...

I would love to gather more feedback! cc @falcosecurity/falco-maintainers

incertum: Maybe we need to just make everything configurable, for example, we could add sub metrics configs similar to below:

```
metrics:
  enabled: false
  interval: 1h
  # Typically, in production, you only use `output_rule` or `output_file`, but not both. 
  # However, if you have a very unique use case, you can use both together.
  output_rule: true
  # output_file: /tmp/falco_stats.jsonl
  output_prometheus:
    # Will only be enabled when `metrics.enabled` is set to true
    enabled: false
    # Take a new metrics snapshot automatically after each Prometheus server scrape
    auto_refresh_enabled: false
    # Alternative to `auto_refresh_enabled` refreshes /metrics with new metrics snapshot every x interval; supersedes and is separate from the global `metrics.interval` setting
    refresh_interval: 5m
  resource_utilization_enabled: true
  state_counters_enabled: true
  kernel_event_counters_enabled: true
  libbpf_stats_enabled: true
  convert_memory_to_mb: true
  include_empty_values: false
```

falco.yaml
```diff
@@ -982,6 +982,7 @@ metrics:
   libbpf_stats_enabled: true
   convert_memory_to_mb: true
   include_empty_values: false
+  prometheus_enabled: false
```
FedeDP, line 985: Given that this is indeed exposed by the webserver, shouldn't it live under `webserver:`; perhaps we might want to specify that it is useful only when `metrics` are enabled.

sgaist: Good point and agreed 

One related question: should the endpoint name also be configurable ? It usually is `/metrics` but...

FedeDP: I'd leave `/metrics` for now. We can always make int configurable in the future if there is demand for it!

sgaist:
There's something that is escaping me with regard to the `sinsp` object initialization. It seems that when the webserver is started, which happens very late in the process, said `sinsp` objects are not fully initialized.

sgaist:
> Samuel, pulled your branch and wanted to test drive it, but it doesn't compile. Mind fixing the build? Thank you! I'll review the changes then. In addition, we still need to agree on the design wrt refreshing the metrics, @leogr to compile a final proposal based on suggestions we have so far. Thanks.

I knew I forgot to commit the cmake changes when I read your message 😅
However I missed the test build failure as I was sure that they were enabled.

All fixed now

incertum:
Great, if you look at the existing stats code you see we need a bunch of Ifdefs because of wasm, macos, windows ... etc that should fix those builds, but we can do it later as well.

sgaist:
Builds are fixed however I am wondering which configurations are officially supported on non Linux platforms.

From the looks of it, on macOS, the non minimal build won't build.

userspace/falco/webserver.h
```diff
@@ -36,12 +38,8 @@ class falco_webserver
 	falco_webserver& operator = (const falco_webserver&) = delete;
 	virtual void start(
 		const std::shared_ptr<sinsp>& inspector,
-		uint32_t threadiness,
-		uint32_t listen_port,
-		std::string& list_address,
-		std::string& healthz_endpoint,
-		std::string &ssl_certificate,
-		bool ssl_enabled);
+		const falco_configuration::webserver_config& configuration,
```
incertum, line 41: Sounding like a broken record also here it would be better to name it very specific `webserver_config` instead of just a generic `configuration`.

userspace/falco/falco_metrics.h
```diff
@@ -0,0 +1,40 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+#pragma once
+
+#include "configuration.h"
+
+#include <libsinsp/sinsp.h>
+
+namespace falco {
```
incertum, line 23:
```suggestion
namespace falco::app
{
```

userspace/falco/falco_metrics.h
```diff
@@ -0,0 +1,40 @@
...
+class falco_metrics
+{
+public:
+	falco_metrics(falco::app::state& state);
+	bool is_enabled() const { return m_is_enabled; };
+	std::string to_text() const;
+
+private:
+	bool m_is_enabled = false;
+	std::vector<sinsp*> m_inspectors;
```
incertum, line 38: @sgaist would you mind getting me up to speed why we need a vector with inspectors? Thanks!
Plus you still have an open question around sinsp and config init sequences etc?

userspace/falco/falco_metrics.cpp
```diff
@@ -0,0 +1,114 @@
...
+											METRIC_VALUE_UNIT_TIME_NS_COUNT,
+											METRIC_VALUE_METRIC_TYPE_MONOTONIC,
+											agent_info->start_ts_epoch));
+		static_metrics.push_back(libs_metrics_collector.new_metric("falco_host_boot_ts",
+											METRICS_V2_LIBBPF_STATS,
+											METRIC_VALUE_TYPE_U64,
+											METRIC_VALUE_UNIT_TIME_NS_COUNT,
+											METRIC_VALUE_METRIC_TYPE_MONOTONIC,
+											machine_info->boot_ts_epoch));
+		static_metrics.push_back(libs_metrics_collector.new_metric("falco_host_num_cpus",
```
incertum, line 88: correction + rename to `host_num_cpus` without falco_: 
```
METRICS_V2_MISC,
METRIC_VALUE_TYPE_U32,
METRIC_VALUE_UNIT_COUNT,
METRIC_VALUE_METRIC_TYPE_NON_MONOTONIC_CURRENT,
machine_info->num_cpus));
```

userspace/falco/falco_metrics.cpp
```diff
@@ -0,0 +1,114 @@
...
+		const scap_machine_info* machine_info = inspector->get_machine_info();
+
+		libs::metrics::libs_metrics_collector libs_metrics_collector(inspector, 0);
+
+		prometheus_text += converter.convert_metric_to_text_prometheus("falco_version", "falcosecurity", "falco", {{"falco_version", FALCO_VERSION}});
+		prometheus_text += converter.convert_metric_to_text_prometheus("kernel_release", "falcosecurity", "falco", {{"kernel_release", agent_info->uname_r}});
+		prometheus_text += converter.convert_metric_to_text_prometheus("evt_hostname", "falcosecurity", "falco", {{"evt_hostname", machine_info->hostname}});
+
+		std::vector<metrics_v2> static_metrics;
+		static_metrics.push_back(libs_metrics_collector.new_metric("start_ts",
```
incertum, line 76:
```
```
.new_metric("start_ts",
METRICS_V2_MISC, 
METRIC_VALUE_TYPE_U64,
METRIC_VALUE_UNIT_TIME_TIMESTAMP_NS, 
METRIC_VALUE_METRIC_TYPE_NON_MONOTONIC_CURRENT, 
agent_info->start_ts_epoch));
```

**Comment by incertum** [comment]
> Builds are fixed however I am wondering which configurations are officially supported on non Linux platforms.
> 
> From the looks of it, on macOS, the non minimal build won't build.

Agreed it's not very clear ... I believe this should be made more transparent in the future. Out of scope for this PR.

File: falco.yaml [hunk 10]
```diff
@@ -695,6 +695,7 @@ webserver:
   # Can be an IPV4 or IPV6 address, defaults to IPV4
   listen_address: 0.0.0.0
   k8s_healthz_endpoint: /healthz
+  metrics_enabled: false
```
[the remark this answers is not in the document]
sgaist: I thought that the same webserver would serve both endpoints.
This brings the following question: should we have different web server instance for the metrics part ?

userspace/falco/falco_metrics.cpp
```diff
@@ -0,0 +1,114 @@
...
+#include "app/state.h"
+
+#include <libsinsp/sinsp.h>
+
+falco_metrics::falco_metrics(falco::app::state& state)
+{
+	falco_configuration::webserver_config webserver_config = state.config->m_webserver_config;
+	m_is_enabled = state.config->m_metrics_enabled && webserver_config.m_metrics_enabled;
+
+	if (m_is_enabled)
```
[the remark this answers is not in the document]
sgaist: Done

userspace/falco/falco_metrics.cpp
```diff
@@ -0,0 +1,114 @@
...
+	{
+		return "";
+	}
+
+	static const char* all_driver_engines[] = {
+		BPF_ENGINE, KMOD_ENGINE, MODERN_BPF_ENGINE,
+		SOURCE_PLUGIN_ENGINE, NODRIVER_ENGINE, GVISOR_ENGINE };
+
+
+	libs::metrics::prometheus_metrics_converter converter;
```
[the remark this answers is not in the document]
sgaist: Done

userspace/falco/falco_metrics.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+	for (auto metrics_collector: m_metrics_collectors)
+	{
+		metrics_collector.snapshot();
+		auto metrics_snapshot = metrics_collector.get_metrics();
+
+		for (auto& metric: metrics_snapshot)
+		{
+		   converter.convert_metric_to_unit_convention(metric);
+		   prometheus_text += converter.convert_metric_to_text_prometheus(metric, "falcosecurity", "scap");
```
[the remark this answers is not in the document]
sgaist: I misread the examples, thanks for the correction.

userspace/falco/falco_metrics.cpp
```diff
@@ -0,0 +1,114 @@
...
+		prometheus_text += converter.convert_metric_to_text_prometheus("evt_hostname", "falcosecurity", "falco", {{"evt_hostname", machine_info->hostname}});
+
+		std::vector<metrics_v2> static_metrics;
+		static_metrics.push_back(libs_metrics_collector.new_metric("start_ts",
+											METRICS_V2_LIBBPF_STATS,
+											METRIC_VALUE_TYPE_U64,
+											METRIC_VALUE_UNIT_TIME_NS_COUNT,
+											METRIC_VALUE_METRIC_TYPE_MONOTONIC,
+											agent_info->start_ts_epoch));
+		static_metrics.push_back(libs_metrics_collector.new_metric("falco_host_boot_ts",
```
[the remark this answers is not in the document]
sgaist: All done, thanks for the corrections.

userspace/falco/falco_metrics.h
sgaist: Done

userspace/falco/falco_metrics.cpp
sgaist: Done

userspace/falco/falco_metrics.h
sgaist: This is related to the initialization state of the sinsp objects when starting the webserver.

In my original implementation, all static stuff was created in the constructor of the `falco_metrics` but since the `machine_info` and `agent_info` objects are null pointers at that point, it's not doable. Hence I am currently keeping them as they cannot be retrieved through the converter class.

It's a temporary design to have something working that we can discuss about.

userspace/falco/webserver.h
sgaist: No problem with such records :-)

Done

falco.yaml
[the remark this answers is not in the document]
incertum: Fair, you actually asked me at the beginning of we should just use the existing webserver and I said yes as for me I was referring to existing code ... What would you suggest? Or what is better practice if you want to expose multiple ports? @FedeDP what's you take on this?

I personally never set up webservers, instead if anything I tried to break into them in the past 🙃 .

userspace/falco/falco_metrics.h
incertum: Wonder if we can just adjust the order of initialization a bit? I can check later.

falco.yaml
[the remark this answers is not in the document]
FedeDP: I think we should reuse the same webserver we already use; having everything on the same ports on different endpoints is ok from my PoV, indeed it is better to remember "right, on port X there is Falco webserver".

FedeDP:
cc @falcosecurity/falco-maintainers please review the changes and let's try to focus on the general design.
If we want this to fully land for 0.38.0 (consider the grafana chart part), we should review (and perhaps have a first merge) asap IMHO; then we can build more improvements on top of this, but we need to agree and merge the general design.

falco.yaml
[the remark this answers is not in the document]
incertum: ok then it seems settled, we keep it as is.

userspace/falco/falco_metrics.h
incertum: @FedeDP @Andreagit97 plz advice, do we really need the vectors here because of the various source_infos. https://github.com/falcosecurity/falco/pull/3140/files#diff-0ebdd484eb8f6b10eb0df365c052ec7a91640bd6502ac3a7af1da82d9e62f3b1R31. We don't distinguish between the source info for the stats_writer either.

falco.yaml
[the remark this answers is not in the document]
incertum: Re naming see https://github.com/falcosecurity/falco/pull/3140#issuecomment-2031651529, plz use `prometheus_metrics_enabled` and update the comments, so folks folks understand how to use this config.

userspace/falco/falco_metrics.cpp
```diff
@@ -0,0 +1,120 @@
...
+		}
+
+		const scap_agent_info* agent_info = inspector->get_agent_info();
+		const scap_machine_info* machine_info = inspector->get_machine_info();
+
+		libs::metrics::libs_metrics_collector libs_metrics_collector(inspector, 0);
+
+		prometheus_text += prometheus_metrics_converter.convert_metric_to_text_prometheus("falco_version", "falcosecurity", "falco", {{"falco_version", FALCO_VERSION}});
+		prometheus_text += prometheus_metrics_converter.convert_metric_to_text_prometheus("kernel_release", "falcosecurity", "falco", {{"kernel_release", agent_info->uname_r}});
+		prometheus_text += prometheus_metrics_converter.convert_metric_to_text_prometheus("evt_hostname", "falcosecurity", "falco", {{"evt_hostname", machine_info->hostname}});
```
incertum, line 73:
```suggestion
		prometheus_text += prometheus_metrics_converter.convert_metric_to_text_prometheus("evt_hostname", "falcosecurity", "evt", {{"hostname", machine_info->hostname}});
```

If we wanted to be super consistent I guess. Add here the "evt" "source" metric that is missing, see https://github.com/falcosecurity/falco/pull/3140/files#r1543632512

userspace/falco/falco_metrics.h
FedeDP: I'd cc @jasondellaluce on this point, he might be able to help us :)

jasondellaluce: Can't `falco_metrics::to_text` just take `falco::app::state&` as an argument? That way we avoid maintaining raw pointers all around the place. Also, why is `falco::app::state` declared with a forward declaration few lines above. What's the intention there?

userspace/falco/webserver.cpp
```diff
@@ -68,6 +65,13 @@ void falco_webserver::start(
             res.set_content(versions_json_str, "application/json");
         });
 
+    if (metrics.is_enabled())
+    {
+        m_server->Get("/metrics",
+            [metrics](const httplib::Request &, httplib::Response &res) {
+                res.set_content(metrics.to_text(), "text/plain; version=0.0.4");
```
jasondellaluce, line 72: What is the `0.0.4` version? If it's some constant out of our control, can we define it as a constant somewhere?

userspace/falco/webserver.h
```diff
@@ -25,6 +25,8 @@ limitations under the License.
 #include <memory>
 #include <thread>
 
+class falco_metrics;
```
jasondellaluce, line 28: Is there a reason for the forward declaration?

userspace/falco/webserver.cpp
sgaist: It's the [text based exposition format](https://prometheus.io/docs/instrumenting/exposition_formats/#text-based-format).
Agreed, I'll make it a documented constant.

userspace/falco/falco_metrics.h
sgaist: That's an interesting question/suggestion.

Just to make things clear for future readers, static, in this context, is meant as non-changing and is not related to the `static` C++ keyword.

My original design idea was the following:
1) pass the state to the constructor
2) build a set of static metrics (start up timestamps, etc.)
3) build a set of static textual metrics (falco version, kernel version etc.)
4) build a set of collectors for the dynamic values
5) build the final text in `to_text`

For 2 and 3, while both are static, they use different means to build the metrics text.

The goals for that design were:
1) generate metrics object for values that will not change only once
2) reduce the number of object creation/destruction cycle when requesting metrics (which is a direct follow up of 1).
3) allow to later change the collector class without having to rewrite the whole logic as it is basically: go through both sets of static metrics, go through dynamic metrics and voila, your metrics text is built.

This would make the metrics test generation faster although it would use a bit more memory to keep the static metrics values.

However, as noted earlier, for some reason I don't know (I currently don't know the libs code that well), it seems that the sinsp objects are not fully initialized when reaching the constructor otherwise, I would not have kept this vector of sinsp pointers.

As for the forward declaration, in this case, it's mainly out of habit from other projects I contribute to (only include where you use). Hence I am surprised I didn't do that for `sinsp` as well.

Note that I don't mind changing the design as it will mainly move the collectors creation to `to_text`.

userspace/falco/webserver.h
sgaist: Mostly out of habits from other projects I work on.

falco.yaml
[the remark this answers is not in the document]
incertum: Understood, however if we support multiple standards, end users would need to have a way to choose which one. In addition, right now we don't even have a comment mentioning "Prometheus", how would you guess that it is Prometheus?

userspace/falco/falco_metrics.cpp
```diff
@@ -0,0 +1,114 @@
...
+											METRIC_VALUE_METRIC_TYPE_MONOTONIC,
+											machine_info->boot_ts_epoch));
+		static_metrics.push_back(libs_metrics_collector.new_metric("falco_host_num_cpus",
+											METRICS_V2_LIBBPF_STATS,
+											METRIC_VALUE_TYPE_U64,
+											METRIC_VALUE_UNIT_TIME_NS_COUNT,
+											METRIC_VALUE_METRIC_TYPE_MONOTONIC,
+											machine_info->num_cpus));
+
+		for (auto metrics: static_metrics)
```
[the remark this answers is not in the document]
incertum: > One thing that comes to mind: should we have a list to properly filter out all the metrics that should not be returned for Prometheus ?

Sure, at the same time I believe the ones we don't need for Prometheus are derived ones, hence we can also just skip creating them altogether. For example all of the libs metrics collector metrics are exposed, it's just a handful of derived ones we don't need for Prometheus, for instance the event rates or drop percentages.

falco.yaml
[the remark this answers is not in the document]
sgaist: For multiple standards, that's what my `metrics_backend` suggestion was for.

I agree that a comment in the configuration file is a good starting point.

So just to solve this point, should we:
- Use `prometheus_metrics_enabled` (which is pretty self explanatory)
- Use `metrics_enabled` with a comment stating that it's currently Prometheus that is the backend

?

userspace/falco/falco_metrics.h
sgaist: @FedeDP @jasondellaluce just a small ping around this design question.

I don't have any issue with re-implementing this part passing the application state as parameter of the `to_text` method, I would just like to have an agreement about which architecture we would like to adopt.

falco.yaml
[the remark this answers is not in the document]
incertum: Just noticing now that I didn’t read the metrics_backend properly, apologies. Yeah that works for me, perhaps metrics_backend should be an array of strings similar to the plugins config setup.

userspace/falco/falco_metrics.h
incertum: @FedeDP @jasondellaluce @leogr +1 would appreciate final guidance on what you prefer so that we can get this over the finish line. Thanks.

falco.yaml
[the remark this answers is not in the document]
leogr: > I have used `metrics_enabled` as I though that, while using Prometheus currently, it would be subject to change so rather than having the name hard coded in the option name, having just "web server will also provide a metrics end point" would be simpler for admins in the long run.

I agree.


> Maybe providing a `metrics_backend` at a later stage to allow switching would be cleaner and clearer.
> * Use `metrics_enabled` with a comment stating that it's currently Prometheus that is the backend

@sgaist This works for me too :+1: 

> perhaps metrics_backend should be an array of strings similar to the plugins config setup.

@incertum We noticed that an array of strings does not play well with the `-o` command line flag. So, I'd recommend using multiple `_enabled` options so that it's easy to do `-o foo.bar_enabled=true`. 
Again, I strongly believe we have to re-audit all config options and make them consistent, but it's out of scope for this PR, so I'd leave this decision for later.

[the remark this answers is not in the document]
FedeDP: Perhaps we might want to call the config `prometheus_enabled`?
Anyway, `metrics_enabled` works fine for me too. I'd avoid the complexity to deal with possible multiple metrics backend atm.
We can always introduce the new feat as `Incubating` to be able to play with it and be free to drop it/rename it in the future: https://github.com/falcosecurity/falco/blob/master/proposals/20231220-features-adoption-and-deprecation.md#maturation-levels

userspace/falco/falco_metrics.h
leogr: I don't have strong opinions, but I agree with this :point_down: 
> avoid maintaining raw pointers all around the place

falco.yaml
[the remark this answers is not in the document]
incertum: Right @leogr re arrays configs. @FedeDP  would then just prometheus_metrics_enabled be ok? I believe it could be beneficial for this to show up when string searching for metrics. In addition, its even clearer for newbies to metrics.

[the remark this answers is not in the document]
FedeDP: +1! Looks fine to me!

leogr:
@sgaist

I'm just double-checking to help you get to the finish line.

 I've just tracked these pending items:
- a few metrics are still missing https://github.com/falcosecurity/falco/pull/3140/files#r1543632512
- awaiting consensus here (impl. detail) https://github.com/falcosecurity/falco/pull/3140#discussion_r1559288321
- awaiting consensus for the config option name https://github.com/falcosecurity/falco/pull/3140#discussion_r1559285656 
- some CI tests are failing

Is there anything else missing? Or other blockers? :thinking:

falco.yaml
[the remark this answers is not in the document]
LucaGuerra: I took a look at this. To make the config file easier to use I think it's important to highlight that you need both `metrics_enabled` and `metrics.enabled` to true in order for this to have any effect. So I would suggest `prometheus_metrics_enabled` with a comment that mentions that this only does something if `metrics.enabled` is true.

Technically I would also be ok with setting `prometheus_metrics_enabled` to `true` since metrics are disabled by default but maybe let's do it after the support is stable.

userspace/falco/falco_metrics.h
LucaGuerra: Re. the raw pointers, generally non-owning raw pointers are fine but the problems we can have is when these objects are free'd or reallocated by their owners.

I don't completely remember the complete mechanism that happens in the restart case, but will this work if Falco is restarted, e.g. when a rule is replaced and Falco auto-reloads (therefore rebuilding inspectors)? iirc it should but could you please double check?

jasondellaluce: > I don't have any issue with re-implementing this part passing the application state as parameter of the to_text method, I would just like to have an agreement about which architecture we would like to adopt.

I see your point about creating non-mutating info at construction time, however I think that these non-owning raw pointers can create more harm than benefit. The issue is that they rely on leaked knowledge about:
- The order in which inspectors are allocated in the state, which is transitively dependent on how event sources are ordered, taking also for granted that this order won't change (which is like so now, but who knows later on)
- The lifecycle of these inspectors, meaning at "which point" in their initialization process they are (plugins, configs, captures being open, etc..), which is strongly dependent on the execution flow and the app actions ordering

So IMO it's not just the raw pointer (that could be solved by making it owner through a shared_ptr instance), but more that we leak implicit knowledge about other parts of the codebase. IMO, moving everything in `to_text` is clearer, meaning that the invoker of that method has responsibility of having everything in the right preconditions before invocation time, and doesn't have to care about small parts being retained. If re-computation is a concern, you can consider stateful lazy caching mechanisms (e.g. after getting a non-mutating metric the first time, save it in the state and don't re-compute it again).

I'm not gonna fight for this of course, but this is my point overall.

sgaist: In addition to your excellent points, I did not take into account the dynamic reloading use case which also goes in the favor of your suggested implementation so I will do that.

I think the lazy caching is a good idea for a follow up PR.

userspace/falco/falco_metrics.cpp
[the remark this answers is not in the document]
sgaist: I might be missing something, aren't the missing metrics only available to the stats writer ?
In which case, it seems that there's a need for a deeper refactoring to make the information available in the webserver as well.

sgaist (commented):
> @sgaist
> 
> I'm just double-checking to help you get to the finish line.
> 
> I've just tracked these pending items:
> 
>     * a few metrics are still missing https://github.com/falcosecurity/falco/pull/3140/files#r1543632512
Not done yet, see my answer to this comment 
>     * awaiting consensus here (impl. detail) [feat(webserver): implement metrics endpoint #3140 (comment)](https://github.com/falcosecurity/falco/pull/3140#discussion_r1559288321)
Done
>     * awaiting consensus for the config option name [feat(webserver): implement metrics endpoint #3140 (comment)](https://github.com/falcosecurity/falco/pull/3140#discussion_r1559285656)
Done
>     * some CI tests are failing
It looked like a download issue unrelated to the patch itself
> 
> Is there anything else missing? Or other blockers? 🤔

I think you had them all

LucaGuerra:
I think there's one last issue with this, and it's in `falco::app::run()` in `app.cpp`. I mentioned earlier that I did not remember how the restarting worked and I did go check. Essentially, it waits for all "running" steps to complete and then executes some teardown steps. Problem is, the return of `falco::app::actions::process_events` is called before `falco::app::actions::stop_webserver`. Before completing, `process_events()` calls `inspector->close()` for all registered data sources (syscalls, plugins etc) which would probably not completely free the inspector but it will definitely invalidate many of the pointers inside, and at that point the server is still serving requests and accessing data, meaning that it may trigger an use after free when trying to collect stats and data.

I see two potential ways of solving this:
* maybe we could, instead of calling `inspector->close()`, call `inspector->stop_capture()` and then close it only as a last teardown step; I'm not 100% sure about the consequences but it should be OK I think.
* libs could return null pointers for every invalid piece of data on a closed inspector and the stats webserver will need to check for it, but it will spew garbage stats at that time (missing data essentially). Not a fan.

@jasondellaluce @FedeDP what do you think?

userspace/falco/falco_metrics.cpp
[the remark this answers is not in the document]
incertum: `evt_source` -> available

`outputs_queue_num_drops` -> this is part of `falco_outputs` and not unique to stats_writer, search for method get_outputs_queue_num_drops, aka just need to make `falco_outputs` accessible, should be possible right?

`num_evts` -> this is sinsp number of events not to be confused with n_evts that are the scap number of events, it's counted within `do_inspect`, mind checking what would need to be done to make this accessible?

duration_sec -> calculate just like done in the stats writer - we have everything `(uint64_t)((now - agent_info->start_ts_epoch) / ONE_SECOND_IN_NS);`

LucaGuerra:
I briefly spoke with @jasondellaluce and @FedeDP about the issue highlighted above. They agree a good solution is closing inspectors as a last step. I'm proposing a patch (https://github.com/falcosecurity/falco/pull/3169) so you can rebase on top of that once it's merged.

userspace/falco/falco_metrics.cpp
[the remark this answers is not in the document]
sgaist: `outputs_queue_num_drops`: done
`duration_sec`: done

For `evt_source`, from a metrics point of view, it's a bit more nebular. There should be `syscall` available by default plus anything that configured plugin provides, so there might be more than one. Do you mean that the metrics should return an entry for each event source from an inspector ?

for `num_events` I am still looking for an elegant solution to retrieve or set the value.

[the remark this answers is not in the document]
incertum: Thanks @sgaist!

Re num_evts I suspected that it may be a bit more tricky.

Re evt_source agreed and it's also one open discussion item since right now you loop over the event sources vs in the outputs rule metrics we do not. I would say right now metrics caters more to one event source only. Something we should improve and disentangle in the future ... out of scope for this PR. CC @jasondellaluce since if I recall correctly you asked to add it back then. For now I would add it for parity with the existing metrics. Lastly maybe we still have time for Falco 0.38.0 to address the event source for metrics across the board in a follow up PR.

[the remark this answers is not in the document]
sgaist: As a temporary solution, I am publishing all the sources for each inspector.

I am wondering, from an architecture point of view, if the metrics and stats data should be stored in a common "model" and then the writers and metrics backend would be views that get their data from that model to do their job.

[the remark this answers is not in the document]
incertum: @sgaist +1 a common metrics model makes the most sense. Re metrics for the different sources ... we need to rethink this more as for Prometheus we can surely only host one metrics view. For now let's accept that it's only working for syscalls source well. Most metrics are for the syscalls event source in the first place. For the case where we solely have a plugin source some metrics (CPU usage) are right now slightly broken because of a scap refactor that introduced a regression.

incertum:
@sgaist few super minor change requests:

- falcosecurity_falco_falco_version_info -> correct to raw metrics name "version" as we do it for the metrics output rule, that way it should resolve to the final name `falcosecurity_falco_version_info`
- falcosecurity_falco_duration_timestamp_nanoseconds -> use metrics type `METRIC_VALUE_UNIT_TIME_S_COUNT` so it should resolve to the final name `falcosecurity_falco_duration_seconds_total`

falco.yaml
```diff
@@ -695,6 +695,9 @@ webserver:
   # Can be an IPV4 or IPV6 address, defaults to IPV4
   listen_address: 0.0.0.0
   k8s_healthz_endpoint: /healthz
+  # Enable the metrics endpoint providing Prometheus values
+  # It will only have an effect if metrics.enabled is set to true as well.
+  prometheus_metrics_enabled: false
```
incertum, line 700: In addition @sgaist would you mind modifying the `metrics` config text and refer to setting this value when wanting to export Prometheus metrics?

sgaist: Sure thing !
Done

leogr:
> > @sgaist few super minor change requests:
> > ```
> > * falcosecurity_falco_falco_version_info -> correct to raw metrics name "version" as we do it for the metrics output rule, that way it should resolve to the final name `falcosecurity_falco_version_info`
> > 
> > * falcosecurity_falco_duration_timestamp_nanoseconds -> use metrics type `METRIC_VALUE_UNIT_TIME_S_COUNT` so it should resolve to the final name `falcosecurity_falco_duration_seconds_total`
> > ```
> 
> All fixed, good catch !

Great job @sgaist !

I want to thank you for all the effort you're putting into this, we really appreciated! :pray:

userspace/falco/falco_metrics.cpp
```diff
@@ -0,0 +1,149 @@
...
+																	METRIC_VALUE_METRIC_TYPE_NON_MONOTONIC_CURRENT,
+																	state.outputs->get_outputs_queue_num_drops()));
+
+		auto now = std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
+
+		static_metrics.push_back(libs_metrics_collector.new_metric("duration_sec",
+																	METRICS_V2_MISC,
+																	METRIC_VALUE_TYPE_U64,
+																	METRIC_VALUE_UNIT_TIME_S_COUNT,
+																	METRIC_VALUE_METRIC_TYPE_NON_MONOTONIC_CURRENT,
```
incertum, line 121: Mind changing to monontonic? Same for the outputs queue count. I can check all metrics more carefully later.

falco.yaml
```diff
@@ -967,6 +970,9 @@ syscall_event_drops:
 # beneficial for exploring the data schema and ensuring that fields with empty
 # values are included in the output.
 #
+# If metrics are enabled, the web server can be configured to activate the
+# corresponding Prometheus endpoint using webserver.prometheus_metrics_enabled.
+#
```
incertum, line 975: We can now remote the respective todo item 🙃

userspace/falco/falco_metrics.cpp
sgaist: Done

incertum (dismissed):
/approve 

Let's defer further changes to a follow up PR. Thanks @sgaist!

incertum:
Now just need to check CI fixes.

userspace/falco/falco_metrics.cpp
```diff
@@ -0,0 +1,149 @@
...
+				break;
+			}
+		}
+
+		const scap_agent_info* agent_info = inspector->get_agent_info();
+		const scap_machine_info* machine_info = inspector->get_machine_info();
+
+		libs::metrics::libs_metrics_collector libs_metrics_collector(inspector, 0);
+
+		prometheus_text += prometheus_metrics_converter.convert_metric_to_text_prometheus("falco_version", "falcosecurity", "falco", {{"version", FALCO_VERSION}});
```
incertum, line 81:
```suggestion
		prometheus_text += prometheus_metrics_converter.convert_metric_to_text_prometheus("version", "falcosecurity", "falco", {{"version", FALCO_VERSION}});
```

@sgaist mind re-pushing this fix?

sgaist: Not at all !
Looks like you have better eyes than me to catch these :-)

incertum (approved):
/approve

leogr (commented):
SGTM too.

I am waiting for someone of the previous reviewer to take another look before giving the final approval

falco.yaml
FedeDP, line 700: @leogr do we need to enforce a feat status for this? Like `Incubating` ?

leogr: It would be preferable, IMO. It is not a blocker for this PR anyway. We may reaudit all options later, but still before the 0.38 release. Does it make sense?

incertum: @leogr proposing to merge it as is and address the follow up items in a new PR. I believe we need another touch up PR anyways.

FedeDP (approved):
/approve

FedeDP:
Thank you very much Samuel! You did a terrific job!

PR #3143, opened 2024-03-19
Add kernelversion and kernelrelease options to falco driver loader entrypoint

Sryther:
**What type of PR is this?**

/kind feature

/area proposals

Automatic detection of Kernel version and Kernel release can be wrong in various cases. This PR aims to give access to falcoctl options already available when running falco-driver-loader in a sidecar.

None.

NONE

### DIFF
--- a/docker/driver-loader/docker-entrypoint.sh
+++ b/docker/driver-loader/docker-entrypoint.sh
@@ -28,12 +28,14 @@ print_usage() {
 	echo "  ebpf           eBPF probe"
 	echo ""
 	echo "Options:"
-	echo "  --help           show this help message"
-	echo "  --clean          try to remove an already present driver installation"
-	echo "  --compile        try to compile the driver locally (default true)"
-	echo "  --download       try to download a prebuilt driver (default true)"
- 	echo "  --http-insecure	 enable insecure downloads"
-	echo "  --print-env      skip execution and print env variables for other tools to consume"
+	echo "  --help                   show this help message"
+	echo "  --clean                  try to remove an already present driver installation"
+	echo "  --compile                try to compile the driver locally (default true)"
+	echo "  --download               try to download a prebuilt driver (default true)"
+	echo "  --kernel-release <value> set the kernel release"
+	echo "  --kernel-version <value> set the kernel version"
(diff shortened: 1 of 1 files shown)


docker/driver-loader/docker-entrypoint.sh
```diff
@@ -109,9 +121,6 @@ while test $# -gt 0; do
     shift
 done
 
-if [ -z "$has_opts" ]; then
```
FedeDP, line 112: Why was this removed?

docker/driver-loader/docker-entrypoint.sh
```diff
@@ -109,9 +121,6 @@ while test $# -gt 0; do
...
-	ENABLE_COMPILE="true"
-	ENABLE_DOWNLOAD="true"
-fi
+echo "Extra args: $extra_args"
```
FedeDP, line 124: I think this is a debug `echo` right? We can drop it!

docker/driver-loader/docker-entrypoint.sh
Sryther: Gosh! My bad..

FedeDP (approved):
/approve

docker/driver-loader/docker-entrypoint.sh
```diff
@@ -85,12 +89,20 @@ while test $# -gt 0; do
 			;;
 		--http-insecure)
 			HTTP_INSECURE="true"
-			;;   
+			;;
 		--source-only)
-		    >&2 echo "Support dropped in Falco 0.37.0."
+			>&2 echo "Support dropped in Falco 0.37.0."
```
FedeDP, line 94: While you are here, can you remove `--source-only` case, since it is no more supported (we just added the deprecation NOTE for Falco 0.37.0, but we can remove it now).

FedeDP: Otherwise i can do that in a followup PR :)

Sryther: Done! :slightly_smiling_face:

FedeDP (approved):
/approve
Thanks!

leogr (approved):

PR #3178, opened 2024-04-23
new(falco): add rule selection configuration in falco.yaml

### LINKED ISSUE #3174: Proposal/discussion: More flexible ways to enable/disable rules from configuration
**Motivation**

Currently, we have several rules files and many users that wish to enable and disable rules without having to go through each rule file and changing the `enabled` field. We now have the following options:

* `-D <substring>` Turn off any rules with names having the substring `<substring>`. This option can be passed multiple times. It cannot be mixed with `-t`.
* `-T <tag>` Turn off any rules with a `tag=<tag>`. This option can be passed multiple times. This option can not be mixed with `-t`.
* `-t <tag>` Only enable those rules with a `tag=<tag>`. This option can be passed multiple times. This option can not be mixed with `-T`/`-D`.

There are two main issues here:
* Lack of flexibility, for instance if I want to enable an entire tag _except_ one rule I cannot do so: https://github.com/falcosecurity/falco/issues/2631#issuecomment-1588766322 
* These are only command line switches and cannot be specified in the configuration file `falco.yaml` or from the configmap

For this reason I would like to propose another way of accessing this data. The design is up for discussion and feedback is absolutely welcome.

**Feature Proposal**

I thought about a design that fixes the above two points. A new key in the `falco.yaml` version called `rules` (name can be discussed but I would like to keep it short) that allows to specify a sequence of "commands" to enable and disable rules. It would accept a triplet of `<command>:<object>:<name>`

example (rule and tag names are made up)
```yaml
rules:
  - only:tag:network
  - enable:rule:shell_in_container
  - enable:rule:k8s_*
  - disable:rule:connection_to_suspicious_server
```

At the end of this example, Falco would only consider networking tagged rules except for one `connection_to_suspicious_server`, plus all rules starting with `k8s_` and the `shell_in_container` rule.

The same configuration could also be specified on the command line with something like:

```
$ falco -o rules[]=only:tag:network -o rules[]=enable:rule:shell_in_container -o rules[]=enable:rule:k8s_* -o rules[]=disable:rule:connection_to_suspicious_server
```

Note that the syntax `rules[]` which allows to add an element to a sequence does not exist today and would need to be introduced.

Note that while in my opinion this mechanism provides a good degree of flexibility it may not be ideal and the design could be changed a lot before implementation. Comments are very welcome!

LucaGuerra:
**What type of PR is this?**

/kind feature

/area engine

This is a rather straightforward implementation of https://github.com/falcosecurity/falco/issues/3174#issuecomment-2064152163 . Essentially, it introduces two new ways of enabling/disabling rules without changing the rule files. Following the example:

```yaml
rules:
  - disable:
      rule: *
  - enable:
      tag: network
  - enable:
      rule: Directory traversal monitored file
  - enable:
      rule: k8s_*
  - disable:
      rule: k8s_noisy_rule
```

This means: disable everything, enable all rules tagged `networking`, also enable the rule called `Directory traversal monitored file`, then enable any rule matching the wildcard pattern `k8s_*` and disable `k8s_noisy_rule`.

You can achieve the same via the CLI

```
falco -o "rules[].disable.rule=*" -o "rules[].enable.tag=network" -o "rules[].enable.rule=Directory traversal monitored file
" -o "rules[].enable.rule=k8s_*" -o "rules[].disable.rule=k8s_noisy_rule"
```

The new syntax `[]` allows to append a new element at the end of sequences, which is how the CLI works in this case.

At this point, rule names support wildcard while tag names do not. I am a bit unsure about what to do with tag names. On one side, what you want to do is enable and disable them one by one so wildcards seem a bit too much there. On the other hand we currently allow to "intersect" the tags we want, such as only `networking` AND `exec`, which is not supported here. Perhaps we could add a `tags` option for that?

Fixes #3174 

```release-note
new(falco): allow selecting which rules to load from the configuration file or command line
```

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -177,7 +177,7 @@ rules_files:
   - /etc/falco/falco_rules.local.yaml
   - /etc/falco/rules.d
 
-# [Experimental] `rules`
+# [Incubating] `rules`
 #
 # --- [Description]
 #
--- a/falco.yaml
+++ b/falco.yaml
@@ -177,6 +177,39 @@ rules_files:
   - /etc/falco/falco_rules.local.yaml
   - /etc/falco/rules.d
 
+# [Experimental] `rules`
+#
+# --- [Description]
+#
+# Falco rules can be enabled or disabled by name (with wildcards *) and/or by tag.
+#
+# This configuration is applied after all rules files have been loaded, including
+# their overrides, and will take precedence over the enabled/disabled configuration
+# specified or overridden in the rules files.
+#
+# The ordering matters and selections are evaluated in order. For instance, if you
+# need to only enable a rule you would first disable all of them and then only
+# enable what you need, regardless of the enabled status in the files.
+#
+# --- [Examples]
+#
+# Only enable two rules:
+#
+# rules:
+#   - disable:
+#       rule: "*"
+#   - enable:
+#       rule: Netcat Remote Code Execution in Container
+#   - enable:
+#       rule: Delete or rename shell history
+#
+# Disable all rules with a specific tag:
+#
+# rules:
+#   - disable:
+#       tag: network
+#
+
 ################
 # Falco engine #
 ################
--- a/userspace/falco/configuration.cpp
+++ b/userspace/falco/configuration.cpp
@@ -85,6 +85,13 @@ void falco_configuration::init(const std::vector<std::string>& cmdline_options)
 	load_yaml("default");
 }
 
+void falco_configuration::init_from_content(const std::string& config_content, const std::vector<std::string>& cmdline_options)
+{
+	config.load_from_string(config_content);
+	init_cmdline_options(cmdline_options);
+	load_yaml("default");
+}
+
 void falco_configuration::init(const std::string& conf_filename, std::vector<std::string>& loaded_conf_files,
 			       const std::vector<std::string> &cmdline_options)
 {
@@ -549,6 +556,8 @@ void falco_configuration::load_yaml(const std::string& config_name)
 
 	m_metrics_convert_memory_to_mb = config.get_scalar<bool>("metrics.convert_memory_to_mb", true);
 	m_metrics_include_empty_values = config.get_scalar<bool>("metrics.include_empty_values", false);
+
+	config.get_sequence<std::vector<rule_selection_config>>(m_rules_selection, "rules");
 
 	std::vector<std::string> load_plugins;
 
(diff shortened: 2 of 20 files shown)


leogr:
> 
> 
> /hold

I guess this is a false positive.

incertum (approved):
/approve

leogr (commented):
SGTM

cc @jasondellaluce @FedeDP for a second look

falco.yaml
```diff
@@ -177,6 +177,39 @@ rules_files:
   - /etc/falco/falco_rules.local.yaml
   - /etc/falco/rules.d
 
+# [Experimental] `rules`
```
leogr, line 180:
```suggestion
# [Incubating] `rules`
```

As per the upcoming [adoption policy](https://github.com/falcosecurity/falco/blob/master/proposals/20231220-features-adoption-and-deprecation.md#adoption-policy), this should be labeled as "Incubating".

This is not a blocker for this PR, but maybe it is worth starting using new labels.

FedeDP: Agree

FedeDP:
As suggested by Andrea, i'd deprecate `-t,-T,-D` options so that we can remove them in Falco 0.39: https://github.com/falcosecurity/falco/issues/3174#issuecomment-2078841689

userspace/falco/configuration.cpp
```diff
@@ -90,6 +90,13 @@ void falco_configuration::init(const std::vector<std::string>& cmdline_options)
 	load_yaml("default");
 }
 
+void falco_configuration::init_from_content(const std::string& config_content, const std::vector<std::string>& cmdline_options)
```
FedeDP, line 93: I'd rather add the `config_content=""` optional argument to the old `init` method (the one above) instead of adding a completely new method.

FedeDP: I think they are both just used by tests anyway.

LucaGuerra: I was a bit nervous about adding an optional argument that completely changes the method behavior. Perhaps to make it cleaner I could split that into `init_from_file()` and `init_from_content()`, wdyt?

FedeDP: >  init_from_file() and init_from_content()

Like it!

LucaGuerra: Done! :)

FedeDP (approved):
/approve

incertum (approved):
/approve

PR #3184, opened 2024-05-05
refactor: smart pointer usage

federico-sysdig:
**What type of PR is this?**
/kind cleanup

No specific area.

There are some usages of smart pointers that can be simplified a but and, occasionally, made more efficient.

1. Use the overloaded dereference operator directly on the smart pointer without invoking get(), i.e. prefer *ptr rather than *(ptr.get()).
2. Create objects stored in smart pointers with make_unique or make_shared rather than using new. For shared_ptrs this has the performance advantage of storing the pointer's control block with the object itself, improving memory locality, beneficial for cache reasons, and using a single allocation call.
3. Prefer using automatic objects - i.e. stored on the stack - than heap allocated ones. This effectively removes the usage of the smart pointer.

Fixes #

```release-note
NONE
```

### DIFF
--- a/unit_tests/engine/test_alt_rule_loader.cpp
+++ b/unit_tests/engine/test_alt_rule_loader.cpp
@@ -263,11 +263,11 @@ static void load_rules(sinsp& inspector,
 	rule_loader::collector collector;
 	rule_loader::compiler compiler;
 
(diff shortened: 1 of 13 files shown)


jasondellaluce (approved):

FedeDP (approved):
/approve

PR #3187, opened 2024-05-08
new(metrics): add file sha256sum metrics for loaded config and rules files

incertum:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind release

> /area build

/area engine

> /area tests

> /area proposals

> /area CI

This PR adds the sha256sum for each loaded config and rules file as individual metric. These metrics complement existing informational metrics such as the Falco version or kernelrelease of the host and especially help to track deployment upgrade convergence and integrity. When hot reloading Falco and having watch config files setting enabled, the state is re-initialized and as such the new sha256sum is calculated.

Note: This PR only adds the new metrics, thus deferring future metrics code consolidations to the next release dev cycle.

Fixes #

```release-note
new(metrics): add file sha256sum metrics for loaded config and rules files
```

### DIFF
--- a/userspace/falco/app/actions/load_rules_files.cpp
+++ b/userspace/falco/app/actions/load_rules_files.cpp
@@ -84,7 +84,7 @@ falco::app::run_result falco::app::actions::load_rules_files(falco::app::state& 
 		{
 			falco_logger::log(falco_logger::level::WARNING,res->as_string(true, rc) + "\n");
 		}
-#if defined(__linux__)
+#if defined(__linux__) and !defined(MINIMAL_BUILD) and !defined(__EMSCRIPTEN__)
 		s.config->m_loaded_rules_filenames_sha256sum.push_back(falco::utils::calculate_file_sha256sum(filename));
 #endif
 	}
--- a/userspace/falco/app/actions/load_rules_files.cpp
+++ b/userspace/falco/app/actions/load_rules_files.cpp
@@ -85,7 +85,7 @@ falco::app::run_result falco::app::actions::load_rules_files(falco::app::state& 
 			falco_logger::log(falco_logger::level::WARNING,res->as_string(true, rc) + "\n");
 		}
 #if defined(__linux__) and !defined(MINIMAL_BUILD) and !defined(__EMSCRIPTEN__)
-		s.config->m_loaded_rules_filenames_sha256sum.push_back(falco::utils::calculate_file_sha256sum(filename));
+		s.config->m_loaded_rules_filenames_sha256sum.insert({filename, falco::utils::calculate_file_sha256sum(filename)});
 #endif
 	}
 
--- a/userspace/falco/app/actions/load_rules_files.cpp
+++ b/userspace/falco/app/actions/load_rules_files.cpp
@@ -17,6 +17,7 @@ limitations under the License.
 
 #include "actions.h"
 #include "helpers.h"
+#include "falco_utils.h"
 
 #include <libsinsp/plugin_manager.h>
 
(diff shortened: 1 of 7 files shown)


userspace/falco/configuration.cpp
```diff
@@ -160,6 +160,11 @@ void falco_configuration::merge_configs_files(const std::string& config_name, st
 			}
 		}
 	}
+
+	for(auto &filename : m_loaded_configs_filenames)
```
incertum, line 164: @FedeDP randomly asking you since you worked on this part recently. For `m_loaded_rules_filenames` etc we first clear the list, but we do not do this for the configs filenames list, is this intentional?

incertum:
TODO: Still need to check how to support recalculating the sha256sum for the configs hot reload cases to ensure it is always up to date.

incertum:
This is now ready for review.

userspace/falco/app/actions/load_rules_files.cpp
```diff
@@ -83,6 +84,9 @@ falco::app::run_result falco::app::actions::load_rules_files(falco::app::state&
 		{
 			falco_logger::log(falco_logger::level::WARNING,res->as_string(true, rc) + "\n");
 		}
+#if defined(__linux__) and !defined(MINIMAL_BUILD) and !defined(__EMSCRIPTEN__)
+		s.config->m_loaded_rules_filenames_sha256sum.push_back(falco::utils::calculate_file_sha256sum(filename));
```
FedeDP, line 88: What if `m_loaded_rules_filenames_sha256sum` (and `m_loaded_configs_filenames_sha256sum`) was instead a `unoredered_set<string>string`, where key is filename and value is sha256sum? 
This would allow metrics and stats writer to just iterate over `m_loaded_rules_filenames_sha256sum` (and `m_loaded_configs_filenames`) keys.

FedeDP (commented):
Question: why is all of this linux only?

userspace/falco/app/actions/load_rules_files.cpp
incertum: If that's more desirable, no problems to change it.

FedeDP: I think it would simplify a bit the falco_metrics and stats_writer loops, making them more readable.

incertum:
> Question: why is all of this linux only?

Main purpose of metrics is at runtime when running Falco on Linux. I am not sure why we would need it when we load a capture file on macOS or Windows. In addition, most metrics already only work on Linux. More thoughts?

Support for macOS or Windows likely requires  a different approach as that openssl lib I am using is not available.

userspace/falco/app/actions/load_rules_files.cpp
incertum: Ok.

FedeDP:
I was thinking if running Falco on eg: windows with plugins and their rules, one could still want the shasums in the metrics.
> Support for macOS or Windows likely requires a different approach as that openssl lib I am using is not available.

No problem then, we can introduce it later if someone needs it!

incertum:
> I was thinking if running Falco on eg: windows with plugins and their rules, one could still want the shasums in the metrics.
> 
> > Support for macOS or Windows likely requires a different approach as that openssl lib I am using is not available.
> 
> No problem then, we can introduce it later if someone needs it!

Yeah, right now it's actually not even working well for plugins only even on Linux. Needless to say, metrics still requires lots of work over the next n releases.

FedeDP (approved):
/approve

FedeDP:
>  CI Build / test-dev-packages / test-packages (pull_request) Failing after 4m 

I am investigating the `TestFalco_Legacy_KernelUpgrade` failure since it also fails on https://github.com/falcosecurity/falco/pull/3191

FedeDP (approved):
/approve

leogr (approved):

leogr:
> 
> 
> /hold

false positive
/unhold

PR #3190, opened 2024-05-09
fix(userspace/falco): fix state inizialization.

### LINKED ISSUE #3188: Bug: segfault on hot reload of configs in current falco master
When hot reloading a Falco rules file, Falco segfaults while actually reloading the config file aka at the very beginning of the (re-)init phase.

```
Thread 3 "falco" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffefe006c0 (LWP 119895)]
yaml_helper::load_from_file (path="../falco.yaml", this=0x390)
    at /falco/userspace/falco/yaml_helper.h:97
97			m_root = load_from_file_int(path);
Missing separate debuginfos, use: dnf debuginfo-install elfutils-libelf-0.191-2.fc39.x86_64 glibc-2.38-16.fc39.x86_64 libcap-2.48-9.fc39.x86_64 libstdc++-13.2.1-6.fc39.x86_64 sssd-client-2.9.4-1.fc39.x86_64 systemd-libs-254.10-1.fc39.x86_64 zlib-1.2.13-4.fc39.x86_64
(gdb) 
```

To repro just run current Falco master and then modify and save a rules file that was initially loaded.

Note: I tested it with tag 0.37.1 and there were no issues, therefore the bug was recently introduced. 
@FedeDP mind helping? Thanks!

This bug is blocking the following PR: https://github.com/falcosecurity/falco/pull/3187

FedeDP:
**What type of PR is this?**

/kind bug

/area engine

#3184 introduced a bug that causes Falco to segfault

Fixes #3188 

```release-note
fix(userspace/falco): fix state inizialization avoid a crash during hot reload
```

### DIFF
--- a/userspace/falco/app/state.h
+++ b/userspace/falco/app/state.h
@@ -70,10 +70,10 @@ struct state
     {
     }
 
-    state(const std::string& cmd, const falco::app::options& opts):
-        cmdline(cmd),
-        options(opts)
+    state(const std::string& cmd, const falco::app::options& opts): state()
     {
+	    cmdline = cmd;
+	    options = opts;
     }
 
     std::string cmdline;


userspace/falco/app/state.h
```diff
@@ -70,10 +70,10 @@ struct state
     {
     }
 
-    state(const std::string& cmd, const falco::app::options& opts):
-        cmdline(cmd),
-        options(opts)
+    state(const std::string& cmd, const falco::app::options& opts): state()
```
FedeDP, line 73: We weren't calling proper `state()` ctor anymore.

incertum (approved):
/approve

🎉

jasondellaluce (approved):

Issue #3132, opened 2024-03-08
[TRACKING] Expand Falco metrics w/ rules counters -> deprecate falco-exporter + new Grafana dashboard for Falco metrics


incertum:
For context on current Falco metrics refactor please see https://github.com/falcosecurity/falco/issues/3131.

`libs` will already feature a new flag called `METRICS_V2_RULE_COUNTERS`. Proposing to already prepare for this new metrics category in the upcoming Falco metrics refactor to fully support Prometheus @sgaist. Goal is to deprecate https://github.com/falcosecurity/falco-exporter.

In addition, we need to decide who implements this. I won't have time for Falco 0.38.0 🙃 .

@leogr @alacuku @Issif re Grafana dashboard: How do we want to go about the design? Would you want to collect feedback from various folks in order to distill the most useful metrics? Happy to post a list of the most critical metrics from my perspective. 

It would be cool to upstream our new Grafana dashboard to the CNCF Environmental Sustainability Technical Advisory Group (TAG ENV) - Green Reviews Working Group initiatives.

- https://github.com/falcosecurity/cncf-green-review-testing/issues/12
- https://github.com/falcosecurity/cncf-green-review-testing/discussions/14

incertum:
Dashboards still pending, cc @leogr and @alacuku perhaps track that in a new issue?

PR #3192, opened 2024-05-13
new(metrics): add `rules_counters_enabled` option

incertum:
Intended to replace https://github.com/falcosecurity/falco-exporter when used with Prometheus output

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind release

> /area build

/area engine

> /area tests

> /area proposals

> /area CI

new(metrics): add rules_counters_enabled option

Intended to replace https://github.com/falcosecurity/falco-exporter when used with Prometheus output.
CC @leogr and @jasondellaluce (current maintainers of https://github.com/falcosecurity/falco-exporter).

Fixes https://github.com/falcosecurity/falco/issues/3132

Fixes #

```release-note
new(metrics): add `rules_counters_enabled` option
```

### DIFF
--- a/userspace/engine/falco_utils.cpp
+++ b/userspace/engine/falco_utils.cpp
@@ -151,6 +151,18 @@ std::string calculate_file_sha256sum(const std::string& filename)
 }
 #endif
 
+std::string sanitize_metric_name(const std::string& name)
+{
+	std::string sanitized_name = name;
+	RE2::GlobalReplace(&sanitized_name, "[^a-zA-Z0-9_:]", "_");
+	RE2::GlobalReplace(&sanitized_name, "_+", "_");
+	if (!sanitized_name.empty() && sanitized_name.back() == '_')
+	{
+		sanitized_name.pop_back();
+	}
+	return sanitized_name;
+}
+
 std::string wrap_text(const std::string& in, uint32_t indent, uint32_t line_len)
 {
 	std::istringstream is(in);
--- a/userspace/engine/stats_manager.h
+++ b/userspace/engine/stats_manager.h
@@ -63,6 +63,24 @@ public:
 		const indexed_vector<falco_rule>& rules,
 		std::string& out) const;
 
+	// Getter functions
+	inline const std::atomic<uint64_t>& get_total() const
+	{
+		return m_total;
+	}
+
+	inline const std::vector<std::unique_ptr<std::atomic<uint64_t>>>& get_by_priority() const
+	{
+		return m_by_priority;
+	}
+
+	inline const std::vector<std::unique_ptr<std::atomic<uint64_t>>>& get_by_rule_id() const
+	{
+		return m_by_rule_id;
+	}
+
+
+private:
 	std::atomic<uint64_t> m_total;
 	std::vector<std::unique_ptr<std::atomic<uint64_t>>> m_by_priority;
 	std::vector<std::unique_ptr<std::atomic<uint64_t>>> m_by_rule_id;
--- a/userspace/engine/stats_manager.h
+++ b/userspace/engine/stats_manager.h
@@ -63,7 +63,6 @@ public:
 		const indexed_vector<falco_rule>& rules,
 		std::string& out) const;
 
-private:
 	std::atomic<uint64_t> m_total;
 	std::vector<std::unique_ptr<std::atomic<uint64_t>>> m_by_priority;
 	std::vector<std::unique_ptr<std::atomic<uint64_t>>> m_by_rule_id;
--- a/userspace/falco/falco_metrics.cpp
+++ b/userspace/falco/falco_metrics.cpp
@@ -166,7 +166,8 @@ std::string falco_metrics::to_text(const falco::app::state& state)
 
 			prometheus_metrics_converter.convert_metric_to_unit_convention(metric);
 			prometheus_text += prometheus_metrics_converter.convert_metric_to_text_prometheus(metric, "falcosecurity", "falco");
-			for (size_t i = 0; i < rule_stats_manager.get_by_rule_id().size(); i++)
+			const std::vector<std::unique_ptr<std::atomic<uint64_t>>>& rules_by_id = rule_stats_manager.get_by_rule_id();
+			for (size_t i = 0; i < rules_by_id.size(); i++)
 			{
 				auto rule = rules.at(i);
 				std::string rules_metric_name = "rules." + rule->name;
@@ -176,7 +177,7 @@ std::string falco_metrics::to_text(const falco::app::state& state)
 																	METRIC_VALUE_TYPE_U64,
 																	METRIC_VALUE_UNIT_COUNT,
 																	METRIC_VALUE_METRIC_TYPE_MONOTONIC,
-																	rule_stats_manager.get_by_rule_id()[i]->load());
+																	rules_by_id[i]->load());
 				prometheus_metrics_converter.convert_metric_to_unit_convention(metric);
 				const std::map<std::string, std::string>& const_labels = {
 					{"priority", std::to_string(rule->priority)},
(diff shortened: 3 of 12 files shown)


incertum:
Proposing

/milestone 0.38.0

so that we can start the deprecation cycle for https://github.com/falcosecurity/falco-exporter @leogr .

incertum:
@sgaist this PR includes super minor touch ups re variable naming for the Prometheus webserver code. 
In general once we start the dev cycle for Falco 0.39.0 we should chat re how to consolidate and simplify the Falco metrics model even more.

leogr:
> Proposing
> 
> /milestone 0.38.0
> 
> so that we can start the deprecation cycle for https://github.com/falcosecurity/falco-exporter @leogr .

If we are able to get the PR reviewed in time, +1 for me. I'll take a look as soon as I can.

/assign
/assign @FedeDP 
(as discussed)

userspace/falco/falco_metrics.cpp
```diff
@@ -86,26 +86,59 @@ std::string falco_metrics::to_text(const falco::app::state& state)
...
+																	METRIC_VALUE_UNIT_COUNT,
+																	METRIC_VALUE_METRIC_TYPE_MONOTONIC,
+																	rule_stats_manager.m_total.load()));
+
+			for (size_t i = 0; i < rule_stats_manager.m_by_rule_id.size(); i++)
+			{
+				auto rule = rules.at(i);
+				std::string rules_metric_name = "rules." + rule->name;
+				// Separate processing of rules counter metrics given we add extra tags
+				auto metric = libs_metrics_collector.new_metric(rules_metric_name.c_str(),
```
FedeDP, line 107: Out of scope for this PR; but since we only instantiate `libs_metrics_collector` to add metrics to it, that, given the code: https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/metrics_collector.h#L260
seems like it can be made static, if we made the change in libs (ie: make `new_metric` static), can we drop the `libs_metrics_collector` instance?

userspace/falco/falco_metrics.cpp
```diff
@@ -86,26 +86,59 @@ std::string falco_metrics::to_text(const falco::app::state& state)
 		{
 			prometheus_text += prometheus_metrics_converter.convert_metric_to_text_prometheus("evt_source", "falcosecurity", "falco", {{"evt_source", source}});
 		}
-		std::vector<metrics_v2> static_metrics;
-		static_metrics.push_back(libs_metrics_collector.new_metric("start_ts",
+		std::vector<metrics_v2> falco_metrics;
```
FedeDP, line 89: It would be nice to split this method into multiple logical pieces, like:
* scap_engines_to_text()
* falco_rules_to_text()
* agent_info_to_text()
*
...and so on.
Do you think it is feasible?

userspace/engine/stats_manager.h
```diff
@@ -63,7 +63,6 @@ class stats_manager
 		const indexed_vector<falco_rule>& rules,
 		std::string& out) const;
 
-private:
```
sgaist, line 66: I think this is the wrong move. Currently, this class is the sole manager of these values by providing a single entry point to update them. Making them all public means that any part of the code which has access to the stats_manager can change them which may have unpredictable side effects.
Having a dedicated read API would keep the responsibility within this class.

userspace/engine/stats_manager.h
```diff
@@ -63,7 +63,6 @@ class stats_manager
...
 	std::atomic<uint64_t> m_total;
 	std::vector<std::unique_ptr<std::atomic<uint64_t>>> m_by_priority;
 	std::vector<std::unique_ptr<std::atomic<uint64_t>>> m_by_rule_id;
```
sgaist, line 68: Outside of this PR scope but why unique pointers to atomic rather than just atomics ?

leogr: cc @jasondellaluce ?

userspace/engine/stats_manager.h
incertum: I almost anticipated that comment from someone 😆 sure I expand the API instead.

userspace/falco/falco_metrics.cpp
incertum: Of course. Bog question is now or wait until a more major cleanup follows post Falco 0.38.0, cc @sgaist?

userspace/falco/falco_metrics.cpp
incertum: Yup, I'll create an issue later to track all the post Falco 0.38.0 metrics items.

userspace/falco/falco_metrics.cpp
FedeDP: Much better, thanks!

FedeDP (approved):
/approve

userspace/falco/falco_metrics.cpp
```diff
@@ -134,34 +137,77 @@ std::string falco_metrics::to_text(const falco::app::state& state)
...
+			{
+				auto rule = rules.at(i);
+				std::string rules_metric_name = "rules." + rule->name;
+				// Separate processing of rules counter metrics given we add extra tags
+				auto metric = libs_metrics_collector.new_metric(rules_metric_name.c_str(),
+																	METRICS_V2_RULE_COUNTERS,
+																	METRIC_VALUE_TYPE_U64,
+																	METRIC_VALUE_UNIT_COUNT,
+																	METRIC_VALUE_METRIC_TYPE_MONOTONIC,
+																	rule_stats_manager.get_by_rule_id()[i]->load());
```
sgaist, line 179: What about minimizing the number of calls to get_by_rule_id ?

```suggestion
			auto rules_by_id = rule_stats_manager.get_by_rule_id();
			for (size_t i = 0; i < rules_by_id.size(); i++)
			{
				auto rule = rules.at(i);
				std::string rules_metric_name = "rules." + rule->name;
				// Separate processing of rules counter metrics given we add extra tags
				auto metric = libs_metrics_collector.new_metric(rules_metric_name.c_str(),
																	METRICS_V2_RULE_COUNTERS,
																	METRIC_VALUE_TYPE_U64,
																	METRIC_VALUE_UNIT_COUNT,
																	METRIC_VALUE_METRIC_TYPE_MONOTONIC,
																	rules_by_id[i]->load());
```

sgaist (changes requested):

userspace/falco/falco_metrics.cpp
incertum: Incorporated, thanks.

incertum:
Note that just occurred to me: Let me check in the morning if white spaces in the rules names cause issues for Prometheus. Perhaps let's replace white spaces of the rules names with underscore? Other ideas?

sgaist:
> Note that just occurred to me: Let me check in the morning if white spaces in the rules names cause issues for Prometheus. Perhaps let's replace white spaces of the rules names with underscore? Other ideas?

Based on the [OpenMetrics text format](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#text-format) the metric name should follow these rules:

```
metricname = metricname-initial-char 0*metricname-char

metricname-char = metricname-initial-char / DIGIT
metricname-initial-char = ALPHA / "_" / ":"
```
So white spaces would indeed be wrong. As a safety measure, I would recommend turning anything that does not fit the metric name definition into an underscore.

userspace/engine/falco_utils.cpp
```diff
@@ -151,6 +151,13 @@ std::string calculate_file_sha256sum(const std::string& filename)
 }
 #endif
 
+std::string sanitize_metric_name(const std::string& name)
+{
+	std::string sanitized_name = name;
+	RE2::GlobalReplace(&sanitized_name, "\\s+", "_");
```
sgaist, line 157: Are we sure that only white spaces are the problematic characters that can be found in a metric name ?

incertum: i read your comment after I pushed lol, repushed just now.

incertum:
@sgaist and @FedeDP ok added the proper helper (also for the config and rules sha256 metric names, just in case). I was about to go to sleep yesterday and then that thought came into my head. Good we fix that before merging indeed ...

incertum:
A rule name like `Testing rule   2 (CVE-2244)` becomes `falco.rules.Testing_rule_2_CVE_2244` now.

FedeDP (approved):
/approve

sgaist (approved):

PR #3206, opened 2024-05-20
chore(docs): added falco.yaml section about config keys maturity.

FedeDP:
**What type of PR is this?**

/kind cleanup

Also, rename `Experimental` -> `Incubating` and move `prometheus_metrics_enabled` to `Incubating`.

Fixes #

See the proposal: https://github.com/falcosecurity/falco/blob/master/proposals/20231220-features-adoption-and-deprecation.md 

```release-note
chore(docs): apply features adoption and deprecation proposal to config file keys
```

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -25,7 +25,7 @@
 #
 #     (Falco command-line arguments)
 #     (Falco environment variables)
-# Falco config files
+# Falco configs files
 #     configs_files
 # Falco rules files
 #     rules_files
@@ -69,6 +69,19 @@
 #     base_syscalls
 # Falco libs
 #     falco_libs
+
+########################
+# Config maturity tags #
+########################
+
+# As per features adoption and deprecation proposal we support 4 levels of configuration keys maturity:
+# * Sandbox -> Experimental/alpha features, not recommended for production use, can be removed at any time without further notice.
+# * Incubating -> Beta features, long-term support is not guaranteed.
+# * Stable -> General Availability (GA) features for which long-term support is expected.
+# * Deprecated -> Deprecated keys, soon to be removed.
+#
+# For more info, please take a look at the proposal: https://github.com/falcosecurity/falco/blob/master/proposals/20231220-features-adoption-and-deprecation.md.
+
 
 ################################
 # Falco command-line arguments #
@@ -138,7 +151,7 @@ configs_files:
 
 # [Stable] `rules_files`
 
-NOTICE: Before Falco 0.38, this config key was `rules_file` (singular form), which is now deprecated in favor of `rules_files` (plural form).
+# NOTICE: Before Falco 0.38, this config key was `rules_file` (singular form), which is now deprecated in favor of `rules_files` (plural form).
 #
 # Falco rules can be specified using files or directories, which are loaded at
 # startup.
@@ -525,7 +538,7 @@ json_include_tags_property: true
 # output mechanism. By default, buffering is disabled (false).
 buffered_outputs: false
 
-# [Experimental] `rule_matching`
+# [Incubating] `rule_matching`
 #
 # The `rule_matching` configuration key's values are:
 #  - `first`: Falco stops checking conditions of rules against upcoming event
@@ -749,6 +762,8 @@ webserver:
   # Can be an IPV4 or IPV6 address, defaults to IPV4
   listen_address: 0.0.0.0
   k8s_healthz_endpoint: /healthz
+  # [Incubating] `prometheus_metrics_enabled`
+  #
   # Enable the metrics endpoint providing Prometheus values
   # It will only have an effect if metrics.enabled is set to true as well.
   prometheus_metrics_enabled: false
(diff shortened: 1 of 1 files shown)


falco.yaml
```diff
@@ -749,6 +762,8 @@ webserver:
   # Can be an IPV4 or IPV6 address, defaults to IPV4
   listen_address: 0.0.0.0
   k8s_healthz_endpoint: /healthz
+  # [Incubating] `prometheus_metrics_enabled`
```
FedeDP, line 765: Am not sure about how to integrate the feature level label here, for a sub-option of a `Stable` config key.

falco.yaml
```diff
@@ -525,7 +538,7 @@ json_include_tags_property: true
 # output mechanism. By default, buffering is disabled (false).
 buffered_outputs: false
 
-# [Experimental] `rule_matching`
+# [Incubating] `rule_matching`
```
FedeDP, line 541: Experimental -> Incubating

FedeDP:
I was also wondering whether we could improve somehow the way maturity labels are attached, ie: perhaps moving them to the index at the top? Or moving them just the line before the configuration key, like:
```
# `configs_files`
#
# Falco will load additional configs files specified here.
# Their loading is assumed to be made *after* main config file has been processed,
# exactly in the order they are specified.
# Therefore, loaded config files *can* override values from main config file.
# Also, nested include is not allowed, ie: included config files won't be able to include other config files.
#
# Like for 'rules_files', specifying a folder will load all the configs files present in it in a lexicographical order.
configs_files:
#
# [Stable] 
  - /etc/falco/config.d
```

falco.yaml
```diff
@@ -25,7 +25,7 @@
 #
 #     (Falco command-line arguments)
 #     (Falco environment variables)
-# Falco config files
+# Falco configs files
```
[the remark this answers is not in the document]
FedeDP: Am not sure though, since we have `watch_config_files` too, and indeed we have multiple configuration (singular) files (plural), so `config_files` is not that bad after all, even if it somehow clashes with `rules_files` (but in this case, we have multiple files with multiple rules inside).
WDYT? Ie i am proposing to rename `configs_files` to `config_files`.
cc @leogr

falco.yaml
```diff
@@ -1170,7 +1185,7 @@ base_syscalls:
 # Falco libs #
 ##############
 
-# [Experimental] `falco_libs` - Potentially subject to more frequent changes
+# [Incubating] `falco_libs` - Potentially subject to more frequent changes
```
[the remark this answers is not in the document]
incertum: Agreed, remove it.

falco.yaml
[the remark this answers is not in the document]
incertum: So we keep `config` (singular form)? Works for me. Could we re-audit if we are consistent everywhere?

falco.yaml
incertum: urgh ok here we should maybe also introduce a synonym `rules_matching` @leogr?

incertum:
> I was also wondering whether we could improve somehow the way maturity labels are attached, ie: perhaps moving them to the index at the top? Or moving them just the line before the configuration key, like:
> 
> ```
> # `configs_files`
> #
> # Falco will load additional configs files specified here.
> # Their loading is assumed to be made *after* main config file has been processed,
> # exactly in the order they are specified.
> # Therefore, loaded config files *can* override values from main config file.
> # Also, nested include is not allowed, ie: included config files won't be able to include other config files.
> #
> # Like for 'rules_files', specifying a folder will load all the configs files present in it in a lexicographical order.
> #
> # [Stable] 
> configs_files:
>   - /etc/falco/config.d
> ```

@leogr we haven't responded to this yet. 

- Adding also to the index at the top, big +1
- Keeping `# [Stable] configs_files` was good would say, we can also add to the specific sub-config if needed. Don't have a strong opinion here. You can decide.

falco.yaml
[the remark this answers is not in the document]
FedeDP: Ok Leo is on board with changing `configs_files` to `config_files`. I will add a new commit for that.

falco.yaml
FedeDP: I think singular is ok, in this case; it is `rule-matching algorithm`, feels ok `rule_matching: foo` to me.

FedeDP:
Should've done everything:
* `configs_files` -> `config_files`
* `experimental` -> `incubating`
* removed `- Potentially subject to more frequent changes` as @leogr suggested
* Added maturity level to config index
* merged under same section `config_files` and `watch_config_files` settings

falco.yaml
```diff
@@ -25,50 +25,64 @@
...
-#     engine
+#     engine [Stable]
 # Falco plugins
-#     load_plugins
-#     plugins
-# Falco config files
-#     watch_config_files
+#     load_plugins [Stable]
+#     plugins [Stable]
 # Falco outputs settings
```
FedeDP, line 40: Note for eg: 1.0.0: we might add an `outputs.` namespace for these settings.
Similarly, for all the others, eg:
* `services.{grpc,webserver}`
* log.{stderr,syslog,level,libs}`
...

incertum (approved):
/approve

leogr (approved):
cc @LucaGuerra pls keep track of this, it may be useful in release comms.

PR #3208, opened 2024-05-20
refactor(userspace): move falco logger under falco engine

jasondellaluce:
**What type of PR is this?**

/kind bug

/kind cleanup

/area engine

Since introducing the method `evttype_index_ruleset::print_enabled_rules_falco_logger()`, we created an implicit dependency between the `falco_engine` target and the falco logger, which instead lives under the `falco_application` target. The two targets don't depend on each other and the linking succeeds just because they're both included by the `falco` cmake executable target. Pedantic linker checks will have this failing instead, and is fundamentally wrong even if it works by luck.

There are many ways for overcome this, but my proposal is to just move the falco logger as part of the falco engine's code. There are many instances in which logging would be useful and it hasn't been done exactly because of this uncomfortable dependency, so moving the component seems like an easy and appropriate fix IMO.

I'll let the reviewers decide on the right milestone. IMO this may be 0.38.0 material for better code quality.

```release-note
refactor(userspace): move falco logger under falco engine
```


jasondellaluce:
/unhold

False positive

FedeDP:
Question, what about creating an entire new target `falco_utils` (or whatever) that will be then liked by both `falco_application` and `falco_engine`?
I mean, aren't we still creating a link between `falco_engine` and `falco_application` with your solution?
Also, i can see people struggling to find source code for the falco logger that imo does not seem to belong to the engine, at a first glance at least.

FedeDP (approved):
/approve

FedeDP:
Yep let's go with this!
/milestone 0.38.0

leogr (approved):

Issue #2821, opened 2023-09-21
Regression: host metrics are not available running only source plugins 


Andreagit97:
**Describe the bug**

After this change, https://github.com/falcosecurity/libs/pull/1362 host metrics will no longer be available when running a source plugin like `k8saduit`. This is a regression and we need to find a way to address it.

Metrics extracted with dummy plugin in Falco 0.36.0-rc3

```json
{
	"output_fields": {
		"evt.source": "dummy",
		"evt.time": 1695381599561603364,
		"falco.container_memory_used": 0,
		"falco.cpu_usage_perc": 0.0,
		"falco.duration_sec": 1695381599,
		"falco.host_boot_ts": 0,
		"falco.host_num_cpus": 0,
		"falco.hostname": "",
		"falco.kernel_release": "",
		"falco.memory_pss": 25,
		"falco.memory_rss": 29,
		"falco.memory_vsz": 1779,
		"falco.num_evts": 18641280,
		"falco.num_evts_prev": 0,
		"falco.outputs_queue_num_drops": 0,
		"falco.start_ts": 0,
		"falco.version": "0.36.0-rc3",
		"scap.engine_name": "source_plugin"
	},
	"sample": 1
}
{
	"output_fields": {
		"evt.source": "dummy",
		"evt.time": 1695381604561553388,
		"falco.container_memory_used": 0,
		"falco.cpu_usage_perc": 0.1,
		"falco.duration_sec": 1695381604,
		"falco.evts_rate_sec": 3573155.7144059963,
		"falco.host_boot_ts": 0,
		"falco.host_num_cpus": 0,
		"falco.hostname": "",
		"falco.kernel_release": "",
		"falco.memory_pss": 25,
		"falco.memory_rss": 29,
		"falco.memory_vsz": 1908,
		"falco.num_evts": 36506880,
		"falco.num_evts_prev": 18641280,
		"falco.outputs_queue_num_drops": 0,
		"falco.start_ts": 0,
		"falco.version": "0.36.0-rc3",
		"scap.engine_name": "source_plugin"
	},
	"sample": 2
}

```

incertum:
Pitching one approach to get out of this regression here: https://github.com/falcosecurity/libs/pull/1362#issuecomment-2238352359

gnosek:
> Oh hey @gnosek
> 
> > Special notes for your reviewer:
> > This is definitely not the first time around the block with this, so I'm waiting for the inevitable "revert this" PR ;)
> 
[...]

PR #3230, opened 2024-05-31
fix(userspace/falco): fixed `falco_metrics::to_text` implementation when running with plugins

### LINKED ISSUE #3229: Scraping prometheus metrics endpoint crashes falco process
**Describe the bug**

Followed the prometheus support section in the docs to enable the /metrics endpoint. A soon as you make a request to this endpoint the falco process crashes without any indication on stdout/stderr what went wrong.

**How to reproduce it**

```
engine:
  kind: nodriver

metrics:
  enabled: true
  interval: 1h (also tried with 1m and waiting ~ 5 mins before /metrics request)
  output_rule: true
  rules_counters_enabled: false (tried with only this one enabled)
  resource_utilization_enabled: false (tried with only this one enabled)
  state_counters_enabled: false
  kernel_event_counters_enabled: false
  libbpf_stats_enabled: false
  convert_memory_to_mb: true
  include_empty_values: false

webserver:
  enabled: true
  listen_port: 8765
  prometheus_metrics_enabled: true
```

Enabled debug logging as well:
```
libs_logger:
  enabled: true
  severity: debug
log_level: debug
```

This doesn't give any output when requesting the /metrics endpoint. So, it is not helping narrowing down what it happening just before it crashes.

This is what happens with a port forward to the falco pod:
```
% curl localhost:8765/healthz
{"status": "ok"}

% curl localhost:8765/metrics
curl: (52) Empty reply from server
```

After the `/metrics` request has been done, Kubernetes shows pod termination:
```
      lastState:
        terminated:
          exitCode: 139
          reason: Error
```

**Expected behaviour**

Expected to see some metrics, or at least not to crash the entire falco process.

**Screenshots**

**Environment**

- Falco version: 0.38.0 (x86_64)

- System info:

- OS:

- Kernel:

- Installation method: Kubernetes (docker.io/falcosecurity/falco-no-driver image)

**Additional context**

This is a dedicated github plugin instance of falco (running as a k8s pod), i.e. not using syscall at all (`--disable-source syscall`).

```
2024-05-31T09:43:33+0000: Loaded event sources: syscall, github
2024-05-31T09:43:33+0000: Enabled event sources: github
2024-05-31T09:43:33+0000: Opening event source 'github'
```

No clue if this has anything to do with the crash though.

FedeDP:
**What type of PR is this?**

/kind bug

/area engine

This PR does 3 things:
* avoid crashing when scap_machine_info or scap_agent_info are NULL (ie: with non-linux scap platform, ie: while running plugins)
* properly use a `shared_ptr` for inspectors vector, so that we know they are kept alive while `to_text` runs and we can dereference the pointer
* only expose metrics for actually enabled sources (not loaded ones). Before, in `nodriver` mode with syscall source disabled, we would've tried to emit metrics for the syscall source/inspector too, but that is wrong since it is not enabled (and it would crash Falco too!)

Fixes #3229 

```release-note
fix(userspace/falco): fixed `falco_metrics::to_text` implementation when running with plugins
```

### DIFF
--- a/userspace/falco/falco_metrics.cpp
+++ b/userspace/falco/falco_metrics.cpp
@@ -53,20 +53,20 @@ std::string falco_metrics::to_text(const falco::app::state& state)
 		BPF_ENGINE, KMOD_ENGINE, MODERN_BPF_ENGINE,
 		SOURCE_PLUGIN_ENGINE, NODRIVER_ENGINE, GVISOR_ENGINE };
 
-	std::vector<sinsp*> inspectors;
+	std::vector<std::shared_ptr<sinsp>> inspectors;
 	std::vector<libs::metrics::libs_metrics_collector> metrics_collectors;
 
-	for (const auto& source_info: state.source_infos)
+	for (const auto& source: state.enabled_sources)
 	{
-		sinsp *source_inspector = source_info.inspector.get();
+		auto source_info = state.source_infos.at(source);
+		auto source_inspector = source_info->inspector;
 		inspectors.emplace_back(source_inspector);
-		metrics_collectors.emplace_back(libs::metrics::libs_metrics_collector(source_inspector, state.config->m_metrics_flags));
+		metrics_collectors.emplace_back(libs::metrics::libs_metrics_collector(source_inspector.get(), state.config->m_metrics_flags));
 	}
-
 	libs::metrics::prometheus_metrics_converter prometheus_metrics_converter;
 	std::string prometheus_text;
 
-	for (auto* inspector: inspectors)
+	for (auto inspector: inspectors)
 	{
 		// Falco wrapper metrics
 		//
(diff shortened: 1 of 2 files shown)


incertum (approved):
Right yes we knew that it was not working for plugin only since the scap regression.
Thanks @FedeDP for adding the safety checks.

incertum (dismissed):
/approve

incertum:
@FedeDP do we also check for null in the output rule code? If not maybe we should add some checks in there too given we have the scap regression?

incertum:
yes @FedeDP please also adjust below code 🙏 since we will continue having the scap regression for a while.

```
void stats_writer::collector::get_metrics_output_fields_wrapper(
		nlohmann::json& output_fields,
		const std::shared_ptr<sinsp>& inspector,
		const std::string& src, uint64_t num_evts,
		uint64_t now, double stats_snapshot_time_delta_sec)
{
	static const char* all_driver_engines[] = {
		BPF_ENGINE, KMOD_ENGINE, MODERN_BPF_ENGINE,
		SOURCE_PLUGIN_ENGINE, NODRIVER_ENGINE, GVISOR_ENGINE };
	const scap_agent_info* agent_info = inspector->get_agent_info();
	const scap_machine_info* machine_info = inspector->get_machine_info();
```

FedeDP:
> yes @FedeDP please also adjust below code 🙏 since we will continue having the scap regression for a while.

For some reason it was not failing on that locally :/ weird, but yep i add the extra checks!

incertum:
> indeed i can confirm that both agent_info and machine_info are not NULL here.

they are not initialized when running Falco with plugin only, see https://github.com/falcosecurity/falco/issues/2821 :/

incertum (approved):
/approve

userspace/falco/falco_metrics.cpp
```diff
@@ -53,20 +53,20 @@ std::string falco_metrics::to_text(const falco::app::state& state)
 		BPF_ENGINE, KMOD_ENGINE, MODERN_BPF_ENGINE,
 		SOURCE_PLUGIN_ENGINE, NODRIVER_ENGINE, GVISOR_ENGINE };
 
-	std::vector<sinsp*> inspectors;
+	std::vector<std::shared_ptr<sinsp>> inspectors;
 	std::vector<libs::metrics::libs_metrics_collector> metrics_collectors;
 
-	for (const auto& source_info: state.source_infos)
+	for (const auto& source: state.enabled_sources)
```
FedeDP, line 59: Oh well it turns out the issue was actually here: the old code also used the syscall inspector that was not even inited i guess, and thus had NULL agent and machine infos.
That's why stats_writer was not crashing.

I think leaving the checks is ok anyway, more protection for free :)

FedeDP: cc @incertum

incertum: Ahh understood, greta job spotting this and agreed more checks help here and also make it clearer that these 2 structs can be NULL in some circumstances.

FedeDP:
/milestone 0.39.0
(will probably become 0.38.1)

LucaGuerra (approved):

PR #3251, opened 2024-06-13
new(userspace/engine): generalize indexable ruleset

mstemm:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind release

> /area build

/area engine

> /area tests

> /area proposals

> /area CI

Now that custom rules loading implementations (and related, custom rulesets) can be swapped into falco in a customizable way, there is some functionality in evttype_index_ruleset that could be used by other rulesets, specifically the part that segregates filters by ruleset and enables/disables filters based on name substring + tags. This PR makes a base class for that. See the commit comments for more details.

Fixes #

NO

```release-note
NONE
```

### DIFF
--- a/userspace/engine/evttype_index_ruleset.cpp
+++ b/userspace/engine/evttype_index_ruleset.cpp
@@ -17,183 +17,18 @@ limitations under the License.
 
 #include "evttype_index_ruleset.h"
 
-#include "falco_utils.h"
-
 #include "logger.h"
 
 #include <algorithm>
 
 evttype_index_ruleset::evttype_index_ruleset(
-	std::shared_ptr<sinsp_filter_factory> f): m_filter_factory(f)
+	std::shared_ptr<sinsp_filter_factory> f):
+	m_filter_factory(f)
 {
 }
 
 evttype_index_ruleset::~evttype_index_ruleset()
 {
-}
-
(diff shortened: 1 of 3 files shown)


userspace/engine/evttype_index_ruleset.cpp
```diff
@@ -230,180 +66,48 @@ void evttype_index_ruleset::on_loading_complete()
...
-		m_rulesets.emplace_back(std::make_shared<ruleset_filters>());
-	}
-
-	for(const auto &wrap : m_filters)
-	{
-		bool matches;
-		std::string::size_type pos;
-
-		switch(match)
+		std::shared_ptr<wrapper> derived = std::dynamic_pointer_cast<wrapper>(wrap);
```
jasondellaluce, line 73: A dynamic cast for every rules index bucket, for every event, seems like a bit overkill for performance to me. I like this modularization effort, but can we choose a more performant route? For example, maybe the wrapper type can be a template parameter?

mstemm: I can static_cast it instead, the wrapper will always be the derived type anyway. I switched it to static_cast.

jasondellaluce (approved):
/approve

FedeDP (approved):
/approve

Issue #3258, opened 2024-06-20
Create `container_engines` configs in `falco.yaml`


incertum:
**Motivation**

See https://github.com/falcosecurity/falco/issues/3243#issuecomment-2174224867 and additional comments after that comment.

The current favorite seems to expose new `container_engines` in the following format:

```
container_engines:
    docker:
      enabled: true
    cri:
      enabled: true
      cri: ["/run/containerd/containerd.sock", "/run/crio/crio.sock", "/run/k3s/containerd/containerd.sock"]
      disable-cri-async: false
```

This feature will allow end users to explicitly disable some container engines, plus it can help in deployment scenarios where the existing `--cri` and `--disable-cri-async` CLI flags are more difficult to configure.

PR #3266, opened 2024-06-27
new(config): add `container_engines` config to falco.yaml

incertum:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind release

> /area build

/area engine

> /area tests

> /area proposals

> /area CI

Fixes https://github.com/falcosecurity/falco/issues/3258

Fixes #

```release-note
new(config): add `container_engines` config to falco.yaml
```


userspace/falco/configuration.cpp
```diff
@@ -616,6 +619,41 @@ void falco_configuration::load_yaml(const std::string& config_name)
 	}
 
 	m_watch_config_files = config.get_scalar<bool>("watch_config_files", true);
+
+	m_container_engines_mask = 0;
```
FedeDP, line 623: No need for this, right? Since it is already inited to 0.
```suggestion
```

falco.yaml
```diff
@@ -1206,33 +1206,44 @@ base_syscalls:
...
+# to disable asynchronous fetching. Note that missing fields may still occasionally occur.
+# 
+# The equivalent (stable) CLI args are `--cri` or `--disable-cri-async`.
+
+container_engines:
+  docker:
+    enabled: true
+  cri:
+    enabled: true
+    cri: ["/run/containerd/containerd.sock", "/run/crio/crio.sock", "/run/k3s/containerd/containerd.sock"]
```
FedeDP, line 1238:
```suggestion
    sockets: ["/run/containerd/containerd.sock", "/run/crio/crio.sock", "/run/k3s/containerd/containerd.sock"]
```

falco.yaml
```diff
@@ -1206,33 +1206,44 @@ base_syscalls:
...
+    disable-cri-async: false
```
FedeDP, line 1239:
```suggestion
    disable-async: false
```

FedeDP:
I'd also add a debug print of enabled container engines during the startup of Falco ;)

FedeDP:
> And / or likely more tests in libs make more sense given we just call some API here?

Yes, agree! Also, you could add some tests to https://github.com/falcosecurity/testing/blob/main/tests/falco/config_test.go; for example you could test that enabling each container engine yields a debug log (aforementioned) with the requested engine enabled.

falco.yaml
```diff
@@ -1206,33 +1206,44 @@ base_syscalls:
...
+  podman:
+    enabled: true
+  lxc:
+    enabled: true
+  libvirt_lxc:
+    enabled: true
+  rocket:
+    enabled: true
```
leogr, line 1247: AFAIK `rocket` was the initial name, but its official name is `rkt`

Still, `rkt` was discontinued in 2020 (https://github.com/rkt/rkt), so I don't see any compelling reason to keep it.

@falcosecurity/libs-maintainers wdyt? :thinking:

incertum: Oh I didn't know, yes can change it to `rkt`.

falco.yaml
incertum: @leogr? Re both names we ok with having them different to the CLI args?

userspace/falco/app/actions/init_inspectors.cpp
```diff
@@ -29,18 +29,41 @@ static void init_syscall_inspector(falco::app::state& s, std::shared_ptr<sinsp>
...
+			falco_logger::log(falco_logger::level::DEBUG, "Enabled container runtime socket at '" + p + "' via CLI args");
 		}
 	}
 
 	// Decide whether to do sync or async for CRI metadata fetch
 	inspector->set_cri_async(!s.options.disable_cri_async);
 
+	if(s.options.disable_cri_async || s.config->m_container_engines_disable_cri_async)
+	{
+		falco_logger::log(falco_logger::level::DEBUG, "Disabling async lookups for 'CRI'");
```
incertum, line 64: Would be nice to extend the API and add a `is_cri_async` and a `get_cri_socket_paths` methods. If you also think it would be good I can do it during this dev cycle, but no rush.

falco.yaml
FedeDP: > @falcosecurity/libs-maintainers wdyt? 🤔

Agree, we can clean it up; +1 from me!

incertum: Removed it. It's not relevant for Falco as you pointed out.

FedeDP:
Aside from :point_up: , LGTM!

incertum:
> So, `TestFalco_Legacy_TimeIso8601` is failing because it sees:
> 
> ```
> Fri Jun 28 21:37:49 2024: Enabled container engine 'docker'
>         	            	Fri Jun 28 21:37:49 2024: Enabled container engine 'podman'
>         	            	Fri Jun 28 21:37:49 2024: Enabled container engine 'CRI'
>         	            	Fri Jun 28 21:37:49 2024: Enabled container engine 'lxc'
>         	            	Fri Jun 28 21:37:49 2024: Enabled container engine 'libvirt_lxc'
>         	            	Fri Jun 28 21:37:49 2024: Enabled container engine 'bpm'
> ```
> 
> log before the initial:
> 
> > ```
> >     	            	2024-06-28T21:37:49+0000: Falco version: 0.39.0-33+0dfb7b8 (x86_64)
> > ```
> 
> string. I don't get it though, since the `Falco version` one is logged by `load_config` that is the first `app::run` step, while your new logs come from `init_inspectors` step.

The test should not be dependent on the order? If so, we should adjust the test anyways. Above ones are generated while parsing the config.

FedeDP:
> The test should not be dependent on the order? 

Yes but the test enforces that the very first log line should be `Falco version...`, and that makes sense from an UX perspective IMHO. 
I'd move the logs here from config loading to inspector init time (also if you think that makes sense because in that case, we would only log when the syscall inspector is actually in use, but we won't log if we are eg: plugin only).

FedeDP:
We need to fix #3270 in order to fix CI, since centos7 is eol.

leogr:
> We need to fix #3270 in order to fix CI, since centos7 is eol.

Is https://github.com/falcosecurity/falco/pull/3274 enough? :thinking:

FedeDP:
Hey Melissa, we were finally able to bump libs in Falco master, can you rebase this one? :)

userspace/falco/app/actions/init_inspectors.cpp
incertum: https://github.com/falcosecurity/libs/issues/2022 also added a label "good first issue" if someone wanted to get started helping us.

incertum:
> Hey Melissa, we were finally able to bump libs in Falco master, can you rebase this one? :)

Thanks @FedeDP

userspace/falco/configuration.cpp
```diff
@@ -82,7 +82,10 @@ falco_configuration::falco_configuration():
 	m_metrics_output_file(""),
 	m_metrics_flags(0),
 	m_metrics_convert_memory_to_mb(true),
-	m_metrics_include_empty_values(false)
+	m_metrics_include_empty_values(false),
+	m_container_engines_mask(0),
+	m_container_engines_cri_socket_paths({"/run/containerd/containerd.sock", "/run/crio/crio.sock","/run/k3s/containerd/containerd.sock"}),
```
FedeDP, line 87: Not for now but i'd expose these defaults as static strings from libs cri namespace, like: `libsinsp::container_engines::cri::s_default_socket_paths`.

userspace/falco/configuration.cpp
FedeDP, line 1: The schema_json in this file must be updated to match the newly added config key.

FedeDP: This will also fix CI of course.

LucaGuerra (approved):

PR #3273, opened 2024-07-04
new(docker): streamline docker images

### LINKED ISSUE #3165: [Proposal] Streamlining docker images
**Motivation**

The modern eBPF probe is gaining prominence, prompting our decision to adopt it as the [default driver](https://falco.org/docs/roadmap/#drivers). This shift offers significant advantages to our distribution system. By moving to the eBPF probe, we eliminate the need to include the full driver-building toolchain in the default Falco distribution. Users who prefer alternative setups can still opt for a different container image. Consequently, this allows us to transition the Falco default image to a ["no-driver"/"distroless" configuration](https://falco.org/docs/roadmap/#distribution), streamlining our deployment and enhancing system simplicity.

As a result of this decision, we have re-audited all Docker images, gathered feedback, and are now proposing to streamline the Falco Docker images to better meet current needs and enhance their functionality. This proposal outlines the desired state of these improvements.

**Feature**

## New supported images

| Image Name | Tag (aliases) | Description |
|---|---|---|
| falcosecurity/falco | _x.y.z_ (latest) | Distroless image without driver building toolchain support, based on the latest released tar.gz of Falco.  No tools or `falcoctl` not included.|
| falcosecurity/falco | _x.y.z_-debian  (latest-debian) | Debian-based image without driver building toolchain support, based on the latest released Deb of Falco (similar to the old _falcosecurity/falco-no-driver_ image). May include some tools (ie. `jq`, `curl`), but not `falcoctl`.  |
| falcosecurity/falco-driver-loader | _x.y.z_  (latest) | Based on `falcosecurity/falco:x.y.z-debian` (see above) plus the driver building toolchain support and also shipping the latest version of `falcoctl`. Recommended only when modern eBPF is not supported.   |
| falcosecurity/falco-driver-loader | _x.y.z_-buster  (latest) | Similar to `falcosecurity/falco-driver-loader` (see above) but based on a legacy Debian image (i.e. `buster` ).  Recommended only for old kernel versions  | 

## Deprecated images

| Image Name | Reason |
|---|---|
|falcosecurity/falco-distroless|Deprecated in favor of `falcosecurity/falco:x.y.z` |
|falcosecurity/falco-no-driver| Deprecated in favor of `falcosecurity/falco:x.y.z-debian` (basically the same image with a new name)|

Other legacy or almost not-used images included in https://hub.docker.com/u/falcosecurity may be deprecated, too.

**Alternatives**

We have already evaluated different alternatives, but no valid alternatives emerged (or were discarded early). 
A still valid alternative would be not to implement some image variants (i.e., `-debian` and `-buster`) or add others variants. This, however, may reduce the Falco compatibility range.

**Additional context**

This needs to be shipped before Falco 1.0 as per our roadmap.

Tentatively for
/milestone 0.38.0

Probably for 0.39.0

FedeDP:
**What type of PR is this?**

/kind feature

/area build

Fixes #3165

TODO:
- [x] falcoctl must not be shipped with Falco -> for now workarounded by using `rm -rf` (only in the Falco image)
- [ ] falco packages shall not have any dep (split packages in falco-kmod, falco-ebpf, and just falco (modern)), otherwise falco-debian image will ship gcc/dkms... too as deps of the Falco package
- [ ] update docs

```release-note
new(docker): streamline docker images
```


FedeDP:
For 
> falco packages shall not have any dep (split packages in falco-kmod, falco-ebpf, and just falco (modern)), otherwise falco-debian image will ship gcc/dkms... too as deps of the Falco package

We might want to install the Falco package without installing any deps, perhaps. I am not sure whether it is going to work though, but i guess it should.

docker/falco/Dockerfile
```diff
@@ -1,67 +1,36 @@
...
+    https://download.falco.org/packages/${VERSION_BUCKET}/$(uname -m)/falco-${FALCO_VERSION_URLENCODED}-$(uname -m).tar.gz && \
+    tar -xvf falco.tar.gz && \
+    rm -f falco.tar.gz && \
+    mv falco-${FALCO_VERSION}-$(uname -m) falco && \
+    rm -rf /falco/usr/src/falco-* && \
+    cp -r /falco/* / && \
+    rm -rf /falco
 
-COPY ./docker-entrypoint.sh /
+RUN sed -i -e 's/time_format_iso_8601: false/time_format_iso_8601: true/' /etc/falco/falco.yaml
```
leogr, line 31: Is there any compelling reason not to put `time_format_iso_8601: true` in the config file by default?
Likely, yes, but I can't remember why we need to do this anymore. 

@falcosecurity/falco-maintainers any opinion in this regard?

docker/falco/Dockerfile
```diff
@@ -1,67 +1,36 @@
...
-	zstd \
-	&& rm -rf /var/lib/apt/lists/*
-
-RUN curl -s https://falco.org/repo/falcosecurity-packages.asc | apt-key add - \
-	&& echo "deb https://download.falco.org/packages/${VERSION_BUCKET} stable main" | tee -a /etc/apt/sources.list.d/falcosecurity.list \
-	&& apt-get update -y \
-	&& if [ "$FALCO_VERSION" = "latest" ]; then FALCO_DRIVER_CHOICE=none apt-get install -y --no-install-recommends falco; else FALCO_DRIVER_CHOICE=none apt-get install -y --no-install-recommends falco=${FALCO_VERSION}; fi \
-	&& apt-get clean \
-	&& rm -rf /var/lib/apt/lists/*
+RUN apk update && apk add curl ca-certificates jq libelf libstdc++
```
leogr, line 17: Do we want to ship `curl` and `jq` in the main container image?

It may make sense since they can be used by the `program` output. 

If the consensus is to keep them, I will document this (so that we won't remove these two tools anymore from the image).

Otherwise, if we don't want to keep them, I recommend using a multi-stage build and not including them in the last stage.

docker/falco/Dockerfile
```diff
@@ -1,67 +1,36 @@
-FROM debian:bookworm
+FROM cgr.dev/chainguard/wolfi-base
 
 LABEL maintainer="cncf-falco-dev@lists.cncf.io"
 LABEL org.opencontainers.image.source="https://github.com/falcosecurity/falco"
 
-LABEL usage="docker run -i -t --privileged -v /var/run/docker.sock:/host/var/run/docker.sock -v /dev:/host/dev -v /proc:/host/proc:ro -v /boot:/host/boot:ro -v /lib/modules:/host/lib/modules:ro -v /usr:/host/usr:ro -v /etc:/host/etc --name NAME IMAGE"
+LABEL usage="docker run -i -t --privileged -v /var/run/docker.sock:/host/var/run/docker.sock -v /dev:/host/dev -v /proc:/host/proc:ro --name NAME IMAGE"
```
leogr, line 6:
```suggestion
LABEL usage="docker run -i -t --privileged -v /var/run/docker.sock:/host/var/run/docker.sock -v /proc:/host/proc:ro  -v /etc:/host/etc:ro falcosecurity/falco:latest"
```

Two suggestions here:
- Now, we use the modern eBPF by default, so I've aligned it with the [official doc](https://falco.org/docs/install-operate/running/#docker-privileged)
- `--name NAME IMAGE` is perhaps confusing. On the other hand, the full command with the correct image name can help to try it out

Further thought: If we consistently make these small changes across all `usage` labels, we may rely on the system to automatically generate the documentation.

docker/driver-loader/docker-entrypoint.sh
```diff
@@ -136,3 +136,5 @@ else
 fi
 
 /usr/bin/falcoctl driver install --compile=$ENABLE_COMPILE --download=$ENABLE_DOWNLOAD --http-insecure=$HTTP_INSECURE --http-headers="$FALCOCTL_DRIVER_HTTP_HEADERS" $extra_args
+
+exec /usr/bin/falco
```
leogr, line 140: Why can't we use `CMD` in the docker file? :thinking:

leogr (commented):
Great job :clap: 

Left a few comments, just nitpicking :angel:

leogr:
Tentatively for 
/milestone 0.39.0

docker/falco/Dockerfile
FedeDP: I don't really know, i think we can enable it by default without many troubles.

docker/falco/Dockerfile
FedeDP: `curl` (and `ca-certificates`) is also needed to download falco.tar.gz in the subsequent step though.

FedeDP: We can remove it afterwards, perhaps?

docker/driver-loader/docker-entrypoint.sh
FedeDP: How is it supposed to work? Unless `exec "$@"` as the entrypoint latest command; is this what you are referring to?

docker/falco/Dockerfile
FedeDP: Done!

docker/driver-loader/docker-entrypoint.sh
FedeDP: Done (using `exec "$@"` in the entrypoint).

docker/driver-loader-buster/Dockerfile
```diff
@@ -123,7 +123,12 @@ RUN curl -L -o binutils_2.30-22_${TARGETARCH}.deb https://download.falco.org/dep
 	&& curl -L -o binutils-common_2.30-22_${TARGETARCH}.deb https://download.falco.org/dependencies/binutils-common_2.30-22_${TARGETARCH}.deb \
 	&& dpkg -i *binutils*.deb \
 	&& rm -f *binutils*.deb
+	
+# Install latest falcoctl
+RUN curl -L -o falcoct.tar.gz $(curl -s "https://api.github.com/repos/falcosecurity/falcoctl/releases/latest" | jq -r '.assets[] | select(.name|test(".linux_$TARGETARCH.tar.gz")) | .browser_download_url') && tar -xvf falcoctl.tar.gz && mv falcoctl /usr/bin
```
leogr, line 128:
```suggestion
```

Since we installed Falco using .deb (which already includes `falcoctl`), why should we replace it with the latest version?

In the end, falcoctl and Falco versions provided by the deb package are guaranteed to work together, whereas the latest falcoctl version might not.

Unless I'm missing something, I'd say not to overwrite falcoctl.

docker/falco/Dockerfile
```diff
@@ -1,67 +1,36 @@
-FROM debian:bookworm
+FROM cgr.dev/chainguard/wolfi-base
```
leogr, line 1: Is there any compelling reason not to use a multi-stage build here? :thinking: 

The original [docker/no-driver/Dockerfile.distroless](https://github.com/falcosecurity/falco/pull/3273/files#diff-150a50aeb7255c76b8e25342f071be4d122cc40f1441edc16d4530c319c39078) did so

leogr (commented):
I did a second pass, and I think we are close :) 

See comments below.

docker/driver-loader-buster/Dockerfile
FedeDP: Makes sense!

docker/driver-loader/Dockerfile
```diff
@@ -1,14 +1,52 @@
...
+	llvm \
+	make \
+	netcat-openbsd \
+	patchelf \
+	xz-utils \
+	zstd \
+	&& rm -rf /var/lib/apt/lists/*
+	
+# Install latest falcoctl
+RUN curl -L -o falcoct.tar.gz $(curl -s "https://api.github.com/repos/falcosecurity/falcoctl/releases/latest" | jq -r '.assets[] | select(.name|test(".linux_$TARGETARCH.tar.gz")) | .browser_download_url') && tar -xvf falcoctl.tar.gz && mv falcoctl /usr/bin
```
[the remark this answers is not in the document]
FedeDP: In this case, we need to download falcoctl because `falco-debian` does a `rm -rf /usr/bin/falcoctl /etc/falcoctl/` because we do not want to ship falcoctl with the Falco image.
Perhaps, for now, we can leave `falcoctl` in the falco-debian image, and just `rm -rf` it from the `falco` image?

docker/falco/Dockerfile
FedeDP: It would make sense if we decided to drop `jq`, `curl` and `ca-certificates` packages from the final image; but we decided to keep them since they are useful for Falco output_program.

docker/driver-loader/Dockerfile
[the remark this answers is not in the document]
leogr: got it, you're right 👍 

> Perhaps, for now, we can leave falcoctl in the falco-debian image, and just rm -rf it from the falco image?
Good idea, this is the simplest solution for now, IMO. We can keep it and the we will see.

Thank you!

docker/README.md
```diff
@@ -4,15 +4,9 @@ This directory contains various ways to package Falco as a container and related
...
+| [falcosecurity/falco-driver-loader:latest](https://hub.docker.com/repository/docker/falcosecurity/falco-driver-loader), [falcosecurity/falco-driver-loader:_tag_](https://hub.docker.com/repository/docker/falcosecurity/falco-driver-loader), [falcosecurity/falco-driver-loader:master](https://hub.docker.com/repository/docker/falcosecurity/falco-driver-loader)                      | docker/driver-loader        | Based on falcosecurity/falco:x.y.z-debian (see above) plus the driver building toolchain support and also shipping the latest version of falcoctl. Recommended only when modern eBPF is not supported.                     |
```
leogr, line 11:
```suggestion
| [falcosecurity/falco-driver-loader:latest](https://hub.docker.com/repository/docker/falcosecurity/falco-driver-loader), [falcosecurity/falco-driver-loader:_tag_](https://hub.docker.com/repository/docker/falcosecurity/falco-driver-loader), [falcosecurity/falco-driver-loader:master](https://hub.docker.com/repository/docker/falcosecurity/falco-driver-loader)                      | docker/driver-loader        | Based on falcosecurity/falco:x.y.z-debian (see above) plus the driver building toolchain support and falcoctl. This is intended to be used as an installer or an init container when modern eBPF cannot be used.                     |
```

leogr (commented):
Hey @FedeDP 

I've realized that running Falco within the driver loader differs from the original proposal's spirit. Also, using `CMD` would not have allowed arguments to be passed to the driver loader.

So, I've reverted those changes and reviewed command line usage examples.

I've also updated the documentation to reflect this.


Please take a look at the the suggested changes below.

leogr (dismissed):

leogr (approved):

docker/README.md
```diff
@@ -4,15 +4,10 @@ This directory contains various ways to package Falco as a container and related
...
+| [falcosecurity/falco:latest](https://hub.docker.com/repository/docker/falcosecurity/falco), [falcosecurity/falco:_tag_](https://hub.docker.com/repository/docker/falcosecurity/falco), [falcosecurity/falco:master](https://hub.docker.com/repository/docker/falcosecurity/falco)                                                                                                          | docker/falco                | Distroless image based on the latest released tar.gz of Falco. 
+No tools are included in the image.                                                                              |
```
ekoops, line 10: Shouldn't this line be in the description of the previous raw?

docker/driver-loader-buster/Dockerfile
```diff
@@ -126,4 +126,4 @@ RUN curl -L -o binutils_2.30-22_${TARGETARCH}.deb https://download.falco.org/dep
 
 COPY ./docker-entrypoint.sh /
 
-ENTRYPOINT ["/docker-entrypoint.sh"]
+ENTRYPOINT ["/docker-entrypoint.sh"]
```
ekoops, line 129: Could you please add a newline here?

ekoops (changes requested):
Just left a couple of comments.

docker/README.md
leogr: good catch - fixing

leogr, line 10:
```suggestion
| [falcosecurity/falco:latest](https://hub.docker.com/repository/docker/falcosecurity/falco), [falcosecurity/falco:_tag_](https://hub.docker.com/repository/docker/falcosecurity/falco), [falcosecurity/falco:master](https://hub.docker.com/repository/docker/falcosecurity/falco)                                                                                                          | docker/falco                | Distroless image based on the latest released tar.gz of Falco. No tools are included in the image.                                                                              |
```

leogr (dismissed):

sgaist (dismissed):

docker/falco-debian/Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM debian:12-slim
+
+LABEL maintainer="cncf-falco-dev@lists.cncf.io"
+LABEL org.opencontainers.image.source="https://github.com/falcosecurity/falco/docker/falco-debian"
+
+LABEL usage="docker run -i -t -v /var/run/docker.sock:/host/var/run/docker.sock -v /dev:/host/dev -v /proc:/host/proc:ro falcosecurity/falco:latest-debian"
```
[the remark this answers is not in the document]
leogr: fixed

leogr (approved):

sgaist (approved):

Issue #3270, opened 2024-07-02
CI pipeline uses CentOS 7 which is EOL


johananl:
The following CI pipeline uses CentOS 7 which is **EOL** now:

https://github.com/falcosecurity/falco/blob/c25ded8f393ddd89f46f330f6e53c806d5bf2423/.github/workflows/reusable_build_packages.yaml#L55

In addition to being EOL, CentOS 7 uses RPM `4.11.x` which uses an old signing method in the RPM binaries it produces which can cause problems when installing Falco in environments with strict security requirements.

Relevant info in the RPM release notes: https://rpm.org/wiki/Releases/4.14.0

It would be great if the pipeline could be updated to some current RPM-based distro (no strong opinions on the specific distro as long as it contains RPM >= `4.14.x` which uses the modern signing method).

Thanks! :pray:

FedeDP:
Hi! Thanks for opening this issue! I agree, we need to get rid of `centos:7`. At the same time, we really care and need the oldest possible glibc version (and moving to something newer could break usage for people on old glibc versions).
We need to dig into this to find a non-breaking alternative.
cc @falcosecurity/falco-maintainers

FedeDP:
/milestone 0.39.0
We need to fix this ASAP because it is breaking our CI builds.

FedeDP:
So, centos7 used glibc 2.17; ubuntu 18.04 uses glibc 2.27, ubuntu 16.04 uses glibc 2.23.

FedeDP:
For now, we switched to use `vault.centos.org` mirrors to keep centos7 CI running.
We need to figure out either a glibc bump or something else.

FedeDP:
A Better solution is being developed in #3307 .
/milestone 0.40.0

johananl:
Thank you for addressing this!

PR #3274, opened 2024-07-04
fix(ci): use vault.centos.org for centos:7 CI build.

FedeDP:
**What type of PR is this?**

/kind bug

/area CI

Enforce the usage of the vault.centos.org mirrors for centos:7 CI. This should fix the CI and let us gain time to design a proper solution to drop centos7.

Also, recently github actions enforced the usage of node20 runners, breaking our `actions/checkout` step when ran inside centos:7 container (glibc version mismatch); workaround that issue too, by adding the env `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true` to the workflow.
See https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/.

Fixes #

```release-note
fix(ci): use vault.centos.org for centos:7 CI build.
```

### DIFF
--- a/.github/workflows/reusable_build_packages.yaml
+++ b/.github/workflows/reusable_build_packages.yaml
@@ -57,7 +57,18 @@ jobs:
       # Always install deps before invoking checkout action, to properly perform a full clone.
       - name: Install build dependencies
         run: |
+          # fix broken mirrors
+          sed -i s/mirror.centos.org/vault.centos.org/g /etc/yum.repos.d/*.repo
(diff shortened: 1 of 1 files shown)


FedeDP:
Refs #3270

FedeDP:
It seems like arm64 centos-release-scl repo is missing from vault.centos.org :/

LucaGuerra (approved):

leogr (approved):

PR #3305, opened 2024-08-26
feat(cmake): add conditional builds for falcoctl and rules paths

tembleking:
There are systems, like Nix derivations where the build process does not have network access in order to enforce reproducibility. This patch allows people building Falco to optionally skip the build of falcoctl with `-DADD_FALCOCTL_DEPENDENCY=OFF` and point to their own self-backed, or pre-fetched rules files with `-DFALCOSECURITY_RULES_FALCO_PATH=<some-path>` and `-DFALCOSECURITY_RULES_LOCAL_PATH=<some-path>`.

For context, I needed to apply these patches while building the project with Nix in https://github.com/tembleking/falco-nix but I think that would be benefitial for the community to have also these options open, and that would also make Falco feasible to be added to the nixpkgs repository at https://github.com/nixos/nixpkgs

/kind feature
/kind release
/area build

```release-note
- When building Falco, users can now disable the build of `falcoctl` with the engine, by specifying the CMake flag `-DADD_FALCOCTL_DEPENDENCY=OFF`.
- When building Falco, users can now specify the source paths of both the rules files, and the local rules files with the CMake flags `-DFALCOSECURITY_RULES_FALCO_PATH=<some-path>` and `-DFALCOSECURITY_RULES_LOCAL_PATH=<some-path>`.
```

### DIFF
--- a/cmake/modules/falcoctl.cmake
+++ b/cmake/modules/falcoctl.cmake
@@ -14,9 +14,14 @@
 
 include(ExternalProject)
 
+option(ADD_FALCOCTL_DEPENDENCY "Add falcoctl dependency while building falco" ON)
+
+if(ADD_FALCOCTL_DEPENDENCY)
 string(TOLOWER ${CMAKE_HOST_SYSTEM_NAME} FALCOCTL_SYSTEM_NAME)
 
 set(FALCOCTL_VERSION "0.8.0")
+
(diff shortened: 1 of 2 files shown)


FedeDP:
/milestone 0.39.0

I really like this small patch; it's great to have Falco building in as many env as we can :)

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #3307, opened 2024-08-28
new(ci): use `zig` compiler instead of relying on centos7.

### LINKED ISSUE #3270: CI pipeline uses CentOS 7 which is EOL

FedeDP:
**What type of PR is this?**

/kind cleanup

/area CI

This PR drops centos7 from our CI by instead relying on zig compiler to provide glibc 2.17 compatible builds for us.

Fixes #3270

Linked libs PRs:
* https://github.com/falcosecurity/libs/pull/2034 -> new c-ares builds with zig 
* https://github.com/falcosecurity/libs/pull/2036 -> libanl is not present under zig; fixes build
* https://github.com/falcosecurity/libs/pull/2037 -> title says it all
* https://github.com/falcosecurity/libs/pull/2043 -> avoids Falco linking issue `undefined symbol: google::protobuf::internal::InternalMetadata::~InternalMetadata()`
* https://github.com/falcosecurity/libs/pull/2045 -> fixes arm64 `Trace/breakpoint trap` issue for threadinfo and k8saudit plugin opening (source_plugin.c UB)
* https://github.com/falcosecurity/libs/pull/2049 -> cmake patches for libelf and grpc to support zig build
* https://github.com/falcosecurity/libs/pull/2088 -> more UBs fixes for libs

TODO:
- [x] sanitizer build fails at cmake configure step, because of: https://github.com/ziglang/zig/issues/11403. **Workaround: disable `zig` for sanitizer builds, ie: only use it for to-be-shipped artifacts.**
- [x] arm64 build is failing while building abseil randen_hwaes.cc file, since their cmake tries to detect platform features and enforces `-march=armv8-a+crypto`, but `armv8` is unsupported by zig. **Workaround: patch https://github.com/abseil/abseil-cpp/blob/1031858eb2b87a5d47f236d8eac11d8b05c535d5/absl/copts/GENERATED_AbseilCopts.cmake#L225 to use a supported arch name.** 
- [x] fix `test-dev-packages-arm64` -> it seems like a `Trace/breakpoint trap` is caught during `sinsp::open...` methods calls:
```
Tue Sep  3 10:46:35 2024: Loaded event sources: syscall
Tue Sep  3 10:46:35 2024: Enabled event sources: syscall
Tue Sep  3 10:46:35 2024: Opening 'syscall' source with Kernel module
Trace/breakpoint trap
```
- [x] fix `test-dev-packages-arm64` -> it seems like a `Trace/breakpoint trap` is caught related to stats_writer line: `output_fields["falco.host_boot_ts"] = machine_info->boot_ts_epoch;` -> bca6f31e94f5417ecf07975ab65e520c46656fac
- [x] fix `test-dev-packages-arm64` -> it seems like a `Trace/breakpoint trap` is caught related to k8saudit opening 
- [x] libelf is built bundled

```release-note
new(ci): use `zig` compiler instead of relying on centos7.
```

### DIFF
--- a/.github/workflows/reusable_build_packages.yaml
+++ b/.github/workflows/reusable_build_packages.yaml
@@ -52,44 +52,17 @@ jobs:
           retention-days: 1
 
   build-packages:
-    env:
-      ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
     # See https://github.com/actions/runner/issues/409#issuecomment-1158849936
     runs-on: ${{ (inputs.arch == 'aarch64' && 'actuated-arm64-8cpu-16gb') || 'ubuntu-latest' }}
     needs: [build-modern-bpf-skeleton]
-    container: centos:7
     steps:
       # Always install deps before invoking checkout action, to properly perform a full clone.
-      - name: Fix mirrors to use vault.centos.org
-        run: |
-          sed -i s/mirror.centos.org/vault.centos.org/g /etc/yum.repos.d/*.repo
-          sed -i s/^#.*baseurl=http/baseurl=https/g /etc/yum.repos.d/*.repo
-          sed -i s/^mirrorlist=http/#mirrorlist=https/g /etc/yum.repos.d/*.repo
-
-      - name: Install scl repos
-        run: |
-          yum -y install centos-release-scl
-
-      - name: Fix new mirrors to use vault.centos.org
-        run: |
-          sed -i s/mirror.centos.org/vault.centos.org/g /etc/yum.repos.d/*.repo
-          sed -i s/^#.*baseurl=http/baseurl=https/g /etc/yum.repos.d/*.repo
-          sed -i s/^mirrorlist=http/#mirrorlist=https/g /etc/yum.repos.d/*.repo
-
-      - name: Fix arm64 scl repos to use correct mirror
-        if: inputs.arch == 'aarch64'
-        run: |
-          sed -i 's/vault.centos.org\/centos/vault.centos.org\/altarch/g' /etc/yum.repos.d/CentOS-SCLo-scl*.repo
-
       - name: Install build deps
         run: |
-          yum -y install devtoolset-9-gcc devtoolset-9-gcc-c++
-          source /opt/rh/devtoolset-9/enable
-          yum install -y wget git make m4 rpm-build elfutils-libelf-devel perl-IPC-Cmd devtoolset-9-libasan-devel devtoolset-9-libubsan-devel
+          sudo apt update && sudo apt install -y --no-install-recommends ca-certificates cmake curl wget build-essential git pkg-config autoconf automake libtool libelf-dev m4 rpm
 
       - name: Checkout
-        # It is not possible to upgrade the checkout action to versions >= v4.0.0 because of incompatibilities with centos 7's libc.
-        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
+        uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608 # v4.1.0
 
       - name: Download skeleton
         uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
@@ -97,14 +70,12 @@ jobs:
           name: bpf_probe_${{ inputs.arch }}.skel.h
           path: /tmp
 
-      - name: Install updated cmake
-        run: |
-          curl -L https://github.com/Kitware/CMake/releases/download/v3.22.5/cmake-3.22.5-linux-$(uname -m).tar.gz \
-            | tar --directory=/usr --strip-components=1 -xzp
+      - name: Install zig
+        if: inputs.sanitizers == false
+        uses: falcosecurity/libs/.github/actions/install-zig@master
 
       - name: Prepare project
         run: |
-          source /opt/rh/devtoolset-9/enable
           cmake -B build -S . \
               -DCMAKE_BUILD_TYPE=${{ inputs.build_type }} \
               -DUSE_BUNDLED_DEPS=On \
@@ -117,12 +88,10 @@ jobs:
 
       - name: Build project
         run: |
-          source /opt/rh/devtoolset-9/enable
           cmake --build build --target falco -j6
 
       - name: Build packages
         run: |
-          source /opt/rh/devtoolset-9/enable
           cmake --build build --target package
 
       - name: Upload Falco tar.gz package
--- a/cmake/modules/falcosecurity-libs.cmake
+++ b/cmake/modules/falcosecurity-libs.cmake
@@ -42,9 +42,9 @@ else()
 	# version (or branch, or commit) just pass the variable - ie., `cmake
 	# -DFALCOSECURITY_LIBS_VERSION=dev ..`
 	if(NOT FALCOSECURITY_LIBS_VERSION)
-		set(FALCOSECURITY_LIBS_VERSION "0.18.1")
+		set(FALCOSECURITY_LIBS_VERSION "aeb87935dcdd6b3bffb50302269353f2468f612c")
 		set(FALCOSECURITY_LIBS_CHECKSUM
-			"SHA256=1812e8236c4cb51d3fe5dd066d71be99f25da7ed22d8feeeebeed09bdc26325f"
+			"SHA256=fc2f18ed2f9e4bdcb93f32eee326a813a3614af50b82d30a7ccf96ac2a7cfbf0"
 		)
 	endif()
 
@@ -73,7 +73,6 @@ set(LIBS_PACKAGE_NAME "falcosecurity")
 
 if(CMAKE_SYSTEM_NAME MATCHES "Linux")
 	add_definitions(-D_GNU_SOURCE)
-	add_definitions(-DHAS_CAPTURE)
 endif()
 
 if(MUSL_OPTIMIZED_BUILD)
--- a/userspace/falco/stats_writer.cpp
+++ b/userspace/falco/stats_writer.cpp
@@ -360,7 +360,9 @@ void stats_writer::collector::get_metrics_output_fields_wrapper(
 		output_fields["evt.hostname"] =
 		        machine_info->hostname; /* Explicitly add hostname to log msg in case hostname rule
 		                                   output field is disabled. */
-		output_fields["falco.host_boot_ts"] = machine_info->boot_ts_epoch;
+		// This line generates a SIGTRAP in zig debug builds if the casting is removed.
+		// It seems caused by the pragma pack for the scap_machine_info structure.
+		output_fields["falco.host_boot_ts"] = (uint64_t)machine_info->boot_ts_epoch;
 		output_fields["falco.host_num_cpus"] = machine_info->num_cpus;
 	}
 	output_fields["falco.outputs_queue_num_drops"] =
--- a/cmake/modules/driver.cmake
+++ b/cmake/modules/driver.cmake
@@ -35,9 +35,9 @@ else()
 	# FALCOSECURITY_LIBS_VERSION. In case you want to test against another driver version (or
 	# branch, or commit) just pass the variable - ie., `cmake -DDRIVER_VERSION=dev ..`
 	if(NOT DRIVER_VERSION)
-		set(DRIVER_VERSION "7.3.0+driver")
+		set(DRIVER_VERSION "aeb87935dcdd6b3bffb50302269353f2468f612c")
 		set(DRIVER_CHECKSUM
-			"SHA256=8f572d9a83feda635a3fa53b859d61e37af127c241e35068aadee3bc50d212c0"
+			"SHA256=fc2f18ed2f9e4bdcb93f32eee326a813a3614af50b82d30a7ccf96ac2a7cfbf0"
 		)
 	endif()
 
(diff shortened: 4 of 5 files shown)


FedeDP:
So, we need to disable the build with shared libelf, because `zig` statically links all specified libraries; see: https://github.com/ziglang/zig/issues/7094
Indeed the CI complains it cannot find system libelf.
Since this was an :hot_pepper: topic for the CNCF, i am going to hold this PR; at the same time, i think we should actually find a way to drop centos:7 while maintaining same glibc requirements.


Anyway, even trying with BUNDLED_LIBELF, gives a build error...

/hold

submodules/falcosecurity-rules
FedeDP, line 1: As always, a `git commit -sam` committed this :rofl:

.github/workflows/reusable_build_packages.yaml
```diff
@@ -94,17 +82,12 @@ jobs:
...
-          curl -L https://github.com/Kitware/CMake/releases/download/v3.22.5/cmake-3.22.5-linux-$(uname -m).tar.gz \
-            | tar --directory=/usr --strip-components=1 -xzp
-
       - name: Prepare project
         run: |
-          source /opt/rh/devtoolset-9/enable
           cmake -B build -S . \
               -DCMAKE_BUILD_TYPE=${{ inputs.build_type }} \
               -DUSE_BUNDLED_DEPS=On \
+              -DUSE_BUNDLED_LIBELF=On \
```
FedeDP, line 90: Enabled bundled libelf to build with zig. Of course this is not desiderable.

FedeDP:
Now failing the build on 
> ld.lld: error: undefined symbol: arc4random_buf
>>> referenced by ares_rand.c:270

That's probably because c-ares does some magic to detect that symbol: https://github.com/c-ares/c-ares/blob/main/CMakeLists.txt#L456; that symbol was added in glibc2.36; most probably it is detecting the symbol even if `zig` is not actually supporting it since we are targeting a build with glibc-2.17.

FedeDP:
I tried to build `c-ares` from scratch (ie: without Falco involved at all) with `zig` and it worked fine with the same `export`s we are doing in the CI.

FedeDP:
Update: i saw that updating c-ares to 1.33.1 (latest release) fixed the issue. I am going to bump it in libs (if possible, ie: if grpc does not complain :/ )

FedeDP:
So, i now bumped this branch to use libs [`chore/zig_build` branch HEAD](https://github.com/falcosecurity/libs/tree/chore/zig_build); that is a branch i pushed upstream with all zig related fixes on it:
* https://github.com/falcosecurity/libs/pull/2034
* https://github.com/falcosecurity/libs/pull/2036
* https://github.com/falcosecurity/libs/pull/2037

Moreover, i found out that CMAKE_{C,CXX}_COMPILER truncates `CC` and `CXX` strings at first whitespace, leading with cmake using `zig` as compiler for both, instead of `zig cc` and `zig c++`.
I created 2 small wrapper scripts in CI to workaround this issue:
```
cat > zig-linux-${{ inputs.arch }}-${ZIG_VERSION}/zig-cc << EOF
#!/bin/bash
zig cc -target ${{ inputs.arch }}-linux-gnu.2.17 "$@"
EOF
chmod +x zig-linux-${{ inputs.arch }}-${ZIG_VERSION}/zig-cc
echo "CC=zig-cc" >> $GITHUB_ENV

// same for zig-c++
```

FedeDP:
So, we now have artifacts working (passing tests too!); here are the differences for arm64:
* normal Falco:
```
ldd falco
linux-vdso.so.1 (0x0000e4ff9035f000)
	libelf.so.1 => /lib/aarch64-linux-gnu/libelf.so.1 (0x0000e4ff902f0000)
	librt.so.1 => /lib/aarch64-linux-gnu/librt.so.1 (0x0000e4ff902c0000)
	libanl.so.1 => /lib/aarch64-linux-gnu/libanl.so.1 (0x0000e4ff90290000)
	libdl.so.2 => /lib/aarch64-linux-gnu/libdl.so.2 (0x0000e4ff90260000)
	libpthread.so.0 => /lib/aarch64-linux-gnu/libpthread.so.0 (0x0000e4ff90230000)
	libz.so.1 => /lib/aarch64-linux-gnu/libz.so.1 (0x0000e4ff901f0000)
	libstdc++.so.6 => /lib/aarch64-linux-gnu/libstdc++.so.6 (0x0000e4ff8fe00000)
	libm.so.6 => /lib/aarch64-linux-gnu/libm.so.6 (0x0000e4ff90140000)
	libgcc_s.so.1 => /lib/aarch64-linux-gnu/libgcc_s.so.1 (0x0000e4ff90100000)
	libc.so.6 => /lib/aarch64-linux-gnu/libc.so.6 (0x0000e4ff8fc40000)
	/lib/ld-linux-aarch64.so.1 (0x0000e4ff90322000)
	libzstd.so.1 => /lib/aarch64-linux-gnu/libzstd.so.1 (0x0000e4ff8fb80000)

objdump -T  falco | grep -Eo 'GLIBC_\S+' | sort -u
GLIBC_2.17)

du -h falco
26M	falco
```

* zig Falco:
```
ldd falco
linux-vdso.so.1 (0x0000e17bf7aff000)
	libm.so.6 => /lib/aarch64-linux-gnu/libm.so.6 (0x0000e17bf7a20000)
	libpthread.so.0 => /lib/aarch64-linux-gnu/libpthread.so.0 (0x0000e17bf79f0000)
	libc.so.6 => /lib/aarch64-linux-gnu/libc.so.6 (0x0000e17bf7830000)
	libdl.so.2 => /lib/aarch64-linux-gnu/libdl.so.2 (0x0000e17bf7800000)
	librt.so.1 => /lib/aarch64-linux-gnu/librt.so.1 (0x0000e17bf77d0000)
	/lib/ld-linux-aarch64.so.1 (0x0000e17bf7ac2000)
	
objdump -T falco | grep -Eo 'GLIBC_\S+' | sort -u
GLIBC_2.17)

du -h falco
124M	falco
```
So, the zig Falco:
* links less things, and that's expected since libelf is now statically built, libanl is no more linked, there is no libgcc/libstdc++ and libzstd; moreover libz dynamically linked by libelf is gone too.
* has same glibc requirements (2.17+)
* is much bigger (at least in the Debug build of arm64; the x86_64 debug build is much similar, 29M to 24M).

.github/workflows/reusable_build_packages.yaml
```diff
@@ -69,13 +69,13 @@ jobs:
 
           cat > zig-linux-${{ inputs.arch }}-${ZIG_VERSION}/zig-cc <<EOF
           #!/bin/bash
-          exec zig cc -target ${{ inputs.arch }}-linux-gnu.2.17 -mcpu=baseline "\$@"
+          exec zig cc -L/usr/lib/${{ inputs.arch }}-linux-gnu/ -target ${{ inputs.arch }}-linux-gnu.2.17 -mcpu=baseline "\$@"
```
FedeDP, line 72: Needed to find system libelf.

cmake/modules/falcosecurity-libs-repo/patches/zig_libelf.patch
FedeDP, line 1: Basically, we need to patch libs libelf cmake module to:
* add the `HINTS` about the possible location of the libelf library. Normally, cmake would auto-detect that it is running under debian/ubuntu and add that folder to its search list; under zig that is not true and we need to specifically add it
* to avoid including the whole `/usr/include`, we just copy libelf headers to `CMAKE_CURRENT_BINARY_DIR` and include it. This is needed to avoid including the whole `/usr/include` during the compilation that would lead to compilation issues since we would mix the system toolchain with the zig one

cmake/modules/falcosecurity-libs-repo/patches/zig_aarch64_grpc.patch
FedeDP, line 1: Abseil-cpp is smart enough to detect certain hw capabilities to improve its perf, in this specific case we are talking about the `ABSL_RANDOM_HWAES_ARM64_FLAGS`: https://github.com/abseil/abseil-cpp/blob/master/absl/copts/GENERATED_AbseilCopts.cmake#L225.
Setting `armv8` is not a supported `march` for zig; therefore we patch the grpc cmake module (that builds abseil-cpp) to PATCH aforementioned abseil source file to name that `march` in another, zig-compatible, way.

cmake/modules/falcosecurity-libs.cmake
```diff
@@ -50,13 +50,17 @@ else()
...
+  if (ENABLE_ZIG_WORKAROUNDS)
+    add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules/falcosecurity-libs-repo/patches)
+  endif()
 endif()
 
 set(LIBS_PACKAGE_NAME "falcosecurity")
 
 if(CMAKE_SYSTEM_NAME MATCHES "Linux")
   add_definitions(-D_GNU_SOURCE)
-  add_definitions(-DHAS_CAPTURE)
```
FedeDP, line 59: HAS_CAPTURE was dropped since ages :D

userspace/falco/stats_writer.cpp
```diff
@@ -336,7 +336,9 @@ void stats_writer::collector::get_metrics_output_fields_wrapper(
 	if (machine_info)
 	{
 		output_fields["evt.hostname"] = machine_info->hostname; /* Explicitly add hostname to log msg in case hostname rule output field is disabled. */
-		output_fields["falco.host_boot_ts"] = machine_info->boot_ts_epoch;
+		// This line generates a SIGTRAP in zig debug builds if the casting is removed.
+		// It seems caused by the pragma pack for the scap_machine_info structure.
+		output_fields["falco.host_boot_ts"] = (uint64_t)machine_info->boot_ts_epoch;
```
FedeDP, line 341: As weird as it seems, this stupid cast fixed a SIGTRAP generated on debug builds by zig.
If i removed the `pragma pack` from the scap_machine_info struct, the issue went away. Most probably this is just a bug somewhere in how zig/clang compile the sources.

cmake/modules/falcosecurity-libs-repo/patches/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+file(GLOB patches "*.patch")
+file(MAKE_DIRECTORY ${PROJECT_BINARY_DIR}/libs-patches/)
+foreach(p ${patches})
+    configure_file("${p}" "${PROJECT_BINARY_DIR}/libs-patches/" @ONLY)
```
FedeDP, line 22: libelf patch uses `@CMAKE_SYSTEM_PROCESSOR@` thus it needs to be configured.
`@ONLY` to avoid touching also normal cmake variables like `${FOO}`.

cmake/modules/CompilerFlags.cmake
FedeDP, line 1: Can revert this one but it is not a nice behavior to erase pre-defined `CMAKE_{C,CXX}_FLAGS` IMHO.

FedeDP:
The PR is ready; idea is to merge this right after the Falco release, after https://github.com/falcosecurity/libs/pull/2034 and https://github.com/falcosecurity/libs/pull/2037 are merged in Falco libs (i'll then take care of update this PR to use libs master of course).

Then, we will have 4 months to discover any new issue and fix it.
Also, as soon as i'll have some spare time, i'd like to benchmark the new binary (eg: by locally building sinsp with zig and running `make bench` on it, since we now have a bunch of google benchmarks in libs, and doing the same from Falco master).

FedeDP:
/unhold
Everything is fixed and we are using libs master.

FedeDP:
/hold
Don't get why 
>  CI Build / test-dev-packages-arm64 / test-packages (pull_request) Failing after 4m 

is failing on 2 tests because cannot find `.scap` files :/

leogr (approved):

leogr (approved):

leogr:
> 
> 
> /hold

False positive

FedeDP:
Yep, tomorrow will rebase on latest libs master that fixes arm64 test-dev-packages and unhold this one :)

FedeDP:
/unhold
Ready!

leogr (approved):

FedeDP:
:loudspeaker: it might happen that master build fails because the cached key (ie: zig version) is no more available and `zig` moved to a newer master version, thus we cannot find the correct zig version to be downloaded.
In that case, i'll update zig to latest in libs and restart the CI here.

FedeDP:
/hold 
until next week.

jasondellaluce (approved):

FedeDP:
I'll keep this on hold until we release Falco 0.39.1 to avoid issues with the new CI.

FedeDP:
/unhold
Let's see if master CI complains (i guess so :D )

PR #3299, opened 2024-08-17
fixed the token-permission and pinned-dependencies issue

harshitasao:
/area CI

**What this PR does / why we need it**:
fixed the token-permission and pinned-dependencies issue which are reported by [scorecard](https://scorecard.dev/viewer/?uri=github.com/falcosecurity/falco).

**Which issue(s) this PR fixes**:
Part of #3297 

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
None
```

PR #3317, opened 2024-09-10
fix(ci): restore master and release CI workflow permissions.

FedeDP:
**What type of PR is this?**

/kind bug

/area CI

https://github.com/falcosecurity/falco/pull/3299 broke master CI triggering (if you see latest master commits did not trigger master CI, only codeql and scorecard).

This is an attempt to fix them.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/master.yaml
+++ b/.github/workflows/master.yaml
@@ -6,10 +6,7 @@ on:
 # Checks if any concurrent jobs is running for master CI and eventually cancel it
 concurrency:
   group: ci-master
-  cancel-in-progress: true  
-
-permissions:  
-  contents: read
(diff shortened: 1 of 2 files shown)


Andreagit97 (approved):
it would be great to understand why we cannot apply the permissions on this but I understand that we cannot have the master CI broken

/approve

FedeDP:
> it would be great to understand why we cannot apply the permissions on this but I understand that we cannot have the master CI broken

We can actually, but we can't set smaller permissions set on the caller workflow when some called workflows (eg: reusable_publish_X workflows) require larger permissions set.

LucaGuerra (approved):

PR #3322, opened 2024-09-12
update(engine): modify append_output format

LucaGuerra:
**What type of PR is this?**

/kind feature

/area engine

Having discussed this feature with @leogr and @Issif we came to the conclusion that it could be better to modify the format of `append_output` to something like this example:

```yaml
append_output:
  - match:
      source: syscall
      tags: 
        - bar
        - foo
      rule: rule A
    extra_output: "gparent=%proc.aname[2] ggparent=%proc.aname[3] gggparent=%proc.aname[4]"
    extra_fields:
      - ka.verb
      - home_directory: "${HOME}"
      - my_field: "this is event number %evt.num"
```

`tags`, `source`, `rule` are ANDed together and also when multiple tags are specified those are all ANDed together as well.

Fixes #

```release-note
NONE
```


leogr (dismissed):

userspace/engine/rule_loader_compiler.cpp
```diff
@@ -501,9 +502,18 @@ void rule_loader::compiler::compile_rule_infos(
 				continue;
 			}
 
-			if (extra.m_tag != "" && r.tags.count(extra.m_tag) == 0)
+			if (extra.m_tags.size() != 0)
```
FedeDP, line 505: Exctract an helper method since the same check is also used below?

FedeDP:
I think we need to update configuration.cpp json schema too!
/hold

LucaGuerra:
@FedeDP damn I forgot 🤣 , will fix. thanks

leogr:
> I think we need to update configuration.cpp json schema too! /hold

Good catch!

side topic: it would be nice to introduce some automatic check

FedeDP:
> side topic: it would be nice to introduce some automatic check

There is an automatic check that loads the falco.yaml from the CMAKE_CURRENT_SOURCE_DIR; but in this case, the new option is not present by default in the Falco yaml thus the test passes fine.

leogr (approved):

FedeDP (approved):
/approve

PR #3310, opened 2024-09-03
new(falco): enable CLI options with -o key={object}

LucaGuerra:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

/kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

/area engine

**What this PR does / why we need it**:

In https://github.com/falcosecurity/falco/pull/3308 we add a new configuration option such as:
```yaml
append_output:
  - source: syscall
    tag: persistence
    rule: some rule name
    format: "gparent=%proc.aname[2] ggparent=%proc.aname[3] gggparent=%proc.aname[4]"
```

Given its format, it is not possible to specify it on the command line with a `-o` option unless we enable a syntax like `-o key={object}`, or `-o key[]={object}`. Example: `-o append_output[]={"source": "syscall", "tag": "persistence", "rule": "some rule name", "format": "gparent=%proc.aname[2] ggparent=%proc.aname[3] gggparent=%proc.aname[4]"}`. This PR does exactly that.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

For now, we are not enabling lists here but they can be allowed as well.

**Does this PR introduce a user-facing change?**:

```release-note
new(falco): enable CLI options with -o key={object}
```

PR #3363, opened 2024-10-01
fix(engine): disable comma separated vectors in cxxopts

LucaGuerra:
**What type of PR is this?**

/kind bug

/area engine

There was a missed configuration option in this PR https://github.com/falcosecurity/falco/pull/3310 , which sadly made the `-o append_output[]={"source": "syscall", "tag": "persistence", "rule": "some rule name", "format": "gparent=%proc.aname[2] ggparent=%proc.aname[3] gggparent=%proc.aname[4]"}` style options non functional due to the fact that cxxopts is trying to parse commas in the json object. Sadly I only realized after release. This PR fixes it.

Fixes #

```release-note
fix(engine): fix parsing issues in -o key={object} when the object definition contains a comma
```

### DIFF
--- a/userspace/falco/app/options.cpp
+++ b/userspace/falco/app/options.cpp
@@ -19,6 +19,9 @@ limitations under the License.
 #include "../configuration.h"
 #include "config_falco.h"
 
+// disable cxxopts vector delimiter, meaning that
+// -o test1,test2,test3 won't be treated like -o test1 -o test2 -o test3
+#define CXXOPTS_VECTOR_DELIMITER '\0'
 #include <cxxopts.hpp>
 
 #include <fstream>


LucaGuerra:
/milestone 0.40.0

buf if there is a 0.39.1 this would get in.

leogr:
> /milestone 0.40.0
> 
> buf if there is a 0.39.1 this would get in.

:+1: 
We are investigating about a recently discovered bug, so - since an hotfix release is very likely to happen - I've created the milestone

/milestone 0.39.1

userspace/falco/app/options.cpp
```diff
@@ -19,6 +19,9 @@ limitations under the License.
 #include "../configuration.h"
 #include "config_falco.h"
 
+// disable cxxopts vector delimiter, meaning that
+// -o test1,test2,test3 won't be treated like -o test1 -o test2 -o test3
+#define CXXOPTS_VECTOR_DELIMITER '\0'
```
leogr, line 24: This may be seen as a (minor) breaking change in the CLI context. Is there any other available alternative?

FedeDP (approved):
/approve

userspace/falco/app/options.cpp
LucaGuerra: It does not look like it as it happens unconditionally for `vector<T>` values: https://github.com/jarro2783/cxxopts/blob/63d1b65a694cfceafc20863afa75df49dfbe6b2a/include/cxxopts.hpp#L1100

leogr (approved):

PR #3388, opened 2024-10-22
new(userspace): honor new plugins exposed suggested output formats

FedeDP:
**What type of PR is this?**

/kind feature

/area engine

Since https://github.com/falcosecurity/libs/pull/2116, the plugin API now exposes a way for extractor plugins to declare that certain fields should be used as output formats.
Honor the request, and add a `suggested_formats` config option to eventually disable the suggested formats.

Fixes #

```release-note
new(userspace,cmake): honor new plugins exposed suggested output formats
```

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -571,12 +571,6 @@ buffered_outputs: false
 # deploying it in production.
 rule_matching: first
 
-# [Incubating] `suggested_formats`
-#
-# When enabled, Falco will honor requests by extractor plugins
-# that suggest certain fields to be part of outputs.
-suggested_formats: true
-
 # [Stable] `outputs_queue`
 #
 # Falco utilizes tbb::concurrent_bounded_queue for handling outputs, and this parameter
@@ -624,6 +618,7 @@ outputs_queue:
 #             affect the regular Falco message in any way. These can be specified as a
 #             custom name with a custom format or as any supported field
 #             (see: https://falco.org/docs/reference/rules/supported-fields/)
+#   `suggested_output`: enable the use of extractor plugins suggested fields for the matching source output.
 #
 # Example:
 # 
@@ -639,6 +634,13 @@ outputs_queue:
 # at the end telling the CPU number. In addition, if `json_output` is true, in the "output_fields"
 # property you will find three new ones: "evt.cpu", "home_directory" which will contain the value of the
 # environment variable $HOME, and "evt.hostname" which will contain the hostname.
+
+# By default, we enable suggested_output for any source.
+# This means that any extractor plugin that indicates some of its fields
+# as suggested output formats, will see these fields in the output
+# in the form "foo_bar=$foo.bar"
+append_output:
+  - suggested_output: true
 
 
 ##########################
--- a/falco.yaml
+++ b/falco.yaml
@@ -571,6 +571,12 @@ buffered_outputs: false
 # deploying it in production.
 rule_matching: first
 
+# [Incubating] `suggested_formats`
+#
+# When enabled, Falco will honor requests by extractor plugins
+# that suggest certain fields to be part of outputs.
+suggested_formats: true
+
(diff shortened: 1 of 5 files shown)


FedeDP:
CI is failing because there is some problem with the install-zig action :/

userspace/falco/app/actions/init_falco_engine.cpp
```diff
@@ -45,6 +58,25 @@ void configure_output_format(falco::app::state& s) {
...
 
+	// Add suggested filtercheck formats to each source output
+	if(s.config->m_suggested_formats) {
+		for(auto& src : s.loaded_sources) {
+			auto src_info = s.source_infos.at(src);
+			auto& filterchecks = *src_info->filterchecks;
+			std::vector<const filter_check_info*> fields;
+			filterchecks.get_all_fields(fields);
+			for(const auto& fld : fields) {
+				if(fld->m_flags & EPF_FORMAT_SUGGESTED) {
```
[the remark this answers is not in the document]
FedeDP: Oh right we already have this!

falco.yaml
```diff
@@ -634,6 +635,13 @@ outputs_queue:
 # property you will find three new ones: "evt.cpu", "home_directory" which will contain the value of the
 # environment variable $HOME, and "evt.hostname" which will contain the hostname.
 
+# By default, we enable suggested_output for any source.
+# This means that any extractor plugin that indicates some of its fields
+# as suggested output formats, will see these fields in the output
+# in the form "foo_bar=$foo.bar"
+append_output:
```
FedeDP, line 642: By default, we enable `suggested_output` for all rules, tags and sources.

userspace/falco/app/actions/init_falco_engine.cpp
```diff
@@ -32,6 +66,17 @@ void configure_output_format(falco::app::state& s) {
 			                                  false);
 		}
 
+		// Add suggested filtercheck formats to each source output
+		if(eo.m_suggested_output) {
```
FedeDP, line 70: If `suggested_output` is enabled:
* if any matching source is specified, we only list fields for that source, and add the extra output format for it
* else, we add the extra output format for all loaded sources

`Tags` and `Rule` are still managed normally.

userspace/falco/configuration.h
```diff
@@ -100,6 +100,7 @@ class falco_configuration {
 		std::set<std::string> m_tags;
 		std::string m_rule;
 		std::string m_format;
+		bool m_suggested_output = false;
```
FedeDP, line 103: When missing, default value is false.

leogr: Is there any particular reason why it is enabled by default in the config but disabled if missing? 

Not saying this is wrong, I'm just wondering why.

FedeDP: That's because i want it to be effective only when explicitly enabled; i don't want it to have any effect if you specify eg:
```
append_output:
   - match:
       source: syscall
     extra_output: "on CPU %evt.cpu"
```
Ie: in this example, i only want `on CPU %evt.cpu` to be appended to output, not the plugin suggested fields, unless `suggested_output: true` is also set.

But i am also open for more opinions.

leogr (approved):
LGTM

leogr (approved):

userspace/falco/app/actions/init_falco_engine.cpp
LucaGuerra: Why does this need special treatment? If you don't specify any source the extra format already applies to all sources.

LucaGuerra: ... it needs special treatment because every source has its own fields and thus its own format so in order to select the correct one we need to iterate through all the sources. Makes sense. Thanks @FedeDP

LucaGuerra (approved):

PR #3426, opened 2024-12-06
fix(userspace/falco): use correct filtercheck_field_info.

FedeDP:
**What type of PR is this?**

/kind bug

/area engine

Fix on top of #3388: `filter_check_info->m_name` is the global name of the list of checks, eg: `container (plugin)`.
Instead, we need to loop on each field info.

Fixes #

```release-note
NONE
```


Andreagit97 (approved):
/approve

LucaGuerra (approved):

PR #3428, opened 2024-12-13
new(build): reintroduce static build

LucaGuerra:
**What type of PR is this?**

/kind feature

/area build

We can now link libelf (from BSD elftoolchain) statically. So, the static build can come back!

Fixes #

```release-note
new(build): reintroduce static build
```


cmake/modules/cpp-httplib.cmake
```diff
@@ -18,6 +18,10 @@ option(USE_BUNDLED_CPPHTTPLIB "Enable building of the bundled cpp-httplib" ${USE
 if(USE_BUNDLED_CPPHTTPLIB)
 	set(HTTPLIB_USE_BROTLI_IF_AVAILABLE OFF)
 	set(HTTPLIB_REQUIRE_BROTLI OFF)
+	set(HTTPLIB_USE_ZLIB_IF_AVAILABLE OFF)
+	set(HTTPLIB_REQUIRE_ZLIB OFF)
+	set(HTTPLIB_USE_OPENSSL_IF_AVAILABLE OFF)
+	set(HTTPLIB_REQUIRE_OPENSSL OFF)
```
LucaGuerra, line 24: cpp-httplib strikes again! I don't think we use compression or encryption in `webserver.cpp`

sgaist: For compression, you are likely correct. The versions endpoint is `application/json` and the metrics endpoint is currently `text/plain; version=0.0.4` using the prometheus exposition format.

However, for encryption, there's the SSL Enabled option.

LucaGuerra: Makes sense, I have re-enabled OpenSSL. Fortunately it does recognize the static OpenSSL build. Thanks a lot @sgaist !

LucaGuerra:
This PR supersedes https://github.com/falcosecurity/falco/pull/3425, https://github.com/falcosecurity/falco/pull/3430, https://github.com/falcosecurity/falco/pull/3431

sgaist (dismissed):
Looks good to me

.github/workflows/reusable_build_packages.yaml
```diff
@@ -139,10 +238,10 @@ jobs:
       - name: Prepare project
         run: |
           cmake -B build -S . \
-                -DCMAKE_BUILD_TYPE=${{ inputs.build_type }} \
+                -DCMAKE_BUILD_TYPE=Release \
```
FedeDP, line 241: Do we want to enforce jemalloc for musl build?

FedeDP:
/hold

Will give it a look later today :)

.github/workflows/reusable_build_packages.yaml
```diff
@@ -166,7 +265,7 @@ jobs:
 
   build-wasm-package:
     if: ${{ inputs.arch == 'x86_64' }}
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
FedeDP, line 268: Same fix we applied in https://github.com/falcosecurity/libs/pull/2203

.github/workflows/reusable_build_packages.yaml
```diff
@@ -99,37 +98,137 @@ jobs:
...
-    if: false
+    if: ${{ inputs.arch == 'x86_64' }}
     runs-on: ubuntu-latest
     container: alpine:3.17
     steps:
       # Always install deps before invoking checkout action, to properly perform a full clone.
       - name: Install build dependencies
         run: |
-          apk add g++ gcc cmake make git bash perl linux-headers autoconf automake m4 libtool elfutils-dev libelf-static patch binutils bpftool clang
+          apk add g++ gcc cmake make git bash perl linux-headers autoconf automake m4 libtool elfutils-dev libelf-static patch binutils clang llvm
```
FedeDP, line 227: Question: do we need `elfutils-dev libelf-static` deps here? Considering we are using a bundled build.

.github/workflows/reusable_build_packages.yaml
```diff
@@ -99,37 +98,137 @@ jobs:
...
+  build-packages-sanitizers:
+    # See https://github.com/actions/runner/issues/409#issuecomment-1158849936
+    runs-on: ${{ (inputs.arch == 'aarch64' && 'oracle-aarch64-4cpu-16gb') || 'ubuntu-latest' }}
+    if: ${{ inputs.enable_sanitizers == true }}
+    needs: [build-modern-bpf-skeleton]
+    steps:
+      # Always install deps before invoking checkout action, to properly perform a full clone.
+      - name: Install build deps
+        run: |
+          sudo apt update && sudo apt install -y --no-install-recommends ca-certificates cmake curl wget build-essential git pkg-config autoconf automake libtool libelf-dev m4 rpm
```
FedeDP, line 179: Q: do we need `libelf-dev` here?

FedeDP: And in above jobs?

userspace/falco/webserver.cpp
FedeDP, line 1: Can you comment about this change?

CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ cmake_minimum_required(VERSION 3.5.1)
 project(falco)
 
 option(USE_BUNDLED_DEPS "Bundle hard to find dependencies into the Falco binary" ON)
-option(USE_DYNAMIC_LIBELF "Dynamically link libelf" ON)
+option(USE_DYNAMIC_LIBELF "Dynamically link libelf" OFF)
```
FedeDP, line 20: Do we want to support this "feature" still?

cmake/modules/cpp-httplib.cmake
FedeDP, line 1: Can you expand about why was this needed? I think because we don't want cpp-httplib to dynamically link libz from the system, right?

.github/workflows/reusable_build_packages.yaml
LucaGuerra: yep, wouldn't build without it

.github/workflows/reusable_build_packages.yaml
LucaGuerra: I think they are required to build `bpftool`.

userspace/falco/webserver.cpp
LucaGuerra: ASan was complaining about an use-after-return for the `failed` variable. If you look at the code, there's no guarantee that the calling thread is still in the `start()` method when when the `std::thread` that is launched by it is writing to the variable

cmake/modules/cpp-httplib.cmake
LucaGuerra: Yes that is correct. Its cmake will prioritize a dynamic library no matter what, meaning that you'll have a linking error with musl

.github/workflows/reusable_build_packages.yaml
LucaGuerra: Guess not. Let's try and see if the CI is happy.

.github/workflows/reusable_build_packages.yaml
LucaGuerra: Not sure but I'm not a jemalloc expert. What do you think?

userspace/falco/webserver.cpp
FedeDP: Thanks!

FedeDP (approved):
Thank you!
/approve

.github/workflows/reusable_build_packages.yaml
FedeDP: I think that all of our release artifact should spot the same behavior, therefore i'd say yes!

leogr (approved):

sgaist (approved):

PR #3440, opened 2024-12-23
new(build): add RelWithDebInfo target

### LINKED ISSUE #2268: RelWithDebInfo build
**Motivation**

Hey team, I realize that Falco has only "Debug" and "Release" build types.
My team and org were wondering if it was possible to enable a `RelWithDebInfo` build?

RelWithDebInfo builds are usually built with all the performance optimizations of the release build, but with debug symbols turned on. This might be possible with `CPACK_STRIP_FILES=Off` and build type release, but I'm otherwise unfamiliar with the available cmake/cpack build options.

**Feature**

RelWithDebInfo build type; release-optimized, but with debug symbols

**Alternatives**

Possibly `CPACK_STRIP_FILES=Off` ? I need to investigate further if this satisfies the requirements of my team.

**Additional context**

https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#default-and-custom-configurations
^ Does this imply that cmake supports `RelWithDebInfo` by default and will compile + build accordingly?

shane-lawrence:
Signed-off-by: Shane Lawrence <shane@lawrence.dev>

**What type of PR is this?**
/kind feature

/area build

Debug binaries can behave differently from release binaries, which can make it hard to reproduce Release bugs in a Debug environment for certain edge cases. However, the release target doesn't produce debug symbols so it is difficult to trace certain kinds of bugs.

This PR adds the cmake-standard RelWithDebInfo target so we can create a binary that behaves exactly the same as the release target but has debug symbols available in a separate file.

Fixes #2268.

This PR requires additional testing to ensure that the binary really is the same as the release binary. This first attempt might also need additional changes to conform to the existing style and good coding practices for this project.

NO

```release-note
NONE
```


shane-lawrence:
I've only tried to remove debug symbols from the original binary, but I'm seeing a substantial change in file size compared to the release binary. Building with bundled deps and no specified CMAKE_BUILD_TYPE (defaults to Release), I get a userspace/falco/falco binary that's 398MB. Building with CMAKE_BUILD_TYPE=RelWithDebInfo, I get a userspace/falco/falco.debug file that's 222MB and a userspace/falco/falco binary that's 90% smaller at only 26MB. The binary seems to work the same though.

FedeDP:
Hi! Thanks for this PR! I really like the idea and i think we could start shipping (perhaps just as github release artifacts) the debug symbols files, wdyt?

> Building with bundled deps and no specified CMAKE_BUILD_TYPE (defaults to Release), I get a userspace/falco/falco binary that's 398MB. Building with CMAKE_BUILD_TYPE=RelWithDebInfo, I get a userspace/falco/falco.debug file that's 222MB and a userspace/falco/falco binary that's 90% smaller at only 26MB. The binary seems to work the same though.

Mmmh the 0.39.2-x86_64 Falco released binary is:
```
du -h falco
25M	falco
```
thus we are pretty near.
Latest from master is instead 29.9 Mb but we switched to zig (that uses clang) thus a small difference is expected.

shane-lawrence:
Yes that would be great. Just need the symbols to come from the same build as the running binary to use them.

FedeDP:
> Just need the symbols to come from the same build as the running binary to use them.

Yep exactly!
Sorry, why did you close this one? It seemed good to me :D

shane-lawrence:
Oh, I closed it by mistake.

FedeDP:
That's a classic :rofl:

FedeDP:
Can you remove last commit?
Also, i will update the PR title to reflect our https://github.com/falcosecurity/.github/blob/main/CONTRIBUTING.md#commit-convention :)

FedeDP:
What i see is:
* i test this one
* we merge it
* i'll open a PR to port our release pipeline to use the new target
* i will also update the release workflow to upload the debug info artifacts

This is something that has been on my mind for so long, but never scratched my head around it :) Thank you so much!
/milestone 0.20.0

cmake/modules/CompilerFlags.cmake
```diff
@@ -85,6 +88,17 @@ if(NOT MSVC)
 	set(CMAKE_C_FLAGS_RELEASE "-O3 -fno-strict-aliasing -DNDEBUG")
 	set(CMAKE_CXX_FLAGS_RELEASE "-O3 -fno-strict-aliasing -DNDEBUG")
 
+	set(CMAKE_C_FLAGS_RELWITHDEBINFO "-O3 -fno-strict-aliasing -DNDEBUG -g")
+	set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O3 -fno-strict-aliasing -DNDEBUG -g")
```
FedeDP, line 92:
```suggestion
	set(CMAKE_C_FLAGS_RELWITHDEBINFO "${CMAKE_C_FLAGS_RELEASE} -g")
	set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELEASE} -g")
```
to make it explicit that we follow same flags as release + `-g`.

userspace/falco/CMakeLists.txt
```diff
@@ -194,6 +194,24 @@ if(MUSL_OPTIMIZED_BUILD AND CMAKE_BUILD_TYPE STREQUAL "release")
 	)
 endif()
 
+if(CMAKE_BUILD_TYPE STREQUAL "relwithdebinfo" AND NOT WIN32)
```
FedeDP, line 197: So, what happens on win32 here?

shane-lawrence: We'll have to figure out how to do the equivalent of objcopy in Windows if we want to separate out the debug info properly. I'm only familiar with one way of doing so, and it only works with Visual Studio so I'd like to come up with a more generic method.

FedeDP: Thanks for the insight! For now it is not a big deal since we don't ship windows artifacts (ie: Falco is not officially distributed for windows; we only want to make sure that we build there), perhaps we can leave a TODO comment there?

FedeDP:
Yes we are having issues with the arm64 cncf provided runners, that's not your fault :)

shane-lawrence:
I added a TODO comment for win32 and opened an issue to track it https://github.com/falcosecurity/falco/issues/3445. Is there anything else missing before we can merge this?

FedeDP:
Nope, we just need to fix arm64 runners :)

FedeDP:
@shane-lawrence can you rebase on latest master? It will fix arm64 CI ;)
Thanks!

FedeDP (approved):
/approve

shane-lawrence:
Thanks @FedeDP!

FedeDP:
Thanks to you for opening this PR :D

leogr (approved):
Thank you for this! 🙏

PR #3449, opened 2025-01-15
chore(docs): update `plugins_hostinfo` config file comment.

FedeDP:
**What type of PR is this?**

/kind cleanup

Fixes #

```release-note
NONE
```


falco.yaml
```diff
@@ -480,9 +480,12 @@ plugins:
...
     library_path: libjson.so
 
+# [Sandbox] `plugins_hostinfo`
+#
 # Uncomment to disable host info support for source plugins
-# that DO NOT generate raw events from the libscap event table,
-# dropping the `hostPath` volume requirement for them.
+# that DO NOT generate raw events from the libscap event table
+# or for plugins that DO NOT parse raw events generated by drivers,
+# effectively dropping the `hostPath` volume requirement for them.
```
sgaist, line 488: I think there should be a link to the "hostPath volume requirement".

Because, as a user, reading this, I would search in this file for more information and since there's no other mention of hostPath, would have to start digging through the doc (or maybe even the code) for what that might be.

sgaist (changes requested):

falco.yaml
FedeDP: You are right, will put a link to the charts section!

FedeDP: Done :)

leogr (approved):

sgaist (commented):
Much better !

One last question: AFAICS the pod template does not make the `proc-fs` hostPath optional so it is a bit confusing. Did I miss something ?

falco.yaml
```diff
@@ -480,9 +480,13 @@ plugins:
...
 
+# [Sandbox] `plugins_hostinfo`
+#
 # Uncomment to disable host info support for source plugins
-# that DO NOT generate raw events from the libscap event table,
-# dropping the `hostPath` volume requirement for them.
+# that DO NOT generate raw events from the libscap event table
+# or for plugins that DO NOT parse raw events generated by drivers,
+# effectively dropping the `proc-fs` hostPath volume requirement for them:
+# https://github.com/falcosecurity/charts/blob/master/charts/falco/templates/pod-template.tpl#L302
```
leogr, line 489:
```suggestion
# https://github.com/falcosecurity/charts/blob/bd57711e7c8e00919ea288716e0d9d5fdad8867e/charts/falco/templates/pod-template.tpl#L302-L304
```

very nit: permalink would be better, IMO

FedeDP:
> Did I miss something ?

@sgaist you did not miss anything! The next charts version will make it optional with a flag :) cc @alacuku

FedeDP:
> In that case, should this change be merged after the one on the chart ?

Unfortunately, the chart gets updated very very near the release, and it can even get released a couple of days after the release :( 
Also, this change was already merged, we are only fixing the comment right now!

sgaist:
> > In that case, should this change be merged after the one on the chart ?
> 
> Unfortunately, the chart gets updated very very near the release, and it can even get released a couple of days after the release :( Also, this change was already merged, we are only fixing the comment right now!

Understood
At least we have this ticket in case people are wondering for the mismatch.

leogr (approved):

PR #3450, opened 2025-01-16
update(build): relax dkms dependencies to suggestions in DEB and RPM packages

jthiltges:
Using Falco with the modern_ebpf driver does not require kernel build dependencies. This PR updates the cmake configuration to make the `dkms` and `kernel-devel` package requirements into suggestions.

**What type of PR is this?**

/kind cleanup

/area build

This makes the `kernel-devel` and `dkms` dependencies optional when installing the `falco` package, significantly reducing the number of packages installed when running in modern_ebpf mode.

On Alma 8, installing Falco 0.39.2 from RPM requires `dkms` and `kernel-devel`, which adds many dependencies for a total of 76 packages on my minimal test VM. After converting the `dkms` and `kernel-devel` requirements to suggestions with rpmrebuild, no additional dependencies beyond `falco` itself were required, and Falco appears to run fine with modern_ebpf. This PR should have the same effect on packages generated by cmake.

YES

```release-note
update(build): DEB and RPM package requirements for dkms and kernel-devel are now suggestions
```

### DIFF
--- a/cmake/modules/CPackConfig.cmake
+++ b/cmake/modules/CPackConfig.cmake
@@ -72,7 +72,7 @@ if(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "aarch64")
 	set(CPACK_DEBIAN_PACKAGE_ARCHITECTURE "arm64")
 endif()
 set(CPACK_DEBIAN_PACKAGE_HOMEPAGE "https://www.falco.org")
-set(CPACK_DEBIAN_PACKAGE_DEPENDS "dkms (>= 2.1.0.0)")
+set(CPACK_DEBIAN_PACKAGE_SUGGESTS "dkms (>= 2.1.0.0)")
 set(CPACK_DEBIAN_PACKAGE_CONTROL_EXTRA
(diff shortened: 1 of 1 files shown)


FedeDP:
Hi! Thanks for this PR!
Indeed, we would love to improve Falco packages by splitting them for-driver, so that one can install `falco-kmod` that would install a Falco configured to run with the kernel module driver and with all kmod deps, a `falco-ebpf` and normal `falco` that would just drop all unneeded deps since it runs with modern ebpf driver.

This is something on our mind since ~1yr, but we couldn't find the time to work on it.
IMHO for now, this can be a good solution since the default Falco use case does not need any additional dep.

/cc @leogr @LucaGuerra

leogr (approved):
> Hi! Thanks for this PR! Indeed, we would love to improve Falco packages by splitting them for-driver, so that one can install `falco-kmod` that would install a Falco configured to run with the kernel module driver and with all kmod deps, a `falco-ebpf` and normal `falco` that would just drop all unneeded deps since it runs with modern ebpf driver.
> 
> This is something on our mind since ~1yr, but we couldn't find the time to work on it. IMHO for now, this can be a good solution since the default Falco use case does not need any additional dep.
> 
> /cc @leogr @LucaGuerra

I totally agree with Fede, and I'm ok with merging this PR now, so approved! and
/milestone 0.40.0

Then the only possible issue I see is the documentation. I can then work to update the docs.

@jthiltges thank you for contribution! :pray:

FedeDP (approved):
/approve

PR #3482, opened 2025-02-04
new(build,userspace): switch to use container plugin

### LINKED ISSUE #3403: Reimplement containers metadata support as a plugin
**Motivation**

Historically, Falco container's information retrieval has always caused headaches to users; moreover, the current code does cause headaches to developers too.
While not always working fine, it also has some limitations:
* There is no support for multiple container runtimes on a single node: https://github.com/falcosecurity/falco/issues/3279
* Since the container info retrieval is done asynchronously at first `clone/execve/fork` event, we lose container info enrichment for the very first syscall events generated by threads spawned in a new container
* As previously stated, the code is also very hard to extend; there has been some previous attempts to improve it but for now that's what we got

**Feature**

Rewrite container info support as a plugin, and completely drop libsinsp internal implementation.
The plugin will leverage container engines SDKs to get notified when new containers are created/deleted; the container creation happens **before** the container is even started thus it should give us a time advantage and should avoid event without container metadata.
It will also be responsible of extracting all [`container`](https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/sinsp_filtercheck_container.cpp) and remaining [`k8s`](https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/sinsp_filtercheck_k8s.cpp) related filterchecks.

Main idea is to drop container manager from sinsp, and container_id from threadinfo.
The container_id will be added as a foreign key leveraging the libsinsp state table API by the plugin, that will take care of filling it.
Basically, sinsp without the plugin loaded will be completely unaware of containers.

All of the above should be done **without breaking changes** for the end users. In Falco 0.39 we added a `container_engines` configuration key: https://github.com/falcosecurity/falco/blob/master/falco.yaml#L1312; since it is in incubating state, we can drop it (since we will rely on the plugin `init_config` of course). That should be the only user-facing change.

I already have a private repo with the plugin source code i have been actively working on since mid october, and hope to be able to release it in time for Falco 0.40.0.
I also have an upstream libs branch where i am playing with the big cleanup: https://github.com/falcosecurity/libs/tree/cleanup/drop_container_manager.

I will add the link to the plugin once i am satisfied with it and open source it :) Will share its architecture too.

Main issues: 

**Alternatives**

No clear alternative.

**Additional context**

Listing related issues:
* https://github.com/falcosecurity/falco/issues/3279
* https://github.com/falcosecurity/falco/issues/2708
* https://github.com/falcosecurity/libs/issues/2022
* https://github.com/falcosecurity/libs/issues/1708
* https://github.com/falcosecurity/libs/issues/1589
* https://github.com/falcosecurity/libs/issues/1115
* https://github.com/falcosecurity/libs/issues/2057 (might be a good time to fix it?)

Next step would be to move `user/group` manager to a plugin too.

FedeDP:
**What type of PR is this?**

/kind cleanup
/kind feature

/area build
/area engine

This PR bumps Falco to HEAD of https://github.com/falcosecurity/libs/pull/2207, enforcing the usage of the container plugin.
Keeping it `wip` until all issues are resolved.

TODO:
- [x] fix CI issues -> `test_falco_engine.extra_format_do_not_replace_container_info`
- [x] the config now loads the container plugin by default. But the container plugin itself is not supported on windows and osx (and musl of course) therefore we either use a new `falco_others.yaml` config (that we install on windows,osx and musl in place of falco.yaml), or we use something like `load_plugins: [@FALCO_DEFAULT_PLUGINS@]` and customize `FALCO_DEFAULT_PLUGINS` from within cmake -> we decided to add the plugin configuration in the main falco.yaml file but then install an override config file that loads it only on non-musl linux installations
- [x] what to do with musl? We won't have any way to runtime load the plugin... So the musl build won't support containers anymore -> not a big deal for now; eventually if user demand is high, we might want to statically link the plugin somehow
- [x] fixup Falco yaml `container_engines:` key: drop in favor of plugin config -> implement TODOs or drop? It is not `sandbox` level unfortunately, but `Incubating` thus theoretically we'd need a deprecation period. -> before 1.0.0 `incubating` features can be dropped without deprecation period: https://github.com/falcosecurity/falco/blob/master/proposals/20231220-features-adoption-and-deprecation.md#before-falco-10
- [ ] Update rules? They will need to require container plugin 

While it should be up to falcoctl to install the `container` plugin, since it was previously a core feature of libraries, we decided to keep it installed by the Falco package so that host installations through packages won't lose such a feature by default (ie: without playing with falcoctl to install the plugin).
Another solution could be to let rpm/deb install scripts leverage `falcoctl` to install the plugin, BUT the `tar.gz` tarball would still be hit by the feature loss.

Refs #3403 

The cares,grpc,openssl and curl cmake files were copy pasted from libs repo.

```release-note
new(build,userspace): switch to use container plugin
```


FedeDP:
Going to remove the wip commit once https://github.com/falcosecurity/testing/pull/69 gets merged.

FedeDP:
Now the arm64 test-dev-packages is down to 4 errors, 2 of them are related to the usage of the full branch name for libs and driver deps; they will be fixed once we properly use a commit hash.

The remaining 2 are actually linked to the container plugin: it seems replaying a capture file we are seeing 2 container events instead of 3:
```
  TestFalco_Legacy_ContainerPrivileged

{"deadline":180000000000,"level":"info","msg":"running falco with runner","time":"2025-02-13T16:41:57Z"}
{"cmd":"/usr/bin/falco -c /etc/falco/falco.yaml -o append_output.suggested_output=false -o json_output=true -r falco_rules.yaml -o engine.kind=replay -o engine.replay.capture_file=container-privileged.scap -A -o json_include_output_property=false -o json_include_tags_property=false -o log_level=debug -o log_stderr=true -o log_syslog=false -o stdout_output.enabled=true","level":"debug","msg":"executing command","time":"2025-02-13T16:41:57Z"}
    legacy_test.go:2498: 
        	Error Trace:	/home/runner/work/_actions/falcosecurity/testing/main/legacy_test.go:2498
        	Error:      	Not equal: 
        	            	expected: 3
        	            	actual  : 2
        	Test:       	TestFalco_Legacy_ContainerPrivileged

  TestFalco_Legacy_ContainerSensitiveMount

{"deadline":180000000000,"level":"info","msg":"running falco with runner","time":"2025-02-13T16:41:56Z"}
{"cmd":"/usr/bin/falco -c /etc/falco/falco.yaml -o append_output.suggested_output=false -o json_output=true -r falco_rules.yaml -o engine.kind=replay -o engine.replay.capture_file=container-sensitive-mount.scap -A -o json_include_output_property=false -o json_include_tags_property=false -o log_level=debug -o log_stderr=true -o log_syslog=false -o stdout_output.enabled=true","level":"debug","msg":"executing command","time":"2025-02-13T16:41:56Z"}
    legacy_test.go:2517: 
        	Error Trace:	/home/runner/work/_actions/falcosecurity/testing/main/legacy_test.go:2517
        	Error:      	Not equal: 
        	            	expected: 3
        	            	actual  : 2
        	Test:       	TestFalco_Legacy_ContainerSensitiveMount
```

FedeDP:
The failing tests come from the fact that we are not able to extract `container` fields from the `container` events. Will look into it.

FedeDP:
About `test-dev-packages (static)`: we need to somehow ensure that the testing framework does not use `container.foo` fields in its rules; but this is impossible since many rules use `container.id != host` as a condition.
We know that the musl build won't support loading the container plugin (at least not dynamically and not in 0.41.0), thus i think best we can do is to disable the required check for that job, for now. Or disable the test entirely (to avoid having a red cross CI forever and ever).

FedeDP:
Couldn't reproduce the `x86_64` ASSERT using a sinsp-example debug build from https://github.com/falcosecurity/libs/pull/2207 HEAD. :/

FedeDP:
A quick introspection using gdb and the x86_64 sanitizer artifact, shows no weird issues; basically here `get_main_thread` returns a nullptr: https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/threadinfo.h#L515
And that's because:
* `is_main_thread` returns false (https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/threadinfo.h#L216)
* `m_tginfo` is NOT nullptr
* `m_tginfo->get_first_thread()` correctly returns first thread
* `possible_main->is_main_thread()` is false, thus returning nullptr: https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/threadinfo.h#L228

```
(gdb) p is_main_thread()
$10 = false
(gdb) p m_tginfo
$11 = std::shared_ptr<thread_group_info> (use count 2, weak count 0) = {
  get() = 0x506000168510}
(gdb) p m_tginfo->get_first_thread()
$12 = (sinsp_threadinfo *) 0x51c0003c7880
(gdb) p m_tginfo->get_first_thread()->is_main_thread()
$13 = false
```

Also `m_tid`, `m_pid` and `m_fd_table` seems ok-ish (i mean, we don't have weird memory bugs):
```
(gdb) p m_tid 
$1 = 749869
(gdb) p m_pid
$2 = 749865
(gdb) p m_fdtable
$3 = {<libsinsp::state::built_in_table<long>> = {<libsinsp::state::table<long>> = {<libsinsp::state::base_table> = {
        _vptr.base_table = 0x5555581aaae0 <vtable for sinsp_fdtable+16>, 
        m_key_info = {m_name = 0x55555761a700 "int64", 
          m_type_id = SS_PLUGIN_ST_INT64, m_size = 8, 
          m_construct = 0x555556011806 <libsinsp::state::typeinfo::_construct<long, std::allocator<long> >(void*)>, 
          m_destroy = 0x555556011970 <libsinsp::state::typeinfo::_destroy<long, std::allocator<long> >(void*)>}}, <No data fields>}, 
    m_this_ptr = 0x51c0003c7b00, m_name = "file_descriptors", 
    m_static_fields = 0x55555859b4c0 <s_fdtable_static_fields>, 
    m_field_list = std::vector of length 0, capacity 0, 
    m_field_accessors = std::unordered_map with 0 elements, 
    m_dynamic_fields = std::shared_ptr<libsinsp::state::dynamic_struct::field_infos> (use count 7139, weak count 0) = {get() = 0x50b000008190}}, 
  m_inspector = 0x51b000004690, m_table = std::unordered_map with 0 elements, 
  m_sinsp_stats_v2 = std::shared_ptr<sinsp_stats_v2> (use count 1902, weak count 0) = {get() = 0x50d000015320}, m_last_accessed_fd = -1, 
  m_last_accessed_fdinfo = std::shared_ptr<sinsp_fdinfo> (empty) = {
    get() = 0x0}, m_tid = 749869, 
  m_nullptr_ret = std::shared_ptr<sinsp_fdinfo> (empty) = {get() = 0x0}}
```

So, this seems a legit `ASSERT` trigger (ie: we might want to drop that `ASSERT`). At the same time, 2 questions remain:
* why only bpf and kmod are hit?
* why the very same thing does not happen using libs master? (eg: here https://github.com/falcosecurity/falco/pull/3466 the test CI pass fine)

FedeDP:
So, it seems like the culprit process is `podman`:
```
(gdb) p m_exe
$2 = "/usr/bin/podman"
(gdb) p m_cwd
$3 = ""
(gdb) p m_comm
$4 = "podman"
(gdb) p m_args
$5 = std::vector of length 3, capacity 3 = {"--log-level=info", "system", 
  "service"}
```

FedeDP:
> why only bpf and kmod are hit?

It seems modern ebpf is hit too:
```
grep podman modern.txt 
NULL fd_table_ptr: 'podman' 'podman --log-level=info system service'
```
After adding: 
```C
sinsp_fdinfo* sinsp_threadinfo::add_fd(int64_t fd, std::unique_ptr<sinsp_fdinfo> fdinfo) {
	sinsp_fdtable* fd_table_ptr = get_fd_table();
	if(fd_table_ptr == NULL) {
		printf("NULL fd_table_ptr: '%s' '%s'\n", m_comm.c_str(), m_cmd_line.c_str());
		return NULL;
	}
```

> why the very same thing does not happen using libs master? 

Because the culprit is the podman service started by the podman SDK withing the go-worker of the plugin. Thus no plugin, no issue.

I added a commit on top if my libs PR to drop the `ASSERT()`. Note that `get_fd_table` is used multiple times inside `threadinfo.cpp` and the most recent methods using it, already did not use the `ASSERT`. That's because it can happen that the get_fd_table returns null, given its impl:
```
inline sinsp_fdtable* get_fd_table() {
		if(!(m_flags & PPM_CL_CLONE_FILES)) {
			return &m_fdtable;
		} else {
			sinsp_threadinfo* root = get_main_thread();
			return (root == nullptr) ? nullptr : &(root->get_fdtable());
		}
	}
```

Probably there was a time where we always returned `&m_fdtable` (ie: when we never enforced the `PPM_CL_CLONE_FILES` flag. Nowadays, we are always enforcing that flag, as stated in the comment above the function:
> /* Note that `fd_table` should be shared with the main thread only if `PPM_CL_CLONE_FILES`
	 * is specified. Today we always specify `PPM_CL_CLONE_FILES` for all threads.
	 */

FedeDP:
1 failure left on x86_64:
* TestFalco_Legacy_KernelUpgrade

No idea why this is failing.

2 failures on arm64:
* TestFalcoLegacyBPF
* TestFalcoKmod

Both of them are present in https://github.com/falcosecurity/falco/actions/runs/13330903373 too (ie: they do not seem to depend on the container work).

Many failures on x86_64 musl; that's because we cannot load the plugin on musl, as already stated in the PR body. I think we will just need to remove the required check on it (or just stop testing musl build).

FedeDP:
https://github.com/falcosecurity/testing/pull/71 fixes arm64 kmod and legacyBpf failures. 
Still investigating x86_64 TestFalco_Legacy_KernelUpgrade.

cmake/modules/cares.cmake
```diff
@@ -0,0 +1,76 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2023 The Falco Authors.
```
leogr, line 3:
```suggestion
# Copyright (C) 2025 The Falco Authors.
```

cmake/modules/curl.cmake
```diff
@@ -0,0 +1,100 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2023 The Falco Authors.
```
leogr, line 3:
```suggestion
# Copyright (C) 2025 The Falco Authors.
```

cmake/modules/grpc.cmake
```diff
@@ -0,0 +1,274 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2023 The Falco Authors.
```
leogr, line 3:
```suggestion
# Copyright (C) 2025 The Falco Authors.
```

cmake/modules/openssl.cmake
```diff
@@ -0,0 +1,81 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2023 The Falco Authors.
```
leogr, line 3:
```suggestion
# Copyright (C) 2025 The Falco Authors.
```

userspace/falco/app/actions/init_falco_engine.cpp
```diff
@@ -95,8 +95,8 @@ void configure_output_format(falco::app::state& s) {
 
 	// See https://falco.org/docs/rules/style-guide/
 	const std::string container_info =
-	        "container_id=%container.id container_image=%container.image.repository "
-	        "container_image_tag=%container.image.tag container_name=%container.name";
+	        "container_image=%container.image.repository "
+	        "container_image_tag=%container.image.tag";
```
leogr, line 99: Note for ourselves: we need to update the documentation https://falco.org/docs/rules/style-guide/
cc @LucaGuerra

userspace/engine/falco_engine_version.h
```diff
@@ -20,7 +20,7 @@ limitations under the License.
 
 // The version of this Falco engine
 #define FALCO_ENGINE_VERSION_MAJOR 0
-#define FALCO_ENGINE_VERSION_MINOR 46
+#define FALCO_ENGINE_VERSION_MINOR 47
```
leogr, line 23: Should we bump the major? :thinking: 
cc @jasondellaluce

leogr (commented):
:rocket: 

Just nits :sweat_smile:

LucaGuerra (approved):

userspace/engine/falco_engine_version.h
LucaGuerra: I agree with bumping the major

userspace/falco/app/actions/init_falco_engine.cpp
LucaGuerra: Once this is merged let's open an issue

userspace/engine/falco_engine_version.h
FedeDP: No problem for me; if we all agree, i'll push the change.
cc @jasondellaluce

FedeDP: /hold while we wait for response.

leogr: > We never bumped the major, but we did remove fields in the past 🤔 should be ok with minor I think

That's precisely why I'm asking for this. 
I'm okay with bumping the minor since, even for semver, backward incompatible changes can be introduced at any time when the major is `0`. Just wanted to double-check.

FedeDP:
I'd avoid re-pushing just to fix some licenses year (since it means to wait ~50minutes for the CI).

userspace/engine/falco_engine_version.h
leogr: Briefly discussed with other maintainers, let's keep the minor bump for now.

PR #3484, opened 2025-02-05
refactor(falco/app): apply early return pattern in actions code

ekoops:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind release

> /area build

/area engine

> /area tests

> /area proposals

> /area CI

This PR simplifies the falco app actions code by applying early return pattern.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/falco/app/actions/init_inspectors.cpp
+++ b/userspace/falco/app/actions/init_inspectors.cpp
@@ -147,13 +147,14 @@ falco::app::run_result falco::app::actions::init_inspectors(falco::app::state& s
 	std::string err;
 	std::unordered_set<std::string> used_plugins;
 	const auto& all_plugins = s.offline_inspector->get_plugin_manager()->plugins();
+	const bool is_capture_mode = s.is_capture_mode();
 
 	for(const auto& src : s.loaded_sources) {
 		auto src_info = s.source_infos.at(src);
 
 		// in capture mode, every event source uses the offline inspector.
 		// in live mode, we create a new inspector for each event source
-		if(s.is_capture_mode()) {
+		if(is_capture_mode) {
 			src_info->inspector = s.offline_inspector;
 		} else {
 			src_info->inspector =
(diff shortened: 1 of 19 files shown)


sgaist (approved):

userspace/falco/app/actions/init_inspectors.cpp
```diff
@@ -224,17 +227,18 @@ falco::app::run_result falco::app::actions::init_inspectors(falco::app::state& s
...
-			const auto& sources = src_info->inspector->event_sources();
-			if(sources.size() == 0 || sources.size() > 2 ||
-			   sources[0] != falco_common::syscall_source) {
-				err.clear();
-				for(const auto& source : sources) {
-					err += (err.empty() ? "" : ", ") + source;
-				}
-				return run_result::fatal("Illegal sources setup in live inspector for source '" +
-				                         src + "': " + err);
+		if(is_capture_mode) {
```
sgaist, line 230: Shouldn't the comment just above the if be moved below close the source size check ?
It describe the logic in the body quite well however reading it and then having the capture mode check and continue statement makes things harder to reason about.

ekoops: Yeah you are right. I was not sure which of the two styles to prefer but I guess your alternative is better. Gonna fix it here and in other places where the same logic applies.

userspace/falco/app/actions/init_inspectors.cpp
```diff
@@ -194,22 +195,24 @@ falco::app::run_result falco::app::actions::init_inspectors(falco::app::state& s
...
-				// inspector if we're in capture mode
-				if(!s.is_capture_mode() || used_plugins.find(p->name()) == used_plugins.end()) {
-					if(!plugin->init(config->m_init_config, err)) {
-						return run_result::fatal(err);
-					}
-				}
-				if(is_input) {
-					auto gen_check = src_info->inspector->new_generic_filtercheck();
-					src_info->filterchecks->add_filter_check(std::move(gen_check));
+			if(!plugin) {
```
sgaist, line 200: Same issue as before in this case.

sgaist (changes requested):

userspace/falco/app/actions/init_inspectors.cpp
ekoops: Mh... I didn't move it since I think that this comment is more related to the reason why we are `continue`ing there. My interpretation is: "don't initialize the plugin if it hasn't been registered into an inspector". Maybe I can rephrase the comment this way...

ekoops: Applied suggestion

sgaist (dismissed):
LGTM

FedeDP (approved):
/approve

leogr (approved):

PR #3488, opened 2025-02-06
new(docker,scripts,ci): use an override config file to enable ISO 8601 output timeformat on docker images

FedeDP:
**What type of PR is this?**

/kind cleanup

/area build
/area CI

This PR goes in the "tailor Falco to specific scenario" direction; in this first round, we move the ugly `sed ...` of the Falco config file to enable ISO 8601 output timeformat in docker images to an override config file that gets installed by docker images.

Fixes #

```release-note
new(docker,scripts,ci): use an override config file to enable ISO 8601 output timeformat on docker images
```


.github/workflows/reusable_build_docker.yaml
```diff
@@ -43,8 +43,7 @@ jobs:
 
       - name: Build falco image
         run: |
-          cd ${{ github.workspace }}/docker/falco/
-          docker build -t docker.io/falcosecurity/falco:${{ inputs.arch }}-${{ inputs.tag }} \
+          docker build -f docker/falco/Dockerfile -t docker.io/falcosecurity/falco:${{ inputs.arch }}-${{ inputs.tag }} \
```
FedeDP, line 46: Since we need to access `scripts/falco/` directory, we need our context to be the whole repo.

LucaGuerra (dismissed):

sgaist (changes requested):

FedeDP:
Thanks Samuel, didn't know about that :)
I think that a good content should be:
```
*
!scripts/falco/falco_iso8601_timeformat.yaml
!docker/
```

WDYT?

leogr (commented):
Hey

These changes SGTM! 
The only recommendation I can give is to move 
`scripts/falco/falco_iso8601_timeformat.yaml`
:point_down: 
`config/falco.iso8601_timeformat.yaml`

The rationale for this is:
- The `config` folder can represent a more abstract concept for additional configs in the future; if we also add a dedicated README into the folder, it will give contributors a clear view of the purpose of the folder
- `falco.<config-name>.yaml` pattern would reassemble the convention used for `falco_rules.local.yaml`; we may want to rely on this convention to clearly indicate that such files are intended as _overrides_.

FedeDP:
@leogr makes sense! Will apply asap :)

FedeDP:
@leogr done!

FedeDP:
@sgaist PTAL :pray:

sgaist (commented):
Overall this looks fine
I was just double checking with the documentation and I am wondering whether adopting the structure from the example shown in the [dockerignore part](https://docs.docker.com/build/concepts/context/#dockerignore-files) (using multiple files rather than multiple directories) would make things simpler ?

FedeDP:
Mmh i don't expect our dockerignore list to grow that big, but we are always able to make that change in the future if we need to.
Also, i don't expect the dockerfiles to need different dockerignores, but again we can always improve it later if needed.

sgaist:
> Mmh i don't expect our dockerignore list to grow that big, but we are always able to make that change in the future if we need to. Also, i don't expect the dockerfiles to need different dockerignores, but again we can always improve it later if needed.

I was rather thinking of having one directory with the various `XYZ.Dockerfile` in it rather than one Dockerfile within a sub-directory.
For the `.dockerignore` file, yes, I agree.

FedeDP:
> I was rather thinking of having one directory with the various XYZ.Dockerfile in it rather than one Dockerfile within a sub-directory.

Ah! But the entrypoint is different for each of them (when present); it is surely feasible (we can rename the docker-entrypoint.sh scripts) but at that point i feel it is better the way it is now with subfolders.

sgaist (approved):
Fair enough, it's a good base in any case.

PR #3493, opened 2025-02-13
fix(userspace/falco): init cmdline options after loading all config files

FedeDP:
**What type of PR is this?**

/kind bug

/area engine

Since I introduced Falco sub-config files, since sub-config files were loaded **after** considering the `-o` cmdline override flags, their overridden config entries would win on the `-o` flags.
This PR fixes the behavior.

Fixes #

```release-note
fix(userspace/falco): init cmdline options after loading all config files
```

### DIFF
--- a/unit_tests/falco/test_configuration_config_files.cpp
+++ b/unit_tests/falco/test_configuration_config_files.cpp
@@ -466,6 +466,9 @@ TEST(Configuration, configuration_config_files_cmdline) {
 	std::vector<std::string> cmdline_config_options;
 	cmdline_config_options.push_back((yaml_helper::configs_key + "=conf_2.yaml"));
 
+	// Override foo2 value from cli
+	cmdline_config_options.push_back(("foo2=bar22"));
+
 	falco_configuration falco_config;
 	config_loaded_res res;
 	ASSERT_NO_THROW(res = falco_config.init_from_file("main.yaml", cmdline_config_options));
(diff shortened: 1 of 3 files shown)


userspace/falco/configuration.cpp
```diff
@@ -133,12 +133,20 @@ config_loaded_res falco_configuration::init_from_file(
 		std::cerr << "Cannot read config file (" + conf_filename + "): " + e.what() + "\n";
 		throw e;
 	}
-	init_cmdline_options(cmdline_options);
 
 	// Only report top most schema validation status
 	res[conf_filename] = validation_status[0];
 
+	// Load any `-o config_files=foo.yaml` cmdline additional option
+	load_cmdline_config_files(cmdline_options);
```
FedeDP, line 141: We first load any `-o config_files...` override CLI option.

unit_tests/falco/test_configuration_config_files.cpp
```diff
@@ -466,6 +466,9 @@ TEST(Configuration, configuration_config_files_cmdline) {
 	std::vector<std::string> cmdline_config_options;
 	cmdline_config_options.push_back((yaml_helper::configs_key + "=conf_2.yaml"));
 
+	// Override foo2 value from cli
+	cmdline_config_options.push_back(("foo2=bar22"));
```
FedeDP, line 470: The updated test avoids future similar issues.

sgaist (approved):

PR #3507, opened 2025-02-28
cleanup(falco.yaml): drop `verbosity` from container plugin init config.

FedeDP:
**What type of PR is this?**

/kind cleanup

/area build

Next plugin release will make use of libs logger instead of spdlog; this means we don't need a different verbosity flag, we will just use `libs_logger.severity`.

Also, this PR aims at enabling `libs_logger` by default, with `info` level. This way, plugins that use the provided logger will see their messages logged by default.

Fixes #

```release-note
chore(falco.yaml): enable libs_logger by default with info level
```

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -905,8 +905,8 @@ log_level: info
 # library specifically, providing more granular control over the logging
 # behavior of the underlying components used by Falco. Only logs of a certain
 # severity level or higher will be emitted. Supported levels: "fatal",
-# "critical", "error", "warning", "notice", "info", "debug", "trace". It is not
-# recommended to use "debug" and "trace" for production use.
+# "critical", "error", "warning", "notice", "info", "debug", "trace".
+#  It is not recommended to use "debug" and "trace" for production use.
 libs_logger:
   enabled: true
   severity: info
--- a/falco.yaml
+++ b/falco.yaml
@@ -470,7 +470,6 @@ plugins:
     # For a summary of config option, see https://github.com/FedeDP/container_plugin?tab=readme-ov-file#configuration
     library_path: libcontainer.so
     init_config:
-      verbosity: warning
       label_max_len: 100
       with_size: false
 # We use default config values for engine key.
--- a/falco.yaml
+++ b/falco.yaml
@@ -906,10 +906,10 @@ log_level: info
 # behavior of the underlying components used by Falco. Only logs of a certain
 # severity level or higher will be emitted. Supported levels: "fatal",
 # "critical", "error", "warning", "notice", "info", "debug", "trace". It is not
-# recommended for production use.
+# recommended to use "debug" and "trace" for production use.
 libs_logger:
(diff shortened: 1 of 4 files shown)


falco.yaml
```diff
@@ -907,10 +906,10 @@ log_level: info
 # behavior of the underlying components used by Falco. Only logs of a certain
 # severity level or higher will be emitted. Supported levels: "fatal",
 # "critical", "error", "warning", "notice", "info", "debug", "trace". It is not
-# recommended for production use.
+# recommended to use "debug" and "trace" for production use.
```
sgaist, line 909: Out of curiosity, why are these two values recommended for production use ?

From old habits, I would use them when having an issue so for debugging.

FedeDP: >  It is **not**
> recommended to use "debug" and "trace" for production use.

:laughing:

sgaist: 😅 My bad !

sgaist, line 909: To avoid other people falling in the same trap I did, I would suggest:
```suggestion
# "critical", "error", "warning", "notice", "info", "debug", "trace". It is 
# not recommended to use "debug" and "trace" in production.
```

sgaist (changes requested):

falco.yaml
FedeDP: Pushed the change, thanks!

sgaist (approved):

unit_tests/falco/test_configuration.cpp
```diff
@@ -143,6 +143,9 @@ TEST(Configuration, configuration_webserver_ip) {
 		EXPECT_NO_THROW(falco_config.init_from_content("", cmdline_config_options));
 
 		ASSERT_EQ(falco_config.m_webserver_config.m_listen_address, address);
+
+		// Reset sinsp logger
+		falco_logger::set_sinsp_logging(false, "", "");
```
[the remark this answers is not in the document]
FedeDP: Another solution would be to add that line in falco_configuration ctor, and create a new falco_configuration at each iteration, but i preferred to only touch the tests.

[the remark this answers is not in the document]
FedeDP: Mmmh no that does not fix it for all cases.
I think i will just drop the assert from libs, is not a big deal.

FedeDP:
Opened https://github.com/falcosecurity/libs/pull/2310.

unit_tests/falco/test_configuration.cpp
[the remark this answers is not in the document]
FedeDP: Updated libs and driver to latest master that dropped the assert.

FedeDP:
Updated libs and driver to latest master that dropped the `assert`.

leogr (approved):
LGTM!

LucaGuerra (approved):

PR #3529, opened 2025-04-09
update(userspace/engine): relax validation for `values` in `exceptions`

leogr:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind release

> /area build

/area engine

> /area tests

> /area proposals

> /area CI

Not really a bug, still.
Defining `exceptions` with empty `values` is a legitimate use case since the values can be added to another rules file. Even when values are not populated elsewhere, Falco can work without issues; that's the reason why the `values` field is not required. With this change, we avoid emitting useless validation warnings in situations where exceptions are just defined but not actually used because values are not being provided.

Fixes #

/milestone 0.41.0

```release-note
update: empty `values` in `exceptions` won't emit a warning anymore
```

### DIFF
--- a/userspace/engine/rule_json_schema.h
+++ b/userspace/engine/rule_json_schema.h
@@ -119,8 +119,7 @@ const char rule_schema_string[] = LONG_STRING_CONST(
                 "values": {}
             },
             "required": [
-                "name",
-                "values"
+                "name"
             ],
             "title": "Exception"
         },


FedeDP (approved):
/approve
Makes sense.

LucaGuerra (approved):

PR #3538, opened 2025-04-24
new(userspace,unit_tests)!: add a way to specify merge-strategy for `config_files`

FedeDP:
**What type of PR is this?**

/kind release

/area engine
/area tests

By default we now use the `append` merge-strategy:
* existing sequence keys will be appended
* existing scalar keys will be overridden
* non-existing keys will be added

We also have an `override` merge-strategy:
* existing keys will be overridden
* non-existing keys will be added

Finally, there is an `add-only` merge-strategy:
* existing keys will be ignored
* non-existing keys will be added

To switch between strategies, one must prepend special characters to `config_files` entries: `@` to enable `append` strategy, or `+` to enable `add-only` strategy.

This is a breaking change because the old default behavior was `override` one. But, since we now ship a `falco.container_plugin.yaml` to load the `container` plugin, we want that config to append to `load_plugins`, not to override, otherwise it would kill any loaded plugins by the user.
Also, it seems a more sensible default.

Fixes #

```release-note
new(userspace,unit_tests)!: add a way to specify merge-strategy for `config_files`
```

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -144,22 +144,30 @@
 #
 # Like for 'rules_files', specifying a folder will load all the configs files present in it in a lexicographical order.
 #
-# 3 merge-strategies are available: `append`, `override` and `add-only`.
-# The default merge-strategy is `append`:
+# 3 merge-strategies are available:
+# `append` (default):
 #   * existing sequence keys will be appended
 #   * existing scalar keys will be overridden
 #   * non-existing keys will be added
-# To enable the `override` merge-strategy, `@` must be prepended to the entry, eg: "@/$HOME/override.yaml".
+# `override`:
 #   * existing keys will be overridden
 #   * non-existing keys will be added
-#   Please note that since `@` is a reserved character by yaml spec, you need to quote the whole string.
-# To enable the `add-only` merge-strategy, `+` must be prepended to the entry, eg: "+/$HOME/append.yaml".
+# `add-only`:
 #   * existing keys will be ignored
 #   * non-existing keys will be added
 #
-# When a merge-strategy is enabled for a folder entry, all of the included config files will use that merge-strategy.
+# Each item on the list can be either a yaml map or a simple string.
+# The simple string will be interpreted as the config file path, and the `append` merge-strategy will be enforced.
+# When the item is a yaml map instead, it will be of the form: `   path: foo\n   strategy: X`.
+# When `strategy` is omitted, once again `append` is used.
+#
+# When a merge-strategy is enabled for a folder entry, all the included config files will use that merge-strategy.
 config_files:
   - /etc/falco/config.d
+  # Example of config file specified as yaml map with strategy made explicit.
+  #- path: $HOME/falco_local_configs/
+  #  strategy: add-only
+
 
 # [Stable] `watch_config_files`
 #
--- a/falco.yaml
+++ b/falco.yaml
@@ -143,6 +143,21 @@
 # Also, nested include is not allowed, ie: included config files won't be able to include other config files.
 #
 # Like for 'rules_files', specifying a folder will load all the configs files present in it in a lexicographical order.
+#
+# 3 merge-strategies are available: `append`, `override` and `add-only`.
+# The default merge-strategy is `append`:
+#   * existing sequence keys will be appended
+#   * existing scalar keys will be overridden
+#   * non-existing keys will be added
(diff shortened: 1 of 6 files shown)


userspace/falco/configuration.cpp
```diff
@@ -194,8 +198,11 @@ void falco_configuration::merge_config_files(const std::string &config_name,
 				}
 			}
 			std::sort(v.begin(), v.end());
-			for(const auto &f : v) {
-				m_config.include_config_file(f, m_config_schema, &validation_status);
+			for(auto f : v) {
```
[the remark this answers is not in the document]
FedeDP: Oh you are right, this was done while i was scratching. Will remove it!

sgaist (dismissed):

falco.yaml
```diff
@@ -143,6 +143,21 @@
...
+#   * non-existing keys will be added
+# To enable the `override` merge-strategy, `@` must be prepended to the entry, eg: "@/$HOME/override.yaml".
+#   * existing keys will be overridden
+#   * non-existing keys will be added
+#   Please note that since `@` is a reserved character by yaml spec, you need to quote the whole string.
+# To enable the `add-only` merge-strategy, `+` must be prepended to the entry, eg: "+/$HOME/append.yaml".
+#   * existing keys will be ignored
+#   * non-existing keys will be added
+#
+# When a merge-strategy is enabled for a folder entry, all of the included config files will use that merge-strategy.
```
leogr, line 160: Although I love the idea of having several merge strategies and this lean solution, `@` and `+` are allowed characters in filenames (across all common OSes, i.e. Linux, Mac, Windows). Since the path can be relative, filenames like the following are legit:
- `@file.txt`
- `+file.txt`

For this reason, I think that prepending symbols is not a viable solution :sweat: 

A more structured (and cumbersome - I know) solution would be something like:





```yaml
config_files:
  # default merge strategy (i.e., `append`)
  - /etc/falco/config.d
  # `append` strategy (explicit)
  - path: /etc/falco/config.d/file1.yaml
    strategy: append
  # `override` strategy
  - path: /etc/falco/config.d/file2.yaml
    strategy: override
  # `add-only` strategy
  - path: /etc/falco/config.d/file3.yaml 
    strategy: add-only
```

wdyt? :thinking:

leogr: PS
JSON schema allow to have different schemeas depending on which type the key is, for example:

```json
"oneOf": [
        { "type": "string" },
        {
          "type": "object",
          "properties": {
            "path": { "type": "string" },
            "strategy": { "type": "string" },
          },
```

FedeDP: We can do that too! Will work on it asap

FedeDP: Done. Note that most of lines changed comes from config json schema being re-formatted.

userspace/engine/yaml_helper.h
```diff
@@ -85,6 +85,12 @@ class yaml_helper {
 	inline static const std::string validation_failed = "failed";
 	inline static const std::string validation_none = "none";
 
+	enum config_files_strategy {
+		STRATEGY_APPEND,  // append to existing sequence keys, override scalar keys and add new ones
+		STRATEGY_OVERRIDE,  // override existing keys (sequences too) and add new ones
+		STRATEGY_ADDONLY    // only add new keys and ignore existing ones
+	};
+
```
sgaist, line 93: It might be overkill but shouldn't we have something here that allows to convert this enum from/to string ?
This would keep all the conversions in one place close to the enum.

sgaist (commented):
The new format was a very good idea !

Just one question about the enum handling but otherwise looks good.

FedeDP:
Rebased on top of new master.

userspace/engine/yaml_helper.h
FedeDP: > It might be overkill but shouldn't we have something here that allows to convert this enum from/to string

Yep i agree! Will do that in a moment!

FedeDP: Done!

userspace/falco/config_json_schema.h
```diff
@@ -38,7 +38,30 @@ const char config_schema_string[] = LONG_STRING_CONST(
                 "config_files": {
                     "type": "array",
                     "items": {
-                        "type": "string"
+                        "oneOf": [
```
FedeDP, line 41: Either string or an object.
In case of object, `path` is required, while `strategy` must be an enum between `append,override,add-only`.

sgaist (approved):
LGTM

Issue #3049, opened 2024-02-01
Follow CLI args standards in Falco


incertum:
Proposing to follow common and standard Program Argument Syntax Conventions for Falco's CLI options, for example see this [resource](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html). For example [-pk](https://github.com/falcosecurity/falco/blob/40f4ce008ab2b67ef30129036129d121fe7e938d/userspace/falco/app/options.cpp#L193) and similar options seem to fall outside of these conventions.

GLVSKiriti:
@Andreagit97 @incertum I am interested in this issue!
Can I work on this?

Andreagit97:
Not clear to me what was the plan @incertum had in mind when she opened this issue, so I'll let her answer here.
My 2 cents, this issue seems related to possible breaking changes in the Falco UX, maybe something that we want to tackle for Falco 1.0.0, not sure this is a "good first issue" but let's see what others think about that

incertum:
Goal is to correct the `-pc`, `-pk` or `-pcontainer`, `-pkubernetes` flags that do not conform to standard CLI practices. Usually `-pc` implies a combination of `-p` and `-c`.

https://falco.org/docs/outputs/formatting/
https://github.com/falcosecurity/falco/blob/99781f79364264411fc4e95d8dcd1edf2629ddff/userspace/falco/app/options.cpp#L187

I would propose them being:
- `-pc` -> `--print-container` and  `--print-container-gvisor` or similar
- `-pk` -> `--print-kubernetes`  and `--print-kubernetes-gvisor`  or similar

Suggesting to sync with @leogr.

leogr:
> Usually `-pc` implies a combination of `-p` and `-c`.

This is the main issue, likely.

If we follow [POSIX's recommendation](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html):

> Multiple options may follow a hyphen delimiter in a single token if the options do not take arguments. Thus, ‘-abc’ is equivalent to ‘-a -b -c’.

So, `-pc` is not ok.

But:

> An option and its argument may or may not appear as separate tokens. (In other words, the whitespace separating them is optional.) Thus, -o foo and -ofoo are equivalent.

So `-pcontainer` is equivalent to `-p container` and it would be ok (and it is what's happening [here](https://github.com/falcosecurity/falco/blob/master/userspace/falco/app/actions/init_falco_engine.cpp#L34)).

[...]

PR #3543, opened 2025-04-30
cleanup(userspace): deprecate print container info

FedeDP:
**What type of PR is this?**

/kind cleanup

/area engine
/area tests

PR is split in logical commits:
* deprecate usage of `%container.info` in rules. Right now it is replaced by empty string (thus useless); its support is kept for now to avoid breaking users rules. Falco default ruleset will ship without it in the next version (https://github.com/falcosecurity/rules/pull/280)
* deprecate `-p` options: `-pc` and `-pk` are useless since container plugin takes care of suggesting the same output formats; only gvisor still makes sense (until it will be ported as a plugin); at the same time, the not-so-new `-o append_output` option is already doing the same thing in a more customizable way
* dropped `replace_container_info` extra format option; we don't need it anymore
* fixed unit_tests

Fixes #

```release-note
cleanup(userspace)!: deprecate print of `container.info`
```


unit_tests/engine/test_extra_output.cpp
```diff
@@ -92,32 +92,7 @@ TEST_F(test_falco_engine, extra_format_by_tag_rule) {
 	EXPECT_EQ(get_compiled_rule_output("a_third_rule"), "out 3 extra 1 extra 3");
 }
 
-TEST_F(test_falco_engine, extra_format_replace_container_info) {
```
FedeDP, line 95: No need to test this since `replace_container_info` has been dropped.

leogr:
cc @LucaGuerra (see https://github.com/falcosecurity/falco/issues/3049#issuecomment-2587237384 )

leogr:
> 
> 
> /hold

false positive
/unhold

leogr (approved):
LGTM!

userspace/falco/app/options.cpp
```diff
@@ -117,7 +117,7 @@ void options::define(cxxopts::Options& opts)
...
+		("p,print",                  "DEPRECATED: use -o append_output... instead. Print (or replace) additional information in the rule's output.\nUse -pc or -pcontainer to append container details to syscall events.\nUse -pk or -pkubernetes to add both container and Kubernetes details to syscall events.\nIf using gVisor, choose -pcg or -pkg variants (or -pcontainer-gvisor and -pkubernetes-gvisor, respectively).\nIf a syscall rule's output contains %container.info, it will be replaced with the corresponding details. Otherwise, these details will be directly appended to the rule's output.\nAlternatively, use -p <output_format> for a custom format. In this case, the given <output_format> will be appended to the rule's output without any replacement to all events, including plugin events.", cxxopts::value(print_additional), "<output_format>")
```
LucaGuerra, line 120: This still mentions `%container.info`, but this PR removes support for it

LucaGuerra:
Generally speaking this LGTM. The only concern is that this is not just a deprecation but a (slightly breaking) change: since rules no longer support `%container.info` that string won't be replaced (maybe we should double check that but there is a test so I trust it 😅 ) and so users should be notified a bit better that this does not work anymore since otherwise they may (rightly!) open a bug report because they might be seeing spurious `%container.info` strings in their output.

userspace/falco/app/options.cpp
FedeDP: `%container.info` is still supported even if it does nothing (ie: replaces it with empty string)

FedeDP: Pushed, hopefully it is now good to go :)

LucaGuerra (approved):

leogr (approved):

PR #3551, opened 2025-05-09
cleanup(engine)!: only consider .yaml/.yml rule files

LucaGuerra:
**What type of PR is this?**

/kind cleanup

/area engine

Normally, Falco would try to load any file or directory specified in `rules_files` or with the `-r` option. However, if you are trying to load a non-yaml file or a directory which contains both yaml and other types of files Falco would emit a fatal error on the first file that fails to load. Since it makes sense to put also other files in the rule directory (such as a `README`) this PR restricts the files that are considered to `.yaml` and `.yml`, ignoring the rest. Likewise, single files that do not end in `.yml` or `.yaml` will not be loaded.

Thanks to @tks98 for noticing the behavior and suggesting this improvement!

/milestone 0.41.0

Fixes #

```release-note
cleanup(engine)!: only consider .yaml/.yml rule files
```


LucaGuerra:
/unhold

false positive

FedeDP (approved):
/approve

leogr (approved):

PR #3557, opened 2025-05-15
wip: new(userspace): static fields

### LINKED ISSUE #3554: Static event fields for falco daemonsets
**Motivation**

Falco has support for [events fields that you can use in conditions and outputs of Falco Rules](https://falco.org/docs/reference/rules/supported-fields/), those fields are related to different aspects of the runtime: evt,  process, user, group, container, k8s, etc. 

Falco instrumentation, like every other instrumentation, should allow the user to pass event fields that are static and not tied to the runtime and more importantly, fields that are only know at deployment time. A good set of examples are:

- https://github.com/openzipkin/zipkin-go/blob/master/middleware/http/client.go#L35
- https://github.com/open-telemetry/opentelemetry-go/discussions/2211
- https://istio.io/latest/docs/tasks/observability/distributed-tracing/telemetry-api/#customizing-tracing-tags
- https://opentelemetry.io/docs/languages/go/instrumentation/#initialize-metrics (search for resource.NewWithAttributes)

Notice all this attributes, tags or fields are static, known on beforehand and not resolvable in runtime. How/where do you know about this key here because falco rules can be related to this fields that are known statically providing a finer grained rules combining deploy time and runtime fields.

**Feature**

Falco deployments to support fields **that can be declared in charts** and can be used rules and also shown in output fields.

If you have a platform build on top of K8s and you need to give more granularity to rules, e.g. `platform_cluster.name = CI` to relax conditions in nested rules.

**Alternatives**

- Output fields can not be used in rules.
- Using tags for rules isn't ideal because it would imply duplication of rules in some cases where the rule is complex and has nested `if`s. Having to split the rule in two rules for a tag is a burden in maintenance.

**Additional context**

FedeDP:
**What type of PR is this?**

/kind feature

/area engine

The PR adds a new `static_fields` configuration object; each `{key,value}` pair will be used as static field that will be automatically appended to the output (if `append_output.suggested_output` is enabled), and can be used as rule condition.

Fixes #3554 

This PR depends upon https://github.com/falcosecurity/libs/pull/2405

```release-note
new(userspace): static fields
```


FedeDP:
Relevant PRs have been merged in libs master; once https://github.com/falcosecurity/falco/pull/3592 is merged i will rebase this one and get it ready.

userspace/falco/app/actions/init_inspectors.cpp
```diff
@@ -82,6 +84,10 @@ static bool populate_filterchecks(const std::shared_ptr<sinsp>& inspector,
 		filterchecks.add_filter_check(sinsp_plugin::new_filtercheck(plugin));
 		used_plugins.insert(plugin->name());
 	}
+
```
jasondellaluce, line 87: Looking at the check above, we should make `static_fields` part of the checks too otherwise we risk adding static fields that are overlapping with builtin ones, which would make Falco behave in a completely unpredictable way during rules matching potentially.

FedeDP: The static fields are enforced the `static.` prefix in libs!

FedeDP: Ok through a private chat i got it. Fixed in latest commit.

jasondellaluce (approved):

jasondellaluce (approved):

leogr (approved):

PR #3562, opened 2025-05-18
update(metrics): improve restart/hot_reload conditions inspection

incertum:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind release

> /area build

/area engine

> /area tests

> /area proposals

> /area CI

Fixes #

Proposing the following minor improvements:
- Introduce an immediate initial metrics msg (output_rule or output_file) upon start/restart/hot_reload -> this may be useful to directly inspect deployment convergence and eliminates the need to wait until the first two ticks happened
- Introduce a restart ts metric for all metrics output channels (including prometheus) -> this may be useful to statistically inspect potential issues wrt falco crashing and restarting and/or whether the hot config reload was successful and similar conditions

```release-note
update(metrics): improve restart/hot_reload conditions inspection
```

### DIFF
--- a/userspace/falco/app/actions/load_config.cpp
+++ b/userspace/falco/app/actions/load_config.cpp
@@ -66,6 +66,10 @@ falco::app::run_result falco::app::actions::load_config(const falco::app::state&
 		}
 	}
 
+	s.config->m_falco_restart_ts = (int64_t)std::chrono::duration_cast<std::chrono::nanoseconds>(
+	                                       std::chrono::system_clock::now().time_since_epoch())
+	                                       .count();
+
 	s.config->m_buffered_outputs = !s.options.unbuffered_outputs;
 
 	return apply_deprecated_options(s);
--- a/userspace/falco/app/actions/load_config.cpp
+++ b/userspace/falco/app/actions/load_config.cpp
@@ -66,9 +66,9 @@ falco::app::run_result falco::app::actions::load_config(const falco::app::state&
 		}
 	}
 
-	s.config->m_falco_restart_ts = (int64_t)std::chrono::duration_cast<std::chrono::nanoseconds>(
-	                                       std::chrono::system_clock::now().time_since_epoch())
-	                                       .count();
+	s.config->m_falco_reload_ts = (int64_t)std::chrono::duration_cast<std::chrono::nanoseconds>(
+	                                      std::chrono::system_clock::now().time_since_epoch())
+	                                      .count();
 
 	s.config->m_buffered_outputs = !s.options.unbuffered_outputs;
 
--- a/userspace/falco/configuration.h
+++ b/userspace/falco/configuration.h
@@ -161,7 +161,7 @@ public:
 
 	bool m_watch_config_files;
 	bool m_buffered_outputs;
-	int64_t m_falco_restart_ts;
+	int64_t m_falco_reload_ts;
 	size_t m_outputs_queue_capacity;
 	bool m_time_format_iso_8601;
 	bool m_buffer_format_base64;
--- a/userspace/falco/configuration.h
+++ b/userspace/falco/configuration.h
@@ -161,7 +161,6 @@ public:
 
 	bool m_watch_config_files;
 	bool m_buffered_outputs;
-	int64_t m_falco_reload_ts;
 	size_t m_outputs_queue_capacity;
 	bool m_time_format_iso_8601;
 	bool m_buffer_format_base64;
@@ -215,7 +214,15 @@ public:
 	gvisor_config m_gvisor = {};
 
 	yaml_helper m_config;
+
+	//
+	// Runtime-Generated values (not user-configurable)
+	//
+
+	// JSON schema generated from a hardcoded string
 	nlohmann::json m_config_schema;
+	// Timestamp of most recent configuration reload
+	int64_t m_falco_reload_ts;
 
 private:
 	void merge_config_files(const std::string& config_name, config_loaded_res& res);
--- a/userspace/falco/configuration.h
+++ b/userspace/falco/configuration.h
@@ -222,7 +222,7 @@ public:
 	// JSON schema generated from a hardcoded string
 	nlohmann::json m_config_schema;
 	// Timestamp of most recent configuration reload
-	int64_t m_falco_reload_ts;
+	int64_t m_falco_reload_ts{0};
 
 private:
 	void merge_config_files(const std::string& config_name, config_loaded_res& res);
--- a/userspace/falco/configuration.h
+++ b/userspace/falco/configuration.h
@@ -161,6 +161,7 @@ public:
 
 	bool m_watch_config_files;
 	bool m_buffered_outputs;
+	int64_t m_falco_restart_ts;
 	size_t m_outputs_queue_capacity;
 	bool m_time_format_iso_8601;
 	bool m_buffer_format_base64;
--- a/userspace/falco/falco_metrics.cpp
+++ b/userspace/falco/falco_metrics.cpp
@@ -144,12 +144,12 @@ std::string falco_metrics::to_text(const falco::app::state& state) {
 		std::vector<metrics_v2> additional_wrapper_metrics;
 
 		additional_wrapper_metrics.emplace_back(libs::metrics::libsinsp_metrics::new_metric(
-		        "restart_ts",
+		        "reload_ts",
 		        METRICS_V2_MISC,
 		        METRIC_VALUE_TYPE_S64,
 		        METRIC_VALUE_UNIT_TIME_TIMESTAMP_NS,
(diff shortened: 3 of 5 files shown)


userspace/falco/stats_writer.cpp
```diff
@@ -246,7 +246,7 @@ void stats_writer::worker() noexcept {
 
 		// this helps waiting for the first tick
 		tick = stats_writer::get_ticker();
-		if(first_tick != tick) {
+		if(m_first_run || first_tick != tick) {
```
incertum, line 249: Can't be `static` since we have to account for the hot reload conditions.

userspace/falco/falco_metrics.cpp
```diff
@@ -143,6 +143,14 @@ std::string falco_metrics::to_text(const falco::app::state& state) {
 		}
 		std::vector<metrics_v2> additional_wrapper_metrics;
 
+		additional_wrapper_metrics.emplace_back(libs::metrics::libsinsp_metrics::new_metric(
```
incertum, line 146: Example:

```
# HELP falcosecurity_falco_restart_timestamp_nanoseconds https://falco.org/docs/metrics/
# TYPE falcosecurity_falco_restart_timestamp_nanoseconds gauge
falcosecurity_falco_restart_timestamp_nanoseconds 1747587061694208044
```

incertum:
@FedeDP @sgaist and @leogr proposing

/milestone 0.41.0

if it's alright.

userspace/falco/falco_metrics.cpp
incertum: Actually, perhaps we should name it `reload` instead of restart since technically the falco process is not restarting in case of a config hot reload ... WDYT?

FedeDP: Yep i agree, i'd call it `reload`.

FedeDP:
Yes 0.41.0 is alrights! Thanks for the PR @incertum !
This one LGTM, left a comment about `restart`->`reload` thing.

userspace/falco/falco_metrics.cpp
incertum: Thank you, and just pushed the changes!

userspace/falco/stats_writer.cpp
FedeDP: I am wondering whether it would be easier to instead declare eg:
```
auto first_tick = -1;
```
so that
```
tick = stats_writer::get_ticker();
if(first_tick != tick) {
```
is always true on first iteration.
This way, we could avoid the `m_first_run` bool entirely. I double checked, i don't see `first_tick` used elsewhere, i think it was meant for the purpose of skipping the very first tick.
Perhaps we can get rid of `first_tick` entirely too.

userspace/falco/stats_writer.cpp
```diff
@@ -659,7 +661,10 @@ void stats_writer::collector::collect(const std::shared_ptr<sinsp>& inspector,
 			                                  num_evts,
 			                                  now,
 			                                  stats_snapshot_time_delta_sec);
-			get_metrics_output_fields_additional(output_fields, stats_snapshot_time_delta_sec);
+			if(!m_first_run) {
+				get_metrics_output_fields_additional(output_fields, stats_snapshot_time_delta_sec);
+			}
+			m_first_run = false;
```
FedeDP, line 667: Oh didn't see we use `m_first_run` here too!

incertum: You prefer any changes?

userspace/falco/stats_writer.cpp
incertum: Overlooked this earlier today @FedeDP - I just re-pushed the co-authored cleanup commit. Yes I agree this is more clever and as far as I can tell it works just fine! Thanks for the suggestion!

userspace/falco/stats_writer.cpp
incertum: I overlooked the comment above earlier today. Regardless I believe not making the first log special is better, the code seems robust enough, tested everything once more.

FedeDP:
Thanks Melissa! No problem :D Left another suggestion then it's good to go in my opinion!

userspace/falco/configuration.h
```diff
@@ -161,6 +161,7 @@ class falco_configuration {
 
 	bool m_watch_config_files;
 	bool m_buffered_outputs;
+	int64_t m_falco_reload_ts;
```
sgaist, line 164: There something that feels a bit odd here. The reload timestamp is a generated value rather than a configuration per se.

I would rather put it in its own section in this file with a comment. That will make things clearer/cleaner.

sgaist (changes requested):

userspace/falco/stats_writer.h
```diff
@@ -82,7 +82,9 @@ class stats_writer {
 		                                          double stats_snapshot_time_delta_sec);
 
 		std::shared_ptr<stats_writer> m_writer;
-		stats_writer::ticker_t m_last_tick = 0;
+		// Init m_last_tick w/ invalid value to enable metrics logging immediately after
+		// startup/reload
+		stats_writer::ticker_t m_last_tick = -1;
```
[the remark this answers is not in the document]
incertum: I'll check, and I am not sure why the ticks where checked in two places in the initial architecture etc ...

userspace/falco/configuration.h
incertum: Thanks Sam, I'll take a look!

userspace/falco/stats_writer.h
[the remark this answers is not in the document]
incertum: @FedeDP it is needed, this ticker appears to be the one responsible to wait for the specified time period between snapshots. Removing it results in endless metrics logs. Perhaps a future architecture change could be useful as it is indeed confusing how everything ticker related is setup.

userspace/falco/configuration.h
incertum: @sgaist pushed a suggestion, let me know if you would like a different approach. Thank you!

userspace/falco/app/actions/load_config.cpp
```diff
@@ -66,6 +66,10 @@ falco::app::run_result falco::app::actions::load_config(const falco::app::state&
 		}
 	}
 
+	s.config->m_falco_reload_ts = (int64_t)std::chrono::duration_cast<std::chrono::nanoseconds>(
```
FedeDP, line 69: What if we set this value in the configuration object constructor? After all it is bound to the lifetime of the config object, right?
It should work fine for our use case i think.

FedeDP:
My main suggestion is still to drop
> IMHO i'd just get rid of first_tick altogether since it was added just to prevent the very first tick.

Then we can proceed :) Thanks for your patience Melissa!

userspace/falco/configuration.h
sgaist: Looks way better ! Thanks

incertum:
> My main suggestion is still to drop
> 
> > IMHO i'd just get rid of first_tick altogether since it was added just to prevent the very first tick.
> 
> Then we can proceed :) Thanks for your patience Melissa!

I had looked at the wrong tick comparison and was confused about your suggestion there. Now looked at the right file and re-pushed last 2 commits.

userspace/falco/app/actions/load_config.cpp
incertum: Since it's not user configurable I would leave it w/ the comment for now and perhaps we can re-evaluate the entire metrics approach for the next release? In Falco it is still a bit rough around the edges, I believe in libs we have cleaned it up a lot and it's in a better state there.

FedeDP (approved):
/approve

FedeDP:
@sgaist PTAL :) 

@leogr it seems like https://github.com/falcosecurity/testing/pull/48 broke tests on this PR :/ 
I am gonna fix them; `--gvisor-generate-config` is only available on amd64 (https://github.com/falcosecurity/libs/blob/d1f550a596b2f04db7b5853bb92c68baeeae8cb1/cmake/modules/engine_config.cmake#L26)

sgaist (commented):
I think the points brought by copilot are worth fixing.

Otherwise, LGTM

userspace/falco/configuration.h
```diff
@@ -214,7 +214,15 @@ class falco_configuration {
...
 	yaml_helper m_config;
+
+	//
+	// Runtime-Generated values (not user-configurable)
+	//
+
+	// JSON schema generated from a hardcoded string
 	nlohmann::json m_config_schema;
+	// Timestamp of most recent configuration reload
+	int64_t m_falco_reload_ts;
```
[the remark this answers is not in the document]
incertum: Initializing non-const variable in the header file isn't recommended.

leogr (approved):

Issue #3639, opened 2025-07-09
inotify handler created when no watches are enabled


Proximyst:
**Describe the bug**

When disabling the config watching feature, the inotify handler is still created, and will crash if it cannot be made.

**How to reproduce it**

Set your inotify limit to 0 or use all handlers, then start Falco. It will fail with:

```
Error: could not initialize inotify handler
```

**Expected behaviour**

Either to let us disable the behaviour, or to not fail when there is nothing to watch.

**Environment**

- Falco version: 0.41.3

- System info: `2025-07-09T13:48:43+0000: System info: Linux version 6.6.87+ (builder@0fae150a673b) (Chromium OS 17.0_pre498229-r33 clang version 17.0.0 (/var/cache/chromeos-cache/distfiles/egit-src/external/github.com/llvm/llvm-project 14f0776550b5a49e1c42f49a00213f7f3fa047bf), LLD 17.0.0) #1 SMP PREEMPT_DYNAMIC Tue Jun 10 08:55:45 UTC 2025`

- Cloud provider or hardware configuration: AWS EKS, GCP GKE, Azure AKS. All x86_64 machines.
- OS: Uses the Helm chart, so Wolfi. Hosts use various Linux distros, all glibc-based.

- Kernel: e.g. 6.6.87

- Installation method: Kubernetes (Helm)

**Additional context**

PR #3640, opened 2025-07-09
fix(restart_handler): disable if there is no work

Proximyst:
When there is no work to do, i.e. when all config watching is disabled, there is no need to keep the restart_handler running. Disable it in this case.

This is helpful to do on nodes where there is little to no headroom in terms of open inotify watches (as per the inotify/max_user_instances configuration), as can happen on nodes populated with other software that also watch the filesystem for changes. If Falco is run on such a node, it may fail to start due to functionality the app does not even intend on using.

This has one change in terms of behaviour, however: the dry-run restarts will no longer occur. As there is still never going to happen a real restart, I understand it as unlikely for there to be a proper need for dry-run restarts.

Fixes: #3639

/kind bug
/area engine

**Does this PR introduce a user-facing change?**: YES

```release-note
Dry-run restarts are disabled if Falco runs with config-watching disabled.
```

### DIFF
--- a/userspace/falco/app/restart_handler.cpp
+++ b/userspace/falco/app/restart_handler.cpp
@@ -38,7 +38,9 @@ limitations under the License.
 
 falco::app::restart_handler::~restart_handler() {
 	stop();
-	close(m_inotify_fd);
+	if(m_inotify_fd != -1) {
+		close(m_inotify_fd);
+	}
 	m_inotify_fd = -1;
 }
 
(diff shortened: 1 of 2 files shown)


FedeDP:
Ehy @Proximyst thanks for the PR! It LGTM; i am trying to understand why the CI is complaining about the engine checksum and version (you did not touch anything related!).

leogr:
> i am trying to understand why the CI is complaining about the engine checksum and version (you did not touch anything related!).

@FedeDP, the CI complaint should be unrelated. A rebase may fix the issue. Anyway, we can merge the PR and check it later, eventually.

FedeDP (approved):
For future reference, we don't need to touch `falco::app::restart_handler::stop()` because https://en.cppreference.com/w/cpp/thread/thread/joinable.html:
> Specifically, returns true if get_id() != [std::thread::id](https://en.cppreference.com/w/cpp/thread/thread/id.html)(). So a default constructed thread is not joinable.

/approve

Issue #3664, opened 2025-08-22
Memory leak with short-lived containers


leogr:
**Describe the bug**

Falco appears to have a memory leak when running in an environment where many short-lived containers are being created. This might (or not) be related to some memory consumption issue recently reported (including https://github.com/falcosecurity/falco/issues/3643 and the very last comments of https://github.com/falcosecurity/falco/issues/2495).

I've reproduced the problem only with Falco 0.41.3 (not sure about previous versions), with and without the container plugin.
Unfortunately, the problem is present in both cases, so the root cause shouldn't be within the container plugin.

**How to reproduce it**

Create `/tmp/falco_mem.sh` as follow:
```bash
#!/bin/bash

ps -C falco -o pid=,rss= | awk '{rss+=$2} END {printf "Memory: %.2f MB\n", rss/1024}'
docker run -it --rm ubuntu echo ciao
```
Then `chmod +x` it.

Then run `falco` and `watch /tmp/falco_mem.sh` in parallel. Wait a bit and you can see the memory growing up. To accelerate the effect, you can run multiple `watch /tmp/falco_mem.sh` in parallel.

**By running 12 instances of the script above, Falco reaches 2GB of memory consumption in ~15 mins on my environment.**

**Environment**

- Falco version:
```
❯ falco --version
Fri Aug 22 14:19:40 2025: Falco version: 0.41.3 (x86_64)
Fri Aug 22 14:19:40 2025: Falco initialized with configuration files:
Fri Aug 22 14:19:40 2025:    /etc/falco/config.d/falco.container_plugin.yaml | schema validation: ok
Fri Aug 22 14:19:40 2025:    /etc/falco/falco.yaml | schema validation: ok
Fri Aug 22 14:19:40 2025: System info: Linux version 6.16.1-arch1-1 (linux@archlinux) (gcc (GCC) 15.2.1 20250813, GNU ld (GNU Binutils) 2.45.0) #1 SMP PREEMPT_DYNAMIC Fri, 15 Aug 2025 16:04:43 +0000
Falco version: 0.41.3
Libs version:  0.21.0
Plugin API:    3.11.0
Engine:        0.50.0
Driver:
  API version:    8.0.0
  Schema version: 3.6.0
  Default driver: 8.1.0+driver
```
- OS: Arch Linux

[...]

PR #3666, opened 2025-08-31
refactor: run the webserver unprivileged

### LINKED ISSUE #3485: No longer run webserver as root under main falco process
**Motivation**

We want to collect prometheus metrics from all servers.

**Feature**

I noticed the webserver apparently runs as root, under the same process that Falco binary does:

Netstat:

``` tcp        0      0 0.0.0.0:8765            0.0.0.0:*               LISTEN      0          1    426859/falco```

PS:

```root      426859  0.0  0.7 584040 27648 ?        Ssl  Jan20  20:19 /usr/bin/falco -o engine.kind=modern_ebpf```

Would it be possible to split this responsibility, having a separate webserver process with less privileges than the main Falco process?

**Alternatives**

Not that im aware of.

**Additional context**

Separation of concern, least privilege, and other best practices

### PR DESCRIPTION
**Pull request by sgaist**
**What type of PR is this?**

> /kind bug

/kind cleanup

/kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind release

> /area build

> /area engine

> /area tests

> /area proposals

> /area CI

The webserver is running in a thread with the same privileges as the falco process. In order to make running it more secure, use fork rather than thread and then change the user and group id to unprivileged values.

Fixes #3485 

```release-note
webserver: the webserver component now runs in an unprivileged fork. The user id and group id used to run the fork can be configured through the webserver configuration block.
```

### DIFF
--- a/userspace/falco/webserver.cpp
+++ b/userspace/falco/webserver.cpp
@@ -21,6 +21,7 @@ limitations under the License.
 #include "app/state.h"
 #include "versions_info.h"
 #include <atomic>
+#include <signal.h>
 
 falco_webserver::~falco_webserver() {
 	stop();
@@ -58,47 +59,62 @@ void falco_webserver::start(const falco::app::state &state,
 		              res.set_content(versions_json_str, "application/json");
 	              });
 
-	// run server in a separate thread
 	if(!m_server->is_valid()) {
 		m_server = nullptr;
 		throw falco_exception("invalid webserver configuration");
 	}
 
 	m_failed.store(false, std::memory_order_release);
-	m_server_thread = std::thread([this, webserver_config] {
+
+	// fork the server
+	m_pid = fork();
+
+	if(m_pid == 0) {
+		falco_logger::log(falco_logger::level::INFO, "Webserver: forked\n");
+		int res = setgid(webserver_config.m_uid);
+		if(res != NOERROR) {
+			throw falco_exception("Webserver: an error occurred while setting group id: " +
+			                      std::to_string(errno));
+		}
+		res = setuid(webserver_config.m_gid);
+		if(res != NOERROR) {
+			throw falco_exception("Webserver: an error occurred while setting user id: " +
+			                      std::to_string(errno));
+		}
+		falco_logger::log(falco_logger::level::INFO,
+		                  "Webserver: fork running as " + std::to_string(webserver_config.m_uid) +
+		                          ":" + std::to_string(webserver_config.m_gid) + "\n");
 		try {
 			this->m_server->listen(webserver_config.m_listen_address,
 			                       webserver_config.m_listen_port);
 		} catch(std::exception &e) {
 			falco_logger::log(falco_logger::level::ERR,
-			                  "falco_webserver: " + std::string(e.what()) + "\n");
+			                  "Webserver: " + std::string(e.what()) + "\n");
+			m_failed.store(true, std::memory_order_release);
 		}
-		this->m_failed.store(true, std::memory_order_release);
-	});
-
-	// wait for the server to actually start up
-	// note: is_running() is atomic
-	while(!m_server->is_running() && !m_failed.load(std::memory_order_acquire)) {
-		std::this_thread::yield();
-	}
-	m_running = true;
-	if(m_failed.load(std::memory_order_acquire)) {
-		stop();
-		throw falco_exception("an error occurred while starting webserver");
+	} else if(m_pid < 0) {
+		throw falco_exception("Webserver: an error occurred while forking webserver");
 	}
 }
 
 void falco_webserver::stop() {
-	if(m_running) {
+	if(m_pid > 0) {
+		falco_logger::log(falco_logger::level::INFO, "Webserver: stopping server\n");
 		if(m_server != nullptr) {
 			m_server->stop();
 		}
-		if(m_server_thread.joinable()) {
-			m_server_thread.join();
+		falco_logger::log(falco_logger::level::INFO, "Webserver: killing fork\n");
+		int res = kill(m_pid, SIGTERM);
+		if(res != 0) {
+			throw falco_exception("Webserver: an error occurred while killing fork: " +
+			                      std::to_string(errno));
 		}
-		m_server = nullptr;
-		m_running = false;
+		m_pid = 0;
+		falco_logger::log(falco_logger::level::INFO, "Webserver: stopping fork done\n");
 	}
+
+	m_server = nullptr;
+	m_running = false;
 }
 
 void falco_webserver::enable_prometheus_metrics(const falco::app::state &state) {
--- a/userspace/falco/webserver.cpp
+++ b/userspace/falco/webserver.cpp
@@ -75,13 +75,15 @@ void falco_webserver::start(const falco::app::state &state,
 		falco_logger::log(falco_logger::level::INFO, "Webserver: forked\n");
 		int res = setgid(webserver_config.m_uid);
 		if(res != NOERROR) {
-			throw falco_exception("Webserver: an error occurred while setting group id: " +
-			                      std::to_string(errno));
+			throw falco_exception(
+			        std::string("Webserver: an error occurred while setting group id: ") +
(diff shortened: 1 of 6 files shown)


sgaist:
Note for reviewers: it's the first time I am taking a shot at fork so I might have implemented things in a wrong fashion.
I will welcome hints to documentation/best practice for it.

leogr:
Hey @sgaist 

I'm unsure if we have enough time to put this into the upcoming release, but it seems really interesting!

So tentatively for
/milestone 0.42.0

Also, I have a side question. While investigating https://github.com/falcosecurity/falco/issues/3664, I had the feeling that when the Prometheus metrics and web server are enabled, the memory consumption slightly increases by simply repeatedly refreshing the metrics web page. However, I can't reliably repro this. So, it would be great if, along with this initiative, we can also investigate more about possible issues with the web server (and the dependent metrics) feature(s).

leogr:
Since we are likely too late for the upcoming release, moving to 
/milestone 0.43.0

userspace/falco/webserver.cpp
```diff
@@ -58,47 +59,62 @@ void falco_webserver::start(const falco::app::state &state,
...
+
+	// fork the server
+	m_pid = fork();
+
+	if(m_pid == 0) {
+		falco_logger::log(falco_logger::level::INFO, "Webserver: forked\n");
+		int res = setgid(webserver_config.m_uid);
+		if(res != NOERROR) {
+			throw falco_exception("Webserver: an error occurred while setting group id: " +
+			                      std::to_string(errno));
```
ekoops: Could we use `std::strerror()` here and in the other occurrences?

userspace/falco/webserver.cpp
```diff
@@ -58,47 +59,62 @@ void falco_webserver::start(const falco::app::state &state,
...
+		                  "Webserver: fork running as " + std::to_string(webserver_config.m_uid) +
+		                          ":" + std::to_string(webserver_config.m_gid) + "\n");
 		try {
 			this->m_server->listen(webserver_config.m_listen_address,
 			                       webserver_config.m_listen_port);
 		} catch(std::exception &e) {
 			falco_logger::log(falco_logger::level::ERR,
-			                  "falco_webserver: " + std::string(e.what()) + "\n");
+			                  "Webserver: " + std::string(e.what()) + "\n");
+			m_failed.store(true, std::memory_order_release);
```
ekoops: I guess this is not gonna work... The child process updates its own `m_failed` copy. In order to establish a communication between the two processes, we should use an IPC mechanism, like a pipe. In this way, the child process, could communicate to falco that it correctly bootstrapped the webserver or that it failed while attempting.

ekoops:
Hey @sgaist , any update on this?

sgaist:
@ekoops sorry, I just realized that your review got lost in my mail box. I'll check that ASAP.

leogr:
@sgaist any update on this?

sgaist:
I have done some exploration on the IPC side with pipes however I realized that there's no way to send back the information that the webserver was started through this mechanism as by design it enters a blocking loop and there's no callback that could be used to state "started".
Next thing I am thinking about is to simply start a small loop from the parent that tries to talk to the webserver to know that it is started.
What do you guys think ?

userspace/falco/webserver.cpp
sgaist: You're correct, I had forgotten about that rule.

I have implemented the check using `cpp-httplib`'s client.

sgaist:
> Also, I have a side question. While investigating #3664, I had the feeling that when the Prometheus metrics and web server are enabled, the memory consumption slightly increases by simply repeatedly refreshing the metrics web page. However, I can't reliably repro this. So, it would be great if, along with this initiative, we can also investigate more about possible issues with the web server (and the dependent metrics) feature(s).

I am wondering whether this is the result of calling [the metrics collector snapshot function](https://github.com/falcosecurity/libs/blob/da834817e8f36a6bc864260289c91166d2fdf47b/userspace/libsinsp/metrics_collector.cpp#L383).

userspace/falco/webserver.cpp
```diff
@@ -58,47 +61,85 @@ void falco_webserver::start(const falco::app::state &state,
...
+	m_pid = fork();
+
+	if(m_pid < 0) {
+		throw falco_exception("Webserver: an error occurred while forking webserver");
+	} else if(m_pid == 0) {
+		falco_logger::log(falco_logger::level::INFO, "Webserver: forked\n");
+		int res = setgid(webserver_config.m_uid);
+		if(res != NOERROR) {
+			throw falco_exception(
+			        std::string("Webserver: an error occurred while setting group id: ") +
```
ekoops: Here we are calling setgid but we are providing m_uid. If the keep setgid here, then the exception message is ok; otherwise we also need to change it

userspace/falco/webserver.cpp
```diff
@@ -58,47 +61,85 @@ void falco_webserver::start(const falco::app::state &state,
...
 void falco_webserver::stop() {
-	if(m_running) {
+	if(m_pid > 0) {
+		falco_logger::log(falco_logger::level::INFO, "Webserver: stopping server\n");
 		if(m_server != nullptr) {
 			m_server->stop();
 		}
-		if(m_server_thread.joinable()) {
-			m_server_thread.join();
+		falco_logger::log(falco_logger::level::INFO, "Webserver: killing fork\n");
```
ekoops: What about using "webserver process" instead of "fork"? I mean, both here and in the other places... WDYT?

ekoops (changes requested):

userspace/falco/webserver.cpp
```diff
@@ -58,47 +61,85 @@ void falco_webserver::start(const falco::app::state &state,
...
+			                  "Webserver: " + std::string(e.what()) + "\n");
 		}
-		this->m_failed.store(true, std::memory_order_release);
-	});
+	} else {
+		std::string schema = "http";
+		if(webserver_config.m_ssl_enabled) {
+			schema = "https";
+		}
+		std::string url = schema + "://localhost:" + std::to_string(webserver_config.m_listen_port);
```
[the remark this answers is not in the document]
sgaist: Good point

Done

userspace/falco/webserver.cpp
```diff
@@ -58,47 +61,88 @@ void falco_webserver::start(const falco::app::state &state,
...
+		                  "Webserver: process running as " +
+		                          std::to_string(webserver_config.m_uid) + ":" +
+		                          std::to_string(webserver_config.m_gid) + "\n");
 		try {
 			this->m_server->listen(webserver_config.m_listen_address,
 			                       webserver_config.m_listen_port);
 		} catch(std::exception &e) {
 			falco_logger::log(falco_logger::level::ERR,
-			                  "falco_webserver: " + std::string(e.what()) + "\n");
+			                  "Webserver: " + std::string(e.what()) + "\n");
```
ekoops: My concern is what happens if the `m_server->listen(...)` call fail. Here, we are catching the exception, and after we print a log line, we just return from the function. Even if the main falco process will recognize that the son is not running, I guess we should avoid letting the web server process to proceed.

sgaist: Currently, if the server does not start for some reason, it will be detected by the loop below that tries to connect to the server. However, it's not a nice way to detect that.
I have refactored that part.

ekoops:
Hey @sgaist , I guess we broke prometheus metrics: https://github.com/falcosecurity/falco/actions/runs/21926985793?pr=3666

sgaist:
Oh damn, it seems that my answer got lost in the process.

I dug a bit on that part only to realize that the metrics end point activation will happen in the parent process hence the breakage you spotted.

sgaist:
I was able to come back to this and I think it needs a bit more design brainstorming (I was I bit tired when starting this and there are several details that escaped me at that time).

I think that the webserver, as a forked process, should not have anything to do with directly accessing the `state` object to retrieve the stats. We should rather have a proper IPC in place that decouples both.

What do you think ?

PR #3676, opened 2025-09-24
fix(userspace/falco): fix actions taken when events are dropped

terror96:
User can configure a list of actions that are taken when Falco detects a threshold exceeding value in drop statistics.

However, the logic that handles the list of configured actions is designed to process only a single action; it takes only the first action of the list. This approach has the problem that the order of the actions comes as the deciding factor in choosing which action is taken in case there are more than one action.

This fix enables Falco to process all actions on the list.

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind release

> /area build

/area engine

> /area tests

> /area proposals

> /area CI

Configuration as shown in the template at https://github.com/falcosecurity/falco/blob/7fb9986e5ad486a12af33e259da0ab1659724824/falco.yaml#L1092C1-L1099 indicates that it is possible to configure a list of actions. However, the current implementation for handling the action list in https://github.com/falcosecurity/falco/blob/7fb9986e5ad486a12af33e259da0ab1659724824/userspace/falco/event_drops.cpp#L135 processes only the first item on the list. I believe this to be a bug, and a better approach is to be process all actions on the list.

Fixes #

```release-note

```

### DIFF
--- a/userspace/falco/event_drops.cpp
+++ b/userspace/falco/event_drops.cpp
@@ -138,15 +138,16 @@ bool syscall_evt_drop_mgr::perform_actions(uint64_t now,
 	std::string rule = "Falco internal: syscall event drop";
 	std::string msg =
 	        rule + ". " + std::to_string(delta.n_drops) + " system calls dropped in last second.";
+	bool ret = true;
 
 	for(auto &act : m_actions) {
 		switch(act) {
 		case syscall_evt_drop_action::DISREGARD:
-			return true;
+			continue;
 
 		case syscall_evt_drop_action::LOG:
 			falco_logger::log(falco_logger::level::DEBUG, std::move(msg));
-			return true;
+			continue;
 
 		case syscall_evt_drop_action::ALERT: {
 			nlohmann::json output_fields;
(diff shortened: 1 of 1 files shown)


terror96:
One additional thing to consider before merging:

If support for multiple actions is enabled, the configuration template settings are a bit problematic in the sense that if both `log` and `alert` are listed as actions, and if `log_level` is also then changed to `debug`, the drop message will appear multiple times because both `log_stderr` and `log_syslog` are also set as `true` by default. This will lead soon to: `Fri Sep 26 07:28:23 2025: Syscall event drop but token bucket depleted, skipping action` like messages. However, maybe this is okay, because if the user willingly sets `log_level` to `debug` it is safe to assume that there could be a peak in outputted messages.

leogr:
> If support for multiple actions is enabled, the configuration template settings are a bit problematic in the sense that if both `log` and `alert` are listed as actions, and if `log_level` is also then changed to `debug`, the drop message will appear multiple times because both `log_stderr` and `log_syslog` are also set as `true` by default.

IMO, this is legit and expected since `alert`s go through all enabled output channels


> This will lead soon to: `Fri Sep 26 07:28:23 2025: Syscall event drop but token bucket depleted, skipping action` like messages. However, maybe this is okay, because if the user willingly sets `log_level` to `debug` it is safe to assume that there could be a peak in outputted messages.

This is not nice, but acceptable for now, especially if `log_level` is set to `debug`

leogr (approved):

PR #3741, opened 2025-11-17
fix(metrics): Prevent race condition crash during metrics collection on shutdown

### LINKED ISSUE #3739: Segmentation Fault when accessing outputs_queue_num_drops prometheus metrics
**Describe the bug**
Falco crashes when accessing metrics related to outputs_queue_num_drops.

**How to reproduce it**

Enable   state_counters_enabled: true in falco.yaml and shutdown falco and scrape metrics post.

**Expected behaviour**
No crash when accessing metrics related to outputs_queue_num_drops

Coredump

`#0  0x000055a08d682d40 in falco_outputs::get_outputs_queue_num_drops() ()
[Current thread is 1 (LWP 430540)]
(gdb) bt
#0  0x000055a08d682d40 in falco_outputs::get_outputs_queue_num_drops() ()
#1  0x000055a08d6f2d4c in falco_metrics::falco_to_text_prometheus[abi:cxx11](falco::app::state const&, libs::metrics::prometheus_metrics_converter&, std::vector<metrics_v2, std::allocator<metrics_v2> >&) ()
#2  0x000055a08d6f5343 in falco_metrics::to_text_prometheus[abi:cxx11](falco::app::state const&) ()
#3  0x000055a08d6a5cf6 in std::_Function_handler<void (httplib::Request const&, httplib::Response&), falco_webserver::enable_prometheus_metrics(falco::app::state const&)::{lambda(httplib::Request const&, httplib::Response&)#1}>::_M_invoke(std::_Any_data const&, httplib::Request const&, httplib::Response&) ()
#4  0x000055a08d546d82 in httplib::Server::routing(httplib::Request&, httplib::Response&, httplib::Stream&) ()
#5  0x000055a08d548179 in httplib::Server::process_request(httplib::Stream&, bool, bool&, std::function<void (httplib::Request&)> const&) ()
#6  0x000055a08d5491c4 in httplib::Server::process_and_close_socket(int) ()
#7  0x000055a08d6a9817 in std::thread::_State_impl<std::thread::_Invoker<std::tuple<httplib::ThreadPool::worker> > >::_M_run() ()
#8  0x00007f28bd2d44a3 in std::basic_filebuf<wchar_t, std::char_traits<wchar_t> >::~basic_filebuf (this=0x0, __in_chrg=<optimized out>)
    at /build/reproducible-path/gcc-14-14.2.0/build/x86_64-linux-gnu/libstdc++-v3/include/fstream:262
#9  0x00007f288e7c46c0 in ?? ()
#10 0x0000000000000000 in ?? ()`

**Additional context**

This is a similar issue as Issue #3737  but with a different crash location. 

The issue arises due to a race condition when the Prometheus /metrics endpoint is accessed during shutdown.
While I will be submitting a comprehensive fix for #3739 that will resolve this issue as well, a simple nullptr check should be added as a defensive check nonetheless.

adduali1310:
This fixes a segmentation fault that occurs when /metrics endpoint is accessed during Falco shutdown. The crash happens as there is a very minute window, wherein the webserver continues serving /metrics requests after outputs have been destroyed.

The race condition is timing-dependent and occurs in this window:
1. [process_events()](https://github.com/falcosecurity/falco/blob/master/userspace/falco/app/actions/process_events.cpp#L655)
2. Webserver still running and can receive `/metrics` requests
3. Metrics code accesses NULL outputs elements→ crash

Changes:

- Create cleanup_outputs action to handle outputs destruction
- Reorder teardown steps to stop webserver before destorying outputs
- Move outputs.reset() from process_events to cleanup_outputs()

This eliminates the race condition by ensuring the webserver stops accepting requests before any subsystems are destroyed. The synchronisation behaviour of output.reset() block till queue flushed is preserved.

**What type of PR is this?**

/kind bug

/area engine

Fixes a race condition that causes a segmentation fault during Falco shutdown when prometheus metrics endpoint is accessed. The crash occurs in a very short time window before the webserver is stopped but after outputs are destroyed.
Due to the destruction, accessing them leads to a SEGV and invalid access.
The fix reorders the shutdown sequence to stop the webserver before destroying the subsystems.

Fixes #3739 

Created a simple bash script to reproduce the issue.

```
#!/bin/bash

FALCO_BIN="Binary Path"
METRICS_URL="http://localhost:3018/metrics"

echo "Starting race condition test..."

$FALCO_BIN -c /etc/falco/falco.yaml &
FALCO_PID=$!

sleep 3

echo "Falco started (PID: $FALCO_PID), launching 20 parallel scrapers..."

for i in {1..20}; do
    while true; do
        curl -s $METRICS_URL > /dev/null 2>&1
    done &
done

sleep 2

echo "Sending SIGTERM to Falco..."
kill -SIGTERM $FALCO_PID

sleep 3

echo "Cleaning up scrapers..."

pkill -P $$ curl 2>/dev/null
killall curl 2>/dev/null


wait $FALCO_PID 2>/dev/null
EXIT_CODE=$?

echo "Falco exit code: $EXIT_CODE"

if [ $EXIT_CODE -eq 139 ]; then
    echo "CRASH DETECTED: Segmentation fault (exit code 139)"
    exit 1
elif [ $EXIT_CODE -eq 0 ]; then
    echo "SUCCESS: Clean shutdown"
    exit 0
else
    echo "UNEXPECTED: Exit code $EXIT_CODE"
    exit $EXIT_CODE
fi
```

**Before Fix:**

```
Test run 1
Starting race condition test...
Mon Nov 17 08:08:24 2025: Falco version: 0.42.1-2025.11.2 (x86_64)
Mon Nov 17 08:08:24 2025: Falco initialized with configuration files:
Mon Nov 17 08:08:24 2025:    /etc/falco/falco.yaml | schema validation: ok
Mon Nov 17 08:08:24 2025: Loading rules from:
Mon Nov 17 08:08:24 2025:    /etc/falco/falco_rules.local.yaml | schema validation: none
Mon Nov 17 08:08:24 2025: The chosen syscall buffer dimension is: 8388608 bytes (8 MBs)
Mon Nov 17 08:08:24 2025: Starting health webserver with threadiness 96, listening on 0.0.0.0:3018
Mon Nov 17 08:08:24 2025: Setting metrics interval to 1h, equivalent to 3600000 (ms)
Mon Nov 17 08:08:24 2025: Loaded event sources: syscall
Mon Nov 17 08:08:24 2025: Enabled event sources: syscall
Mon Nov 17 08:08:24 2025: Opening 'syscall' source with modern BPF probe.
Mon Nov 17 08:08:24 2025: One ring buffer every '2' CPUs.
Mon Nov 17 08:08:24 2025: [libs]: Trying to open the right engine!
Falco started (PID: 317455), launching 20 parallel scrapers...
Sending SIGTERM to Falco...
Cleaning up scrapers...
Terminated
Terminated
Terminated
Terminated
Terminated
Mon Nov 17 08:08:44 2025: SIGINT received, exiting...
   - num times actions taken: 0
Events detected: 0
Rule counts by severity:
Triggered rules by rule name:
Falco exit code: 139
CRASH DETECTED: Segmentation fault (exit code 139)
FAILED on run 1
```

**After Fix:**

```
Starting race condition test...
...
Mon Nov 17 08:06:07 2025: Falco initialized with configuration files:
Mon Nov 17 08:06:07 2025:    /etc/falco/falco.yaml | schema validation: ok
Mon Nov 17 08:06:07 2025: Loading rules from:
Mon Nov 17 08:06:07 2025:    /etc/falco/falco_rules.local.yaml | schema validation: none
Mon Nov 17 08:06:07 2025: The chosen syscall buffer dimension is: 8388608 bytes (8 MBs)
Mon Nov 17 08:06:07 2025: Starting health webserver with threadiness 96, listening on 0.0.0.0:3018
Mon Nov 17 08:06:07 2025: Setting metrics interval to 1h, equivalent to 3600000 (ms)
Mon Nov 17 08:06:07 2025: Loaded event sources: syscall
Mon Nov 17 08:06:07 2025: Enabled event sources: syscall
Mon Nov 17 08:06:07 2025: Opening 'syscall' source with modern BPF probe.
Mon Nov 17 08:06:07 2025: One ring buffer every '2' CPUs.
Mon Nov 17 08:06:07 2025: [libs]: Trying to open the right engine!
Falco started (PID: 153669), launching 20 parallel scrapers...
Sending SIGTERM to Falco...
Cleaning up scrapers...
Terminated
Terminated
Terminated
Terminated
Terminated
Terminated
Mon Nov 17 08:06:26 2025: SIGINT received, exiting...
   - num times actions taken: 0
Events detected: 0
Rule counts by severity:
Triggered rules by rule name:
Falco exit code: 0
SUCCESS: Clean shutdown
```

Post the fix in this PR, can confirm haven't seen the crashes. 
Can be confirmed by running the script aggressively as well.

```release-note
None
```

### DIFF
--- a/userspace/falco/CMakeLists.txt
+++ b/userspace/falco/CMakeLists.txt
@@ -48,6 +48,7 @@ add_library(
 	app/actions/start_webserver.cpp
 	app/actions/validate_rules_files.cpp
 	app/actions/close_inspectors.cpp
+	app/actions/cleanup_outputs.cpp
 	app/actions/print_config_schema.cpp
 	app/actions/print_rule_schema.cpp
 	configuration.cpp
--- a/userspace/falco/app/actions/actions.h
+++ b/userspace/falco/app/actions/actions.h
@@ -55,6 +55,7 @@ falco::app::run_result stop_webserver(falco::app::state& s);
 falco::app::run_result unregister_signal_handlers(falco::app::state& s);
 falco::app::run_result validate_rules_files(falco::app::state& s);
 falco::app::run_result close_inspectors(falco::app::state& s);
+falco::app::run_result cleanup_outputs(falco::app::state& s);
 
 };  // namespace actions
 };  // namespace app
--- a/userspace/falco/app/actions/cleanup_outputs.cpp
+++ b/userspace/falco/app/actions/cleanup_outputs.cpp
@@ -1,0 +1,29 @@
+// SPDX-License-Identifier: Apache-2.0
+/*
+Copyright (C) 2025 The Falco Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "actions.h"
+
+using namespace falco::app;
+using namespace falco::app::actions;
+
+falco::app::run_result falco::app::actions::cleanup_outputs(falco::app::state& s) {
+	if(s.outputs) {
+		s.outputs.reset();
+		s.engine->print_stats();
+	}
+	return run_result::ok();
+}
--- a/userspace/falco/app/actions/process_events.cpp
+++ b/userspace/falco/app/actions/process_events.cpp
@@ -646,13 +646,5 @@ falco::app::run_result falco::app::actions::process_events(falco::app::state& s)
 		}
 	}
 
-	// By deleting s.outputs, we make sure that the engine will wait until
-	// regular output has been completely sent before printing stats, avoiding
-	// intermixed stats with output.
-	// Note that this will only work if this is the last reference held by the
-	// shared pointer.
-	s.outputs.reset();
-	s.engine->print_stats();
-
 	return res;
 }
--- a/userspace/falco/app/app.cpp
+++ b/userspace/falco/app/app.cpp
@@ -87,6 +87,8 @@ bool falco::app::run(falco::app::state& s, bool& restart, std::string& errstr) {
 	std::list<app_action> const teardown_steps = {
 	        falco::app::actions::unregister_signal_handlers,
 	        falco::app::actions::stop_webserver,
+	        // Note: calls print_stats internally after resetting outputs.
+	        falco::app::actions::cleanup_outputs,
 	        falco::app::actions::close_inspectors,
 	};
 


leogr:
good catch! thank you

/milestone 0.43.0

ekoops:
Hey @adduali1310 , sorry, but we had experienced some delay due to rate-limiting issues. The change looks good, even if that `print_stats()` embedded in that action could be difficult to spot. Do you mind to move that print in a separate action?

adduali1310:
Apologies for the delay @ekoops . Separated print_stats into a separate action as requested and is also called separately in the teardown sequence.

adduali1310:
Hey @ekoops @leogr 

Checking in as to when do we plan to merge this change?

ekoops:
> Hey @ekoops @leogr
> 
> Checking in as to when do we plan to merge this change?

Hey! Sorry, we were busy with the new release. Now we can work for merging this. I see tests failure in CI, but I haven't had time to go through it. Can you rebase on top of master? I hope this fill fix the issue. If the CI become green, I'll definitely merge this (notice that the failing jobs are marked as "required").

adduali1310:
Hey @ekoops 

So took me a while to get to the bottom of things but quick update on the CI status after rebasing. I ran into a couple of issues.

- After moving print_stats() to a separate teardown action as requested, the check-engine-checksum and check-engine-version CI checks started failing as the teardown steps run on every exit path, including info commands like --list-events and --version. On master, print_stats() only ran at the end of process_events(), which is never reached for those commands. As a teardown action, it was unconditionally appending extra output to --list-events:

```
Events detected: 0
Rule counts by severity:
Triggered rules by rule name:
```
This changed the computed engine checksum. I built locally and confirmed by diffing falco --list-events output between master and my branch.
To fix this, I added an if(s.outputs) guard to both print_stats.cpp and cleanup_outputs.cpp, since s.outputs is only initialized in init_outputs which runs right before process_events and for info commands it remains null.

- While adding the guard did fix the checksum, but TestFalcoKmod and TestFalcoModernBpf are now failing with:

`Error: Expect "" to match "Events detected:"`

These tests run Falco with -M 3 and check stderr for the "Events detected:" string from print_stats(). The problem is a teardown ordering conflict:
1. cleanup_outputs runs first, calling s.outputs.reset() which sets s.outputs to null
2. print_stats runs next, but the if(s.outputs) guard sees null and skips it

Swapping the order would fix the test but breaks the original intent on master, s.outputs.reset() was intentionally called before print_stats() to flush the output queue and avoid intermixed output.

So s.outputs isn't a viable guard since it's always null by the time print_stats needs to run, and there's no clean flag in state to indicate "we actually processed events."

I think the simplest correct fix is to move print_stats() back into process_events() (at the end, before the return) and only keep cleanup_outputs as the separate teardown action. print_stats() itself doesn't have the race condition with the webserver — only s.outputs.reset() does. This preserves the original ordering (reset outputs, then print stats) and the teardown safety.

Unless you have a better or alternative guard in mind to keep print_stats as a separate teardown action let me know, how you would prefer to proceed.

ekoops:
Mmm I see. Yes please, let's bring back `print_stats()` inside the action. However, I would add a comment, where the action is used, that says that internally it calls `print_stats()`... WDYT?

adduali1310:
Hey @ekoops 

Made the requested change. Could you take a look when possible?
Also, all checks pass except test-dev-packages-arm64 / test-packages with the failure for TestFalcoctl_Artifact_InstallPlugin and is unrelated to this change.

I would presume rerunning this test should fix it.

leogr:
> I would presume rerunning this test should fix it.

Job re-launched. Let's see.

adduali1310:
Seems to have worked? @leogr

PR #3746, opened 2025-11-26
Reduce image size

jfcoz:
**What type of PR is this?**

/kind cleanup
/area build

**What this PR does / why we need it**:

`falcoctl` is added in a layer, then deleted (hidden) in another layer.

The `rm falcoctl` is now in the same image layer, this divides the image size by 2.

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #3747, opened 2025-11-26
Add common plugins

jfcoz:
**What type of PR is this?**

/kind cleanup
/kind feature

/area build

Variant of https://github.com/falcosecurity/falco/pull/3746 , but with more plugins by default:
- json
- k8smeta
- k8saudit

they are the most used based on packages download count, and not specific to any cloud provider : https://github.com/orgs/falcosecurity/packages?sort_by=downloads_desc

Each pod should be able to start faster and without reaching this quota problem : https://github.com/falcosecurity/rules/issues/331

Does not fixes totally, but should improve https://github.com/falcosecurity/rules/issues/331

Also this should allow to disable artifact-install and artifact-follow, and prevent supply chain attack.

And this reduce the image size from 212 to 136Mb.

I am quite new on this project, maybe it would be simpler to build theses plugins directly on the falco package which is installed in the Dockerfile.

```release-note
NONE
```

### DIFF
--- a/docker/falco/Dockerfile
+++ b/docker/falco/Dockerfile
@@ -35,6 +35,7 @@ RUN FALCO_VERSION_URLENCODED=$(echo -n ${FALCO_VERSION}|jq -sRr @uri) && \
     rm -rf /falco/usr/src/falco-* && \
     cp -r /falco/* / && \
     rm -rf /falco && \
+    for plugin in json k8smeta k8saudit; do falcoctl artifact install $plugin; done && \
     rm -rf /usr/bin/falcoctl /etc/falcoctl/
 
 


jfcoz:
@leogr , the move on the `rm -rf /usr/bin/falcoctl /etc/falcoctl/` only is in this other PR: https://github.com/falcosecurity/falco/pull/3746

I need to use k8smeta, but would like to not download it at each pod start, how should I do instead ?

Maybe there could be some variants images, with k8saudit ? for each cloud provider with specific plugins ?

leogr:
> @leogr , the move on the `rm -rf /usr/bin/falcoctl /etc/falcoctl/` only is in this other PR: #3746

:+1: I had lost track of it. Now, I've approved it. Thanks!

> I need to use k8smeta, but would like to not download it at each pod start, how should I do instead ?

You may want to create a custom container image that includes k8smeta. It should be easy to do; the only caveat is that you need to publish it in your own container registry. Note that you can just use ghcr.io for free and it's easy to configure. 

> Maybe there could be some variants images, with k8saudit ? for each cloud provider with specific plugins ?

Ideally yes. At a very early stage, when we introduced the plugin system, that was the initial plan. Unfortunately, as the number of plugins grew and the need to ship different kinds of Falco images for other purposes increased, it would have been hard to deal with all these image variants. So we eventually decided not to do that.

So, at the moment, you can do that yourself with only the cost of maintaining your custom image, as mentioned above.
But I'm open to discussing this together with other maintainers.

ekoops:
I agree with what @leogr said regarding creating multiple specialized images. The number of combination is high, and the same would be for maintenance overheads. I suggest you to maintain your own Falco custom image with this trivial addition.

leogr:
I had some random discussions about this with other maintainers and the consensus is not to introduce other plugins in the base image. So I'm going to close this.

That being said, thank again for your proposal and the contribution you made to the project. We really appreciate it.

Thanks

/close

PR #3771, opened 2026-01-16
chore(build): add support for gperftools CPU profiler

irozzo-1A:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind release

/area build

> /area engine

> /area tests

> /area proposals

> /area CI

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/falco/CMakeLists.txt
+++ b/userspace/falco/CMakeLists.txt
@@ -75,6 +75,11 @@ if(USE_JEMALLOC OR USE_MIMALLOC)
 	list(APPEND FALCO_LIBRARIES ${MALLOC_LIB})
 endif()
 
+if(USE_GPERFTOOLS)
+	list(APPEND FALCO_DEPENDENCIES gperftools)
+	list(APPEND FALCO_LIBRARIES "${GPERFTOOLS_PROFILER_LIB}")
+endif()
+
 if(NOT WIN32)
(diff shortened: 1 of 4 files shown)


userspace/falco/CMakeLists.txt
```diff
@@ -198,25 +203,6 @@ if(MUSL_OPTIMIZED_BUILD AND CMAKE_BUILD_TYPE STREQUAL "release")
 	)
 endif()
 
-# TODO: Add win32 support. https://github.com/falcosecurity/falco/issues/3445
-if(CMAKE_BUILD_TYPE STREQUAL "relwithdebinfo" AND NOT WIN32)
```
FedeDP: Mmh this is going to break uploading of debug symbols to Falco github release artifacts right? Why are we removing it?

irozzo-1A:
> Overall looks good to me, I am just wondering whether we should have builds that checks that these are working properly ?
> 
> There's already quite a lot of building happening though.

Thanks @sgaist, in general I would agree with you that having something that is not automatically tested brings a risk, but I would not add complexity and cost to the CI.
This requires a custom build anyway, because it is disabled by default, if it breaks it's annoying but it is not critical.
We may discuss if we want this enable by default or not, but we can also have a follow up on this.

PR #3824, opened 2026-03-17
feat: capture events filesize limits

### LINKED ISSUE #3764: Add more criteria for capture file deadlines
**Motivation**

Falco 0.42.0 added support for saving events to a capture file. You can currently limit the capture based on time, but it might be useful to specify limits based on the number of events captured and output file size as well. Other tools have similar functionality including the [sysdig CLI](https://github.com/draios/sysdig/blob/dev/userspace/sysdig/man/sysdig.md) (`-M` and `-n`), [tcpdump](https://www.tcpdump.org/manpages/tcpdump.1.html) (`-C` and `-c`), and [dumpcap](https://www.wireshark.org/docs/man-pages/dumpcap.html) (`-a`). 

**Feature**

Given that you can specify a capture duration globally using `capture.default_duration` and per-rule using `capture_duration`, it seems to make sense to add corresponding `..._events` and `..._filesize` configuration options like so:

```
capture:
  enabled: false
  path_prefix: /tmp/falco
  mode: rules
  # Specify one or more of the following
  default_duration: 5000 # milliseconds
  default_events: 25000
  default_filesize: 80000 #kB
```

```
- rule: Suspicious File Access
  desc: Detect suspicious file access patterns
  condition: >
    open_read and fd.name startswith "/etc/"
  output: >
    Suspicious file access (user=%user.name command=%proc.cmdline file=%fd.name)
  priority: WARNING
  capture: true
  # Specify one or more of the following
  capture_duration: 5000 # milliseconds
  capture_events: 25000
  capture_filesize: 80000 #kB
```

For the sake of simplicity, if more than one option is specified, their conditions should be ORed, i.e. the first condition to be reached wins.

leogr:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind release

> /area build

/area engine

/area tests

> /area proposals

> /area CI

This adds capture_events and `capture_filesize` (kB) as new stop conditions for capture files, both as per-rule fields and as global config defaults (`default_events`, `default_filesize`). When multiple conditions are configured alongside the existing capture_duration, the first one that is met stops the capture (OR semantics).

Also fixes:
 - pre-existing gaps in the rule Override JSON schema 
 - and adds comprehensive unit tests for all capture-related fields.

Fixes #3764

/milestone 0.44.0

cc @geraldcombs 

```release-note
new(config,rules): add `capture_events` and `capture_filesize` stop conditions for capture files
```

### DIFF
--- a/falco.yaml
+++ b/falco.yaml
@@ -461,11 +461,29 @@ engine:
 # 2. `all_rules`: Captures events when any enabled rule is triggered.
 #
 # When a capture starts, Falco records events from the moment the triggering rule
-# fires until the deadline is reached. The deadline is determined by the rule's
-# `capture_duration` if specified, otherwise the `default_duration` is used.
-# If additional rules trigger during an active capture, the deadline is extended
-# accordingly. Once the deadline expires, the capture stops and data is written
-# to a file. Subsequent captures create new files with unique names.
+# fires until a stop condition is reached. Two stop conditions are available:
+#
+# - Time (per-rule, soft): determined by the rule's `capture_duration` if
+#   specified, otherwise `default_duration` is used. If additional rules trigger
+#   during an active capture, the deadline is extended accordingly (the longest
+#   deadline wins). For this reason, the time limit has "at least" semantics:
+#   the capture is guaranteed to last at least that long, but can last longer if
+#   other rules keep matching.
+#
+# - File size (global, hard): `max_file_size_mb` applies to any capture,
+#   regardless of which rule triggered it. Unlike the time limit, it cannot be
+#   overridden or extended by rules: if a capture reaches this size, it stops.
+#   N.B. The size check uses the dumper's compressed on-disk counter, which is
+#   updated in chunks as zlib flushes its internal buffers. The effective file
+#   size is therefore approximate and may overshoot the configured value by up
+#   to one flush window. For this reason, avoid very small values (under a few
+#   MB), which may be inaccurate, and consider tuning with a healthy margin.
+#
+# The first stop condition met wins (OR semantics). Once a stop condition is
+# met, the capture stops and data is written to a file. Subsequent captures
+# create new files with unique names. When a capture stops because of
+# `max_file_size_mb`, Falco emits an internal INFO message so the truncation
+# is visible in the configured outputs.
 #
 # Captured data is stored in files with a `.scap` extension, which can be
 # analyzed later using:
@@ -483,7 +501,8 @@ engine:
 # Use `capture.mode` to choose between `rules` and `all_rules` modes.
 #
 # Set `capture.default_duration` to define the default capture duration
-# in milliseconds.
+# in milliseconds. Optionally, set `capture.max_file_size_mb` to enforce a
+# hard upper bound on the capture file size in MB (applies to any capture).
 #
 # --- [Suggestions]
 #
@@ -512,6 +531,10 @@ capture:
   mode: rules
   # -- Default capture duration in milliseconds if not specified in the rule.
   default_duration: 5000
+  # -- Global hard cap on capture file size in MB (0 = unlimited).
+  # This limit applies to any capture and cannot be overridden or extended by rules.
+  # The check is approximate (see the section above): prefer values of at least a few MB.
+  # max_file_size_mb: 100
 
 #################
 # Falco plugins #
--- a/unit_tests/engine/test_rule_loader.cpp
+++ b/unit_tests/engine/test_rule_loader.cpp
@@ -1571,3 +1571,137 @@ TEST_F(test_falco_engine, deprecated_evt_dir_via_macro_folded_scalar_condition_s
 	}
 	FAIL() << "no LOAD_DEPRECATED_ITEM warning found: " << m_load_result_string;
 }
+
+// Capture field tests
+
+TEST_F(test_falco_engine, rule_capture_enabled) {
+	std::string rules_content = R"END(
+- rule: test_rule
+  desc: test rule
+  condition: evt.type = close
+  output: user=%user.name
+  priority: WARNING
+  capture: true
+)END";
+
+	std::string rule_name = "test_rule";
+	ASSERT_TRUE(load_rules(rules_content, "rules.yaml")) << m_load_result_string;
+	ASSERT_VALIDATION_STATUS(yaml_helper::validation_ok) << m_load_result->schema_validation();
+
+	auto rule_description = m_engine->describe_rule(&rule_name, {});
+	ASSERT_EQ(rule_description["rules"][0]["info"]["capture"].template get<bool>(), true);
+}
+
+TEST_F(test_falco_engine, rule_capture_disabled_by_default) {
+	std::string rules_content = R"END(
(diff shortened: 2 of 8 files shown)


ekoops (dismissed):
I'd replace `filesize` with `file_size` everywhere, but I'm gonna approve it. Feel free to change or mantain it as is.
/approve

leogr:
> I'd replace `filesize` with `file_size` everywhere, but I'm gonna approve it. Feel free to change or mantain it as is. /approve


good point. Just
/hold

to collect some more feedback about this

falco.yaml
```diff
@@ -483,12 +491,15 @@ engine:
 # Use `capture.mode` to choose between `rules` and `all_rules` modes.
 #
 # Set `capture.default_duration` to define the default capture duration
-# in milliseconds.
+# in milliseconds. Optionally, set `capture.default_events` to limit the
+# number of captured events, and `capture.default_filesize` to limit the
+# capture file size in kB.
```
geraldcombs: Should this be "kiB"? There was a big push to use IEC units for base 2 prefixes years ago, but I'm not sure where best practices stand nowadays.

This looks great otherwise!

leogr: Good point 🤔

I checked, and the Falco project consistently uses SI-style unit names everywhere (e.g., rollover_mb in sinsp_cycledumper, convert_memory_to_mb, and kb for RSS/PSS/VSZ in Falco metrics). There are no IEC units (kiB, MiB) used anywhere in either falco or libs.

I believe sticking with kB here is the right call for consistency with the rest of the codebase, even though technically the code multiplies by 1024.

Likely, it is worth keeping a note for a full review of Falco with respect to the units convention.

Thanks for the review! 🙏

ekoops: Does the documentation for these configuration variables use SI units?

leogr: No, there's no documentation. It's just the convention of the current implementation.

ekoops (dismissed):

falco.yaml
```diff
@@ -461,11 +461,19 @@ engine:
...
+#   the `default_duration` is used.
+# - Event count: determined by the rule's `capture_events` if specified, otherwise
+#   the `default_events` is used.
+# - File size: determined by the rule's `capture_filesize` if specified, otherwise
+#   the `default_filesize` is used.
+#
+# When multiple stop conditions are configured, the first one reached wins (OR
+# semantics). If additional rules trigger during an active capture, the limits
+# are extended accordingly. Once a stop condition is met, the capture stops and
+# data is written to a file. Subsequent captures create new files with unique
```
irozzo-1A, line 475: basically limits of the same type are extended when new rules trigger, but the first stop condition wins.
Would not be more consistent with the extension logic to keep capturing until the last active stop condition is met?

leogr: I'm recapping on the public record, since the discussion evolved on private channels with @irozzo-1A, @geraldcombs, @c2ndev and @ekoops 

The current implementation in this PR has three per-rule limits (capture_duration, capture_filesize, capture_events), all with an "at least" semantics: a matching rule sets or extends the stop condition; the max value wins when multiple rules match; and the capture stops at the first condition met.

Concerns:

  - The "at least" / extension logic is a reasonable fit for the time dimension, but not for size. For size, users probably expect a hard upper bound on the capture file, regardless of how many rules
  fire during the capture.
  - Documenting how the different dimensions interact (and how the first-stop-wins logic plays out across three of them) is not straightforward.
  - The capture_events limit doesn't seem to have a practical use case. Reasoning in terms of "how many events to capture after a trigger" doesn't really map to something users would want to express,
  IMO.

Proposed design:

  - Global size limit (hard cap) - configured once in falco.yaml, applies to any capture. If a capture exceeds it, it stops - no extension, no override from rules.
  - Per-rule time limit (unchanged) - capture_duration stays as-is, with the current "at least" semantic.
  - Drop capture_events entirely.

With this, the interaction becomes simple: the global size cap wins if hit, otherwise the time limit applies as today. Will make this explicit in the docs.

Since we have a consensus now,  I'll refactor the PR accordingly
:+1:

ekoops: Yes. In my opinion, it doesn't really make sense to have per-rule size limit. Moreover, I agree that the semantic of a size limit should be "if the capture hits this limit, stop adding other events to it and ship it". Finally, I agree with the fact that adding a limit on the cardinality of events is not useful.

leogr: done, please review again :pray:

leogr:
@irozzo-1A @ekoops @geraldcombs PTAL again :pray:

userspace/falco/app/actions/process_events.cpp
```diff
@@ -347,14 +349,39 @@ static falco::app::run_result do_inspect(
...
 				dumper->flush();
+				uint64_t written = dumper->written_bytes();
 				dumper->close();
 				dump_started_ts = 0;
 				dump_deadline_ts = 0;
+				if(reason == capture_stop_reason::SIZE_LIMIT) {
+					std::string rule = "Falco internal: capture size limit reached";
+					std::string msg =
+					        rule + ". Configured limit: " + std::to_string(dump_max_file_size_mb) +
+					        " MB.";
```
ekoops: These are constants right? Making them static to avoid reallocation? Just wondering...

leogr: Good point. It should be addressed now. PTAL again :pray:

ekoops: Actually, I meant leaving those variable declarations there but making them `static`. This will avoid them to be initialized if no limits are configured, but will initialize them once if they are

leogr: Ok got it, now, my bad :sweat_smile: 

Reworking it again.

leogr: Done. 

@ekoops please note we can use `static` for `rule` only, since `msg` is interpolated from a config value that can change at runtime (i.e. after an hot-reload).

ekoops (approved):

unit_tests/engine/test_rule_loader.cpp
```diff
@@ -1571,3 +1571,137 @@ TEST_F(test_falco_engine, deprecated_evt_dir_via_macro_folded_scalar_condition_s
...
+  priority: WARNING
+  capture: true
+)END";
+
+	std::string rule_name = "test_rule";
+	ASSERT_TRUE(load_rules(rules_content, "rules.yaml")) << m_load_result_string;
+	ASSERT_VALIDATION_STATUS(yaml_helper::validation_ok) << m_load_result->schema_validation();
+
+	auto rule_description = m_engine->describe_rule(&rule_name, {});
+	ASSERT_EQ(rule_description["rules"][0]["info"]["capture"].template get<bool>(), true);
```
geraldcombs, line 1592: Should there be an assertion for the default duration here?

geraldcombs (approved):
This looks good, and I think making the file size a hard limit makes sense.

irozzo-1A (approved):
/lgtm

PR #3835, opened 2026-03-24
fix: cli `-o` flag escaping

leogr:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

/kind documentation

> /kind failing-test

/kind feature

> /kind release

> /area build

/area engine

> /area tests

> /area proposals

> /area CI

See https://github.com/falcosecurity/falco/issues/3049#issuecomment-2587702909

Part of #3049 
(N.B. With this, the issue is addressed; still, we will keep it open as a reminder to clean up deprecated items)

/milestone 0.44.0

```release-note
new!: add backslash escaping support to `-o` key-path parser for literal dots and brackets in YAML key names (e.g., -o 'my\.dotted\.key=val')
```


leogr:
> 
> 
> /hold

/hold cancel

false positive

johnmichaelll238-cloud:
Hi, please is this issue still available? I'd like to help out and learn along the way

johnmichaelll238-cloud:
Also, I'd like to ask if the issue is suitable for first-time contributors, and if not, could I be possibly be assigned a sub-issue that's beginner-friendly?

userspace/falco/app/options.cpp
```diff
@@ -155,7 +155,7 @@ void options::define(cxxopts::Options& opts)
...
 		("N",                        "Only print field names when used in conjunction with the --list option. It has no effect when used with other options.", cxxopts::value(names_only)->default_value("false"))
-		("o,option",                 "Set the value of option <opt> to <val>. Overrides values in the configuration file. <opt> can be identified using its location in the configuration file using dot notation. Elements of list entries can be accessed via square brackets [].\n    E.g. base.id = val\n         base.subvalue.subvalue2 = val\n         base.list[1]=val", cxxopts::value(cmdline_config_options), "<opt>=<val>")
+		("o,option",                 "Set the value of option <opt> to <val>. Overrides values in the configuration file. <opt> can be identified using its location in the configuration file using dot notation. Elements of list entries can be accessed via square brackets []. Use backslash (\\) to escape literal dots or brackets in key names.\n    E.g. base.id=val\n         base.subvalue.subvalue2=val\n         base.list[1]=val\n         base.dotted\\.key=val", cxxopts::value(cmdline_config_options), "<opt>=<val>")
```
ekoops: What if we want to escape a single `\`? Should we write something like `\\\`?

leogr:
> Also, I'd like to ask if the issue is suitable for first-time contributors, and if not, could I be possibly be assigned a sub-issue that's beginner-friendly?

Hi @johnmichaelll238-cloud 

This is a PR, not an issue. The implementation is done, we are just reviewing it. If you want, your feedback is welcome.

userspace/falco/app/options.cpp
leogr: Good catch. Yes, `\\` is a literal backslash (same as Helm's `--set`).

 So to set a key named back\slash under base:
 ``` 
  -o 'base.back\\slash=val'
```

I'll update the help text to mention this.

Comment (ekoops) [reply userspace/falco/app/options.cpp]: Mmm I don't understand why we are escaping dots with `\\.` and slashes with `\\`... It should be `\\\`... Maybe I'm missing something

Comment (leogr) [reply userspace/falco/app/options.cpp]: @ekoops 

Ah, I think I see what's tripping you up. It's the shell escaping layer mixing with the parser layer. Those are two separate, orthogonal things.

At the parser level, the rule is uniform: a single \ escapes the next char.
  - `\.` = literal `.`
  - `\[` = literal `[`
  - `\\` = literal `\`

At the shell level, how many backslashes you type depends on the quoting:

```
  ┌─────────────────┬───────────────┬─────────────────┐
  │ Parser must see │ Single-quoted │ Unquoted (bash) │
  ├─────────────────┼───────────────┼─────────────────┤
  │ \.              │ '...\.key'    │ ...\\.key       │
  ├─────────────────┼───────────────┼─────────────────┤
  │ \\              │ '...\\slash'  │ ...\\\\slash    │
  └─────────────────┴───────────────┴─────────────────┘
```


The help text documents the parser-level form on purpose - the shell layer varies (bash vs zsh vs PowerShell, etc.), but the parser is invariant. If you wrap the arg in single quotes (which I'd recommend anyway when keys have special chars), what you type equals what the parser sees.

Does this make sense? :thinking:

Comment (ekoops) [reply userspace/falco/app/options.cpp]: Yes, it makes sense :+1:

**Review by ekoops — DISMISSED** [review]
/approve

**Comment by leogr** [comment]
@falcosecurity/falco-maintainers PTAL :pray:

File: userspace/engine/yaml_helper.h [hunk 2]
```diff
@@ -369,24 +369,57 @@ class yaml_helper {
...
+						                         std::to_string(i));
+					}
+					nodeKey += next;
+					++i;
+					// If this escaped char is the last in the key, shift now
+					if(i == key.size() - 1) {
+						if(nodeKey.empty()) {
+							throw std::runtime_error("Parsing error: unexpected character at pos " +
+							                         std::to_string(i));
+						}
```
irozzo-1A: Am I wrong or this condition is never met because of line 409?

userspace/engine/yaml_helper.h
```diff
@@ -369,24 +369,57 @@ class yaml_helper {
 	 * nested nodes, with arbitrary depth. The key string follows
 	 * this regular language:
 	 *
-	 * Key 		:= NodeKey ('.' NodeKey)*
-	 * NodeKey	:= (any)+ ('[' (integer)+? ']')*
+	 * Key      := NodeKey ('.' NodeKey)*
+	 * NodeKey  := (Char | EscSeq)+ ('[' (integer)+? ']')*
```
irozzo-1A, line 373: nit: previously `Char` was defined as `any`

userspace/engine/yaml_helper.h
```diff
@@ -369,24 +369,57 @@ class yaml_helper {
...
+	 * EscSeq   := '\' ('.' | '[' | '\')
```
irozzo-1A: nit: isn/t weird that `]` cannot be escaped?

irozzo-1A (commented):
Only a few nits. As a general remark, this is technically a breaking change. As the impact should be limited I don't think it's a big deal, but it's worth mentioning.

userspace/engine/yaml_helper.h
leogr: Good catch, fixing.

userspace/engine/yaml_helper.h
leogr: I think `(Char | EscSeq)` adds clarity to the intent. I just missed the `Char` definition. I will add it.

userspace/engine/yaml_helper.h
leogr: It doesn't really need to be escaped, but I agree the fact it can't be escaped is weird. I will fix it.

leogr:
> As a general remark, this is technically a breaking change. As the impact should be limited I don't think it's a big deal, but it's worth mentioning.

Correct. Before, backslash was a regular character. Now, it isn't anymore. So I marked it as a breaking change.

I'm fixing other nits.

ekoops (approved):

irozzo-1A (approved):
/lgtm

PR #3839, opened 2026-03-30
feat(metrics): add optional k8s metadata labels to rules_matches

### LINKED ISSUE #3826: Prometheus metrics should include pod name and namespace as labels
**Motivation**

I'm monitoring Falco rule alerts via the Prometheus metric `falcosecurity_falco_rules_matches_total`, but it's difficult to identify which Kubernetes pods or namespaces triggered the alerts. I can see it in the logs of falco but not prometheus.
Was able to see it previously with falco_events from falco-exporter as k8s.ns.name and k8s.pod.name but that's deprecated now

**Feature**
I would like the `falcosecurity_falco_rules_matches_total` metric to include the following additional labels:

    k8s_pod_name
    k8s_ns_name

These should reflect the Kubernetes namespace and pod that triggered the Falco rule match, as already included in the event logs.

Additional context
- https://github.com/falcosecurity/falco/issues/3652

Debasish-87:
**What type of PR is this?**

/kind feature

/area engine

This PR introduces an optional configuration flag `metrics.include_k8s_metadata` to enrich the `falcosecurity_falco_rules_matches_total` Prometheus metric with Kubernetes-related labels.

When enabled, the metric includes:
- `k8s_ns_name`
- `k8s_pod_name`

This addresses the current limitation where Kubernetes context is available in Falco event logs but not exposed in Prometheus metrics, making it difficult to correlate alerts with specific workloads.

Since Kubernetes metadata is not available at the metrics aggregation layer, placeholder values (`"n/a"`) are used. This provides a consistent interface while preserving forward compatibility for future improvements where real metadata may be available.

The feature is disabled by default to avoid introducing high-cardinality metrics unless explicitly enabled by the user.

Fixes #3826

- Fully backward-compatible (feature is opt-in and defaults to `false`)
- No impact on existing metrics unless explicitly enabled
- Avoids high-cardinality risks by design
- Placeholder values are used due to current architectural constraints at aggregation level

```release-note
Add optional `metrics.include_k8s_metadata` configuration to include Kubernetes labels (namespace and pod name) in Prometheus rule match metrics.
```

### DIFF
--- a/userspace/falco/config_json_schema.h
+++ b/userspace/falco/config_json_schema.h
@@ -577,6 +577,10 @@ const char config_schema_string[] = LONG_STRING_CONST(
                 },
                 "jemalloc_stats_enabled": {
                     "type": "boolean"
+                },
+                "include_k8s_metadata": {
+                    "type": "boolean",
+                    "default": false
                 }
             },
             "minProperties": 1,
--- a/userspace/falco/configuration.cpp
+++ b/userspace/falco/configuration.cpp
@@ -96,6 +96,7 @@ falco_configuration::falco_configuration():
         m_metrics_flags(0),
         m_metrics_convert_memory_to_mb(true),
         m_metrics_include_empty_values(false),
+        m_include_k8s_metadata(false),
         m_plugins_hostinfo(true),
         m_capture_enabled(false),
         m_capture_path_prefix("/tmp/falco"),
@@ -563,6 +564,7 @@ void falco_configuration::load_yaml(const std::string &config_name) {
 	m_base_syscalls_all = m_config.get_scalar<bool>("base_syscalls.all", false);
 
 	m_metrics_enabled = m_config.get_scalar<bool>("metrics.enabled", false);
+	m_include_k8s_metadata = m_config.get_scalar<bool>("metrics.include_k8s_metadata", false);
 	m_metrics_interval_str = m_config.get_scalar<std::string>("metrics.interval", "5000");
 	m_metrics_interval = falco::utils::parse_prometheus_interval(m_metrics_interval_str);
 	m_metrics_stats_rule_enabled = m_config.get_scalar<bool>("metrics.output_rule", false);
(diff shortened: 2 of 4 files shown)


Debasish-87:
Hi maintainers,

This PR adds an opt-in configuration to include Kubernetes metadata in Prometheus metrics while preserving backward compatibility and avoiding high-cardinality issues.

Would appreciate your feedback and review.

Thanks!

Debasish-87:
Hi reviewers, gentle ping for review. Happy to make any changes if needed.

leogr:
> Hi reviewers, gentle ping for review. Happy to make any changes if needed.



> Hi reviewers, gentle ping for review. Happy to make any changes if needed.

Hey @Debasish-87 

Thanks for this PR and sorry for the delay.  I just want to let you know that it's under my radar.

leogr (changes requested):
Hey @Debasish-87, thanks for looking into this. The feature request in [#3826](https://github.com/falcosecurity/falco/issues/3826) is valid, and including namespace/pod labels in Prometheus metrics is standard practice in the Kubernetes ecosystem (see [Prometheus instrumentation best practices](https://prometheus.io/docs/practices/instrumentation/)).

However, I believe this PR doesn't address the issue. It adds `k8s_ns_name` and `k8s_pod_name` labels hardcoded to `"n/a"`, which provides no useful information. Constant-value labels violate Prometheus best practices since labels should [distinguish time series](https://prometheus.io/docs/practices/naming/).

The underlying problem is that `rules_matches_total` is currently a per-rule aggregate counter. The `stats_manager::on_event()` only receives a `const falco_rule&` and has no access to per-event data. That said, the `sinsp_evt*` IS available at match time (in `falco_engine::process_event()`), and `falco_outputs::handle_event()` already resolves all output fields (including `k8s.ns.name`, `k8s.pod.name` from the container plugin) into a `map<string, string>`. So the data is there, it just doesn't reach the stats_manager.

A more generic solution would be to allow users to configure which output fields should be promoted to Prometheus labels on `rules_matches_total`. Something like:

```yaml
metrics:
   rules_counters_extra_labels:
     - k8s.ns.name
     - k8s.pod.name
```

This would require changes to the `stats_manager` to track per-(rule, label_values) counters instead of just per-rule counters, but the architecture supports it since the event data is available at the right point in the pipeline.

This is a larger change than what this PR proposes, and I'm still unsure if this is a good idea. I'd suggest opening a design discussion for this approach. wdyt? 🤔 

PR #3844, opened 2026-04-03
build(docker): replace apt-key with keyring; use gnupg instead of gnupg2

parisnakitakejser:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind release

/area build

> /area engine

> /area tests

> /area proposals

> /area CI

gnupg2 is, at the time of this patch, a dummy transitional package. gnupg is the official package name that provides gnupg 2.

This PR updates the Dockerfiles to correct the package name.

Fixes #

```release-note
NONE
```

### DIFF
--- a/docker/falco-debian/Dockerfile
+++ b/docker/falco-debian/Dockerfile
@@ -29,8 +29,8 @@ RUN apt-get -y update && apt-get -y install curl jq ca-certificates gnupg \
 WORKDIR /
 
 RUN curl -fsSL https://falco.org/repo/falcosecurity-packages.asc \
-	| gpg --dearmor -o /usr/share/keyrings/falcosecurity.gpg \
-	&& echo "deb [signed-by=/usr/share/keyrings/falcosecurity.gpg] https://download.falco.org/packages/${VERSION_BUCKET} stable main" | tee -a /etc/apt/sources.list.d/falcosecurity.list \
+	| gpg --dearmor -o /usr/share/keyrings/falco-archive-keyring.gpg \
+	&& echo "deb [signed-by=/usr/share/keyrings/falco-archive-keyring.gpg] https://download.falco.org/packages/${VERSION_BUCKET} stable main" | tee -a /etc/apt/sources.list.d/falcosecurity.list \
 	&& apt-get update -y \
 	&& if [ "$FALCO_VERSION" = "latest" ]; then FALCO_DRIVER_CHOICE=none apt-get install -y --no-install-recommends falco; else FALCO_DRIVER_CHOICE=none apt-get install -y --no-install-recommends falco=${FALCO_VERSION}; fi \
 	&& apt-get clean \
--- a/docker/falco-debian/Dockerfile
+++ b/docker/falco-debian/Dockerfile
@@ -23,13 +23,14 @@ ENV FALCO_VERSION="${FALCO_VERSION}" \
     HOST_ROOT="${HOST_ROOT}" \
     HOME="${HOME}"
 
-RUN apt-get -y update && apt-get -y install curl jq ca-certificates gnupg2 \
+RUN apt-get -y update && apt-get -y install curl jq ca-certificates gnupg \
 	&& apt clean -y && rm -rf /var/lib/apt/lists/*
 
 WORKDIR /
 
-RUN curl -s https://falco.org/repo/falcosecurity-packages.asc | apt-key add - \
-	&& echo "deb https://download.falco.org/packages/${VERSION_BUCKET} stable main" | tee -a /etc/apt/sources.list.d/falcosecurity.list \
+RUN curl -fsSL https://falco.org/repo/falcosecurity-packages.asc \
+	| gpg --dearmor -o /usr/share/keyrings/falcosecurity.gpg \
+	&& echo "deb [signed-by=/usr/share/keyrings/falcosecurity.gpg] https://download.falco.org/packages/${VERSION_BUCKET} stable main" | tee -a /etc/apt/sources.list.d/falcosecurity.list \
 	&& apt-get update -y \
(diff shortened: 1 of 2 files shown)


leogr (changes requested):
Hey @parisnakitakejser, thanks for looking into this!

The `apt-key` modernization and the `gnupg2` 👉  `gnupg` change are good improvements 👍

However, upgrading the base image to `debian:13-slim` and dropping `gcc-11` is something we can't do. The `falco-driver-loader` image exists to compile kernel modules and legacy eBPF probes on-the-fly on user systems. The GCC version must be compatible with the one used to build the running kernel, and different kernel versions require different GCC versions (e.g., kernel 5.x needs `gcc-11`, kernel 6.0-6.4 needs `gcc-12`).

By bumping to Debian 13, the default `gcc` would jump to 14.x (only valid for kernel 6.9+) and `gcc-11` would no longer be available from apt. That would break driver compilation for anyone running kernel 5.x or 6.0-6.8.

N.B. the main `falcosecurity/falco` image is already based on Wolfi (not Debian) and is not affected by Debian CVEs. The `-debian` variant and the `driver-loader` image are Debian-based specifically because of the build toolchain requirements. They use old Debian versions on purpose, and the CVEs reported can't be really exploited because the affected tools are only used to build the driver (during the init phase) and not to run Falco.

May you consider splitting this PR? The `apt-key` and `gnupg` fixes could be submitted separately and would likely be accepted. The base image bump needs a broader discussion since it impacts driver compatibility.

Thanks :pray:

/kind cleanup
/area build

parisnakitakejser:
> Hey @parisnakitakejser, thanks for looking into this!
> 
> The `apt-key` modernization and the `gnupg2` 👉 `gnupg` change are good improvements 👍
> 
> However, upgrading the base image to `debian:13-slim` and dropping `gcc-11` is something we can't do. The `falco-driver-loader` image exists to compile kernel modules and legacy eBPF probes on-the-fly on user systems. The GCC version must be compatible with the one used to build the running kernel, and different kernel versions require different GCC versions (e.g., kernel 5.x needs `gcc-11`, kernel 6.0-6.4 needs `gcc-12`).
> 
> By bumping to Debian 13, the default `gcc` would jump to 14.x (only valid for kernel 6.9+) and `gcc-11` would no longer be available from apt. That would break driver compilation for anyone running kernel 5.x or 6.0-6.8.
> 
> N.B. the main `falcosecurity/falco` image is already based on Wolfi (not Debian) and is not affected by Debian CVEs. The `-debian` variant and the `driver-loader` image are Debian-based specifically because of the build toolchain requirements. They use old Debian versions on purpose, and the CVEs reported can't be really exploited because the affected tools are only used to build the driver (during the init phase) and not to run Falco.
> 
> May you consider splitting this PR? The `apt-key` and `gnupg` fixes could be submitted separately and would likely be accepted. The base image bump needs a broader discussion since it impacts driver compatibility.
> 
> Thanks 🙏
> 
> /kind cleanup /area build

I have tryed to do what you explain, let me know if i miss something thanks :)

parisnakitakejser (commented):
LGTM

irozzo-1A:
Hi @parisnakitakejser, as you noticed I adjusted the title and put back the original template that should be used for the PR description, please fill it to reflect your changes and feel free to ping me if you need assistance.

parisnakitakejser:
> Hi @parisnakitakejser, as you noticed I adjusted the title and put back the original template that should be used for the PR description, please fill it to reflect your changes and feel free to ping me if you need assistance.

If you can help me to wirte the description in the right way, like you did with the title i will be very happy it's look you are better then me to do that part, :)

docker/falco-debian/Dockerfile
```diff
@@ -23,13 +23,14 @@ ENV FALCO_VERSION="${FALCO_VERSION}" \
...
+RUN apt-get -y update && apt-get -y install curl jq ca-certificates gnupg \
 	&& apt clean -y && rm -rf /var/lib/apt/lists/*
 
 WORKDIR /
 
-RUN curl -s https://falco.org/repo/falcosecurity-packages.asc | apt-key add - \
-	&& echo "deb https://download.falco.org/packages/${VERSION_BUCKET} stable main" | tee -a /etc/apt/sources.list.d/falcosecurity.list \
+RUN curl -fsSL https://falco.org/repo/falcosecurity-packages.asc \
+	| gpg --dearmor -o /usr/share/keyrings/falcosecurity.gpg \
+	&& echo "deb [signed-by=/usr/share/keyrings/falcosecurity.gpg] https://download.falco.org/packages/${VERSION_BUCKET} stable main" | tee -a /etc/apt/sources.list.d/falcosecurity.list \
```
leogr:
```suggestion
	| gpg --dearmor -o /usr/share/keyrings/falco-archive-keyring.gpg \
	&& echo "deb [signed-by=/usr/share/keyrings/falco-archive-keyring.gpg] https://download.falco.org/packages/${VERSION_BUCKET} stable main" | tee -a /etc/apt/sources.list.d/falcosecurity.list \
```

Nit. For consistency with our install docs (https://falco.org/docs/setup/packages/#install-with-apt), may you rename the keyring to `falco-archive-keyring.gpg`? It also follows the Debian `*-archive-keyring.gpg` convention :pray:

leogr (commented):
Hey @parisnakitakejser

Thanks for the rework, this addresses my previous feedback nicely 👍 

Just left a tiny nit about the keyring filename (see inline).

Also, the same `apt-key` + `gnupg2` pattern is still present in `docker/driver-loader-buster/Dockerfile`. May you consider updating it here, too, or leave it for a follow-up PR? Since that image is a legacy one, I would not block on it.

Thanks :pray:

leogr (approved):
LGTM now!

Thank you 🙏

sgaist (approved):
I updated the description as well as unless you are aware of the gnupg2 vs gnupg package situation in Debian, it is not necessarily an obvious thing to do.

PR #3858, opened 2026-04-14
fix(engine): show condition text in warning snippet for folded scalar conditions

### LINKED ISSUE #3857: Default rules using YAML folded scalar > causes LOAD_DEPRECATED_ITEM false warning about evt.dir
When Falco loads default rules that use the YAML folded scalar > for multi-line values (condition: >), the parser appears to misinterpret the > character, producing incorrect deprecation warnings referencing evt.dir, example:

```
In rules content: (/etc/falco/falco_rules.yaml:0:0)
    rule 'Fileless execution via memfd_create': (/etc/falco/falco_rules.yaml:1243:2)
    rule condition: (/etc/falco/falco_rules.yaml:1249:13)
------
  condition: >
             ^
------
LOAD_DEPRECATED_ITEM (Used deprecated item: field 'evt.dir'): due to the drop of enter events, 'evt.dir = <' always evaluates to true, and 'evt.dir = >' always evaluates to false. The rule expression can be simplified by removing the condition on 'evt.dir'
```

The affected rules (e.g. Fileless execution via memfd_create) do not contain any reference to evt.dir. The warning points at the > character on the condition.

**Steps to reproduce:**
Deploy Falco via Helm with default rules
Check Falco logs for LOAD_DEPRECATED_ITEM warnings
Inspect the flagged rules — they contain no evt.dir usage

**Expected behavior:**
No warning should be emitted for rules that don't use evt.dir.

**Environment**
Falco version: 0.43.1 (x86_64)
Helm chart version: 8.0.2
Platform: GKE

ssam18:
When a rule condition uses YAML folded scalar syntax (`condition: >`), the warning snippet was pointing at the `>` character, making it look like the YAML indicator itself was causing the deprecated field warning (e.g. `evt.dir`). The fix wraps the warning context with the actual condition text, consistent with how compile errors are already handled in the same code path. Added a regression test covering this case. Fixes #3857 

/kind bug
/area engine

```release-note
fix: show condition text in warning snippet for folded scalar conditions
```

### DIFF
--- a/unit_tests/engine/test_rule_loader.cpp
+++ b/unit_tests/engine/test_rule_loader.cpp
@@ -1541,3 +1541,33 @@ TEST_F(test_falco_engine, deprecated_evt_dir_folded_scalar_condition_snippet) {
 	}
 	FAIL() << "no warning snippet containing 'evt.dir': " << m_load_result_string;
 }
+
+TEST_F(test_falco_engine, deprecated_evt_dir_via_macro_folded_scalar_condition_snippet) {
+	std::string rules_content = R"END(
+- macro: spawned_process
+  condition: evt.type in (execve, execveat) and evt.dir = <
+
+- rule: test_rule
+  desc: test rule
+  condition: >
+    spawned_process and proc.name = bash
+  output: command=%proc.cmdline
+  priority: INFO
+)END";
+
+	ASSERT_TRUE(load_rules(rules_content, "rules.yaml"));
+	ASSERT_VALIDATION_STATUS(yaml_helper::validation_ok) << m_load_result->schema_validation();
+	ASSERT_TRUE(has_warnings());
+	ASSERT_TRUE(check_warning_message("evt.dir")) << m_load_result_string;
+
+	for(const auto& warn : m_load_result_json["warnings"]) {
+		if(warn["code"] != "LOAD_DEPRECATED_ITEM") {
+			continue;
+		}
+		std::string snippet = warn["context"]["snippet"];
+		ASSERT_EQ(snippet.find("condition: >"), std::string::npos) << snippet;
+		ASSERT_NE(snippet.find("spawned_process"), std::string::npos) << snippet;
+		return;
+	}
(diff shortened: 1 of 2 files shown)


unit_tests/engine/test_rule_loader.cpp
```diff
@@ -1516,3 +1516,28 @@ TEST_F(test_falco_engine, no_deprecated_field_warning_in_output) {
 	ASSERT_FALSE(check_warning_message("evt.dir")) << m_load_result_string;
 	EXPECT_EQ(num_rules_for_ruleset(), 1);
 }
+
+TEST_F(test_falco_engine, deprecated_evt_dir_folded_scalar_condition_snippet) {
```
leogr, line 1520: Non-blocking: #3857 describes the warning firing on a rule whose *macros* reference `evt.dir` indirectly (not the rule condition itself). May you add a second test case where `evt.dir` is hidden inside a macro the rule depends on? It would mirror the bug report more closely and guard against divergence if the warning emission path ever gains per-node positions. Not a blocker for this PR 🙏

leogr (commented):
Hey @ssam18, welcome and thanks for the fix 🙏

Nice work. The approach is consistent with how compile errors and compile warnings are already wrapped a few lines below in the same function (see `rule_loader_compiler.cpp:387` and `:398`), which is exactly the right pattern to reuse here. The regression test covers the folded-scalar case well. Left one small non-blocking note inline.

I found just nit (see my comment below). Let me know if you want to fix it in this PR.

leogr:
Also, a friendly reminder for the future: DO NOT remove the PR template, please. We put it there for a reason :sweat_smile: 

I will fix it for this time, so no action is needed from you.

leogr:
> 
> 
> /hold

false positive

/hold cancel

leogr (approved):
Thanks for the follow-up 🙏

LGTM now.

leogr:

sgaist (approved):

PR #3871, opened 2026-05-07
fix(userspace): open pidfile with O_NOFOLLOW to prevent symlink TOCTOU

alexmchughdev:
**What type of PR is this?**

/kind bug

Hardens the pidfile open path in `userspace/falco/app/actions/pidfile.cpp` against a symlink TOCTOU. Previously the pidfile was created via `std::ofstream::open()`, which has no portable `O_NOFOLLOW` equivalent. If an operator configures a pidfile path in a directory writable by an unprivileged user, that user could pre-place a symlink and trick a root-running Falco into clobbering an arbitrary file with the PID on startup.

The pidfile path is operator-controlled and typically lives in a root-owned directory like `/var/run`, so this is not exploitable in the standard threat model — this is a defence-in-depth fix.

The change replaces the `std::ofstream` open with a raw POSIX `open()` using `O_WRONLY | O_CREAT | O_TRUNC | O_NOFOLLOW | O_CLOEXEC` and writes the PID with `dprintf`. The `open()` will now fail with `ELOOP` if the path is a symlink. Error reporting goes through `falco_strerror_r` matching the pattern in `create_signal_handlers.cpp`. Behaviour is unchanged for legitimate users.

Fixes #

- Same path and same `exit(-1)` on failure as before. The written content changes from `<pid>` to `<pid>\n` (master wrote no trailing newline; nothing in-tree reads the file).
- The `O_CLOEXEC` flag is also added so the fd does not leak into any later forked program-output children.
- A short comment explains *why* `O_NOFOLLOW` is there so it is not removed during a future cleanup.

```release-note
fix(userspace): open the pidfile with O_NOFOLLOW to prevent a symlink TOCTOU when the pidfile path is in an attacker-writable directory. The pidfile content now ends with a trailing newline
```

### DIFF
--- a/unit_tests/falco/app/actions/test_pidfile.cpp
+++ b/unit_tests/falco/app/actions/test_pidfile.cpp
@@ -1,0 +1,87 @@
+// SPDX-License-Identifier: Apache-2.0
+/*
+Copyright (C) 2023 The Falco Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "app_action_helpers.h"
+
+#include <filesystem>
+#include <fstream>
+#include <string>
+
+#ifndef _WIN32
+#include <unistd.h>
+#else
+#include <process.h>
+#define getpid _getpid
+#endif
+
+TEST(ActionPidfile, empty_filename_is_noop) {
+	falco::app::state s;
+	s.options.pidfilename = "";
+	EXPECT_ACTION_OK(falco::app::actions::pidfile(s));
+}
+
+TEST(ActionPidfile, dry_run_does_not_create_file) {
+	auto path = std::filesystem::temp_directory_path() / "falco_test_dry_run.pid";
+	std::filesystem::remove(path);
+
+	falco::app::state s;
+	s.options.dry_run = true;
+	s.options.pidfilename = path.string();
+	EXPECT_ACTION_OK(falco::app::actions::pidfile(s));
+	EXPECT_FALSE(std::filesystem::exists(path));
+}
+
+TEST(ActionPidfile, writes_pid_to_file) {
+	auto path = std::filesystem::temp_directory_path() / "falco_test_pidfile.pid";
+
+	// Pre-fill the file with longer content to check it is fully replaced.
+	{
+		std::ofstream prefill(path.string());
+		prefill << "999999999999999999\n";
+	}
+
+	falco::app::state s;
+	s.options.pidfilename = path.string();
+	EXPECT_ACTION_OK(falco::app::actions::pidfile(s));
+
+	std::string content;
+	{
+		std::ifstream in(path.string());
+		content.assign((std::istreambuf_iterator<char>(in)), std::istreambuf_iterator<char>());
+	}
+	EXPECT_EQ(content, std::to_string(getpid()) + "\n");
+
+	std::filesystem::remove(path);
+}
+
+#if defined(GTEST_HAS_DEATH_TEST) && GTEST_HAS_DEATH_TEST && !defined(_WIN32)
+TEST(ActionPidfileDeathTest, refuses_symlinked_path) {
+	auto dir = std::filesystem::temp_directory_path();
+	auto target = dir / ("falco_test_symlink_target_" + std::to_string(getpid()) + ".pid");
+	auto link = dir / ("falco_test_symlink_" + std::to_string(getpid()) + ".pid");
+	std::filesystem::remove(link);
+	std::filesystem::remove(target);
+	std::filesystem::create_symlink(target, link);
+
+	falco::app::state s;
+	s.options.pidfilename = link.string();
+	EXPECT_EXIT(falco::app::actions::pidfile(s), ::testing::ExitedWithCode(255), "");
+	EXPECT_FALSE(std::filesystem::exists(target));
+
+	std::filesystem::remove(link);
+}
+#endif
--- a/userspace/falco/app/actions/pidfile.cpp
+++ b/userspace/falco/app/actions/pidfile.cpp
@@ -23,12 +23,47 @@ limitations under the License.
 #include <cerrno>
 #include <cstdio>
 #include "compat.h"
+#else
+#include <windows.h>
+#include <process.h>
+#include <cstdio>
+#include <string>
 #endif
 
 #include "actions.h"
 
 using namespace falco::app;
 using namespace falco::app::actions;
+
+#ifdef _WIN32
+namespace {
+std::string win32_error_string(DWORD code) {
+	char* msg = nullptr;
+	DWORD n = FormatMessageA(FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM |
+	                                 FORMAT_MESSAGE_IGNORE_INSERTS,
+	                         nullptr,
+	                         code,
+	                         MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
+	                         reinterpret_cast<LPSTR>(&msg),
+	                         0,
+	                         nullptr);
+	std::string out = std::to_string(code);
+	if(n > 0 && msg != nullptr) {
+		std::string text(msg, n);
+		while(!text.empty() && (text.back() == '\n' || text.back() == '\r')) {
+			text.pop_back();
+		}
+		if(!text.empty()) {
+			out += " (" + text + ")";
+		}
+	}
+	if(msg != nullptr) {
+		LocalFree(msg);
+	}
+	return out;
+}
+}  // namespace
+#endif
 
 falco::app::run_result falco::app::actions::pidfile(const falco::app::state& state) {
 	if(state.options.dry_run) {
@@ -69,19 +104,58 @@ falco::app::run_result falco::app::actions::pidfile(const falco::app::state& sta
 
 	::close(fd);
 #else
-	int64_t self_pid = getpid();
-
-	std::ofstream stream;
-	stream.open(state.options.pidfilename);
-
-	if(!stream.good()) {
-		falco_logger::log(
-		        falco_logger::level::ERR,
-		        "Could not write pid to pidfile " + state.options.pidfilename + ". Exiting.\n");
+	// Windows analog of O_NOFOLLOW: refuse to write the pidfile if the path
+	// is a reparse point (symlink/junction). We pre-check with
+	// GetFileAttributesA so an existing reparse point produces a clear error,
+	// and we also pass FILE_FLAG_OPEN_REPARSE_POINT to CreateFile plus a
+	// post-open BY_HANDLE_FILE_INFORMATION check as defence-in-depth against
+	// the small TOCTOU window between the two calls.
+	DWORD attrs = GetFileAttributesA(state.options.pidfilename.c_str());
+	if(attrs != INVALID_FILE_ATTRIBUTES && (attrs & FILE_ATTRIBUTE_REPARSE_POINT)) {
+		falco_logger::log(falco_logger::level::ERR,
+		                  "Refusing to write pidfile " + state.options.pidfilename +
+		                          ": path is a reparse point. Exiting.\n");
 		exit(-1);
 	}
-	stream << self_pid;
-	stream.close();
+
+	HANDLE h = CreateFileA(state.options.pidfilename.c_str(),
+	                       GENERIC_WRITE,
+	                       FILE_SHARE_READ,
+	                       nullptr,
+	                       CREATE_ALWAYS,
+	                       FILE_ATTRIBUTE_NORMAL | FILE_FLAG_OPEN_REPARSE_POINT,
+	                       nullptr);
+	if(h == INVALID_HANDLE_VALUE) {
+		falco_logger::log(falco_logger::level::ERR,
+		                  "Could not open pidfile " + state.options.pidfilename +
+		                          " (error: " + win32_error_string(GetLastError()) +
+		                          "). Exiting.\n");
+		exit(-1);
+	}
+
+	BY_HANDLE_FILE_INFORMATION info{};
+	if(!GetFileInformationByHandle(h, &info) ||
+	   (info.dwFileAttributes & FILE_ATTRIBUTE_REPARSE_POINT)) {
+		CloseHandle(h);
+		falco_logger::log(falco_logger::level::ERR,
+		                  "Refusing to write pidfile " + state.options.pidfilename +
+		                          ": path is a reparse point. Exiting.\n");
+		exit(-1);
+	}
+
+	char buf[32];
+	int len = std::snprintf(buf, sizeof(buf), "%lld\n", (long long)_getpid());
+	DWORD written = 0;
+	if(len < 0 || !WriteFile(h, buf, (DWORD)len, &written, nullptr) ||
+	   written != (DWORD)len) {
+		DWORD err = GetLastError();
+		CloseHandle(h);
+		falco_logger::log(falco_logger::level::ERR,
+		                  "Could not write pid to pidfile " + state.options.pidfilename +
+		                          " (error: " + win32_error_string(err) + "). Exiting.\n");
+		exit(-1);
+	}
+	CloseHandle(h);
 #endif
 
 	return run_result::ok();
--- a/userspace/falco/app/actions/pidfile.cpp
+++ b/userspace/falco/app/actions/pidfile.cpp
@@ -15,9 +15,15 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+#ifndef _WIN32
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <fcntl.h>
+#include <unistd.h>
+#include <cerrno>
+#include <cstdio>
+#include "compat.h"
+#endif
 
 #include "actions.h"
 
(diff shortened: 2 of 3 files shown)


ekoops:
Hey @alexmchughdev , unfortunately this PR breaks windows build. Could you please take a look? https://github.com/falcosecurity/falco/actions/runs/25507407572/job/75014856179?pr=3871

ekoops:
The new solution fixes the issue on non-Windows platform, but leaves it as is on WIndows. It would be great to also fix the issue there.

leogr (commented):
clang-format wants to reformat a couple of lines in the Windows branch. 

@alexmchughdev May you run `pre-commit run --all-files` locally and push a fixup? 🙏

sathvik-shettyy:
Nice hardening improvement. I was wondering whether we currently have any regression tests validating that pidfile creation fails when the configured path is a symlink. It might be useful to add coverage for both a regular pidfile and a symlink target to prevent future regressions.

userspace/falco/app/actions/pidfile.cpp
```diff
@@ -34,19 +75,86 @@ falco::app::run_result falco::app::actions::pidfile(const falco::app::state& sta
...
+		DWORD err = GetLastError();
+		CloseHandle(h);
+		falco_logger::log(falco_logger::level::ERR,
+		                  "Could not write pid to pidfile " + state.options.pidfilename +
+		                          " (error: " + win32_error_string(err) + "). Exiting.\n");
 		exit(-1);
 	}
-	stream << self_pid;
-	stream.close();
+	CloseHandle(h);
```
leogr, line 147:
```suggestion
	// OPEN_ALWAYS rather than CREATE_ALWAYS: FILE_FLAG_OPEN_REPARSE_POINT is
	// documented as incompatible with CREATE_ALWAYS, and we need the flag
	// honoured so the check below inspects the link itself, not its target.
	// OPEN_ALWAYS does not truncate, so we truncate explicitly after writing.
	HANDLE h = CreateFileA(state.options.pidfilename.c_str(),
	                       GENERIC_WRITE,
	                       FILE_SHARE_READ,
	                       nullptr,
	                       OPEN_ALWAYS,
	                       FILE_ATTRIBUTE_NORMAL | FILE_FLAG_OPEN_REPARSE_POINT,
	                       nullptr);
	if(h == INVALID_HANDLE_VALUE) {
		falco_logger::log(falco_logger::level::ERR,
		                  "Could not open pidfile " + state.options.pidfilename + " (error: " +
		                          win32_error_string(GetLastError()) + "). Exiting.\n");
		exit(-1);
	}

	BY_HANDLE_FILE_INFORMATION info{};
	if(!GetFileInformationByHandle(h, &info) ||
	   (info.dwFileAttributes & FILE_ATTRIBUTE_REPARSE_POINT)) {
		CloseHandle(h);
		falco_logger::log(falco_logger::level::ERR,
		                  "Refusing to write pidfile " + state.options.pidfilename +
		                          ": path is a reparse point. Exiting.\n");
		exit(-1);
	}

	char buf[32];
	int len = std::snprintf(buf, sizeof(buf), "%lld\n", (long long)_getpid());
	DWORD written = 0;
	if(len < 0 || !WriteFile(h, buf, (DWORD)len, &written, nullptr) || written != (DWORD)len) {
		DWORD err = GetLastError();
		CloseHandle(h);
		falco_logger::log(falco_logger::level::ERR,
		                  "Could not write pid to pidfile " + state.options.pidfilename +
		                          " (error: " + win32_error_string(err) + "). Exiting.\n");
		exit(-1);
	}

	// Drop anything left over from a longer PID written by a previous run.
	if(!SetEndOfFile(h)) {
		DWORD err = GetLastError();
		CloseHandle(h);
		falco_logger::log(falco_logger::level::ERR,
		                  "Could not truncate pidfile " + state.options.pidfilename +
		                          " (error: " + win32_error_string(err) + "). Exiting.\n");
		exit(-1);
	}
	CloseHandle(h);
```

I have a doubt here 🤔

`CreateFile` documents the flag like this 👉 https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew

> Normal reparse point processing will not occur; CreateFile will attempt to open the reparse point. **This flag cannot be used with the CREATE_ALWAYS flag.**

Since the two are combined here, I believe the flag is not honoured, so the `BY_HANDLE_FILE_INFORMATION` check below can never see `FILE_ATTRIBUTE_REPARSE_POINT`. And with the flag not in effect 👉 https://learn.microsoft.com/en-us/windows/win32/fileio/symbolic-link-effects-on-file-systems-functions

> If **FILE_FLAG_OPEN_REPARSE_POINT** is not specified and: [...] If **CREATE_ALWAYS**, **TRUNCATE_EXISTING**, or **FILE_FLAG_DELETE_ON_CLOSE** are specified, the file affected is the target.

which is exactly the clobber we want to avoid.

To be fair, the `GetFileAttributesA` pre-check still catches a reparse point that is already sitting at the path, so the realistic case is covered either way. What does not hold is the TOCTOU window the comment claims to cover.

The suggestion switches to `OPEN_ALWAYS`, which is the only disposition **not** listed as incompatible with the flag, plus an explicit `SetEndOfFile()` since `OPEN_ALWAYS` does not truncate on its own (otherwise a shorter PID would leave trailing bytes from a longer previous one). I verified it is `clang-format` clean, however I have no Windows box, so may you double-check it? 🙏

That said, I'd be equally happy if you just dropped the flag and the post-open check and left a comment about the residual race. Windows is a `MINIMAL_BUILD` (EXPERIMENTAL) target we do not publish, libscap has no Windows capture engine, and nothing consumes a pidfile there, so I'd rather have less code we cannot test than more.

Your call. wdyt? 🤔

userspace/falco/app/actions/pidfile.cpp
```diff
@@ -34,19 +75,86 @@ falco::app::run_result falco::app::actions::pidfile(const falco::app::state& sta
 		return run_result::ok();
 	}
 
-	int64_t self_pid = getpid();
+#ifndef _WIN32
+	// O_NOFOLLOW makes open() fail with ELOOP if the path is a symlink, so an
+	// unprivileged user who can write to the pidfile's directory cannot
+	// pre-place a symlink and trick a root-running Falco into clobbering an
+	// arbitrary file with the PID.
+	int fd = ::open(state.options.pidfilename.c_str(),
```
leogr, line 83: This is the correct race-free pattern 👍

+1 on the regression test point raised by @sathvik-shettyy above. I believe it is worth having, since this is exactly the kind of hardening that rots silently.

For the record, on what is feasible today:

- the success path is easy with the app-action harness (see `unit_tests/falco/app/actions/` and the `EXPECT_ACTION_OK` helper): set `pidfilename` to a temp path and assert the content
- the symlink-refusal path cannot go there, since `pidfile()` calls `exit(-1)` directly (pre-existing, not your doing) instead of returning `run_result::fatal(...)` like the sibling actions, so `EXPECT_ACTION_FAIL` never sees it

So I'd say the natural place is a black-box test in `falcosecurity/testing`, which already has everything needed: `WithArgs(...)`, `ExitCode()`, and `run.WorkDir()` to plant the symlink.

Three gotchas, so nobody loses time on them:

- `exit(-1)` surfaces as exit code `255`, not `1`
- `--dry-run` cannot be used to reach it, since `pidfile()` returns early on it
- that suite is a separate repo pinned at `@main`, so a test there cannot gate this PR, and it must not assume the new behaviour until this ships

Not a blocker for me, and happy to take it as a follow-up if you prefer 🙏

leogr (commented):
Reviewed using https://github.com/leogr/falco-expert

userspace/falco/app/actions/pidfile.cpp
alexmchughdev: Good catch, you're right that the flag was a no-op with CREATE_ALWAYS, so the post-open check could never fire. Went with your second option: dropped the flag and the post-open check, kept the GetFileAttributesA pre-check, and updated the comment to document the residual race. Agreed that less untestable code is the right trade here. Credited you with a Suggested-by on the commit.

userspace/falco/app/actions/pidfile.cpp
alexmchughdev: Added unit_tests/falco/app/actions/test_pidfile.cpp with the success path (asserts exactly PID + newline, pre-filled with longer content to prove truncation) plus the empty-pidfilename and --dry-run cases. Agreed the symlink path can't be covered there, so I'll do the black-box test in falcosecurity/testing as a follow-up once this ships. Thanks for mapping out the gotchas.

userspace/falco/app/actions/pidfile.cpp
```diff
@@ -34,19 +75,77 @@ falco::app::run_result falco::app::actions::pidfile(const falco::app::state& sta
...
+	                O_WRONLY | O_CREAT | O_TRUNC | O_NOFOLLOW | O_CLOEXEC,
+	                0644);
```
leogr: Not for this PR, just so it is on the record.

`O_NOFOLLOW` rejects symlinks, but it does not reject a FIFO. I checked it: with a pre-existing FIFO at the path, this exact `open()` never returns, it just blocks waiting for a reader. Since `pidfile` is app action 23 of 28, Falco would never reach `process_events`, i.e. a silent detection outage rather than a clobbered file.

It is pre-existing (`std::ofstream::open` blocks the same way), so this PR makes nothing worse. However, it is the same attacker and the same directory the PR description describes, so `O_NOFOLLOW` alone does not fully close that threat model. The complete fix would be an `fstat` + `S_ISREG` check on the fd after the open.

Happy to open a follow-up issue myself, unless you prefer to fold it in here. wdyt? 🤔

unit_tests/falco/app/actions/test_pidfile.cpp
```diff
@@ -0,0 +1,66 @@
...
+	falco::app::state s;
+	s.options.pidfilename = path.string();
+	EXPECT_ACTION_OK(falco::app::actions::pidfile(s));
+
+	std::ifstream in(path.string());
+	std::string content((std::istreambuf_iterator<char>(in)), std::istreambuf_iterator<char>());
+	EXPECT_EQ(content, std::to_string(getpid()) + "\n");
+
+	std::filesystem::remove(path);
+}
```
leogr:
```suggestion
}

#if defined(GTEST_HAS_DEATH_TEST) && GTEST_HAS_DEATH_TEST && !defined(_WIN32)
TEST(ActionPidfileDeathTest, refuses_symlinked_path) {
	auto dir = std::filesystem::temp_directory_path();
	auto target = dir / ("falco_test_symlink_target_" + std::to_string(getpid()) + ".pid");
	auto link = dir / ("falco_test_symlink_" + std::to_string(getpid()) + ".pid");
	std::filesystem::remove(link);
	std::filesystem::remove(target);
	std::filesystem::create_symlink(target, link);

	falco::app::state s;
	s.options.pidfilename = link.string();
	EXPECT_EXIT(falco::app::actions::pidfile(s), ::testing::ExitedWithCode(255), "");
	EXPECT_FALSE(std::filesystem::exists(target));

	std::filesystem::remove(link);
}
#endif
```

Sorry, I steered you wrong in my previous review 😅

I said the symlink case cannot live here because `pidfile()` calls `exit(-1)`. That holds for `EXPECT_ACTION_FAIL`, but I forgot about gtest death tests, which exist precisely for this. So the regression test @sathvik-shettyy asked for can stay in-repo and gate this PR, and the `falcosecurity/testing` follow-up is not needed at all.

I prototyped it outside the Falco tree, against a stub with the same `open()` flags and the same `exit(-1)`:

- with `O_NOFOLLOW` 👉 passes
- without it 👉 `Result: failed to die`, plus `std::filesystem::exists(target)` turning `true`

So it does catch the regression, on both assertions. It is also `clang-format` clean with our `.clang-format`.

Two things I could not check, since I did not build Falco itself:

- we vendor gtest `v1.14.0` and I tested against `1.17.0`. `EXPECT_EXIT`, `ExitedWithCode` and `GTEST_HAS_DEATH_TEST` are unchanged between the two, so I believe it is fine
- `falco::app::state` builds a `sinsp`. If that leaves threads around, gtest prints a death-test warning (not a failure). The `#if` guard already covers wasm, where death tests are not available

May you give it a run before pushing? 🙏

userspace/falco/app/actions/pidfile.cpp
```diff
@@ -34,19 +75,77 @@ falco::app::run_result falco::app::actions::pidfile(const falco::app::state& sta
...
+	if(fd == -1) {
+		char errbuf[256];
+		const char* errstr = falco_strerror_r(errno, errbuf, sizeof(errbuf));
+		falco_logger::log(falco_logger::level::ERR,
+		                  "Could not write pid to pidfile " + state.options.pidfilename +
+		                          " (error: " + errstr + "). Exiting.\n");
+		exit(-1);
+	}
+
+	if(dprintf(fd, "%lld\n", (long long)getpid()) < 0) {
```
leogr, line 95: `master` writes `stream << self_pid;` with no trailing newline 👉 https://github.com/falcosecurity/falco/blob/master/userspace/falco/app/actions/pidfile.cpp#L48

So the file content goes from `1234` to `1234\n`. I'm fine with the change itself, since a trailing newline is the conventional pidfile format and nothing in-tree reads the file. However, the description says "same content (PID + newline)" and the `release-note` block does not mention it, so may you add it there? 🙏

unit_tests/falco/app/actions/test_pidfile.cpp
```diff
@@ -0,0 +1,66 @@
...
+		prefill << "999999999999999999\n";
+	}
+
+	falco::app::state s;
+	s.options.pidfilename = path.string();
+	EXPECT_ACTION_OK(falco::app::actions::pidfile(s));
+
+	std::ifstream in(path.string());
+	std::string content((std::istreambuf_iterator<char>(in)), std::istreambuf_iterator<char>());
+	EXPECT_EQ(content, std::to_string(getpid()) + "\n");
```
leogr, line 63:
```suggestion
	std::string content;
	{
		std::ifstream in(path.string());
		content.assign((std::istreambuf_iterator<char>(in)), std::istreambuf_iterator<char>());
	}
	EXPECT_EQ(content, std::to_string(getpid()) + "\n");
```

This is what turns win32 red 👇

```
[ RUN      ] ActionPidfile.writes_pid_to_file
unknown file: error: C++ exception with description "remove: The process cannot access the file
because it is being used by another process.: "C:\Users\RUNNER~1\AppData\Local\Temp\falco_test_pidfile.pid""
thrown in the test body.
[  FAILED  ] ActionPidfile.writes_pid_to_file (2 ms)
```

`in` is still open when `std::filesystem::remove()` runs below. POSIX lets you unlink an open file, Windows does not, so `remove()` hits a sharing violation and its throwing overload throws. That is also why only win32 fails, since Linux, macOS and wasm are all green.

The suggestion scopes the read the same way the prefill above already does. It is `clang-format` clean, and I checked the equivalent on Linux, but I have no Windows box, so may you confirm the job goes green? 🙏

unit_tests/falco/app/actions/test_pidfile.cpp
leogr, line 66:
```suggestion
}

#if defined(GTEST_HAS_DEATH_TEST) && GTEST_HAS_DEATH_TEST && !defined(_WIN32)
TEST(ActionPidfileDeathTest, refuses_symlinked_path) {
	auto dir = std::filesystem::temp_directory_path();
	auto target = dir / ("falco_test_symlink_target_" + std::to_string(getpid()) + ".pid");
	auto link = dir / ("falco_test_symlink_" + std::to_string(getpid()) + ".pid");
	std::filesystem::remove(link);
	std::filesystem::remove(target);
	std::filesystem::create_symlink(target, link);

	falco::app::state s;
	s.options.pidfilename = link.string();
	EXPECT_EXIT(falco::app::actions::pidfile(s), ::testing::ExitedWithCode(255), "");
	EXPECT_FALSE(std::filesystem::exists(target));

	std::filesystem::remove(link);
}
#endif
```

Sorry, I steered you wrong in my previous review 😅

I said the symlink case cannot live here because `pidfile()` calls `exit(-1)`. That holds for `EXPECT_ACTION_FAIL`, but I forgot about gtest death tests, which exist precisely for this. So the regression test @sathvik-shettyy asked for can stay in-repo and gate this PR, and you can drop the `falcosecurity/testing` follow-up entirely. That is strictly better, since a test in that repo is pinned at `@main` and could never gate a Falco PR anyway.

I prototyped it outside the Falco tree, against a stub with the same `open()` flags and the same `exit(-1)`:

- with `O_NOFOLLOW` 👉 passes
- without it 👉 `Result: failed to die`, plus `std::filesystem::exists(target)` turning `true`

So it does catch the regression, on both assertions. It is also `clang-format` clean with our `.clang-format`.

Two things I could not check, since I did not build Falco itself:

- we vendor gtest `v1.14.0` and I tested against `1.17.0`. `EXPECT_EXIT`, `ExitedWithCode` and `GTEST_HAS_DEATH_TEST` are unchanged between the two, so I believe it is fine
- `falco::app::state` builds a `sinsp`. If that leaves threads around, gtest prints a death-test warning (not a failure). The `#if` guard already covers wasm, where death tests are not available

May you give it a run before pushing? 🙏

leogr (commented):
Thanks for the follow-up 🙏

Both my previous points are addressed, and dropping the flag was the right call.

However, win32 is red, and it is this PR. `build-dev-packages-x86_64 / build-win32-package` was green on `a2b80dfe` and fails on `bef3089d`, in the new test 👉 https://github.com/falcosecurity/falco/actions/runs/32602923944/job/99436758207

The build is fine, it is the `Run unit Tests` step. Diagnosis and a fix in the comment below. Everything else is green (Linux, macOS, wasm, musl, sanitizers, arm64).

Also, I owe you a correction on the testing plan: the symlink case can be covered in-repo after all, so the `falcosecurity/testing` follow-up is not needed 😅

leogr (commented):
Thanks for the follow-up 🙏

Both my previous points are addressed, and dropping the flag was the right call.

However, win32 is red, and it is this PR. `build-dev-packages-x86_64 / build-win32-package` was green on `a2b80dfe` and fails on `bef3089d`, in the new test 👉 https://github.com/falcosecurity/falco/actions/runs/32602923944/job/99436758207

The build is fine, it is the `Run unit Tests` step. Diagnosis and a fix in the comment below. Everything else is green (Linux, macOS, wasm, musl, sanitizers, arm64).

Also, I owe you a correction on the testing plan: the symlink case can be covered in-repo after all, so the `falcosecurity/testing` follow-up is not needed 😅

sgaist:
@alexmchughdev gentle ping :-)

Would you have time to finalize this for the next Falco release ?

leogr:
Hey @alexmchughdev, since we'd like to ship this with `0.45.0`, I took the liberty of applying the two pending suggestions (the `ifstream` scope for win32 and the death test for the symlink path) and force-pushed to your branch, keeping your authorship. I also updated the release note to mention the trailing newline.

Hope that's fine with you 🙏

leogr (approved):
cc @sgaist 
PTAL again 🙏

sgaist (approved):

PR #3899, opened 2026-05-24
fix(userspace): Escape output controls

### LINKED ISSUE #3895: Log injection via unencoded control characters in event field values
**Describe the bug**

Falco does not encode control characters (newlines, ANSI escape sequences) when emitting event field values - such as file paths or process arguments - to text-based outputs (file, syslog, stdout).

An attacker who can influence those field values, for example, by creating a file with `\n` or ANSI escape sequences in its name, can inject content into Falco's output stream.

Concretely, with a rule like Write below etc, an attacker can:
  - Embed a `\n` in a filename to split a single Falco alert across multiple log lines, fabricating fake alert entries surrounding the real one.
  - Pass ANSI escape sequences (e.g. `\x1b[...`) verbatim to downstream log viewers and terminals.

The original alert is still emitted - this is not a silent detection bypass - but the surrounding output stream can be poisoned, and downstream consumers that do not strip control characters may be misled.

**How to reproduce it**

On a host where Falco is running with a rule that includes file paths in its output (e.g. `Write below etc`):

```
# Inject a newline into a filename
touch "/etc/$(printf 'evt\nfake_alert')"
```

The Falco alert line will be split, with the second half appearing as if it were a separate log entry.

**Environment**

- Falco version: 0.43.1

- System info:

- OS:

- Kernel:

- Installation method:

**Additional context**

This is already addressed by the same change that resolved [falcosecurity/libs#2965](https://github.com/falcosecurity/libs/issues/2965). 

The fix lands in Falco 0.44.0.
/milestone 0.44.0

**Credit**

Reported responsibly via the maintainers' private security mailing list by [@raj3shp](https://github.com/raj3shp). Thanks 🙏

mark-liu:
**What type of PR is this?**

/kind bug

/area engine

Escapes control characters in plain text Falco output so event field values cannot inject forged lines or terminal control sequences.

* I kept JSON output on `nlohmann::json::dump()` so structured consumers do not get double-escaped values
* I handled internal plain text messages because they use the same output sinks
* I left the libs formatter path alone because Falco owns the text serialization boundary here

Closes #3895

Assumption: downstream JSON consumers expect decoded `output_fields` values to stay unchanged.

Out of scope: `format_event()` still assumes a non-empty rule output format before indexing `message_format[0]`.

```release-note
Plain text outputs now escape control characters in rendered alert text.

### DIFF
--- a/unit_tests/CMakeLists.txt
+++ b/unit_tests/CMakeLists.txt
@@ -41,6 +41,7 @@ add_executable(
 	engine/test_filter_details_resolver.cpp
 	engine/test_filter_macro_resolver.cpp
 	engine/test_filter_warning_resolver.cpp
+	engine/test_formats.cpp
 	engine/test_plugin_requirements.cpp
 	engine/test_rule_loader.cpp
 	engine/test_rulesets.cpp
@@ -51,6 +52,7 @@ add_executable(
 	falco/test_configuration_env_vars.cpp
 	falco/test_configuration_output_options.cpp
 	falco/test_configuration_schema.cpp
+	falco/test_outputs.cpp
 	falco/app/actions/test_select_event_sources.cpp
 	falco/app/actions/test_load_config.cpp
 )
--- a/unit_tests/engine/test_formats.cpp
+++ b/unit_tests/engine/test_formats.cpp
@@ -1,0 +1,83 @@
+#include <gtest/gtest.h>
+#include <nlohmann/json.hpp>
+
+#include <engine/formats.h>
+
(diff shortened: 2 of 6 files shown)


ekoops:
Hey @mark-liu . Thank you for this contribution, but as mentioned in https://github.com/falcosecurity/falco/issues/3895#issuecomment-4552829604, the issue was already fixed, and the fix was delivered with Falco 0.44.0.
For this reason, I'm gonna close it. Feel free to reopen if you find out some corner-case still needs to be addressed.

PR #3913, opened 2026-06-06
fix(userspace/engine): backtrack in matches_wildcard

tejgokani:
**What type of PR is this?**

/kind bug

/area engine

`falco::utils::matches_wildcard` located the literal segment that follows a `*` using `std::string::find`, which returns the first occurrence. When that first occurrence left the remaining pattern unable to match the rest of the input, the function returned `false` even though a later occurrence would have produced a successful match.

A few concrete failing inputs (all of which should match but returned `false`):

| Pattern | Input | Old result | Expected |
|---|---|---|---|
| `*ab` | `xabab` | `false` | `true` |
| `*.yaml` | `backup.yaml.yaml` | `false` | `true` |
| `*ab*ab` | `ababab` | `false` | `true` |
| `hello*world` | `hello world hello world` | `false` | `true` |

Trace for `*ab` vs `xabab`: the matcher finds `ab` at index 1, recurses with the empty pattern against the remaining `ab`, and returns `false` because the empty pattern is not equal to `"ab"`. The second `ab` at index 3 (which would have matched against the empty remainder) is never tried.

The call sites that consume this helper are:

- [`userspace/falco/configuration.cpp`](https://github.com/falcosecurity/falco/blob/master/userspace/falco/configuration.cpp) — `*.yaml` / `*.yml` filtering when loading rule files from a directory. Filenames containing the literal more than once (e.g. `backup.yaml.yaml`) were silently skipped.
- [`userspace/engine/indexable_ruleset.h`](https://github.com/falcosecurity/falco/blob/master/userspace/engine/indexable_ruleset.h) — wildcard matching of rule names for `enable`/`disable`. Rules whose names contain the literal segment more than once would not be matched.

The fix replaces the single `find` with a loop that tries every occurrence of the literal segment and backtracks on failure. The other branch (literal prefix before the `*`) is unchanged since there is no ambiguity there.

All existing tests in `unit_tests/engine/test_falco_utils.cpp` continue to pass with the new implementation. New tests are added that cover the previously-broken inputs as well as a couple of negative cases to ensure backtracking still respects a non-matching suffix.

Fixes #

- The recursion depth is bounded by the number of `*` separators in the pattern, identical to the prior implementation. Worst-case work is `O(|s| * occurrences-of-literal)` per recursion level, which is the standard cost of greedy backtracking globbing on short inputs like file names and rule names.
- I do not have a maintainer-grade local toolchain (the project requires `clang-format 18.1.8` and a full libsinsp build to run `BUILD_FALCO_UNIT_TESTS`). The fix was matched to the surrounding style by hand and verified by extracting the function into a standalone harness; please let CI re-run the formatter and full unit-test suite.

```release-note
fix(engine): correctly match wildcard patterns when the literal segment that follows a `*` appears more than once in the input (e.g. `*.yaml` against `backup.yaml.yaml`, or wildcard `enable`/`disable` of rule names whose names repeat a literal segment).
```

### DIFF
--- a/unit_tests/engine/test_falco_utils.cpp
+++ b/unit_tests/engine/test_falco_utils.cpp
@@ -101,4 +101,17 @@ TEST(FalcoUtils, matches_wildcard) {
 	ASSERT_FALSE(falco::utils::matches_wildcard("hello*world", "hello new world yes"));
 	ASSERT_FALSE(falco::utils::matches_wildcard("*hello*world", "come on hello this world yes"));
 	ASSERT_FALSE(falco::utils::matches_wildcard("*hello*world*", "come on hello this yes"));
+
+	// Backtracking: when a wildcard segment is followed by a literal that
+	// occurs more than once in `s`, matching the first occurrence is not
+	// always correct. The matcher must try later occurrences before giving up.
+	ASSERT_TRUE(falco::utils::matches_wildcard("*ab", "xabab"));
+	ASSERT_TRUE(falco::utils::matches_wildcard("*.yaml", "backup.yaml.yaml"));
+	ASSERT_TRUE(falco::utils::matches_wildcard("*foo*bar", "foofoobar"));
+	ASSERT_TRUE(falco::utils::matches_wildcard("*ab*ab", "ababab"));
+	ASSERT_TRUE(falco::utils::matches_wildcard("hello*world", "hello world hello world"));
+
+	// Backtracking must still respect a non-matching suffix.
+	ASSERT_FALSE(falco::utils::matches_wildcard("*ab", "xabcd"));
+	ASSERT_FALSE(falco::utils::matches_wildcard("*foo*bar", "foofoobaz"));
 }
--- a/unit_tests/engine/test_falco_utils.cpp
+++ b/unit_tests/engine/test_falco_utils.cpp
@@ -111,6 +111,13 @@ TEST(FalcoUtils, matches_wildcard) {
 	ASSERT_TRUE(falco::utils::matches_wildcard("*ab*ab", "ababab"));
 	ASSERT_TRUE(falco::utils::matches_wildcard("hello*world", "hello world hello world"));
 
+	// Consecutive wildcards (`**`) collapse to a single `*` because the
+	// leading run of stars is consumed in one step by `find_first_not_of`,
+	// so backtracking behaves identically whether the pattern says `*` or `**`.
+	ASSERT_TRUE(falco::utils::matches_wildcard("**ab", "xabab"));
+	ASSERT_TRUE(falco::utils::matches_wildcard("**ab**ab", "ababab"));
+	ASSERT_FALSE(falco::utils::matches_wildcard("**ab", "xabcd"));
+
 	// Backtracking must still respect a non-matching suffix.
 	ASSERT_FALSE(falco::utils::matches_wildcard("*ab", "xabcd"));
 	ASSERT_FALSE(falco::utils::matches_wildcard("*foo*bar", "foofoobaz"));
--- a/userspace/engine/falco_engine_version.h
+++ b/userspace/engine/falco_engine_version.h
@@ -20,7 +20,7 @@ limitations under the License.
 
 // The version of this Falco engine
 #define FALCO_ENGINE_VERSION_MAJOR 0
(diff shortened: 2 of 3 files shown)


tejgokani:
Thanks for the catch. I've pushed [d350844d](https://github.com/falcosecurity/falco/pull/3913/commits/d350844d) bumping `FALCO_ENGINE_VERSION_MINOR` from `62` to `63`.

Per the [Falco Components Versioning section in `RELEASE.md`](https://github.com/falcosecurity/falco/blob/master/RELEASE.md), `FALCO_ENGINE_VERSION` must be bumped when:

1. a backward-incompatible change has been introduced to the rules files syntax and loading logic, and/or
2. `FALCO_ENGINE_CHECKSUM` has changed.

This PR falls under (1): `matches_wildcard` is used by `falco_configuration::read_rules_file_directory` for `*.yaml` / `*.yml` filtering, and by `indexable_ruleset::enable_disable` for wildcard `enable`/`disable` of rule names. The fix makes the matcher strictly more permissive in cases the old implementation missed, which is a behavior change in rule-loading logic.

It does not fall under (2): no rule fields, event types, or driver schema changes are introduced, so `FALCO_ENGINE_CHECKSUM` is left unchanged.

I followed the recent convention of incrementing `MINOR` and keeping `PATCH` at `0` (matching [573871955c](https://github.com/falcosecurity/falco/commit/573871955ce84a420b3cdbfe275ef2e198fb2881) and [59dae06e13](https://github.com/falcosecurity/falco/commit/59dae06e132c0db949881977197a89f9b8789050)). Happy to adjust if you'd prefer a different cadence.

Ready for another look when you have a moment.

tejgokani:
@Kaizhe  and @irozzo-1A  please take a look at this PR and if any changes to be done, I'll me more than happy to resolve it

kunalworldwide (commented):
The backtracking logic looks correct — greedy match on the first occurrence of `to_find`, then recurse on the remainder, advancing `search_from` on failure. This is the standard approach and handles the cases in the new tests well.

A few observations:

**Engine version bump**: Bumping MINOR from 62→63 makes sense since this changes rule-loading behavior (patterns that previously didn't match now will). Worth noting in a release note since existing rules with patterns like `*.yaml` will start matching paths like `backup.yaml.yaml` that they previously missed.

**Edge case — empty `to_find`**: When the pattern has consecutive wildcards like `**`, `to_find` ends up as an empty string. `s.find("", 0)` returns 0, so the `while(true)` loop would recurse with the same `s` minus nothing — but since `remaining_pattern` is `*` (the second wildcard), the recursion should eventually bottom out at the `endsWith("*")` check on line 217. Might be worth adding a test case for `**` to document that it behaves identically to `*`.

**Test coverage**: The 7 new test cases are good — covers both positive backtracking and negative suffix mismatch. Clean.

Looks good to me overall.

tejgokani:
@kunalworldwide thanks for the valuable comments. 

I just wanted to know that it's been few days since my PR is approved but not yet merged. When shall I expect it to be merged.

kunalworldwide:
@tejgokani I'm not a maintainer, just a reviewer — merge decisions are up to the Falco team. You could try pinging in the `#falco` channel on CNCF Slack or adding a `/cc @jasondellaluce` to get maintainer eyes on it. The PR looks ready from a code perspective.

tejgokani:
@poiana  could you please re evaluate this PR.

kunalworldwide (commented):
The backtracking fix looks correct. The previous code used `s.find(to_find)` and returned false if the rest of the pattern did not match from that first occurrence; the new loop tries each occurrence and backtracks, which is the right behavior for `*` wildcards.

A few notes:

1. Please link this PR to the original issue in the PR description (looks like it addresses the wildcard matching bug but the issue number is not filled in).
2. Consider adding an edge-case test for `**` patterns if they are supported by this matcher — consecutive wildcards can interact with the backtracking loop in subtle ways.
3. The worst-case complexity is now O(n²) on the input length, which is acceptable for Falco's short rule patterns but worth noting in a code comment so future maintainers do not try to "optimize" it back to a greedy single pass.

Tests cover the main regression cases well. LGTM once the issue link is added.

sgaist (dismissed):

sgaist (approved):

leogr (approved):
/hold cancel

PR #3918, opened 2026-06-10
fix: load no evttype false positive

### LINKED ISSUE #3883: LOAD_NO_EVTTYPE false positive: rules using statically-unsatisfiable conditions (e.g. the `never_true` idiom)
**Describe the bug**

The static analyzer in `userspace/libsinsp/filter/ppm_codes.cpp` (libs) emits `LOAD_NO_EVTTYPE` on every rule whose top-level condition is statically unsatisfiable. The most prominent case is the `never_true` idiom — 11 rules in `falcosecurity/rules` itself use this pattern as a soft-deprecation / placeholder marker, and every one of them trips the warning at load time.

The warning is technically correct under the visitor's current rules (any non-`evt.type` leaf over-approximates to `all_codes_set()`), but it's misleading for these rules: they can never match anything at runtime, so warning that they'd "match too many `evt.type` values" is the opposite of what's actually happening.

Filing as a discussion before any PR — happy to write the fix if there's alignment on direction.

**How to reproduce it**

Load any rule with `condition: (never_true)` into a recent Falco:

```yaml
- macro: never_true
  condition: (evt.num=0)

- rule: example_dead_rule
  desc: never fires; uses the never_true placeholder
  condition: (never_true)
  output: "should never see this"
  priority: NOTICE
```

→ Falco emits `LOAD_NO_EVTTYPE: Rule matches too many evt.type values. This has a significant performance penalty.` at load.

Same on every rule in `falcosecurity/rules` whose condition is `(never_true)` (11 cases at time of writing). The `never_true` macro itself is defined in `rules/falco_rules.yaml` with this comment:

```yaml
# This macro `never_true` is used as placeholder for tuning negative
# logical sub-expressions, e.g.
#   - macro: allowed_ssh_hosts
#     condition: (never_true)
# Used with double negation `and not allowed_ssh_hosts` to evaluate to true.
- macro: never_true
  condition: (evt.num=0)
```

**Expected behaviour**

The analyzer should not emit `LOAD_NO_EVTTYPE` on rules whose condition is statically unsatisfiable. `(evt.num=0)` is unsatisfiable — `evt.num` is monotonic and ≥ 1 at runtime — so the rule provably never matches anything. The warning gives the author the opposite signal of what's helpful ("rule too broad" rather than "rule never fires").

A reasonable bar: a leaf-level check for `evt.num = N` where `N ≤ 0` returns the empty `code_set_t{}` instead of `all_codes_set()`. That alone clears the threshold and silences the false positive without affecting any rule whose condition is actually broad.

**Screenshots**

N/A — this is a rule-load warning emitted on stderr / via the load-result API, not a UI issue.

**Environment**

This is a static-analyzer false positive at rule-load time, so most of the environment fields don't really apply — the warning fires regardless of OS / kernel / installation method. Verified against:

- Falco version: 0.43.1 (`docker run --rm falcosecurity/falco:latest falco --version`)
- System info: N/A — reproduces in any environment that loads the rule
- Cloud provider or hardware configuration: N/A
- OS: N/A
- Kernel: N/A
- Installation method: any (verified via the `falcosecurity/falco:latest` Docker image)

**Additional context**

### Why the analyzer can't see it today

In `binary_check_expr::visit` (`userspace/libsinsp/filter/ppm_codes.cpp`), only `evt.type` / `syscall.type` checks get narrowed. Everything else falls through to `all_codes_set()`:

```cpp
void visit(const libsinsp::filter::ast::binary_check_expr* e) override {
    m_last_node_has_codes = false;
    if (is_evttype_operator(e->op)) {
        e->left->accept(this);
        if (m_last_node_is_evttype_field) { ... return; }   // narrow path
    }
    m_last_node_codes = all_codes_set();                    // over-approximate
    ...
}
```

That's the full ppm event-code set — well above the `>100 evt.types` threshold in `rule_loader_compiler.cpp:514`, which is what emits `LOAD_NO_EVTTYPE`. There's no `constant_fold` / `satisfiable` / `dead_rule` analysis anywhere in the pipeline.

The over-approximation is intentional for soundness: the analyzer can't introspect arbitrary field semantics, and `userspace/libsinsp/test/filter_ppm_codes.ut.cpp:168` explicitly locks in `proc.name=nginx → all_events`. So this is a baseline to **extend** safely, not a behavior to undo.

### Precedent

[`falcosecurity/libs@4cd3f967`](https://github.com/falcosecurity/libs/commit/4cd3f967) (Falco 0.42.0 milestone) addressed [#3546](https://github.com/falcosecurity/falco/issues/3546) — same family of `LOAD_NO_EVTTYPE` false positive, same visitor, narrow fix that taught `names_to_event_set` about async-event names from loaded plugins. This would be a similarly narrow fix for the statically-unsatisfiable case.

### Proposed direction (small, scoped)

Detect trivially-unsatisfiable comparisons at the leaf level. Initial scope: `evt.num = N` where `N ≤ 0`. Sketch:

```diff
 if (is_evttype_operator(e->op)) {
     e->left->accept(this);
     if (m_last_node_is_evttype_field) { ... return; }
+
+    // Known-unsatisfiable constant comparisons on monotonic fields.
+    // evt.num is >= 1 at runtime, so `evt.num = N` for N <= 0 never matches.
+    if (e->op == "=" && /* left is field_expr for evt.num */ ...) {
+        try {
+            long long n = std::stoll(/* right value */);
+            if (n <= 0) {
+                m_last_node_codes = code_set_t{};   // empty
+                m_last_node_has_codes = true;
+                try_inversion(m_last_node_codes);
+                return;
+            }
+        } catch (...) { /* fall through */ }
+    }
 }
 m_last_node_codes = all_codes_set();
```

Tests would extend `filter_ppm_codes.ut.cpp`:

```cpp
TEST_CODES(filter_ppm_codes, check_statically_false) {
    T t;
    code_set_t empty;
    ASSERT_FILTER_SET_EQ(t, "evt.num=0", empty);                     // never_true
    ASSERT_FILTER_SET_EQ(t, "evt.num=-5", empty);                    // any non-positive
    ASSERT_FILTER_SET_EQ(t, "evt.type=open and evt.num=0", empty);   // intersected
    ASSERT_FILTER_SET_EQ(t, "evt.num=1", t.all_set());               // positive stays unconstrained
    ASSERT_FILTER_SET_EQ(t, "evt.num=42", t.all_set());
}
```

Total touch ~30 lines including tests. No new abstractions, no satisfiability engine.

### Alternatives considered

- **Special-case the literal macro name `never_true` in the analyzer.** Probably not — couples the static analyzer to a rules-pack convention.
- **A general constant-folding / satisfiability pass.** Architecturally cleaner and probably the right long-term direction, but a much bigger change. Happy to scope toward this if maintainers prefer; the small fix above is meant as a stepping stone.

### Empirical verification

Verified against `falcosecurity/falco:latest` (0.43.1) with 8 cases — `<` / `<=` / `>` operators, list comprehensions, multi-level macro chains, range contradictions, and same-field equality contradictions. All 8 trip `LOAD_NO_EVTTYPE` today. Can share the matrix if useful.

### Open questions for the discussion

1. **Suppress vs. relabel?** Keep `LOAD_NO_EVTTYPE` silent on dead rules — or introduce a new warning code (something like `LOAD_DEAD_RULE`) to communicate "this rule never fires" rather than "this rule is too broad"? The two carry pretty different actions for the rule author. I'd lean toward a new code, but it's a UX call.
2. **Scope of the unsatisfiable detector.** `evt.num = <non-positive>` is the obvious first case. Other fields with known runtime invariants worth a small whitelist?
3. **Negation.** `not (evt.num=0)` should resolve back to `all_codes_set()` — `try_inversion` already does this; want to confirm the path stays correct.
4. **Backward compatibility.** Any tests in `falcosecurity/libs` or `falcosecurity/rules` that depend on the current over-approximating behavior for *unsatisfiable* leaves? I didn't find any.

Happy to take a stab at a PR if maintainers want one, on whichever shape (small fix or larger pass) you'd prefer.

leogr:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind release

> /kind chart-release

> /area build

/area engine

> /area tests

> /area proposals

> /area automation

> /area chart

Falco was emitting `LOAD_NO_EVTTYPE` ("Rule matches too many evt.type values...") for rules whose condition is statically unsatisfiable - notably the documented `never_true` idiom (`evt.num=0`) - which is the opposite of what's actually happening: those rules match *nothing*.

Since falcosecurity/libs#3009, the event-code analyzer resolves such conditions to the empty set. This PR bumps libs to latest master and changes the warning gate to fire only on genuinely broad conditions (`size() > 100`), instead of conflating "matches nothing" with "matches too much".

Fixes #3883

- The libs bump and the gate change must ship together: with the new libs only, the common `... and never_true` tuning pattern would start producing empty sets and trigger the old warning.
- Explicit contradictions (e.g. `evt.type=open and evt.type=close`) now load silently too. This is intentional: a dead-rule diagnostic is a separate UX design (with its own suppressibility), out of scope here.
- Minor side effects of the more precise analysis: `falco -V` reports "Ok" instead of "Ok, with warnings" for placeholder-only rules, and their `falco -L` events list is no longer the full set.

/milestone 0.45.0

```release-note
fix(engine): do not emit the `LOAD_NO_EVTTYPE` warning for rules with statically unsatisfiable conditions (e.g. the `never_true` placeholder idiom)
```


ekoops (approved):
/approve

cmake/modules/falcosecurity-libs.cmake
```diff
@@ -42,9 +42,9 @@ else()
 	# version (or branch, or commit) just pass the variable - ie., `cmake
 	# -DFALCOSECURITY_LIBS_VERSION=dev ..`
 	if(NOT FALCOSECURITY_LIBS_VERSION)
-		set(FALCOSECURITY_LIBS_VERSION "0.25.4")
+		set(FALCOSECURITY_LIBS_VERSION "01b5cf7ffb294bc6605bc745b327e43f8cf55b40")
```
sgaist, line 45: Just wondering: is the semantic version number being dropped or is it just for this change to be functioning ?

sgaist (approved):

cmake/modules/falcosecurity-libs.cmake
leogr: This updates the current libs version, bringing in the libs implementation this PR depends on.
The temporary bump is equivalent to https://github.com/falcosecurity/falco/pull/3923

Later, `FALCOSECURITY_LIBS_VERSION` will be bumped with the correct version once a stable libs release is out.

PR #3930, opened 2026-06-25
ADD introducing the "splitFiles" feature from feature request issue #3929

### LINKED ISSUE #3929: Allow splitting N falco rules files into N configMap objects
**Motivation**

By default in the falco helm chart, if you fill the "customRules" section with datas that you want to be splitted in several files, at the end of the helm template rendering process, only one configmap object is created and it contains all the rulesfiles specified and its content.

For operationnal reason in organization with several security team contributing to the global security of clusters in a GITOPS way, it could be a good idea to provide a way to split the files into distinct configmap objects, mapping 1 confiMap to 1 rulesfile.

As an example, imagine you want to maintain 2 rulesfiles by 2 distinct population using a GITOPS way using argoCD agent:

* a 'generic' rulefile applicable for the whole infrastructure (the cluster components, system namespaces, etc...) and for many applications hosted on the cluster. This rulefile is maintained by "cluster-admins" and stored in a "infrastructure" GIT Repository accessible only by this population. "cluster-admins" people have the full rights (kubernetes RBAC) on the cluster to maintain it.

* a custom one tailored for a specific sensitive application that goes deeper in the detection rules but on targetet namespaces and resource types. This rulefile is maintained by "project-admins" and is stored in another "application" GIT Repository accessible only by this population. "project-admins" are security people/security contacts (let's say "security champions") that are integrated in the application/dev team but who are interacts often with "cluster-admins" and security experts and are in charge of implementing the "secured by design" principles at their application level. Theses people are the one that are better suited to either understand how the application works, and how it should be protected and monitored using security tools like falco.

* a "shared" (or a fleet of at least 2) argoCD agent tracks and remediates both the 'generic' and 'custom' rulesfiles that are stored in git repositories. This could ensure the rules are always up-to-date regarding the source of truth (the git repo), that no attacker can deactive/remove the rules by overriding the configmap content directly on the cluster, and allow easy maintenance of each rules-set separatly by each team "cluster-admins" and "project-admins".

With the current default settings (1 single configmap mounted in /etc/falco/rules.d for N rulesfiles), this scenario is not fully possible: splitting rules in separate files is possible, but at the end of helm install process, all rules files are merged into a single configmap object.
Having 1 single configmap means that tools such as argoCD will have to track one single configmap object that may contain all rulesfiles: so it induced it can be stored in only one Git repo, and then managed by only one team (sploiler alert: "cluster-admins" team with the higher cluster privilege level).

Instead, if N rulesfiles are mapped to N configmap objects, argoCD could easily track the 'generic rulesfiles' on one side (plugged to git repo A), and the 'custom' one on the other side (plugged to git repo B), without any issue. Each rulesfile is stored in a git repo in which only people responsible for its content have accesss, which is recommended (segregation of duty).

**Feature**

Provide a way to split rules files and their content in distinct configmap objects. For compatibility reason, it SHOULD still be possible to gather all rules files in a single configmap object for legacy and compatibility reasons. This could be done using a boolean variable/flag, such as "splitFiles: (true|false)".

**Alternatives**

This feature was implemented locally (fork + modification) but it could be shared wider and benefit to community userbase than i.

**Additional context**

Here are the templates files that may be concerned by modifications, and the code section concerned:

* pod-template.tpl
```yaml
[...]
      volumeMounts:
[...]
        {{- if .Values.customRules }}
        - mountPath: /etc/falco/rules.d
          name: rules-volume
        {{- end }}
[...]
  volumes:
[...]
{{- if .Values.customRules }}
    - name: rules-volume
      configMap:
        name: {{ include "falco.fullname" . }}-rules
    {{- end }}
```

* rules-configmap.yaml:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ include "falco.fullname" . }}-rules
  namespace: {{ include "falco.namespace" . }}
  labels:
    {{- include "falco.labels" . | nindent 4 }}
data:
{{- range $file, $content :=  .Values.customRules }}
  {{ $file }}: |-
{{ $content | indent 4}}
{{- end }}
{{- end }}
```

Here is an example of the `customRules` section of a values.yaml that could be provided to override default one during helm install:

* values.yaml:
```yaml
customRules:
  falco_rules.local.yaml: |-
    - required_engine_version: 15
    - required_plugin_versions:
        - name: k8saudit
          version: 0.16.0
        - name: json
          version: 0.7.3
        - name: container
          version: 0.6.3
    [...]
```

FeatureRequest was initially open in falcosecurity/charts repo as issue https://github.com/falcosecurity/charts/issues/1018 and is now moved to current repo due to contribution policy evolution ([CONTRIBUTING.md](https://github.com/falcosecurity/charts/blob/master/CONTRIBUTING.md).

heitzflorian:
ADD introducing the "splitFiles" feature from feature request issue #3929

Ref: https://github.com/falcosecurity/falco/issues/3929

**What type of PR is this?**

/kind feature

/area chart

This implements the "splitFiles" feature in the chart, allowing to generate 1 configmap per rules files provided in the "customRules" section instead of merging all rules files in a single configMap which can be harder to maintain.

See https://github.com/falcosecurity/falco/issues/3929

Fixes #3929

PullRequest was initially open in falcosecurity/charts repo as https://github.com/falcosecurity/charts/pull/1020 and is now moved to current repo due to contribution policy evolution ([CONTRIBUTING.md](https://github.com/falcosecurity/charts/blob/master/CONTRIBUTING.md).

```release-note
NONE
```

### DIFF
--- a/chart/falco/templates/pod-template.tpl
+++ b/chart/falco/templates/pod-template.tpl
@@ -188,8 +188,18 @@ spec:
           name: falco-yaml
           subPath: falco.yaml
         {{- if .Values.customRules }}
+        {{- if .Values.customRules.splitFiles.enabled }}
+        {{- $root := . -}}
+        {{- range $file, $content :=  .Values.customRules.splitFiles.files }}
+        {{- $baseName := trimSuffix ".yaml" $file | replace "." "-" | replace "_" "-"}}
+        - mountPath: /etc/falco/rules.d/{{ $file }}
+          name: {{ include "falco.fullname" $root }}-{{ $baseName}}-rules-volume
+          subPath: {{ $file }}
+        {{- end }}
+        {{- else }}
         - mountPath: /etc/falco/rules.d
           name: rules-volume
+        {{- end }}
         {{- end }}
         {{- if or .Values.certs.existingSecret (and .Values.certs.server.key .Values.certs.server.crt .Values.certs.ca.crt) }}
         - mountPath: /etc/falco/certs
@@ -278,9 +288,19 @@ spec:
         - key: falco.yaml
           path: falco.yaml
     {{- if .Values.customRules }}
+    {{- if .Values.customRules.splitFiles.enabled }}
+    {{- $root := . -}}
+    {{- range $file, $content :=  .Values.customRules.splitFiles.files }}
+    {{- $baseName := trimSuffix ".yaml" $file | replace "." "-" | replace "_" "-"}}
+    - name: {{ include "falco.fullname" $root }}-{{ $baseName}}-rules-volume
+      configMap:
+        name: {{ include "falco.fullname" $root }}-{{ $baseName}}-rules
+    {{- end }}
+    {{- else }}
     - name: rules-volume
       configMap:
         name: {{ include "falco.fullname" . }}-rules
+    {{- end }}
     {{- end }}
     {{- if or .Values.certs.existingSecret (and .Values.certs.server.key .Values.certs.server.crt .Values.certs.ca.crt) }}
     - name: certs-volume
--- a/chart/falco/templates/rules-configmap.yaml
+++ b/chart/falco/templates/rules-configmap.yaml
@@ -1,4 +1,21 @@
 {{- if .Values.customRules }}
+{{- if .Values.customRules.splitFiles.enabled }}
+{{- $root := . -}}
+{{- range $file, $content := .Values.customRules.splitFiles.files }}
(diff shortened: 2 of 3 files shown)


leogr:
Hey @heitzflorian

Thanks for this PR 🙏

We are about to freeze for Falco 0.45.0, and this change needs a bit more discussion first: it changes how existing `customRules` entries are rendered, so we need a path that does not break current users. So I'm moving it to the 0.46.0 milestone, and we will get back to it right after the release.

/milestone 0.46.0

PR #3931, opened 2026-06-28
fix(chart): add revisionHistoryLimit support to DaemonSet controller

### LINKED ISSUE #3925: [falco helm-chart] revisionHistoryLimit not properly wired in deployment.yaml and completely absent in daemonset.yaml
Motivation

The deployment.yaml template in the falco chart references .Values.controller.deployment.revisionHistoryLimit without a with/if guard:

```yaml
revisionHistoryLimit: {{ .Values.controller.deployment.revisionHistoryLimit }}
```

But the corresponding value in values.yaml is commented out by default:

```yaml
# revisionHistoryLimit: 1
```

This means the field is rendered as revisionHistoryLimit: with a null value by default, which Kubernetes interprets as unset and replaces with its own default of 10. The option appears configurable but has no practical effect unless the user manually uncomments it.

The daemonset.yaml file contains no reference to revisionHistoryLimit at all — neither in the template nor in values.yaml. Since daemonset is the default value of controller.kind, most users currently have no way to configure this field whatsoever.

Since Kubernetes always sets revisionHistoryLimit to 10 when it's absent from the applied manifest, GitOps tools like Argo CD report a permanent, unfixable config drift between the rendered Helm manifest and the live object, with no way to either explicitly align it or override it.

Proposed feature

Fix deployment.yaml to use a guarded reference, following the same pattern already used for controller.daemonset.updateStrategy:

```yaml
{{- with .Values.controller.deployment.revisionHistoryLimit }}
revisionHistoryLimit: {{ . }}
{{- end }}
```

Add the equivalent block in daemonset.yaml:

```yaml
{{- with .Values.controller.daemonset.revisionHistoryLimit }}
revisionHistoryLimit: {{ . }}
{{- end }}
```

Alternatives

On the Argo CD side, use ignoreDifferences on spec.revisionHistoryLimit to suppress the drift reporting. This avoids the noise in the UI but doesn't actually let users configure the value, and requires every consumer of the chart to reproduce this workaround in their own Application manifest.

Additional context

N/A

Goutham-Annem:
## Summary

Allow DaemonSet users to configure `controller.daemonset.revisionHistoryLimit`.

- Honor an explicit `0` for both DaemonSet and Deployment controllers.
- Omit the field when unset or `null`, allowing Kubernetes to apply its default.
- Document both optional values and regenerate the chart README.

Fixes #3925

## Test plan

- [x] Regression tests cover unset, `null`, `0`, and a positive value for both controllers. The two zero cases failed before the fix.
- [x] All chart unit tests pass with `go test ./unit/... -count=1`.
- [x] `helm lint` passes for both controller kinds.
- [x] `make chart-docs-check` passes with `helm-docs` v1.11.0.
- [x] `git diff --check` passes.

```release-note
fix(chart): support DaemonSet revision history limits and honor an explicit zero for both DaemonSet and Deployment controllers.
```

### DIFF
--- a/chart/falco/CHANGELOG.md
+++ b/chart/falco/CHANGELOG.md
@@ -5,6 +5,8 @@ numbering uses [semantic versioning](http://semver.org).
 
 ## Unreleased
 
+* Add `revisionHistoryLimit` support to the DaemonSet controller, matching existing deployment behavior
+
 ## v9.0.0
 
 * Drop gRPC output and server support
--- a/chart/falco/CHANGELOG.md
+++ b/chart/falco/CHANGELOG.md
@@ -6,6 +6,7 @@ numbering uses [semantic versioning](http://semver.org).
 ## Unreleased
 
 * Fix Grafana dashboard priority level mappings to match Falco's numeric encoding (0=emergency … 7=debug)
+* Add `revisionHistoryLimit` support to the DaemonSet controller, matching existing deployment behavior
 
 ## v9.0.0
 
--- a/chart/falco/CHANGELOG.md
+++ b/chart/falco/CHANGELOG.md
@@ -6,7 +6,7 @@ numbering uses [semantic versioning](http://semver.org).
 ## Unreleased
 
 * Fix Grafana dashboard priority level mappings to match Falco's numeric encoding (0=emergency … 7=debug)
-* Add `revisionHistoryLimit` support to the DaemonSet controller, matching existing deployment behavior
+- Add `revisionHistoryLimit` support to the DaemonSet controller and honor an explicit zero for both DaemonSet and Deployment controllers
 
 ## v9.0.0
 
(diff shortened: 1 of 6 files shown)


Goutham-Annem:
Hi, just bumping this for visibility. Happy to address any feedback or questions!

Goutham-Annem:
Gentle ping — CI is passing, happy to address any feedback.

Goutham-Annem:
Gentle ping — CI is passing and there are no conflicts. Happy to address any feedback!

Goutham-Annem:
Friendly ping — CI is passing and this PR is ready for review. Would appreciate a look when you have a moment!

leogr:
/close

sgaist (approved):

leogr (approved):
/milestone 0.45.0

PR #3935, opened 2026-07-02
test(engine): cover rejection of empty appended rules

Jvlegod:
**What type of PR is this?**

/kind bug

/area engine

/area tests

This PR adds a unit test covering the rejection of `append: true` rules that do not actually append anything (no `condition` and no `exceptions`).

The rule loader collector already rejects them with `LOAD_ERR_VALIDATE` ("Appended rule must have exceptions or condition property"), but the case had no test coverage. The commented-out reader-level check is left as is, since restoring it would only change the reported error code to `LOAD_ERR_YAML_VALIDATE`, which is not what `falcosecurity/testing` expects.

N.B. Scope reduced to the test on September 7th by the maintainers, as agreed in the review (see the comments below).

N/A

Tested with:

```text
cmake -S . -B build-agent -DMINIMAL_BUILD=On -DBUILD_DRIVER=Off -DBUILD_FALCO_UNIT_TESTS=On
cmake --build build-agent --target falco_unit_tests
sudo ./build-agent/unit_tests/falco_unit_tests
```

**Does this PR introduce a user-facing change?:**

```release-note
NONE
```

### DIFF
--- a/unit_tests/engine/test_rule_loader.cpp
+++ b/unit_tests/engine/test_rule_loader.cpp
@@ -872,8 +872,7 @@ TEST_F(test_falco_engine, no_evttype_warning_for_never_true_macro) {
 	EXPECT_EQ(num_rules_for_ruleset(), 2);
 }
 
-// todo!: Probably we shouldn't allow this syntax
-TEST_F(test_falco_engine, rule_enabled_is_ignored_by_append) {
+TEST_F(test_falco_engine, rule_enabled_is_rejected_by_append) {
 	std::string rules_content = R"END(
 - rule: test_rule
   desc: test rule description
@@ -883,16 +882,30 @@ TEST_F(test_falco_engine, rule_enabled_is_ignored_by_append) {
   enabled: false
 
 - rule: test_rule
-  condition: and proc.name = cat
   append: true
   enabled: true
 )END";
 
-	// 'enabled' is ignored by the append, this syntax is not supported
-	// so the rule is not enabled.
-	ASSERT_TRUE(load_rules(rules_content, "rules.yaml"));
+	ASSERT_FALSE(load_rules(rules_content, "rules.yaml"));
 	ASSERT_VALIDATION_STATUS(yaml_helper::validation_ok) << m_load_result->schema_validation();
-	EXPECT_EQ(num_rules_for_ruleset(), 0);
+	ASSERT_TRUE(check_error_message("Appended rule must have exceptions or condition property"));
+}
+
+TEST_F(test_falco_engine, empty_append_rule_is_rejected) {
+	std::string rules_content = R"END(
+- rule: test_rule
+  desc: test rule description
+  condition: evt.type = close
+  output: user=%user.name command=%proc.cmdline file=%fd.name
+  priority: INFO
+
+- rule: test_rule
+  append: true
+)END";
+
+	ASSERT_FALSE(load_rules(rules_content, "rules.yaml"));
+	ASSERT_VALIDATION_STATUS(yaml_helper::validation_ok) << m_load_result->schema_validation();
+	ASSERT_TRUE(check_error_message("Appended rule must have exceptions or condition property"));
 }
 
 // todo!: Probably we shouldn't allow this syntax
--- a/unit_tests/engine/test_rule_loader.cpp
+++ b/unit_tests/engine/test_rule_loader.cpp
@@ -895,6 +895,23 @@ TEST_F(test_falco_engine, rule_enabled_is_ignored_by_append) {
 	EXPECT_EQ(num_rules_for_ruleset(), 0);
 }
 
+TEST_F(test_falco_engine, empty_append_rule_is_rejected) {
+	std::string rules_content = R"END(
+- rule: test_rule
+  desc: test rule description
+  condition: evt.type = close
+  output: user=%user.name command=%proc.cmdline file=%fd.name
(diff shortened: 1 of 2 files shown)


Jvlegod:
Maybe I don't think `FALCO_ENGINE_VERSION` needs to be bumped.

leogr:
> 
> 
> /hold

False positive
/hold cancel

sgaist (dismissed):

userspace/engine/rule_loader_reader.cpp
```diff
@@ -873,10 +873,9 @@ void rule_loader::reader::read_item(rule_loader::configuration& cfg,
 				read_rule_exceptions(cfg, item, v.exceptions, ctx, true);
 			}
 
-			// TODO restore this error and update testing
-			// THROW((!v.cond.has_value() && !v.exceptions.has_value()),
-			//       "Appended rule must have exceptions or condition property",
-			//       v.ctx);
+			THROW((!v.cond.has_value() && !v.exceptions.has_value()),
+			      "Appended rule must have exceptions or condition property",
+			      v.ctx);
```
leogr, line 878: This is where the error code changes: the `THROW` macro in this file uses `LOAD_ERR_YAML_VALIDATE`, while the collector already throws the same message with `LOAD_ERR_VALIDATE` for exactly this case (`rule_loader_collector.cpp`, `has_any_value()` check in `append()`).

May you drop this hunk, leaving the file as it is on `master`? Otherwise `TestFalco_Legacy_InvalidAppendRuleWithoutCondition` in falcosecurity/testing keeps failing, since it asserts `LOAD_ERR_VALIDATE`.

unit_tests/engine/test_rule_loader.cpp
```diff
@@ -872,8 +872,7 @@ TEST_F(test_falco_engine, no_evttype_warning_for_never_true_macro) {
 	EXPECT_EQ(num_rules_for_ruleset(), 2);
 }
 
-// todo!: Probably we shouldn't allow this syntax
-TEST_F(test_falco_engine, rule_enabled_is_ignored_by_append) {
+TEST_F(test_falco_engine, rule_enabled_is_rejected_by_append) {
```
leogr, line 875: Nit. The new name says that `enabled` is rejected, but what gets rejected here is the missing `condition`/`exceptions` (an append with a `condition` and `enabled: true` still loads fine). This makes it nearly identical to `empty_append_rule_is_rejected` below.

May you restore `rule_enabled_is_ignored_by_append` as it is on `master` (it's the only test documenting that `enabled` is ignored on append), and keep `empty_append_rule_is_rejected` as the new coverage?

leogr (commented):
Hey @Jvlegod

Thanks for this PR, and sorry for the long wait!

I had a closer look at the red `test-packages` jobs. Apart from the 11 UTF-8 tests (unrelated, they fail only because the merge commit predates #3942), there is one real failure caused by this change: `TestFalco_Legacy_InvalidAppendRuleWithoutCondition` in falcosecurity/testing expects `LOAD_ERR_VALIDATE`, while the `THROW` in the reader emits `LOAD_ERR_YAML_VALIDATE`.

For the record, `master` already rejects an appended rule without `condition`/`exceptions`: in the `append` branch the reader reads only those two properties, so the collector's `has_any_value()` check fails and it throws the very same message (see `rule_loader_collector.cpp`), with `LOAD_ERR_VALIDATE`. So re-enabling the check in the reader does not change what is accepted or rejected. It only changes the error code, and the order in which two errors are reported (an append to an unknown rule with no condition would now report "Appended rule must have..." instead of "no rule by that name already exists").

That said, the new unit test is valuable, since we had no coverage for this case. So, may you drop the `rule_loader_reader.cpp` change and keep the tests? They should pass on `master` as is, since `check_error_message()` matches the message text. See also my inline comment about the renamed test.

This way we also avoid a cross-repo dance with falcosecurity/testing, whose `main` is used by the CI of every Falco PR.

Thanks 🙏

Jvlegod:
Of course, I will submit the changes shortly.

leogr:
> Of course, I will submit the changes shortly.

Thanks, let us know! :pray:

leogr:
Hey @Jvlegod

Since we'd like to have this in Falco 0.45.0 and the code freeze is close, I went ahead and applied the change we agreed on Sep 4 (see https://github.com/falcosecurity/falco/pull/3935#pullrequestreview-5113447564), and force-pushed to your branch. You are still the author of the commit, I just added my sign-off next to yours. I hope you don't mind 🙏

What changed, in short:

- dropped the `rule_loader_reader.cpp` hunk, since the collector already rejects an appended rule without `condition`/`exceptions` with the same message and with `LOAD_ERR_VALIDATE`, which is what falcosecurity/testing expects
- kept your new `empty_append_rule_is_rejected` test as is
- restored `rule_enabled_is_ignored_by_append` exactly as it is on `master`
- rebased on the current `master`, so the unrelated UTF-8 test failures go away too
- reworded the commit message to `test(engine): cover rejection of empty appended rules`, since the PR is now test-only, and updated the PR title and description accordingly

The new test is the valuable part here, since we had no coverage for this case. Thanks again for the contribution!

N.B. The force-push dropped the `lgtm`, so the PR needs a fresh round of approvals. Nothing needed on your side.

/kind test
/remove-kind bug

sgaist (approved):

leogr (approved):

PR #3936, opened 2026-07-02
fix(engine): suppress source label for generic evt field class in list_fields [1/2]

### LINKED ISSUE #3714: `falco --list --markdown` sets `syscall` as event source for generic fields
**Describe the bug**

`falco --list --markdown` reports `Event Sources: syscall` for fields applicable to any event type.

**How to reproduce it**

Install the latest Falco release (e.g.: 0.42.0-rc3) and run `falco --list --markdown`.
The output will begin with:
```
## Field Class: evt

These fields can be used for all event types

Event Sources: syscall

Name | Type | Description
:----|:-----|:-----------
`evt.num` | UINT64 | event number.
`evt.time` | CHARBUF | event timestamp as a time string that includes the nanosecond part.
`evt.time.s` | CHARBUF | event timestamp as a time string with no nanoseconds.
`evt.time.iso8601` | CHARBUF | event timestamp in ISO 8601 format, including nanoseconds and time zone offset (in UTC).
`evt.datetime` | CHARBUF | event timestamp as a time string that includes the date.
`evt.datetime.s` | CHARBUF | event timestamp as a datetime string with no nanoseconds.
...
```

**Expected behaviour**

The output should highlight that the subsequent fields are not specific to any event source.

**Screenshots**

**Environment**

- Falco version:

- System info:

- OS:

- Kernel:

- Installation method:

**Additional context**

Wahid7852:
What type of PR is this?

/kind bug

Any specific area of the project related to this PR?

/area engine
/area tests

What this PR does / why we need it:

falco --list --markdown was printing Event Sources: syscall under ## Field Class: evt even though those fields apply to all event sources. list_fields() in falco_engine.cpp collects source names per field class and passes that set to as_markdown(), which emits the line whenever it's non-empty. Since sinsp_filter_check_gen_event is registered in every source, evt.* always picked up at least {"syscall"}.

With is_generic from falcosecurity/libs#3044, the first pass in list_fields() now skips srcs.insert(it.name) for generic classes. The map entry still gets created so the second pass doesn't miss the class, but the set stays empty and as_markdown({}) suppresses the line. Test added in unit_tests/engine/test_list_fields.cpp covering both cases.

Which issue(s) this PR fixes:

Fixes #3714

Special notes for your reviewer:

Depends on falcosecurity/libs#3050.

```release-note
fix(engine): `falco --list` no longer incorrectly labels the `evt` field class with a specific event source
```

### DIFF
--- a/userspace/engine/falco_engine.cpp
+++ b/userspace/engine/falco_engine.cpp
@@ -124,7 +124,10 @@ void falco_engine::list_fields(const std::string &source,
 		}
 
 		for(const auto &fld_class : it.filter_factory->get_fields()) {
-			fieldclass_event_sources[fieldclass_key(fld_class)].insert(it.name);
+			auto &srcs = fieldclass_event_sources[fieldclass_key(fld_class)];
+			if(!fld_class.is_generic) {
+				srcs.insert(it.name);
+			}
 		}
 	}
 
--- a/userspace/engine/falco_engine.cpp
+++ b/userspace/engine/falco_engine.cpp
@@ -113,7 +113,8 @@ void falco_engine::list_fields(const std::string &source,
                                bool names_only,
                                output_format format) const {
 	// Maps from field class name + short desc to list of event
-	// sources for which this field class can be used.
+	// sources for which this field class can be used. Generic field
+	// classes are mapped to an empty set (see below).
 	std::map<std::string, std::set<std::string>> fieldclass_event_sources;
 
 	// Do a first pass to group together classes that are
@@ -124,7 +125,15 @@ void falco_engine::list_fields(const std::string &source,
 		}
 
 		for(const auto &fld_class : it.filter_factory->get_fields()) {
-			fieldclass_event_sources[fieldclass_key(fld_class)].insert(it.name);
+			// Generic field classes (e.g. evt.*) apply to every event
+			// source, so they must not be labeled with the specific
+			// sources they happen to be registered in. The map entry is
+			// created anyway, but its set is left empty: the formatters
+			// omit the "Event Sources" line when the set is empty.
+			auto &srcs = fieldclass_event_sources[fieldclass_key(fld_class)];
+			if(!fld_class.is_generic) {
+				srcs.insert(it.name);
+			}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


userspace/engine/falco_engine.cpp
```diff
@@ -124,7 +124,10 @@ void falco_engine::list_fields(const std::string &source,
 		}
 
 		for(const auto &fld_class : it.filter_factory->get_fields()) {
-			fieldclass_event_sources[fieldclass_key(fld_class)].insert(it.name);
+			auto &srcs = fieldclass_event_sources[fieldclass_key(fld_class)];
+			if(!fld_class.is_generic) {
```
sgaist, line 128: Nitpick: I would add a comment to ensure the knowledge stays within the code about filtering out things like `evt.*`.

sgaist (changes requested):
Just missing some comments otherwise LGTM.

leogr:
Hey @Wahid7852, thanks for this fix (and for the libs side too) 🙏

Since we want it in Falco 0.45.0 and the libs dependency is now satisfied (falcosecurity/libs#3050 is included in the libs commit `master` pins since #3957), I took the liberty to push directly on your branch to get it merged in time (thanks for leaving "allow edits from maintainers" on). Here is what changed 👇

- rebased on `master` and squashed the commits into one (you are still the author, I just added my sign-off)
- added the code comment @sgaist asked for, so the reasoning about generic field classes stays in the code
- the markdown test now finds the `evt` sections by header plus description, since two field classes are named `evt` (the generic one and the syscall-only one), so the test no longer depends on the registration order
- added a test for the syscall-only `evt` class and one for the JSON output, to be sure the fix is driven by the generic flag and not by the class name
- reordered the test fixture members so that the engine is destroyed before the inspector and the filter check list it references
- kept the alphabetical order in `unit_tests/CMakeLists.txt`

Everything else is your work. Thanks again 🙏

N.B. The hold was the automatic reminder about the engine version. No bump is needed here: the change only affects how `--list` prints the field classes, it does not alter the available fields, the event types, or the rules syntax.

@sgaist I addressed your comment while taking over, may you re-review when you have a minute? 🙏

/unhold

leogr (approved):
LGTM 👍

The fix is straightforward: generic field classes get an empty source set, and the formatters already skip the `Event Sources` line (and the `event_sources` JSON key) when the set is empty. Built and ran the unit tests locally against the libs commit `master` pins, all green.

Thanks @Wahid7852 and @sgaist 🙏

sgaist (approved):

PR #3937, opened 2026-07-03
fix(engine): add empty string guards in logger

archy-rock3t-cloud:
**What type of PR is this?**

/kind bug

/area engine

Guard the `copy.back()` calls in `falco_logger::log` (`userspace/engine/logger.cpp`) with empty-string checks. 
`std::string::back()` on an empty string is undefined behavior, but the syslog trailing-newline strip and the stderr trailing-newline append both call `.back()` without first checking the string is non-empty.

`if(!copy.empty() && copy.back() == '\n')` on the syslog path and `if(copy.empty() || copy.back() != '\n')` on the stderr path.

```release-note
NONE
```

### DIFF
--- a/userspace/engine/logger.cpp
+++ b/userspace/engine/logger.cpp
@@ -101,6 +101,10 @@ void falco_logger::log(falco_logger::level priority, const std::string&& msg) {
 		return;
 	}
 
+	if(msg.find_first_not_of('\n') == std::string::npos) {
+		return;
+	}
+
 	std::string copy = msg;
 
 #ifndef _WIN32


ekoops (commented):
Hey, thank you for the contribution. I'm wondering what's the point of writing an empty log line 🤔 . Maybe we should early-return if the user passes an empty string. Is there any usage of this API where the user passes an empty string?

kunalworldwide (commented):
The fix is minimal and directly addresses the undefined behavior. Calling `std::string::back()` on an empty string is UB, and both the syslog trailing-newline strip and the stderr append paths were vulnerable.

The guards are correctly placed:
- Syslog: strip only if non-empty and ends with `\n`
- Stderr: append `\n` only if empty or does not end with `\n`

The change is small enough that a dedicated unit test might be overkill, but if there's an existing logger test, adding an empty-message case would prevent regression.

LGTM.

archy-rock3t-cloud:
@ekoops there is no reason to log an empty line. It's just guarding the `.back()` UB, not an intended behaviour.

ekoops:
> @ekoops there is no reason to log an empty line. It's just guarding the `.back()` UB, not an intended behaviour.

Yes indeed. What I was suggesting is adding an early-return statement at the top of those logger functions in case the line is empty, instead of repeating the check at each branch. WDYT?

archy-rock3t-cloud:
@ekoops updated, please take a look

sgaist (approved):

leogr (approved):

PR #3939, opened 2026-07-06
fix(userspace/falco): always serve forced restart requests

c2ndev:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind release

> /kind chart-release

> /area build

/area engine

> /area tests

> /area proposals

> /area automation

> /area chart

`restart_handler::start()` returns without spawning the watcher thread when there is nothing to watch (i.e. `watch_config_files: false`). The `SIGHUP` handler is still installed and `s.restarter` is still assigned, so a `SIGHUP`, or any programmatic `restarter->trigger()`, only sets the internal `m_forced` flag, which nothing then consumes. The requested hot restart is silently dropped, and `systemctl reload falco` (whose shipped units run `ExecReload=kill -1 $MAINPID`) reports success while doing nothing.

This behavior came in with #3640, which skipped the watcher thread to avoid consuming an inotify instance when config-files watching is disabled.

This PR keeps that saving — no inotify instance is created when there is nothing to watch — but always starts the watcher thread so forced restart requests are served again. The watcher blocks in `select()` on an eventfd (plus the inotify fd when present) and is woken immediately by `trigger()`, which writes the eventfd; the write is async-signal-safe, so it works from the SIGHUP handler and there is no idle polling. The 100ms timeout is now used only to debounce bursts once a restart is pending. It also retries `select()` on `EINTR` (a signal delivered to the watcher thread would otherwise shut it down) and consumes `m_forced` with an atomic exchange so a request racing the check is not lost.

Unit tests cover: forced restart with nothing to watch, a failed dry-run not triggering a restart (and allowing a retry), prompt teardown with nothing to watch, and the watched-file inotify path.

Fixes #

```release-note
fix: serve SIGHUP-triggered hot restart when `watch_config_files` is disabled
```

### DIFF
--- a/userspace/falco/app/restart_handler.cpp
+++ b/userspace/falco/app/restart_handler.cpp
@@ -20,6 +20,7 @@ limitations under the License.
 #include "logger.h"
 
 #include <string.h>
+#include <errno.h>
 #include <fcntl.h>
 #ifdef _WIN32
 #include <io.h>
@@ -29,6 +30,7 @@ limitations under the License.
 #ifdef __linux__
 #include <sys/inotify.h>
 #include <sys/select.h>
+#include <sys/eventfd.h>
 #endif
 
 #if __GLIBC__ == 2 && __GLIBC_MINOR__ < 30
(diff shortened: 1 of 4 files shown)


ekoops (approved):
/approve

userspace/falco/app/restart_handler.cpp
```diff
@@ -38,48 +40,77 @@ limitations under the License.
...
 }
 
 void falco::app::restart_handler::trigger() {
 	m_forced.store(true, std::memory_order_release);
+#ifdef __linux__
+	// eventfd write is async-signal-safe, so this is safe from the SIGHUP handler
+	if(m_event_fd != -1) {
+		uint64_t v = 1;
+		auto ret = write(m_event_fd, &v, sizeof(v));
+		(void)ret;
```
sgaist, line 64: Out of curiosity why not:
```suggestion
		(void) write(m_event_fd, &v, sizeof(v));
```

userspace/falco/app/restart_handler.cpp
```diff
@@ -91,6 +122,12 @@ bool falco::app::restart_handler::start(std::string& err) {
 void falco::app::restart_handler::stop() {
 #ifdef __linux__
 	m_stop.store(true, std::memory_order_release);
+	// wake the watcher in case it is blocked in select() without a timeout
+	if(m_event_fd != -1) {
+		uint64_t v = 1;
+		auto ret = write(m_event_fd, &v, sizeof(v));
+		(void)ret;
```
sgaist, line 129: Same as above:
```suggestion
		(void) write(m_event_fd, &v, sizeof(v));
```

sgaist (changes requested):
I saw that the same pattern is used elsewhere in the file, so it's nothing blocking. Just want to know if there are specific reason to use it.

c2ndev:
> I saw that the same pattern is used elsewhere in the file, so it's nothing blocking. Just want to know if there are specific reason to use it.

Hi @sgaist, 
no strong reason in particular, the current form is just warning-free, whereas `(void) write(...)` triggers `-Wunused-result` on GCC release builds (Falco sets `-D_FORTIFY_SOURCE=2`; GCC ignores the `(void)` cast, Clang doesn't). Same pattern as the existing `read(...)` calls. Happy to change it if you prefer.

ekoops:
> > I saw that the same pattern is used elsewhere in the file, so it's nothing blocking. Just want to know if there are specific reason to use it.
> 
> Hi @sgaist, no strong reason in particular, the current form is just warning-free, whereas `(void) write(...)` triggers `-Wunused-result` on GCC release builds (Falco sets `-D_FORTIFY_SOURCE=2`; GCC ignores the `(void)` cast, Clang doesn't). Same pattern as the existing `read(...)` calls. Happy to change it if you prefer.

If so, the current one is strictly better right? I would keep it as is, if @sgaist agree :smile:

sgaist:
> > > I saw that the same pattern is used elsewhere in the file, so it's nothing blocking. Just want to know if there are specific reason to use it.
> > 
> > 
> > Hi @sgaist, no strong reason in particular, the current form is just warning-free, whereas `(void) write(...)` triggers `-Wunused-result` on GCC release builds (Falco sets `-D_FORTIFY_SOURCE=2`; GCC ignores the `(void)` cast, Clang doesn't). Same pattern as the existing `read(...)` calls. Happy to change it if you prefer.
> 
> If so, the current one is strictly better right? I would keep it as is, if @sgaist agree 😄

Yes it is.
I suspected there would be something like this involved so all good.

sgaist (approved):

PR #3942, opened 2026-07-13
feat(userspace)!: improve invalid UTF-8 sequence handling

ekoops:
**What type of PR is this?**

> /kind bug

> /kind cleanup

/kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind release

> /kind chart-release

/area build

> /area engine

> /area tests

> /area proposals

> /area automation

> /area chart

This PR syncs Falco with the changes introduced by libs in https://github.com/falcosecurity/libs/pull/3051 .

The PR moves to a more consistent model in which conditions are evaluated on raw/original field values instead of sanitized versions of them (that is, versions where both invalid UTF-8 sequences and non-printable UTF-8 sequences are replaced with the UTF-8 replacement character). This allows to improve the matching accuracy, since specific non-printable and invalid sequences can now be matched using traditional string operators.

An exception to the aforementioned matching model is given by how the `regex` operator now behaves. Each invalid UTF-8 sequence is now replaced by the replacement character before entering the regex engine: this is due to the fact that the current regex engine doesn't support matching on invalid bytes, and would return false on each field containing them, completely defeating the purpose of the `regex` operator. From the end user point of view, this means that invalid sequences can now be matched by directly trying to match the replacement character (e.g.: `foo regex "bar�baz" or `foo regex "bar.baz"`).

As a side note, independent from this initiative, take into account that the other string operators operate at byte level, so `foo glob "?"` matches a single (even invalid) byte, and will not match any multibyte character.

Single invalid bytes are now matchable leveraging the new `\xHH` escape sequence. As an example, a string containing an invalid `0xFF` byte can be matched by writing a condition like `foo contains "bar\xFFbaz"`. Notice that this escape sequence cannot be used in regular expressions provided to the `regex` operator. Moreover, there's a limitation on the specific `\x00` escape sequence, representing the NUL byte: this can only be used with byte buffer fields to match NUL bytes. Previously, the only way to match a NUL byte in a byte buffer field was using specialized operators like `bcontains` and `bstartswith`. The usage of any `\xHH` escape sequence in the RHS operands of `bcontains`/`bstartswith` operators is discouraged, as these latter ones already accept a sequence of hexadecimal digits as RHS, and a condition like `foo bcontains "\x0a\x0b"` is not the same as `foo bcontains "0a0b"`: the first fails to compile, as bytes `0x0a` and `0x0b` do not correspond to any valid hexadecimal digit in ASCII encoding, while the second correctly expresses the idea of matching against something containing the `0x0a 0x0b` byte sequence.

Raw bytes coming from filterchecks are now sanitized at the end of the alert creation pipeline, in `userspace/engine/formats.cpp`, instead of being sanitized in `falcosecurity/libs`. Plain text output is now encoded leveraging the same technique used for JSON output encoding: non-printable characters are escaped (e.g.: the new line character is escaped as a literal '\' followed by a literal 'n', the NUL character is escaped as "\^@") instead of being replaced by the UTF-8 replacement character, and invalid UTF-8 sequences are replaced by the replacement character.

The rule "validate" (`--validate`) and "describe" (`-L`) command outputs are now consistent with the aforementioned output model.

Fixes #

/milestone 0.45.0

```release-note
feat(userspace)!: evaluate rule conditions on raw field bytes (invalid and non-printable UTF-8 sequences are now matchable with string operators, single bytes with the new `\xHH` escape sequence), let `regex` operate on sanitized input, and escape non-printable characters / replace invalid UTF-8 sequences at output encoding time; `FALCO_ENGINE_VERSION` bumped from `0.62.0` to `0.63.0`; rules that matched the replacement character with non-regex operators must be reviewed
```

### DIFF
--- a/cmake/modules/driver.cmake
+++ b/cmake/modules/driver.cmake
@@ -35,9 +35,9 @@ else()
 	# FALCOSECURITY_LIBS_VERSION. In case you want to test against another driver version (or
 	# branch, or commit) just pass the variable - ie., `cmake -DDRIVER_VERSION=dev ..`
 	if(NOT DRIVER_VERSION)
-		set(DRIVER_VERSION "10.2.0+driver")
+		set(DRIVER_VERSION "6fbc055dd53eff5ce3ad79e96cb5b21252ad0090")
 		set(DRIVER_CHECKSUM
-			"SHA256=0e585f5fc2b76696ef2cb902f0901ea39a2f2df87e1f091f3348f968b9085f39"
+			"SHA256=5521a80af7ce1105ea23fc14351f7c486671aa7aa4fffda5b8e5f0c67a7a1a14"
 		)
 	endif()
 
--- a/cmake/modules/falcosecurity-libs.cmake
+++ b/cmake/modules/falcosecurity-libs.cmake
@@ -42,9 +42,9 @@ else()
 	# version (or branch, or commit) just pass the variable - ie., `cmake
 	# -DFALCOSECURITY_LIBS_VERSION=dev ..`
 	if(NOT FALCOSECURITY_LIBS_VERSION)
-		set(FALCOSECURITY_LIBS_VERSION "01b5cf7ffb294bc6605bc745b327e43f8cf55b40")
+		set(FALCOSECURITY_LIBS_VERSION "6fbc055dd53eff5ce3ad79e96cb5b21252ad0090")
 		set(FALCOSECURITY_LIBS_CHECKSUM
-			"SHA256=ad190a3bb3f0da29e7cf40e212930e955b2b364a556d16c134049ab3f7b4e7e9"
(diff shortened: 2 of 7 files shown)


leogr (dismissed):

leogr (approved):

sgaist (approved):

leogr:
/release-note-edit
```release-note
feat(userspace)!: evaluate rule conditions on raw field bytes (invalid and non-printable UTF-8 sequences are now matchable with string operators, single bytes with the new `\xHH` escape sequence), let `regex` operate on sanitized input, and escape non-printable characters / replace invalid UTF-8 sequences at output encoding time; `FALCO_ENGINE_VERSION` bumped from `0.62.0` to `0.63.0`
BREAKING CHANGE: conditions are evaluated on raw bytes; regex operates on sanitized input; at output encoding stage, non-printable characters are escaped and invalid UTF-8 sequences replaced. Rules that matched the replacement character with non-regex operators must be reviewed
```

PR #3943, opened 2026-07-14
fix(userspace/falco): avoid untrusted search path when loading the kernel module

BoxStrikesTeam:
**What type of PR is this?**

/kind bug

/area engine

`open_live_inspector()` falls back to `system("modprobe " DRIVER_NAME " ...")`
when the kmod engine's initial `open_kmod()` call throws (i.e. the kernel
module isn't already loaded). Since `system()` resolves the unqualified
`modprobe` binary via `/bin/sh -c` and the process's `PATH`, this is an
instance of CWE-426 (Untrusted Search Path): in deployments where an
attacker can influence the Falco process's `PATH` (or write to a directory
that precedes the legitimate modprobe location in it), an attacker-controlled
binary named `modprobe` would be executed with Falco's privileges.

This PR replaces the `system()` call with `fork()` + `execl()` against a
fixed set of absolute paths (`/usr/sbin/modprobe`, `/sbin/modprobe`),
eliminating the shell and the PATH dependency entirely. No functional
behavior change is intended — only the resolution mechanism.

Note: the maintained systemd flow (`falco-kmod-inject.service`) already
loads the module via an absolute path before Falco starts, so this fallback
is mainly reached in manual, non-systemd, or container invocations.

Reported and confirmed via the private security disclosure process
(cncf-falco-maintainers@lists.cncf.io); classified as low severity /
hardening (defense-in-depth), not tracked as a GHSA. Verified locally
that the fallback resolves only the fixed absolute paths regardless of
`PATH` contents.

```release-note
fix(userspace/falco): use the absolute `modprobe` path instead of a `PATH` lookup when falling back to loading the kernel module (untrusted search path, CWE-426)
```
Signed-off-by:Eneshan Erdoğan Karaca

### DIFF
--- a/userspace/falco/app/actions/helpers_inspector.cpp
+++ b/userspace/falco/app/actions/helpers_inspector.cpp
@@ -17,12 +17,15 @@ limitations under the License.
 
 #include <sys/types.h>
 #include <sys/stat.h>
-#include <sys/wait.h>
 #include <fcntl.h>
+
+#ifdef __linux__
+#include <sys/wait.h>
 #include <unistd.h>
 #include <cerrno>
 #include <cstring>
 #include <array>
+#endif
 
 #include <libsinsp/plugin_manager.h>
 #include <configuration.h>
@@ -32,6 +35,7 @@ limitations under the License.
 using namespace falco::app;
 using namespace falco::app::actions;
 
+#ifdef __linux__
 namespace {
 
 // Loads the Falco kernel module by invoking modprobe through a fixed,
@@ -40,6 +44,10 @@ namespace {
 // attacker-controllable in some deployment scenarios (CWE-426: Untrusted
 // Search Path). No shell is involved, and no user-controlled input is
 // interpolated into the call.
+//
+// POSIX-only (fork/waitpid/WIFEXITED aren't available on Windows), so
+// this is gated behind __linux__; non-Linux builds fall straight through
+// to open_kmod() at the call site, unchanged from prior behavior.
 bool falco_modprobe(const char* module_name) {
 	static constexpr std::array<const char*, 2> modprobe_paths = {
 	        "/usr/sbin/modprobe",
@@ -84,7 +92,16 @@ bool falco_modprobe(const char* module_name) {
 	}
 
 	int status = 0;
-	if(waitpid(pid, &status, 0) < 0) {
+	int wait_ret;
+	do {
+		// Retry on EINTR: a signal (e.g. the SIGALRM used by the metrics
+		// timer, installed without SA_RESTART) landing mid-wait would
+		// otherwise cause us to report a spurious failure even though the
+		// module loaded successfully.
+		wait_ret = waitpid(pid, &status, 0);
+	} while(wait_ret < 0 && errno == EINTR);
+
+	if(wait_ret < 0) {
 		falco_logger::log(falco_logger::level::ERR,
 		                  std::string("waitpid() failed while loading kernel module: ") +
 		                          strerror(errno) + "\n");
@@ -95,6 +112,7 @@ bool falco_modprobe(const char* module_name) {
 }
 
 }  // namespace
+#endif  // __linux__
 
 falco::app::run_result falco::app::actions::open_offline_inspector(falco::app::state& s) {
 	try {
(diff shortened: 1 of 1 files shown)


userspace/falco/app/actions/helpers_inspector.cpp
```diff
@@ -27,6 +32,70 @@ limitations under the License.
 using namespace falco::app;
 using namespace falco::app::actions;
 
+namespace {
```
leogr, line 39: `helpers_inspector.cpp` is compiled on all platforms (it's an unconditional source in `userspace/falco/CMakeLists.txt`, and the `falco` target is built on Windows too), but `<sys/wait.h>`, `fork()`, `waitpid()` and `WIFEXITED()` don't exist on Windows (MSVC), so this will break the `build-win32-package` job as soon as it reaches this file.

May you wrap the POSIX bits in `#ifdef __linux__`, like we already do in `create_signal_handlers.cpp`? That's the new includes (`<sys/wait.h>`, `<unistd.h>`), this `falco_modprobe` block, and the call site below. On non-Linux it then falls through to `open_kmod()` just like today. 🙏

userspace/falco/app/actions/helpers_inspector.cpp
```diff
@@ -27,6 +32,70 @@ limitations under the License.
...
+			dup2(devnull, STDERR_FILENO);
+			close(devnull);
+		}
+		execl(modprobe_bin, modprobe_bin, module_name, (char*)nullptr);
+		// execl only returns on failure.
+		_exit(127);
+	}
+
+	int status = 0;
+	if(waitpid(pid, &status, 0) < 0) {
```
leogr:
```suggestion
	int wait_ret;
	do {
		wait_ret = waitpid(pid, &status, 0);
	} while(wait_ret < 0 && errno == EINTR);
	if(wait_ret < 0) {
```

Nit, non-blocking. `waitpid` here can return `EINTR` if a signal lands mid-wait (e.g. our metrics `SIGALRM`, which we install without `SA_RESTART` in `stats_writer.cpp`). If that happens we'd log a spurious "Unable to load the driver" even though the module actually loaded. Retrying on `EINTR` avoids it. 🙏

leogr (commented):
Hey @BoxStrikesTeam, thanks a lot for this, and welcome! 🎉 

Overall SGTM, I just found an issue (see below).

Note: this PR has been reviewed using https://github.com/leogr/falco-expert

userspace/falco/app/actions/helpers_inspector.cpp
BoxStrikesTeam: Hello, I have applied a new fix for this. Is it suitable?

leogr (dismissed):

userspace/falco/app/actions/helpers_inspector.cpp
```diff
@@ -122,9 +209,11 @@ falco::app::run_result falco::app::actions::open_live_inspector(falco::app::stat
 				falco_logger::log(
 				        falco_logger::level::INFO,
 				        "Trying to inject the Kernel module and opening the capture again...");
-				if(system("modprobe " DRIVER_NAME " > /dev/null 2> /dev/null")) {
+#ifdef __linux__
+				if(!falco_modprobe(DRIVER_NAME)) {
```
sgaist, line 214: Shouldn't the log message also be tied to Linux since the loading only happens on that platform ?
```suggestion
#ifdef __linux__
				falco_logger::log(
				        falco_logger::level::INFO,
				        "Trying to inject the Kernel module and opening the capture again...");

				if(!falco_modprobe(DRIVER_NAME)) {
```

sgaist (changes requested):
One small question/suggestion otherwise looks good

userspace/falco/app/actions/helpers_inspector.cpp
leogr: cc @BoxStrikesTeam

leogr (approved):

leogr:
Hey @sgaist 

PTAL again please :pray:

sgaist (approved):

leogr:
/release-note-edit
```release-note
fix(userspace/falco): use the absolute `modprobe` path instead of a `PATH` lookup when falling back to loading the kernel module (untrusted search path, CWE-426)
```

Issue #3725, opened 2025-10-27
`--validate` option fails at parsing comments in yaml file


NicolasCARPi:
Hello,

Thank you for this great piece of software! I'm submitting this report as I believe it's a clear bug!

------

**Describe the bug**

Using `falco --validate /etc/falco/falco.yaml` results in invalid configuration with errors stemming from the fact that the YAML file contains comments.

**How to reproduce it**

Have a config file that comes with the package (in my case https://falco.org/repo/falcosecurity-rpm.repo).

`$  head /etc/falco/falco.yaml`

~~~yaml
# SPDX-License-Identifier: Apache-2.0
#
# Copyright (C) 2025 The Falco Authors.
#
#
# Licensed under the Apache License, Version 2.0 (the "License");
# you may not use this file except in compliance with the License.
# You may obtain a copy of the License at
#
#     http://www.apache.org/licenses/LICENSE-2.0
~~~

**Expected behaviour**

I can validate the main file and other rules files.

Output of `falco --validate /etc/falco/falco.yaml`

~~~
Mon Oct 27 12:29:10 2025: Falco version: 0.42.0 (x86_64)
Mon Oct 27 12:29:10 2025: Falco initialized with configuration files:
Mon Oct 27 12:29:10 2025:    /etc/falco/config.d/engine-kind-falcoctl.yaml | schema validation: ok
Mon Oct 27 12:29:10 2025:    /etc/falco/config.d/falco.container_plugin.yaml | schema validation: ok
Mon Oct 27 12:29:10 2025:    /etc/falco/falco.yaml | schema validation: ok
Mon Oct 27 12:29:10 2025: System info: Linux version 5.14.0-570.52.1.el9_6.x86_64 (mockbuild@iad1-prod-build001.bld.equ.rockylinux.org) (gcc (GCC) 11.5.0 20240719 (Red Hat 11.5.0-5), GNU ld version 2.35.2-63.el9) #1 SMP PREEMPT_DYNAMIC Wed Oct 15 13:59:22 UTC 2025
Mon Oct 27 12:29:10 2025: Loaded plugin 'container@0.4.0' from file /usr/share/falco/plugins/libcontainer.so
Mon Oct 27 12:29:10 2025: [libs]: container: Enabled 'podman' container engine.
Mon Oct 27 12:29:10 2025: [libs]: container: * enabled container runtime socket at '/run/podman/podman.sock'
[...]

PR #3947, opened 2026-07-19
feat(userspace/falco): clarify rules validation errors

AkashKumar7902:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind release

> /kind chart-release

> /area build

/area engine

/area tests

> /area proposals

> /area automation

> /area chart

When `--validate` receives a top-level YAML mapping such as `falco.yaml`, rules validation currently reports only a structural error. This change adds an actionable hint explaining that `--validate` accepts rules files and that `--dry-run` validates a Falco configuration together with its configured rules.

The hint is added in the `validate_rules_files` application action, so normal rule loading and embedded engine callers are unchanged. JSON stdout remains machine-parseable; the hint stays in the fatal error summary written to stderr.

Related to #3725

This is a draft because #3725 is assigned to @leogr; I would appreciate confirmation that the diagnostic wording and app-layer placement match the intended scope.

Validation:

- Repository `clang-format` 18.1.8 and `cmake-format` 0.6.13 hooks
- CI-matching minimal Linux build of `falco_unit_tests`
- All three focused `ActionValidateRulesFiles` regressions
- Full unit suite: 214/214 tests passed
- Manual CLI E2E: baseline returned the generic error; the patch retained exit code 1 and added the `--dry-run` hint to stdout and stderr
- Text, JSON-preservation, and non-config negative cases

```release-note
feat(userspace/falco): suggest `--dry-run` when `--validate` receives a Falco configuration file
```

### DIFF
--- a/unit_tests/falco/app/actions/test_validate_rules_files.cpp
+++ b/unit_tests/falco/app/actions/test_validate_rules_files.cpp
@@ -1,0 +1,107 @@
+// SPDX-License-Identifier: Apache-2.0
+/*
+Copyright (C) 2026 The Falco Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "app_action_helpers.h"
+
+#include <nlohmann/json.hpp>
+
+#include <cstdio>
+#include <fstream>
+#include <iostream>
+#include <sstream>
+#include <string>
+
+namespace {
+const std::string s_hint =
+        "--validate accepts Falco rules files only. To validate a Falco configuration file "
+        "and its configured rules, use --dry-run.";
+
+class stdout_capture {
+public:
+	stdout_capture(): m_original(std::cout.rdbuf(m_stream.rdbuf())) {}
+	~stdout_capture() { std::cout.rdbuf(m_original); }
+
+	std::string str() const { return m_stream.str(); }
+
+private:
+	std::ostringstream m_stream;
+	std::streambuf* m_original;
+};
+
+falco::app::run_result validate_content(const std::string& filename,
+                                        const std::string& content,
+                                        bool json_output,
+                                        std::string& output) {
+	{
+		std::ofstream file(filename);
+		if(!file.is_open()) {
+			return falco::app::run_result::fatal("Could not create test file " + filename);
+		}
+		file << content;
+	}
+
+	falco::app::state s = {};
+	s.options.validate_rules_filenames = {filename};
+	s.config->m_json_output = json_output;
+
+	falco::app::run_result result;
+	{
+		stdout_capture capture;
+		result = falco::app::actions::validate_rules_files(s);
+		output = capture.str();
+	}
+	EXPECT_EQ(std::remove(filename.c_str()), 0);
+	return result;
+}
+}  // namespace
+
+TEST(ActionValidateRulesFiles, suggests_dry_run_for_config_file) {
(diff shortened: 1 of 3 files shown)


userspace/falco/app/actions/validate_rules_files.cpp
```diff
@@ -106,6 +122,12 @@ falco::app::run_result falco::app::actions::validate_rules_files(falco::app::sta
 			summary += filename + ": Ok, with warnings";
 			falco_logger::log(falco_logger::level::WARNING, res->as_string(true, rc) + "\n");
 		}
+
+		if(!res->successful() && has_yaml_mapping_root(rc.at(filename).get())) {
+			summary +=
+			        "\nHint: --validate accepts Falco rules files only. To validate a Falco "
+			        "configuration file and its configured rules, use --dry-run.";
+		}
```
leogr, line 130:
```suggestion
		if(!res->successful() && has_yaml_mapping_root(rc.at(filename).get())) {
			summary +=
			        "\nHint: --validate expects a rules file, that is a YAML list of rule, macro, "
			        "and list items. If this is a Falco configuration file, validate it and its "
			        "configured rules with `falco -c " +
			        filename + " --dry-run` instead.";
		}
```

Two things about the current wording 👇

**It also fires on rules files.** The heuristic is "the root is a YAML map", and a rules file that is just missing the leading `- ` is exactly that. `rule_loader_reader.cpp` throws `Rules content is not yaml array of objects` for any map root, so a rules author who forgot the dash is told `--validate accepts Falco rules files only` and pointed at `--dry-run`, which is not what they need.

**`--dry-run` alone validates the default config, not the file that was passed.** So the user can get a false green:

```
$ falco --dry-run
Falco initialized with configuration files:
   /etc/falco/falco.yaml | schema validation: ok
$ echo $?
0

$ falco -c /tmp/mycfg.yaml --dry-run
LOAD_ERR_YAML_VALIDATE (...): Rules content is not yaml array of objects
$ echo $?
1
```

The suggestion covers both, since it says what `--validate` expects (so the missing dash case becomes self-explanatory) and it spells out the full command.

N.B. I checked the suggestion against `clang-format` and it builds as is.

unit_tests/falco/app/actions/test_validate_rules_files.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include <cstdio>
+#include <fstream>
+#include <iostream>
+#include <sstream>
+#include <string>
+
+namespace {
+const std::string s_hint =
+        "--validate accepts Falco rules files only. To validate a Falco configuration file "
+        "and its configured rules, use --dry-run.";
```
leogr, line 31:
```suggestion
const std::string s_hint =
        "--validate expects a rules file, that is a YAML list of rule, macro, and list items.";
```

To match the reworded hint, see my other comment.

Also, may you assert that the hint carries the command too, something like `falco -c <filename> --dry-run`? That is the part that actually tells the user what to run, so I believe it is worth covering 🙏

sgaist (commented):
I agree with @leogr's comments

leogr:
Hey @AkashKumar7902

Thanks for this PR! Since we want it in 0.45.0 and the branch was out of date, I took the liberty to push directly on your branch (thanks for leaving "allow edits from maintainers" on). Here is what changed 👇

- rebased on the current `master` (the only conflict was in `unit_tests/CMakeLists.txt`, since `test_pidfile.cpp` landed in the meantime, so I kept both entries)
- applied my two suggestions above: the hint now says what `--validate` expects (a YAML list of rule, macro, and list items) and spells out the full command, i.e. `falco -c <file> --dry-run`, so it also makes sense for a rules file that is just missing the leading `-`
- the unit tests now assert that the hint carries that command too (in the text output and in the error string, but not in the JSON stdout)
- squashed everything in your commit, keeping you as the author, and added my sign-off

Everything else is as you wrote it. Once the CI is green, we can merge it.

Thanks again for your contribution 🙏

leogr (approved):
Both suggestions applied and the command assertion is in, the rest is as the author wrote it. LGTM

leogr:
@sgaist PTAL :pray:

sgaist (approved):

PR #3978, opened 2026-09-09
fix(chart): mount the container engine socket directories, not the socket files

leogr:
**What type of PR is this?**

/kind bug

/area chart

Re-files https://github.com/falcosecurity/charts/pull/1058 by @arpitjain099 against the chart source. The commit is imported as-is (authorship preserved); only the `Chart.yaml` bump and the CHANGELOG hunk were left out, since the chart version is bumped by the chart release PR only.

`falco.containerPluginVolumes` and `falco.containerPluginVolumeMounts` mount each entry of `collectors.containerEngine.engines.*.sockets` as a **file** hostPath. A file bind mount pins the inode: when the container runtime restarts and recreates its socket, the Falco pod keeps the old dead one, and every container created after the restart shows `<NA>` for the container and k8s fields until the pod is restarted.

This is a regression of a 2024 fix that was lost when the container plugin volumes replaced the old per-engine volumes:

- original report: https://github.com/falcosecurity/charts/issues/632
- original fix (directory mount instead of the socket file): https://github.com/falcosecurity/charts/pull/633
- new report: https://github.com/falcosecurity/charts/issues/1052

Both helpers now mount `dir $socket` at `/host<dir>`, so `/host<socket>` still resolves for the plugin configuration, which passes the socket paths unchanged. Deduplication moves to the directory, so two sockets in the same directory share one volume. Volume names and their order are unchanged.

With the default values, the pod now mounts these six host directories instead of the six socket files:

- `/var/run`
- `/run/podman`
- `/run/host-containerd`
- `/run/containerd`
- `/run/crio`
- `/run/k3s/containerd`

N.B. This widens each hostPath from a socket file to its directory. The Falco container already runs privileged with the default values, so it is not a new privilege, but it is worth noting for people reviewing their pod security posture.

Related to #3979 (transferred from https://github.com/falcosecurity/charts/issues/1052)

Re-files https://github.com/falcosecurity/charts/pull/1058

- `make chart-check` passes locally (helm v4.2.2, helm-docs v1.11.0, go 1.27)
- the unit tests in `chart/falco/tests/unit/containerPlugin` (`volumes_test.go`, `volumeMounts_test.go`) now assert the directories, and one more case covers two sockets in the same directory sharing one volume
- I verified the inode pinning with a plain bind mount: after unlinking and recreating the source file, the file bind mount still points at the old inode, while a directory bind mount sees the new file

```release-note
fix(chart): mount the container engine socket directories instead of the socket files, so Falco keeps working after a container runtime restart
```

### DIFF
--- a/chart/falco/CHANGELOG.md
+++ b/chart/falco/CHANGELOG.md
@@ -9,6 +9,7 @@ numbering uses [semantic versioning](http://semver.org).
 * Fix Grafana dashboard priority level mappings to match Falco's numeric encoding (0=emergency … 7=debug)
 - Add `revisionHistoryLimit` support to the DaemonSet controller and honor an explicit zero for both DaemonSet and Deployment controllers
 * Add `serviceAccount.labels` to set custom labels on the ServiceAccount
+* Mount the host directory that holds each container engine socket instead of the socket file, so Falco keeps working after a container runtime restart (regression of falcosecurity/charts#632, originally fixed by falcosecurity/charts#633). With the default values the pod now mounts `/var/run`, `/run/podman`, `/run/host-containerd`, `/run/containerd`, `/run/crio`, and `/run/k3s/containerd` instead of the six socket files
 
 ## v9.0.0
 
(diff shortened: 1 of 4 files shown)


sgaist (dismissed):

sgaist (approved):

c2ndev (approved):
/lgtm

PR #3984, opened 2026-09-10
fix(chart/falco): always mount an emptyDir at /etc/falco/config.d and bump the 0.45 component defaults

leogr:
**What type of PR is this?**

/kind bug

/area chart

Two chart changes for the upcoming 9.2.0 (Falco 0.45.0), found while checking the chart against `0.45.0-rc1`:

1. Always mount an emptyDir at `/etc/falco/config.d` in the Falco container.

The Falco image ships configuration snippets in `/etc/falco/config.d` (`falco.container_plugin.yaml` sets `load_plugins: [container]`). The chart mounted the `specialized-falco-configs` emptyDir over that directory only when the driver loader was enabled, so with `driver.kind=modern_ebpf` (or `driver.loader.enabled=false`) and both falcoctl containers disabled the image snippets stayed visible and were merged on top of the chart configuration (`config_files` uses the `append` strategy). `load_plugins` became `[container, container]` and Falco aborted at startup with `found another plugin with name container`.

N.B. This is pre-existing: it happens with chart 9.1.0 and Falco 0.44.1 too, it is not caused by 0.45.

2. Bump the defaults to the components shipped with Falco 0.45.0: the falcoctl image to `0.14.1`, `collectors.containerEngine.pluginRef` to `container:0.7.4`, and `collectors.kubernetes.pluginRef` to `k8smeta:0.4.2` (which also fixes the k8smeta init failure with purely numeric node names).

The chart version bump to 9.2.0 is left to the chart release PR.

- `make chart-check` passes locally (helm-docs v1.11.0)
- the new unit test `configdVolume_test.go` covers the default values, `modern_ebpf` and `driver.loader.enabled=false` without falcoctl, and the Deployment controller with the driver disabled
- I verified the failure and the fix with both `falco:0.44.1` and `falco:0.45.0-rc1`: the same pod with the empty `config.d` mounted starts fine
- README regenerated with helm-docs

/milestone 0.45.0

```release-note
fix(chart): always mount an emptyDir at /etc/falco/config.d, and bump the default falcoctl image to 0.14.1, the container plugin to 0.7.4, and the k8smeta plugin to 0.4.2
```


c2ndev (approved):
/lgtm

jasondellaluce (approved):
/approve

Issue #3967, opened 2026-09-02
[TRACKING] Falco `0.45.0` release


leogr:
# Falco 0.45.0 Release

Will keep this issue updated with the current status and progress.

## Date

Target release date: September 21st 2026

## Release Steps

The process is described in [this document](https://github.com/falcosecurity/falco/blob/master/RELEASE.md).

## External dependencies

- [ ] falcoctl
  - [x] driverkit `v0.23.2` bumped in falcoctl (CI green) 👉 https://github.com/falcosecurity/falcoctl/pull/1067
  - [x] `google.golang.org/grpc` bump 👉 https://github.com/falcosecurity/falcoctl/pull/1066
  - [x] drop the direct `github.com/docker/docker` dependency 👉 https://github.com/falcosecurity/falcoctl/pull/1065
  - [x] dismiss the 4 `github.com/docker/docker` advisories as not reachable (see the note below) 👉 done on September 3rd, no open Dependabot alerts left in falcoctl
  - [x] release 👉 https://github.com/falcosecurity/falcoctl/releases/tag/v0.14.0 (September 4th; assets, provenance, and `linux/amd64`+`linux/arm64` images verified)
  - [ ] bump `FALCOCTL_VERSION` in `falco/cmake/modules/falcoctl.cmake` (today `0.13.0`, target `0.14.0`)
- [ ] container plugin (**blocks libs**, see the note below)
  - already merged and awaiting a release: `fix(plugins/container): recognize systemd nerdctl cgroups`, the `go.podman.io/podman/v6` + `containerd v2.2.5` migration, and several `go-worker` dependency bumps (`runc`, `buildah`, `fulcio`, `x/net`, `x/crypto`, `grpc`)
  - [x] CI unblocked 👉 https://github.com/falcosecurity/plugins/issues/1510 (GitHub runner image regression, not our code; the test fix from https://github.com/falcosecurity/plugins/pull/1511 plus the image rollback: the re-run of the last failed e2e run on `main` is green on September 4th)
  - [x] safety check of the `main` build (`0.7.2`, commit `fca54d5e7`) against the released Falco `0.44.1`: plugin loads, all engines enabled, modern eBPF probe attached, `Terminal shell in container` fires with full container metadata (id, name, image repository and tag) 👉 done locally on September 4th, not covered by the plugins CI (which validates against `falco:0.41.3` and runs the e2e with `falco:master-debian`)
  - [x] decide which pending PRs go in:
[...]

PR #3990, opened 2026-09-11
update(chart): bump the falcosidekick and falco-talon subchart constraints

leogr:
**What type of PR is this?**

/kind cleanup

/area chart

This PR bumps the subchart dependency constraints of the Falco chart:

- `falcosidekick`: `0.12.*` becomes `0.14.*` (resolves to `0.14.0`)
- `falco-talon`: `0.3.*` becomes `0.4.*` (resolves to `0.4.2`)

Both subcharts keep the same application versions (Falcosidekick `2.31.1` and Falco Talon `0.3.0`). The new chart lines only add opt-in options and template fixes:

- `falcosidekick`: GCP Workload Identity and Gateway API `HTTPRoute` support (both disabled by default), plus the `curlimages/curl` image and the `/healthz` endpoint for the Helm test pod
- `falco-talon`: the missing `namespace` in the config Secret, a `rules-checksum` annotation so that Talon restarts when its rules or `config.rulesOverride` change, and a folder annotation for the Grafana dashboards

`k8s-metacollector` is unchanged. The chart `version` is not bumped here since that will happen in the chart release PR.

Related to #3967

- With the default values, the `helm template` output is identical before and after the bump
- With `falcosidekick.enabled=true` and/or `responseActions.enabled=true`, the only differences are the `helm.sh/chart` labels (and the `checksum/config` annotation derived from them), the test pod image and endpoint, the Talon Secret `namespace`, and the Talon `rules-checksum` annotation
- `make chart-check` passes locally, `ct lint` and `ct install` are left to the CI

```release-note
update(chart): bump the falcosidekick subchart constraint to 0.14.* and the falco-talon subchart constraint to 0.4.*
```

### DIFF
--- a/chart/falco/CHANGELOG.md
+++ b/chart/falco/CHANGELOG.md
@@ -5,6 +5,8 @@ numbering uses [semantic versioning](http://semver.org).
 
 ## Unreleased
 
+* Bump the `falcosidekick` subchart dependency constraint to `0.14.*` (Falcosidekick 2.31.1; adds the GCP Workload Identity and Gateway API `HTTPRoute` options, both disabled by default, and switches the Helm test pod to the `curlimages/curl` image and the `/healthz` endpoint)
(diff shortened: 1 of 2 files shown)


c2ndev (approved):
/lgtm

sgaist (approved):

PR #3994, opened 2026-09-14
fix(packaging): restore services and drivers after upgrades

leogr:
**What type of PR is this?**

/kind bug

/area build
/area tests

Package upgrades can leave Falco stopped or without a loadable kmod. On RPM, the outgoing package's `%preun` stops the service and removes DKMS after the incoming `%post` has started it. On Debian, `--force-confold` retains the old package-generated driver pin, which is incompatible with the RC2 driver API.

This PR fixes the package lifecycle in one place:

- Move RPM driver installation and startup to `%posttrans`, and guard removal cleanup against upgrades. This also handles the already-shipped 0.44.1 scripts.
- Capture the outgoing default before replacing the binary. Migrate matching driver pins, preserve different custom pins, and carry the selected driver through the RPM transaction.
- Verify that `modprobe` can find the selected driver. Repair the cache-only success path in the currently bundled falcoctl through DKMS before starting the service.
- Disable Debian units before deleting their definitions, remove known dangling aliases, and clean up masks recorded as package-created.

Found while testing the 0.44.1 to 0.45.0-rc2 package upgrades. The underlying package-script defects predate RC2.

`FALCO_DRIVER_CHOICE=none` still skips driver setup. A deliberate pin equal to the outgoing default needs an explicit `FALCOCTL_DRIVER_VERSION` during the transaction. Administrator masks and legacy masks without ownership records are retained. RPM generation requires CMake 3.18 or newer for the post-transaction hook.

Validation: 14 package-script tests, formatting checks, and real Ubuntu 24.04 / Rocky 9.8 transactions using the generated scripts with the original RC2 binaries and falcoctl 0.14.1. Both kmod upgrades pass, including Debian `--force-confold`, warm-cache DKMS repair, restart, and reboot. Fresh install/removal checks cover alias cleanup, package-created masks, and administrator-mask preservation.

The standalone installer fix is tracked in https://github.com/falcosecurity/falcoctl/pull/1089. This package fix works with the currently bundled falcoctl.

N.B. The fixture validates CMake/CPack hook generation and package metadata; it is not a full C++ rebuild.

```release-note
fix(packaging): restore Falco services and persistent kernel modules after upgrades, migrate package-default driver pins, and clean up package-created systemd state on removal
```


leogr:
tentatively for 
/milestone 0.45.0

c2ndev (approved):

sgaist (approved):

Issue #4000, opened 2026-09-16
The unsupported syscalls warning is printed also when replaying a capture


leogr:
**Describe the bug**

`configure_interesting_sets` runs for every engine kind, so `check_for_rules_unsupported_events()` prints the `unsupported-evttype` warning also when replaying a capture with `-r`, where the selected syscall set does not matter (the capture contains whatever was recorded).

- https://github.com/falcosecurity/falco/blob/7b7266f8a8b3fdd501e48a27d65edf3a7039d595/userspace/falco/app/app.cpp#L52-L84
- https://github.com/falcosecurity/falco/blob/7b7266f8a8b3fdd501e48a27d65edf3a7039d595/userspace/falco/app/actions/configure_interesting_sets.cpp#L256-L279

**How to reproduce it**

```shell
falco -r <capture> -r <rules with a rule on read>
```

The replay works as expected, but Falco warns that `read` is not activated (e.g. with the `kexec_x86.scap` capture from the libs tests).

**Expected behaviour**

No warning in replay mode, since `base_syscalls` does not apply to a capture.

**Additional context**

- The warning text still suggested the removed `-A` flag, that part is addressed separately
- Found during the Falco `0.45.0` testing, not blocking for it

/kind bug

PR #4001, opened 2026-09-16
fix(falco): drop the removed -A flag from the unsupported syscalls warning

leogr:
**What type of PR is this?**

/kind bug

The warning printed when the loaded rules use syscalls that are not activated, and the related debug log line, still suggest the `-A` flag, which does not exist anymore. This PR only updates the text, so that it points to the current options: `base_syscalls.all` and `base_syscalls.custom_set`.

Related to https://github.com/falcosecurity/falco/issues/4000

Part of https://github.com/falcosecurity/falco/issues/3967

Text-only change, found while testing `0.45.0-rc3`. The same commit goes to `release/0.45.x` too, so that the final release does not suggest a removed flag.

```release-note
fix(falco): the unsupported syscalls warning now points to `base_syscalls.all` instead of the removed `-A` flag
```

/milestone 0.45.0


sgaist (approved):

c2ndev (approved):
/lgtm

PR #4003, opened 2026-09-16
fix(falco): do not warn about unsupported syscalls when replaying a capture

### LINKED ISSUE #4000: The unsupported syscalls warning is printed also when replaying a capture

leogr:
**What type of PR is this?**

/kind bug

When replaying a capture with `-r`, Falco still prints the `unsupported-evttype` warning if the loaded rules use syscalls that are not in the selected set. The selected event set does not apply to a capture file, which contains whatever was recorded, so the warning is just noise there. This PR skips the check in capture mode.

Fixes https://github.com/falcosecurity/falco/issues/4000

Part of https://github.com/falcosecurity/falco/issues/3967

Six lines, early return only. The same commit goes to `release/0.45.x` too. Follows up on the warning text fix 👇
- https://github.com/falcosecurity/falco/pull/4001

```release-note
fix(falco): no unsupported syscalls warning is printed anymore when replaying a capture
```

/milestone 0.45.0


sgaist (approved):

c2ndev (approved):
/lgtm

PR #4005, opened 2026-09-16
feat(falco): expose reload status and Unix socket control

c2ndev:
**What type of PR is this?**

/kind bug
/kind feature

/area engine
/area tests

This feature is designed for Falco Operator integration: request a reload and verify whether the new configuration was applied or rejected. A health check alone cannot distinguish the previous runtime from a successfully reloaded one.

It also fixes SIGHUP handling during runtime teardown by keeping signal reception and pending requests alive across reloads.

Reload control follows an **asynchronous request-reply pattern with status polling**:

- `POST /reload` queues a reload and returns `202 Accepted`, a process identity, a generation baseline, and `Location: /reload`. It is available only through the optional Unix socket listener, disabled by default and supported on non-minimal Linux builds.
- `GET /reload` reports readiness and started/applied/rejected generations over Unix and TCP. TCP observation requires `webserver.enabled` and does not depend on `reload_control.enabled`. HTTP `200` means the status was read; the response body determines the reload outcome.

Readiness is published after all enabled sources have started. Failed validation records a rejection and preserves the active runtime. Manual reloads also work with `watch_config_files: false`.

The Unix listener uses filesystem permissions for access control. Directory validation runs before stopping the active runtime and repeats at startup, without disturbing the live socket during validation.

No linked issue.

Requests may coalesce: generations track runtime progress, not individual jobs. Clients must finish writing configuration before requesting reload and tolerate temporary listener outages. Directory preflight cannot eliminate filesystem races or guarantee successful binding.

```release-note
feat(falco): Expose reload status through GET /reload and add optional reload control over a protected Unix socket on Linux. Preserve SIGHUP reception and pending requests across reloads.
```


userspace/falco/reload_control.cpp
```diff
@@ -0,0 +1,314 @@
...
+			                       {"started_generation", status.started_generation},
+			                       {"applied_generation", status.applied_generation},
+			                       {"rejected_generation", status.rejected_generation},
+			                       {"ready", status.ready}};
+			res.set_header("Cache-Control", "no-store");
+			res.set_content(body.dump(), "application/json");
+		});
+		m_server->Post("/-/reload", [](const httplib::Request& req, httplib::Response& res) {
+			res.set_header("Cache-Control", "no-store");
+			if(!req.body.empty() || req.target.find('?') != std::string::npos) {
```
leogr: `cpp-httplib` stores multipart fields/files in `req.form`, leaving `req.body` empty. A nonempty multipart POST therefore passes this check and queues a reload, despite the empty-body contract. I reproduced the admission check with the pinned `v0.23.1`: an 80-byte form returned `202`.

May you reject multipart requests too and add a regression test checking `400` and an unchanged request counter? Non-blocking, since this does not bypass the socket access controls.

leogr (commented):
Just one non-blocking issue found. Everything else SGTM. I will give it a second look soon.

leogr:
Tentatively for 
/milestone 0.45.0

We need this for the falco-operator.

I will issue a new RC once this gets merged.

leogr (dismissed):
👍

leogr:
@sgaist PTAL :pray:

unit_tests/falco/test_reload_control_runtime.cpp
```diff
@@ -0,0 +1,522 @@
...
+
+extern char** environ;
+
+namespace {
+
+using json = nlohmann::json;
+using clock_type = std::chrono::steady_clock;
+using namespace std::chrono_literals;
+
+void require(bool condition, const std::string& message) {
```
sgaist: This method and the two `require_in_child` share the same implementation.

Should we have that in a single place ?

unit_tests/falco/test_reload_control_runtime.cpp
```diff
@@ -0,0 +1,522 @@
...
+	if(!condition) {
+		throw std::runtime_error(message);
+	}
+}
+
+class scoped_fd {
```
sgaist: The same class is declared twice, maybe have them defined in a single place.

sgaist (changes requested):
Overall, looks very good, just some minor nitpicks.

c2ndev:
Hi @sgaist, could you check it again? TY

leogr (approved):

sgaist (approved):

